Issue #1232, opened 2016-03-09
View in Excel produces error in


MikhailArkhipov:
m <- matrix(paste0(replicate(1000, "n"), collapse = ''), 1, 1)

'Error in x[rows, cols] : subscript out of bounds'

Issue #1227, opened 2016-03-08
'Help on function' doesn't open Help window


JeongPyoHam:
In editor,
f <- function(x) { <something> }

right click on keyword 'function'
The last item in pop-up shows 'Help on function', and then run it.

REPL prompt changes to + and waiting for further input, and R Help window doesn't show (if it was not open before starting repro)

jflam:
Checked in.

Issue #1065, opened 2016-02-29
AF when requesting help on keyword


MikhailArkhipov:
Right click on 'while' in the editor and choose 'Help on while' (eval calls ?while). appears to be ignorable.

![image](https://cloud.githubusercontent.com/assets/12820357/13400203/8e629f02-debb-11e5-8154-c17381e6226c.png)

MikhailArkhipov:
Please verify

JeongPyoHam:
verified

PR #1239, opened 2016-03-09
#1232 #1227 #1065 

MikhailArkhipov:


src/Common/Core/Impl/Extensions/StringExtensions.cs
```diff
@@ -104,5 +104,17 @@ public static class StringExtensions {
             }
             return result;
         }
+
+        public static string SurroundWithBackticks(this string s) {
```
AlexanderSher: I would've call it `EnsureSurroundedWithBackticks` or something similar

int19h: I'd call it `ToRSymbol`.

int19h: Or rather `ToRName`, because that's the terminology that R spec uses.

int19h:
:+1:

src/Package/Impl/Help/ShowHelpOnCurrentCommand.cs
```diff
@@ -32,8 +34,10 @@ internal sealed class ShowHelpOnCurrentCommand : PackageCommand {
         private readonly IRInteractiveWorkflow _workflow;
         private readonly IActiveWpfTextViewTracker _textViewTracker;
 
+        private IActiveRInteractiveWindowTracker _activeReplTracker;
         public ShowHelpOnCurrentCommand(IRInteractiveWorkflow workflow, IActiveWpfTextViewTracker textViewTracker) :
             base(RGuidList.RCmdSetGuid, RPackageCommandId.icmdHelpOnCurrent) {
+            _activeReplTracker = VsAppShell.Current.ExportProvider.GetExportedValue<IActiveRInteractiveWindowTracker>();
```
AlexanderSher: please provide it as ctor parameter so command can be tested without stub injection into VsTestCompositionCatalog

src/Common/Core/Impl/Extensions/StringExtensions.cs
AlexanderSher: Do we have any R-specific code in Common?

int19h: You're right, we don't. So it should be moved to where `ToRStringLiteral` is.

PR #1242, opened 2016-03-09
Several build related fixes

AlexanderSher:
- Fix *proj files so that every project uses its own intermediate folder
- Remove intermediate folder cleanup from build.proj
- Add RenameSetupBundle step to the build.proj
- Change default platform to "any cpu"
- Update submodule reference

### DIFF
--- a/src/build.proj
+++ b/src/build.proj
@@ -13,7 +13,7 @@
     <WixBinSdkDirectory Condition="'$(WixBinSdkDirectory)' == ''">$(WixBinDirectory)sdk\</WixBinSdkDirectory>
 
     <Configuration Condition=" '$(Configuration)'=='' ">Release</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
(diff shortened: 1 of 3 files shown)


src/build.proj
```diff
@@ -13,7 +13,7 @@
     <WixBinSdkDirectory Condition="'$(WixBinSdkDirectory)' == ''">$(WixBinDirectory)sdk\</WixBinSdkDirectory>
 
     <Configuration Condition=" '$(Configuration)'=='' ">Release</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
+    <Platform Condition=" '$(Platform)' == '' ">any cpu</Platform>
```
int19h: But isn't it producing an x86 binary as output?

AlexanderSher: Depends on JIT, though I don't see the reason to limit the output to x86. And we use "any cpu" on build server anyway.

int19h:
:+1:

PR #1246, opened 2016-03-10
Split R.Wix.Settings.targets into separate .settings files for WiX paths and everything else

int19h:


int19h:
Reviewed directly.

PR #1253, opened 2016-03-11
Add functions to return environment list

JeongPyoHam:
- prep for further work to support environment list in UI

### DIFF
--- a/src/Debugger/Impl/rtvs/R/grid.R
+++ b/src/Debugger/Impl/rtvs/R/grid.R
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
     json;
 }
 
+getEnvRepr <- function(env, startEnvLevel) {
+  envRepr <- NULL;
+
+  if (startEnvLevel > 4L) {    # if within call or unknown, try to get call frame
+    nframe <- sys.nframe();
+    if (nframe > 0) {
+      for (i in 1:nframe) {
+        frame <- sys.frame(i)
+        if (identical(env, frame)) {
+          envRepr <- list(name = deparse(sys.call(i)[[1]]), level = 5L, frameIndex = i);
+          break;
+        }
+      }
+    }
+  }
+
+  if (is.null(envRepr)) {
+    if (identical(env, baseenv())) {
+      envRepr <- list(name = 'package:base', level = 2L);
+    } else if (identical(env, globalenv())) {
+      envRepr <- list(name = '.GlobalEnv', level = 4L);
+    } else if (identical(env, emptyenv())) {
+      envRepr <- list(name = 'EmptyEnv', level = 1L);
+    } else {
+      envRepr <- list(name = environmentName(env), level = 3L);
+    }
+  }
+  envRepr;
+}
+
+# return list of environment statck walking from the given environment up.
+# R environment level: empty(1) > base(2) > packages(3) > global(4) > calls(5) > unknown(10)
+getEnvironments <- function(env) {
+  if (missing(env)) {
+    curEnv <- sys.frame(-1);
+  } else {
+    curEnv <- env;
+  }
+  envs <- list();
+
+  prevEnvLevel <- 10L;
+
+  while (!identical(curEnv, emptyenv())) {
+    repr <- getEnvRepr(curEnv, prevEnvLevel);
+    if (repr$name != "Autoloads"
+      && repr$name != "rtvs::graphics::ide") {
+      envs[[length(envs) + 1]] <- repr;
+    }
+    prevEnvLevel <- repr$level;
+    if (prevEnvLevel == 5L) {
+      curEnv <- sys.frame(repr$frameIndex - 1);
+    } else {
+      curEnv <- parent.env(curEnv);
+    }
+  }
+  envs;
+}
+
+# return variable's frame index
+# look up starts from startFrameIndex and up
+getFrameIndex <- function(varName, startFrameIndex) {
+  if (missing(startFrameIndex)) {
+    startFrameIndex <- sys.nframe() - 1;
+  }
+  frameIndex <- startFrameIndex;
+  
+  while (frameIndex >= 0) {
+    if (exists(varName, frame = frameIndex, inherit = FALSE)) {
+      return(frameIndex);
+    }
+    frameIndex <- frameIndex - 1;
+  }
+  
+  -1L;
+}
+
--- a/src/Debugger/Impl/rtvs/R/grid.R
+++ b/src/Debugger/Impl/rtvs/R/grid.R
@@ -75,7 +75,7 @@ grid.dput <- function(obj) {
 getEnvRepr <- function(env, startEnvLevel) {
   envRepr <- NULL;
 
-  if (startEnvLevel > 4L) {    # if within call or unknown, try to get call frame
+  if (startEnvLevel > 4L) {    # 4L means global
     nframe <- sys.nframe();
     if (nframe > 0) {
       for (i in 1:nframe) {
@@ -93,8 +93,6 @@ getEnvRepr <- function(env, startEnvLevel) {
       envRepr <- list(name = 'package:base', level = 2L);
     } else if (identical(env, globalenv())) {
       envRepr <- list(name = '.GlobalEnv', level = 4L);
-    } else if (identical(env, emptyenv())) {
-      envRepr <- list(name = 'EmptyEnv', level = 1L);
     } else {
       envRepr <- list(name = environmentName(env), level = 3L);
     }
@@ -103,7 +101,7 @@ getEnvRepr <- function(env, startEnvLevel) {
 }
 
 # return list of environment statck walking from the given environment up.
-# R environment level: empty(1) > base(2) > packages(3) > global(4) > calls(5) > unknown(10)
+# R environment level: empty(1) > base(2) > packages, etc.(3) > global(4) > calls(5) > unknown(10)
 getEnvironments <- function(env) {
   if (missing(env)) {
     curEnv <- sys.frame(-1);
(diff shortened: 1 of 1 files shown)


src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
     json;
 }
 
+getEnvRepr <- function(env, startEnvLevel) {
+  envRepr <- NULL;
+
+  if (startEnvLevel > 4L) {    # if within call or unknown, try to get call frame
```
int19h: Where does the magic 4L constant come from?

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
...
+        if (identical(env, frame)) {
+          envRepr <- list(name = deparse(sys.call(i)[[1]]), level = 5L, frameIndex = i);
+          break;
+        }
+      }
+    }
+  }
+
+  if (is.null(envRepr)) {
+    if (identical(env, baseenv())) {
```
int19h: Why special-case these? They all have reasonable names.

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
     json;
 }
 
+getEnvRepr <- function(env, startEnvLevel) {
```
int19h: What is the purpose of this function in general? If it's just to get the name of the environment, why not just eval `environmentName` directly? And what is `startEnvLevel`, and `level` in the return value?

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
...
+    } else {
+      envRepr <- list(name = environmentName(env), level = 3L);
+    }
+  }
+  envRepr;
+}
+
+# return list of environment statck walking from the given environment up.
+# R environment level: empty(1) > base(2) > packages(3) > global(4) > calls(5) > unknown(10)
+getEnvironments <- function(env) {
```
int19h: Use `search()` instead to get everything on the search path, and use the existing `DebugSession.GetStackFramesAsync()` to get the call stack.

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
...
+  if (missing(env)) {
+    curEnv <- sys.frame(-1);
+  } else {
+    curEnv <- env;
+  }
+  envs <- list();
+
+  prevEnvLevel <- 10L;
```
int19h: What is the meaning of the magic constant `10L` here, and `5L` below?

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,79 @@ grid.dput <- function(obj) {
...
+    } else {
+      curEnv <- parent.env(curEnv);
+    }
+  }
+  envs;
+}
+
+# return variable's frame index
+# look up starts from startFrameIndex and up
+getFrameIndex <- function(varName, startFrameIndex) {
```
int19h: What is this going to be used for?

src/Debugger/Impl/rtvs/R/grid.R
int19h: Okay, so these are "levels" from the comment above? But then it is wrong. The order that you present is the one that `loadLibrary` / `library` will create, but there's also `data`, and arbitrary environments can be inserted into search path, so this hierarchy of levels does not hold in practice. Besides, there's no reason to distinguish between them at this point. The only worthwhile distinction is search path vs call stack.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: You found the answer yourself already. :)

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: function's enclosed environment doesn't have name

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: It will be more than just name progressively.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: those two are not in same domain. One is in R the other is in VS. Not easy to combine, and doesn't fit exactly to the purpose.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: attached env is in level of package. only global level is meaniningful for now. not a big deal.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: this PR is R side change for environment list

src/Debugger/Impl/rtvs/R/grid.R
int19h: The special-casing here is not for function environments, though.

src/Debugger/Impl/rtvs/R/grid.R
int19h: You will be calling them from VS in any case, and it shouldn't be hard to combine them.

I just don't want us to have two different bits of code written in R that do the same thing (namely, obtain the call stack). If there's any piece of information that is currently missing in the existing code, just add it there.

src/Debugger/Impl/rtvs/R/grid.R
int19h: Please use named constants, at least, to make the intent clear.

src/Debugger/Impl/rtvs/R/grid.R
int19h: Sure. But why would you need to fetch the index of a frame containing a variable by name from VS?

gaebalja1: Fine. I will remove that.

src/Debugger/Impl/rtvs/R/grid.R
gaebalja1: O.K. I will remove Boolean then.

src/Debugger/Impl/rtvs/R/grid.R
gaebalja1: I don't get it.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: Funny. Other account can make comment. (it's public now). 
I will remove the function. You will like that.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: I will change to boolean then.

src/Debugger/Impl/rtvs/R/grid.R
JeongPyoHam: I don't understand the comment.

int19h: I'm trying to figure out why there's a bunch of checks here for specific environments like `baseenv()` and `globalenv()`, and corresponding name assignments. Why not just call `environmentName()`, and use what it returns?

JeongPyoHam: environmentName doesn't return name for frame.

int19h: But it will return the name for `globalenv()`, `baseenv()` etc, which is what is being checked for here.

Better yet, use `search()`, iterate over names it returns, and use `as.environment` to get the corresponding environment objects (this will produce "package:base" rather than "base" for the base environment, among other things).

JeongPyoHam: base, global uses environmentName as you can see.

int19h: It doesn't. It uses a hardcoded "package.base" (on the line below this very comment) and ".GlobalEnv" (on line 95). That is what I'm asking about.

JeongPyoHam: cause it returns base and R_globalenv, and they are not user-friendly.

int19h: Hence the suggestion about `search()` above, then you won't need to hardcode anything.

JeongPyoHam: search doesn't take start environment parameter as you see apparently.

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -72,3 +72,53 @@ grid.dput <- function(obj) {
...
+      envRepr <- list(name = '.GlobalEnv', higherThanGlobal = FALSE);
+    } else {
+      envRepr <- list(name = environmentName(env), higherThanGlobal = FALSE);
+    }
+  }
+  envRepr;
+}
+
+# return list of environment statck walking from the given environment up.
+getEnvironments <- function(env) {
```
AlexanderSher: Do you plan to call this function for any environment, or only for those that are returned by `search()`?

JeongPyoHam: I don't like the UI that I'm seeing here in browser, as I should assume you're asking getEnvironments. The function getEnvironments might get other than search()'s return. I will see more as time goes by.

PR #1278, opened 2016-03-15
Fix #943: Remote Hadoop data sources cause Variable Explorer and debugger to refresh very slowly

### LINKED ISSUE #943: Remote Hadoop data sources cause Variable Explorer and debugger to refresh very slowly
Example of such a data source:
```r
dataSource <- RxXdfData(hdfsData, fileSystem = hdfsFS)
```

int19h:
Remove `DebugEvaluationResultFields.All`, requiring callers to always explicitly specify the fields that they need.

(Computing `DebugEvaluationResult.NameCount` was triggering this behavior, but it wasn't actually used anywhere.)

### DIFF
--- a/src/Debugger/Engine/Impl/AD7Expression.cs
+++ b/src/Debugger/Engine/Impl/AD7Expression.cs
@@ -34,7 +34,7 @@ namespace Microsoft.R.Debugger.Engine {
             _cts = new CancellationTokenSource();
             Task.Run(async () => {
                 try {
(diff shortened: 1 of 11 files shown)


huguesv:
:+1:

int19h:
@JeongPyoHam, can you take a look at the flags and make sure that they ask for everything that VE needs? All the unit tests that passed before the change still pass, so hopefully it is correct, but I'd like a second opinion.

Issue #829, opened 2016-02-05
Reduce full path names to ~/ names like in RS


MikhailArkhipov:
Compare RS and RTVS path output in REPL. RS displays ~/ path while we do all the c:\...

![image](https://cloud.githubusercontent.com/assets/12820357/12856873/e80c68b2-cbfb-11e5-9d24-b1ed1a88ce44.png)

PR #1292, opened 2016-03-17
Improve project system menu

MikhailArkhipov:
image

image


src/Package/Impl/ProjectSystem/Commands/CommandUtilities.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public static string GetSingleNodePath(this IImmutableSet<IProjectTree> nodes) {
+            if (nodes != null && nodes.Count == 1) {
+                return nodes.First().FilePath;
+            }
+            return string.Empty;
+        }
+
+        public static bool IsSingleNodePath(this IImmutableSet<IProjectTree> nodes) {
```
int19h: It seems like this could really just be `IsNullOrEmpty(GetSingleNodePath())`?

MikhailArkhipov: Yep

src/Package/Impl/ProjectSystem/Commands/CopyItemPathCommand.cs
```diff
@@ -0,0 +1,42 @@
...
+        public Task<CommandStatusResult> GetCommandStatusAsync(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, string commandText, CommandStatus progressiveStatus) {
+            if (commandId == RPackageCommandId.icmdCopyItemPath && nodes.IsSingleNodePath()) {
+                return Task.FromResult(new CommandStatusResult(true, commandText, CommandStatus.Enabled | CommandStatus.Supported));
+            }
+            return Task.FromResult(CommandStatusResult.Unhandled);
+        }
+
+        public async Task<bool> TryHandleCommandAsync(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut) {
+            if (commandId == RPackageCommandId.icmdCopyItemPath) {
+                var path = nodes.GetSingleNodePath();
```
int19h: Needs a check for when the node doesn't actually have a path, or it's not a single node.

src/Package/Impl/ProjectSystem/Commands/CopyItemPathCommand.cs
```diff
@@ -0,0 +1,42 @@
...
+                try {
+                    var directory = await SessionUtilities.GetFriendlyDirectoryNameAsync(path);
+                    await VsAppShell.Current.DispatchOnMainThreadAsync(() =>
+                        Clipboard.SetData(DataFormats.UnicodeText, Invariant($"\"{directory}\"")));
+                } catch (Exception) { }
```
int19h: What is it catching here? Can you make it a bit more specific, or if it really has to be so generic, use `IsCriticalException` and log?

src/Package/Impl/ProjectSystem/Commands/SourceFilesCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+                return new CommandStatusResult(true, commandText, CommandStatus.Enabled | CommandStatus.Supported);
+            }
+            return CommandStatusResult.Unhandled;
+        }
+
+        public bool TryHandleCommand(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut) {
+            if (commandId == RPackageCommandId.icmdSourceSelectedFiles) {
+                foreach (var filePath in nodes.GetSelectedNodesPaths()
+                                              .Where(x => Path.GetExtension(x)
+                                                  .Equals("r", StringComparison.OrdinalIgnoreCase))) {
```
int19h: Formatting.

int19h: (Probably better to use an extra variable here to assign the result of the query outside of `foreach`, for readability.

src/Package/Impl/ProjectSystem/Commands/SourceFilesCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.VisualStudio.R.Package.Commands;
+using Microsoft.VisualStudio.R.Package.Repl.Commands;
+
+namespace Microsoft.VisualStudio.R.Package.ProjectSystem.Commands {
+    [ExportCommandGroup("AD87578C-B324-44DC-A12A-B01A6ED5C6E3")]
+    [AppliesTo("RTools")]
+    [OrderPrecedence(200)]
+    internal sealed class SourceFilesCommand : ICommandGroupHandler {
+        public CommandStatusResult GetCommandStatus(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, string commandText, CommandStatus progressiveStatus) {
+            if (commandId == RPackageCommandId.icmdSourceSelectedFiles && !nodes.IsFolder() && nodes.GetSelectedNodesPaths().Count() > 0) {
```
int19h: As written, `IsFolder` will always return `false` for a multiple selection of nodes, even if some of those nodes are folders. So this code will try to source a folder in that scenario.

src/Package/Impl/Repl/Commands/WorkingDirectoryCommand.cs
```diff
@@ -163,24 +164,11 @@ public sealed class WorkingDirectoryCommand : Command, IDisposable {
...
         private async Task UpdateRUserDirectoryAsync() {
             await TaskUtilities.SwitchToBackgroundThread();
             try {
-                using (var evaluation = await _session.BeginEvaluationAsync(false)) {
-                    UserDirectory = await evaluation.GetRUserDirectory();
-                }
-            } catch (TaskCanceledException) {
-            }
+                UserDirectory = await SessionUtilities.GetRUserDirectoryAsync(_interactiveWorkflow);
+            } catch (TaskCanceledException) { }
```
int19h: While you're at it, can you also add `MessageTransportException` here, and change `TaskCanceledException` to `OperationCanceledException` (the latter is a supertype, and we sometimes throw it directly).

int19h: Actually, now that I look at it, `GetRUserDirectoryAsync` already swallows exceptions (should it?).

src/Package/Impl/Utilities/SessionUtilities.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public static async Task<string> GetRWorkingDirectoryAsync(IRInteractiveWorkflow workflow = null) {
+            var session = workflow != null ? workflow.RSession : GetInteractiveSession();
+            if (session.IsHostRunning) {
+                await TaskUtilities.SwitchToBackgroundThread();
+                try {
+                    using (var evaluation = await session.BeginEvaluationAsync(false)) {
+                        return await evaluation.GetWorkingDirectory();
+                    }
+                } catch (TaskCanceledException) { }
```
int19h: `OperationCanceledException`, and should also catch `MessageTransportException`. Same thing in other methods that do `BeginEvaluation` or `BeginInteraction`.

src/Package/Impl/Utilities/SessionUtilities.cs
```diff
@@ -0,0 +1,61 @@
...
+                try {
+                    using (var evaluation = await session.BeginEvaluationAsync(false)) {
+                        return await evaluation.GetRUserDirectory();
+                    }
+                } catch (TaskCanceledException) { }
+            }
+            return null;
+        }
+
+        public static async Task<string> GetFriendlyDirectoryNameAsync(string directory) {
```
int19h: This should be named `GetRShortenedPathNameAsync` or something like that, because it's also useful for files (the fix for #829 will entail calling this function when doing `source`).

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+            var debugger = VsAppShell.Current.GetGlobalService<IVsDebugger>(typeof(SVsShellDebugger));
+            if (debugger != null) {
+                var mode = new DBGMODE[1];
+                if (debugger.GetMode(mode) >= 0) {
+                    bool debugging = mode[0] != DBGMODE.DBGMODE_Design;
+                    var interactiveWorkflowProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>();
+                    var interactiveWorkflow = interactiveWorkflowProvider.GetOrCreate();
+                    var operations = interactiveWorkflow.Operations;
+
+                    operations.ExecuteExpression($"{(debugging ? "rtvs::debug_source" : "source")}({filePath.ToRStringLiteral()})");
```
int19h: If you add a call to `GetRUserDirectoryAsync` here on `filePath`, you will fix #829 as well.

src/Package/Impl/ProjectSystem/Commands/SourceFilesCommand.cs
MikhailArkhipov: I guess we can filter this out in the handler. Probably not worth enumerating and checking if user actually selected 3 folders. Most common case is when they click on a folder node and that's where we want to hide the command

src/Package/Impl/Repl/Commands/WorkingDirectoryCommand.cs
MikhailArkhipov: Well what to do if they do happen?

src/Package/Impl/ProjectSystem/Commands/SourceFilesCommand.cs
smortaz: you also need "Open Explorer" for any item in the directory.  right not
it's only there for Folders.

AlexanderSher:
Please don't merge this change, I need to review it.
________________________________

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
MikhailArkhipov: Sure why not

MikhailArkhipov: You mean  GetRShortenedPathNameAsync

src/Package/Impl/ProjectSystem/Commands/SourceFilesCommand.cs
MikhailArkhipov: There is standard VS command for that

image

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
int19h: Yeah, whatever I asked you to rename it to :)

src/Package/Impl/Repl/Commands/WorkingDirectoryCommand.cs
int19h: Something will have to swallow them in this code path, the only question is where to do it. I don't really have a preference either way.

MikhailArkhipov:
Will resubmit

PR #1293, opened 2016-03-18
Improve project menu

MikhailArkhipov:
Includes fix for #829


MikhailArkhipov:
image

smortaz:
i didnt quite follow the thread - do individual files in a project get:

- Open in File Explorer
- Copy path

?

they really should (or get equivalent)

smortaz:
nice!

MikhailArkhipov:
Multiple selection

image

MikhailArkhipov:
I can hide Add... for multiple selection but not sure

src/Package/Impl/Utilities/ProjectUtilities.cs
```diff
@@ -42,5 +52,95 @@ public static class ProjectUtilities {
 
             return result;
         }
+
+        public static EnvDTE.Project GetActiveProject() {
```
AlexanderSher: Please pass `VsAppShell.Current` as a parameter

AlexanderSher: Actually, maybe `DTE` is better

src/Package/Impl/Utilities/ProjectUtilities.cs
```diff
@@ -42,5 +52,95 @@ public static class ProjectUtilities {
...
+                        }
+                    }
+
+                    // Given path to the project or a folder in it, generate unique file name
+                    var fileName = GetUniqueFileName(destinationPath, name, extension);
+
+                    // Locate folder in the project
+                    var projectFolder = Path.GetDirectoryName(project.FullName);
+                    if (destinationPath.StartsWith(projectFolder, StringComparison.OrdinalIgnoreCase)) {
+                        ProjectItems projectItems = project.ProjectItems;
```
AlexanderSher: Don't change ProjectItems explicitly - file system should handle the change automatically

src/Package/Impl/Utilities/ProjectUtilities.cs
```diff
@@ -42,5 +52,95 @@ public static class ProjectUtilities {
...
+        public static EnvDTE.Project GetActiveProject() {
+            DTE dte = VsAppShell.Current.GetGlobalService<DTE>();
+            var projects = (Array)dte.ActiveSolutionProjects;
+            if (projects != null && projects.Length == 1) {
+                return (EnvDTE.Project)projects.GetValue(0);
+            }
+            return null;
+        }
+
+        public static void AddNewItem(string templateName, string name, string extension, string destinationPath) {
```
int19h: Do we really have to do the whole unzip dance ourselves? I hoped VS would have provided a programmatic way to do the same thing that "Add New Item" does when user clicks OK...

int19h: Actually, looking at what C# and C++ do, it doesn't just add the item right away when you click "Add Class". Instead, it opens the usual "Add New Item" dialog (which lets the user provide the name for the class), but pre-selects the class template. I assume other "Add ..." commands work in a similar fashion.

Both C# and C++ use `IVsAddProjectItemDlg.AddProjectItemDlg` for this purpose - there is a parameter there to pre-select a template.

src/Package/Impl/Utilities/SessionUtilities.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.Common.Core;
+using Microsoft.R.Components.InteractiveWorkflow;
+using Microsoft.R.Host.Client;
+using Microsoft.R.Host.Client.Session;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
+using Microsoft.VisualStudio.R.Package.Shell;
+
+namespace Microsoft.VisualStudio.R.Package.Utilities {
+    public static class SessionUtilities {
+        public static IRSession GetInteractiveSession() {
```
AlexanderSher: We should not create extensions like this one, cause it couples `Microsoft.R.Components` object to the `VsAppShell.Current` instance. That is bad because:
- Any instance of the object from `Microsoft.R.Components` can live longer than `VsAppShell` and created before it. Calling to `VsAppShell.Current` before it is initialized in Package breaks pretty much everything. 
- `VsAppShell.Current` implicitly adds dependency to the IVs* to the code that doesn't really need it (we have dozens of commands in `Package` that can be moved to the `Microsoft.R.Components`).
- Any code that uses `VsAppShell.Current` requires isolated non-parallel run, lots of fakes and VsRHostScript even if the test has no UI involved.

src/Package/Impl/Utilities/SessionUtilities.cs
```diff
@@ -0,0 +1,78 @@
...
+            var workflowProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>();
+            var interactiveWorkflow = workflowProvider.GetOrCreate();
+            return interactiveWorkflow.RSession;
+        }
+
+        public static async Task<string> GetRWorkingDirectoryAsync(IRInteractiveWorkflow workflow = null) {
```
AlexanderSher: This code can be moved to the Microsoft.R.Host.Client:
```{csharp}
public static async Task<string> GetRWorkingDirectoryAsync(this IRSession session) {
    if (session.IsHostRunning) {
        try {
            using (var evaluation = await session.BeginEvaluationAsync(false)) {
               return await evaluation.GetWorkingDirectory();
            }
    }
    return null;
}
```

Also, `IRInteractiveWorkflow` is usually assumed to be not null.

src/Package/Impl/Utilities/SessionUtilities.cs
```diff
@@ -0,0 +1,78 @@
...
+                try {
+                    using (var evaluation = await session.BeginEvaluationAsync(false)) {
+                        return await evaluation.GetRUserDirectory();
+                    }
+                } catch (OperationCanceledException) { }
+            }
+            return null;
+        }
+
+        public static string GetRShortenedPathName(string name, string userDirectory) {
```
AlexanderSher: This belongs to `Microsoft.R.Host.Client.RStringExtensions` I think.

src/Package/Impl/Utilities/SessionUtilities.cs
```diff
@@ -0,0 +1,78 @@
...
+                try {
+                    using (var evaluation = await session.BeginEvaluationAsync(false)) {
+                        return await evaluation.GetWorkingDirectory();
+                    }
+                } catch (OperationCanceledException) { }
+            }
+            return null;
+        }
+
+        public static async Task<string> GetRUserDirectoryAsync(IRInteractiveWorkflow workflow = null) {
```
AlexanderSher: Same with this one

src/Package/Impl/VSAppShell/Current.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace VSAppShell {
+    internal class Current {
```
AlexanderSher: What is this?

src/Package/Impl/ProjectSystem/Commands/CommandUtilities.cs
```diff
@@ -0,0 +1,53 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.ProjectSystem.Designers;
+
+namespace Microsoft.VisualStudio.R.Package.ProjectSystem.Commands {
+    internal static class CommandUtilities {
```
AlexanderSher: I don't like this name. None of these methods is about commands.

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: DTE only support unpacked templates. I could go with the dialog, but then it is as many clicks as regular Add Item. With file system it is simpler to add file and then rename if needed.

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: DTE is hard to mock though.

src/Package/Impl/ProjectSystem/Commands/SetDirectoryHereCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+            return CommandStatusResult.Unhandled;
+        }
+
+        public bool TryHandleCommand(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut) {
+            if (commandId == RPackageCommandId.icmdSetDirectoryHere) {
+                var path = nodes.GetNodeFolderPath();
+                if (!string.IsNullOrEmpty(path)) {
+                    var o = new object();
+
+                    var interactiveWorkflowProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>();
```
AlexanderSher: `SetDirectoryHereCommand` is exported through MEF, so `IRInteractiveWorkflowProvider` this can be obtained in ctor

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: this is not changing, just drilling into folder. DTE.ProjectItems is root, then pick folder from them, get ProjectItems (aka children) from it and so on until the destination is reached.

src/Package/Impl/ProjectSystem/Commands/SetDirectoryHereCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+                    var interactiveWorkflow = interactiveWorkflowProvider.GetOrCreate();
+                    var controller = ReplCommandController.FromTextView(interactiveWorkflow.ActiveWindow.TextView);
+
+                    controller.Invoke(RGuidList.RCmdSetGuid, RPackageCommandId.icmdSetWorkingDirectory, path, ref o);
```
AlexanderSher: Not a part of this change, but it would be great to wire up these `ICommandGroupHandler` and `IAsyncCommandGroupHandler` interfaces to our controller-based invocation in general.

src/Package/Impl/ProjectSystem/Commands/SourceFilesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+                }
+                return new CommandStatusResult(true, commandText, CommandStatus.Enabled | CommandStatus.Supported);
+            }
+            return CommandStatusResult.Unhandled;
+        }
+
+        public bool TryHandleCommand(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut) {
+            if (commandId == RPackageCommandId.icmdSourceSelectedFiles) {
+                var rFiles = nodes.GetSelectedNodesPaths().Where(x =>
+                               Path.GetExtension(x).Equals(".r", StringComparison.OrdinalIgnoreCase) &&
```
AlexanderSher: we have `Microsoft.Common.Core.StringExtensions.EqualsIgnoreCase`.

src/Package/Impl/Utilities/ProjectUtilities.cs
int19h: But then we're not consistent with the rest of VS.

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+            Task.Run(async () => {
+                file = await SessionUtilities.GetRShortenedPathNameAsync(file);
+                VsAppShell.Current.DispatchOnUIThread(() => {
+                    operations.ExecuteExpression($"{(debugging ? "rtvs::debug_source" : "source")}({file.ToRStringLiteral()})");
+                });
+            });
+        }
+
+        private static bool IsDebugging() {
+            var debugger = VsAppShell.Current.GetGlobalService<IVsDebugger>(typeof(SVsShellDebugger));
```
AlexanderSher: We have IDebuggerModeTracker. We can add another property to it.

src/Package/Impl/Utilities/ProjectUtilities.cs
int19h: PTVS has a bunch of DTE mocks that might be of use.

https://github.com/Microsoft/PTVS/tree/0a3d9d17008e8d9b8fe2cb40c6c501f13861ccdb/Common/Tests/MockVsTests

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.R.Components.InteractiveWorkflow;
+using Microsoft.R.Host.Client;
+using Microsoft.VisualStudio.R.Package.Shell;
+using Microsoft.VisualStudio.R.Package.Utilities;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.R.Package.Repl.Commands {
+    internal static class SourceFileHelper {
+        public static void SourceFiles(IEnumerable<string> files) {
+            VsAppShell.Current.AssertIsOnMainThread();
```
AlexanderSher: Why do you need a main thread here?

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.Common.Core;
+using Microsoft.R.Components.InteractiveWorkflow;
+using Microsoft.R.Host.Client;
+using Microsoft.VisualStudio.R.Package.Shell;
+using Microsoft.VisualStudio.R.Package.Utilities;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.R.Package.Repl.Commands {
+    internal static class SourceFileHelper {
+        public static void SourceFiles(IEnumerable<string> files) {
```
AlexanderSher: I think we can add this method to the `IRInteractiveWorkflowOperations` interface

AlexanderSher: And make it async

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
AlexanderSher: Btw, @int19h, can't we get this information from `IRSession` or `DebugSession` directly?

src/Package/Impl/Utilities/SessionUtilities.cs
MikhailArkhipov: I'll make it SessionExtensions

src/Package/Impl/VSAppShell/Current.cs
MikhailArkhipov: Whoops. misclick. Instead of 'add using' apparently clicked 'generate class'

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+            VsAppShell.Current.AssertIsOnMainThread();
+
+            bool debugging = IsDebugging();
+            var operations = GetOperations();
+            var list = new List<string>();
+
+            Task.Run(async () => {
+                foreach (var f in files) {
+                    list.Add(await SessionUtilities.GetRShortenedPathNameAsync(f));
```
AlexanderSher: This is actually  a good example why we have separated `IRSession.BeginEvaluationAsync` into separate method. This code schedules `files.Count()` tasks into `RSession` evaluation queue, without any guarantee that they will be executed one after another.
Instead, this code should call `IRSession.BeginEvaluationAsync` once for the whole loop.

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
int19h: No, because `DebugSession` is always there (its lifetime doesn't correspond to VS debugging or not debugging, since it's also used by VE etc).

int19h: Although I guess the real question that this code should ask is, "are breakpoints enabled", in which case `DebugSession` can track that (it doesn't currently - it just passes it on to R code which does the actual tracking - but it would be easy to add).

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
AlexanderSher: Not to mention that `GetRShortenedPathNameAsync` schedules 2 tasks itself - in `GetRUserDirectoryAsync` and in `GetRShortenedPathName`

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: No all things in VS have to be the same. Besides, we only have two items while C# has like 10.

src/Host/Client/Impl/Extensions/SessionExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Common.Core;
+using Microsoft.R.Host.Client.Session;
+
+namespace Microsoft.R.Host.Client {
+    public static class SessionExtensions {
```
AlexanderSher: We already have class called `RSessionExtensions`
It is in the `Microsoft.R.Host.Client.Session`. We probably should merge them.

src/Package/Impl/ProjectSystem/Commands/CommandUtilities.cs
MikhailArkhipov: ProjectTreeExtensions and move to the project system assembly?

AlexanderSher: Yeah, great idea!

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
int19h: Is it a correctness issue here, or a perf issue?

Anyway, it sounds like you're suggesting that any helper method that performs evals should have an overload that accepts `IRSessionEvaluation`, to reuse an existing one?

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: Since it uses GetGlobalService, it is easily replaceable in tests by mock. Doesn't have to be passed, see
```
        private TestVsAppShell() {
            CompositionService = VsTestCompositionCatalog.Current.CompositionService;
            ExportProvider = VsTestCompositionCatalog.Current.ExportProvider;
            MainThread = UIThreadHelper.Instance.Thread;
            _sp = new TestServiceProvider();
        }
```

so basically just add DTE mock to the TestServiceProvider. Currently it has

```
        public TestServiceProvider() {
            _services.Add(typeof(SVsRegisterProjectTypes), new VsRegisterProjectGeneratorsMock());
            _services.Add(typeof(SVsRegisterEditors), VsRegisterEditorsMock.Create());
            _services.Add(typeof(IMenuCommandService), new MenuCommandServiceMock());
            _services.Add(typeof(SComponentModel), new ComponentModelMock(VsTestCompositionCatalog.Current));
            _services.Add(typeof(SVsTextManager), new TextManagerMock());
            _services.Add(typeof(SVsImageService), VsImageServiceMock.Create());
            _services.Add(typeof(SVsUIShell), new VsUiShellMock());
            _services.Add(typeof(SOleComponentManager), OleComponentManagerMock.Create());
            _services.Add(typeof(SVsSettingsManager), VsSettingsManagerMock.Create());
        }
```

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
MikhailArkhipov: They don't have to be in order

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
MikhailArkhipov: Enqueue/Execute Expression use nteractiveWindow and nteractiveWindow.TextView that are not thread safe and are UI bound.

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
AlexanderSher: User can change his current directory while this loop is running, so part of the paths will be pointing to the wrong place. And we actually don't need to call `GetRUserDirectoryAsync` on every step of the loop.

As for the helper methods, I think that for this type of extensions we should have them for `IRSessionEvaluation` rather than for `IRSession`.

src/Package/Impl/Utilities/ProjectUtilities.cs
int19h: If it's labeled the same, there should be a very good reason for it to not behave the same. In this case, I simply don't see such a reason, especially given the amount of custom code that this unusual behavior requires compared to the regular behavior. And click-wise, it is not only not better, it's actually worse. For example, consider C# again. When I do Add Class, the sequence is as follows:

1. Right-click folder/project -> "Add Class"
2. Type the name of the class (focus will already be in the right field in Add New Item dialog).
3. Press Enter.

With the flow implemented here:

1. Right-click project/folder -> "Add Script"
2. Right-click on the newly added file again -> "Rename" (or click-click it slowly).
3. Type the name of the file.
4. Press Enter.

So basically this workflow _adds_ an extra two clicks in all scenarios, unless I'm okay with "rscript1.R" etc. Which is pretty much never the case outside of hello world type apps.

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
AlexanderSher: But you access Enqueue/ExecuteExpression inside `DispatchOnUIThread`

MikhailArkhipov: IsDebugging calls ```GetGlobalService<IVsDebugger>```

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: PM call. I personally don't like C# behavior, I'd rather add and rename.

AlexanderSher:
Sorry guys, I'm still not feeling well, so I'm done for today.

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: Besides, R more 'scratch' type language (i.e. use file to hold some code for some time) that others so naming files in many cases may not be that important. RS does not show anything, just adds the file.

int19h: @jflam, your call.

src/Package/Impl/Repl/Commands/SourceFileHelper.cs
MikhailArkhipov: Theoretically yes. Practically, it is probably 3-4 files and it is unlikely user will be clicking around while files are getting sourced.

MikhailArkhipov: I'll make batch method extension

MikhailArkhipov:
Addressed most of the feedback, added Open Command Prompt. If we are settled on general interfaces and code locations, I will be working on some tests for the new commands.

MikhailArkhipov:
Ping... We can change Add New Item story based on user feedback.

src/Host/Client/Impl/Session/RSessionExtensions.cs
```diff
@@ -23,5 +24,57 @@ public static class RSessionExtensions {
...
+        }
+
+        public static async Task<string> GetRShortenedPathNameAsync(this IRSession session, string name) {
+            var userDirectory = await session.GetRUserDirectoryAsync();
+            return GetRShortenedPathName(name, userDirectory);
+        }
+
+        public static async Task<IEnumerable<string>> GetRShortenedPathNamesAsync(this IRSession session, IEnumerable<string> names) {
+            var userDirectory = await session.GetRUserDirectoryAsync();
+            var list = new List<string>();
```
AlexanderSher:
```{csharp}
return names.Select(n => GetRShortenedPathName(n, userDirectory));
```

src/Host/Client/Impl/Session/RSessionExtensions.cs
```diff
@@ -23,5 +24,57 @@ public static class RSessionExtensions {
...
+            }
+            return null;
+        }
+
+        public static async Task<string> GetRShortenedPathNameAsync(this IRSession session, string name) {
+            var userDirectory = await session.GetRUserDirectoryAsync();
+            return GetRShortenedPathName(name, userDirectory);
+        }
+
+        public static async Task<IEnumerable<string>> GetRShortenedPathNamesAsync(this IRSession session, IEnumerable<string> names) {
```
AlexanderSher: I don't like the name of the method. Maybe something like `MakeRelativeToUserDirectory`?

MikhailArkhipov: OK

AlexanderSher:
:+1:

src/Package/Impl/Utilities/ProjectUtilities.cs
MikhailArkhipov: John said it is OK to be different.

Issue #1299, opened 2016-03-19
Parser doesn't properly handle loops as values


int19h:
Consider:

``` r
x <- for (i in 1:10) {}
```

This is valid R code - `for` is an expression that always produces `NULL`. However, the editor will squiggle it.

Similar problem exists with `while`, `repeat`, `break` and `next`. All of these are expressions and should be treated as such (basically, everything is an expression).

int19h:
Note that the real scenarios here are more convoluted than simple assignment, and don't actually utilize the return value; but all boil down to loops being expression.  For example, suppose that the body of the loop makes assignments, and we want to localize them. Then:
```r
local(for (i in 1:10) { ... })
```
we currently don't parse it right. Similarly with `eval` and `quote`.

MikhailArkhipov:
Folding into #3642

Issue #1300, opened 2016-03-19
Arbitrary expressions should be legal on the left side of the assignment operator


int19h:
E.g. this is legal R:
```r
x <- y <- 1;
x + y <- 2;
```

Language spec says:
> Notice that the assignment symbols are operators just like the arithmetic, relational, and logical ones. Any expression is allowed also on the target side of an assignment, as far as the parser is concerned (2 + 2 <- 5 is a valid expression as far as the parser is concerned. The evaluator will object, though). Similar comments apply to the model formula operator.

At runtime, the implementation of `<-` will look up a helper function corresponding to the left side. For example, in this case, it will try to look up function named `+<-`. If such a function is defined, it will be used. For example:

```r
`+<-` <- function(...) { print('assigned') }
x <- y <- 1;
print(x + y <- 2)
```
prints:
```
[1] "assigned"
[1] 2
```

MikhailArkhipov:
Also

```1 == 2 == TRUE```

Issue #1298, opened 2016-03-19
Incorrect operator priorities


int19h:
Some operators have priorities assigned to them that do not correspond to the actual evaluation order. In particular:

Function call has higher priority than `@` and `$`. This is important in expressions such as `Class$new(...)`, where `Class$new` is evaluated first, and is the target of the call.

`:` has higher priority than `^` and unary `+`, `-`, `!`. So e.g. `-a:b` is parsed as `-(a:b)`, not as `(-a):b` as it is actually evaluated.

`::` and `:::` have lower priority than almost everything. So `a::b+1` is parsed as `a::(b+1)`, but it's actually evaluated as `(a::b)+1`.

There's some other weirdness in `OperatorPrecedence.cs`. For example, a bunch of `%..%` operators are special-cased to have lower precedence than some other operators, even though per language grammar they all have the same precedence.

R docs have a [precedence table](https://stat.ethz.ch/R-manual/R-devel/library/base/html/Syntax.html), but it's missing some things - notably, function calls.

MikhailArkhipov:
-1:2 produces correct result: operator ':' over -1 and 2. Tree looks like
```
    ExpressionStatement  [-1:2]
        Expression  [-1:2]
            TokenOperator  [: [2...3)]
                NumericalValue  [-1 [0...2)]
                TokenNode  [: [2...3)]
                NumericalValue  [2 [3...4)]
```
i.e. 
```
       :
-1          2
```

int19h:
This is because you're using literals here. My repro was `-a:b`.

By the way, the above tree is also somewhat incorrect, in that the sign in `-1` is actually an operator, and evaluation of `-1` is performed as `1` followed by applying function `-` to it, so redefining `-` will affect the result:
```
> `-` <- function(x) x*2
> -2
[1] 4
```

PR #1303, opened 2016-03-21
 #1298 (operator precedence)

MikhailArkhipov:
Also #1300 (mostly, I think).

1. Fixed operator precedence table numbers
1. Reworked unary operators since before code could not use different priority for unary plus/minus vs. binary add/subtract
1. Fixed associativity so function calls and indexers are grouped properly
1. Added bunch of new tests


src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
```diff
@@ -449,55 +444,32 @@ enum OperationType {
...
-                !(currentOperator.OperatorType == lastOperator.OperatorType && currentOperator.Association == Association.Right)) {
+            if (currentOperator.Precedence < lastOperator.Precedence ||
+                (currentOperator.Precedence == lastOperator.Precedence &&
+                 currentOperator.Associativity == Associativity.Left)) {
                 // New operator has lower or equal precedence. We need to make a tree from
                 // the topmost operator and its operand(s). Example: a*b+c. + has lower priority
                 // and a and b should be on the stack along with * on the operator stack.
                 // Repeat until there are no more higher precendece operators on the stack.
 
+                // If operator cas right associativity (exponent, indexer, function call),
```
int19h: Are indexer and function call actually right-associative? Given that `a[1][2]` is `(a[1])[2]`, and `a(1)(2)` is `(a(1))(2)`.

src/R/Core/Impl/AST/Operators/Associativity.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.R.Core.AST.Operators {
+    // https://en.wikipedia.org/wiki/Operator_associativity
+    /// <summary>
+    /// Associativity. Consider the expression a~b~c. If the operator 
+    /// ~ has left associativity, this expression would be interpreted as (a~b)~c.
+    /// If the operator has right associativity, the expression would be
+    /// interpreted as a~(b~c).
+    /// </summary>
+    public enum Associativity {
```
int19h: It appears that R also has non-associative operators. Try:
```r
1 == 2 == TRUE
```
and it fails, with a parser error at that (can be verified by using `quote` to make sure that no evaluation takes place).

src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
MikhailArkhipov: Wrong comment. Below says

```Left-associated indexer works  
  472   +                // like x[1][2] = (x[1])[2]  
```
Neither Indexer nor FunctionCall override default (left) association. Indexer03, Indexer15, Indexer16 et al parse properly

src/R/Core/Impl/AST/Operators/OperatorPrecedence.cs
```diff
@@ -57,31 +64,35 @@ public static class OperatorPrecedence {
...
+                case OperatorType.Index: // [], [[]]
                 case OperatorType.ListIndex: // $ or @
-                    return 150;
+                    return 170;
 
                 case OperatorType.Namespace: // :: or :::
-                    return 160;
+                    return 180;
 
                 case OperatorType.Group: // ( ) around expression
```
int19h: Does this also include `{}`?

src/R/Core/Impl/AST/Operators/Associativity.cs
MikhailArkhipov: It might, not all edge cases are currently being processed (#1300 is only partial). The reason is (we need to discuss) that design time parsing is often different from run-time parsing. Say, ```x + y <- 1``` may be legal but does user actually want this construct or are they still editing file and forgot to add line break? Ex in HTML browsers are very forgiving but in the editor users prefer strict checks, even when browser allows particular construct.

MikhailArkhipov: Similarly, I am not entirely convinced we have to fix #1299

int19h: This is the reverse case - we treat something as valid R code syntactically, when it doesn't actually get parsed by R itself, so our design-time parsing is more strict than run-time parsing.

Anyway, it's just a random thing that came up while I was playing with operators. I suspect it's not trivial to fix, so I can file a separate issue.

MikhailArkhipov:
```{}``` are parsed as scope (or as lambda), they are not handled as operator

src/R/Core/Impl/AST/Operators/Associativity.cs
MikhailArkhipov: Yes, I think we should check syntax deeper (there is a separate task for async syntax check with better analysis and separate error/warning messages). I.e. we hopefully should be detecting what will fail at runtime but not necessarily allow everything that runtime allows.

int19h:
:+1:

src/R/Core/Impl/AST/Operators/Associativity.cs
int19h: https://github.com/Microsoft/RTVS/issues/1305

PR #1304, opened 2016-03-21
Fix #1081: Debugger data tip

### LINKED ISSUE #1081: Debugger data tip
Implement IVsTextViewFilter.GetDataTipText

int19h:


MikhailArkhipov:
:+1:

src/R/Core/Test/DataTips/DataTipTest.cs
```diff
@@ -0,0 +1,75 @@
...
+        [MemberData("Everywhere", "'a'")]
+        [MemberData("Everywhere", "\"a\"")]
+        [MemberData("Everywhere", "`a`")]
+        [MemberData("Everywhere", "a")]
+        [MemberData("Everywhere", "a$b")]
+        [MemberData("Everywhere", "a@b")]
+        [MemberData("Everywhere", "a::b")]
+        [MemberData("Everywhere", "a:::b")]
+        [MemberData("EverywhereExcept", "a[b]", 2)]
+        [MemberData("EverywhereExcept", "a[[b]]", 3)]
```
AlexanderSher: you should be able to use nameof here

src/R/Core/Impl/DataTips/RDataTip.cs
```diff
@@ -0,0 +1,114 @@
...
+                if (op == OperatorType.Index) {
+                    if (parent.LeftOperand != node) {
+                        break;
+                    }
+                } else if (op != OperatorType.ListIndex && op != OperatorType.Namespace) {
+                    break;
+                }
+
+                node = parent;
+            }
```
AlexanderSher: Let's extract it into private method

int19h: You mean the loop above?

src/R/Core/Impl/AST/AstNode.cs
```diff
@@ -124,10 +124,10 @@ public abstract class AstNode : IAstNode {
         /// <summary>
         /// Finds deepest node that fully encloses given range
         /// </summary>
-        public virtual IAstNode NodeFromRange(ITextRange range) {
+        public virtual IAstNode NodeFromRange(ITextRange range, bool endInclusive = false) {
             IAstNode node = null;
 
-            if (TextRange.Contains(this, range)) {
+            if (endInclusive ? TextRange.ContainsInclusiveEnd(this, range) : TextRange.Contains(this, range)) {
```
AlexanderSher: You can use `Contains(ITextRange range, ITextRange other, bool inclusiveEnd)`

src/R/Core/Test/DataTips/DataTipTest.cs
int19h: Will do.

src/Package/Impl/Debugger/DataTips/DataTipTextViewFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+        private readonly IOleCommandTarget _nextTarget;
+
+        public DataTipTextViewFilter(IWpfTextView textView, IVsDebugger debugger) {
+            if (!textView.TextBuffer.ContentType.IsOfType(RContentTypeDefinition.ContentType)) {
+                return;
+            }
+
+            _textView = textView;
+            _debugger = debugger;
+            _adapterService = ComponentLocator<IVsEditorAdaptersFactoryService>.Import();
```
AlexanderSher: I'd rather use `VsAppShell.Current.ExportProvider.GetExportedValue<IVsEditorAdaptersFactoryService>()`.
`ComponentLocator` calls `SatisfyImportsOnce` which has additional side effects

src/R/Core/Impl/AST/AstNode.cs
int19h: Sure. Weird that I didn't spot it. I'll rename the parameter to `inclusiveEnd` as well, so that it matches.

src/R/Core/Impl/DataTips/RDataTip.cs
AlexanderSher: Yes

---
In reply to: [56885218](https://github.com/Microsoft/RTVS/pull/1304#discussion_r56885218) [](ancestors = 56885218)

int19h: Perhaps it's better to refactor the larger chunk that validates the initial node?

src/Package/Impl/Debugger/DataTips/DataTipTextViewFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            var vsTextView = _adapterService.GetViewAdapter(textView);
```
AlexanderSher: You use this field only in ctor. Variable should be enough

src/Package/Impl/Debugger/DataTips/DataTipTextViewConnectionListener.cs
```diff
@@ -0,0 +1,30 @@
...
+    internal sealed class DataTipTextViewConnectionListener : IWpfTextViewConnectionListener {
+        public DataTipTextViewConnectionListener() {
+        }
+
+        public void SubjectBuffersConnected(IWpfTextView textView, ConnectionReason reason, Collection<ITextBuffer> subjectBuffers) {
+            var debugger = VsAppShell.Current.GetGlobalService<IVsDebugger>();
+            new DataTipTextViewFilter(textView, debugger);
+        }
+
+        public void SubjectBuffersDisconnected(IWpfTextView textView, ConnectionReason reason, Collection<ITextBuffer> subjectBuffers) {
```
AlexanderSher: Don't you need to dispose DataTipTextViewFilter somehow?

src/R/Core/Impl/DataTips/RDataTip.cs
AlexanderSher: Agreed

---
In reply to: [56886434](https://github.com/Microsoft/RTVS/pull/1304#discussion_r56886434) [](ancestors = 56886434)

src/Package/Impl/Debugger/DataTips/DataTipTextViewConnectionListener.cs
MikhailArkhipov: It may be necessary to remove it explicitly with IVsTextView.RemoveCommandFilter depending on the particular view lifetime

int19h: The corresponding code in PTVS never removes it (and I guess the check for content type that's done first thing in the filter before anything else makes it work). 

So, how to do this correctly? Check for content type of the buffer, and if it's not R, remove the filter? Also, where would be the appropriate place to store the reference to the filter so it can be passed to `RemoveCommandFilter`? Does `textView` has some kind of property bag, or let you enumerate all filters so that I could just use `is` to check for mine?

src/Package/Impl/Debugger/DataTips/DataTipTextViewFilter.cs
int19h: I'll move it to the caller. It's inconsistent now in that the caller retrieves debugger service, but this thing retrieves the adapter service.

src/Package/Impl/Debugger/DataTips/DataTipTextViewFilter.cs
int19h: It is used in `Dispose` now, so I'll keep it.

int19h:
Addressed feedback.

AlexanderSher:
:+1:

PR #1314, opened 2016-03-22
Unittest visual tree tests only when control is not collapsed

JeongPyoHam:
- Visual tree dump saves after filter
- Compare properties only when not collapsed

### DIFF
--- a/src/Package/TestApp/Utility/ViewTreeDump.cs
+++ b/src/Package/TestApp/Utility/ViewTreeDump.cs
@@ -17,8 +17,7 @@ namespace Microsoft.VisualStudio.R.Interactive.Test.Utility {
         private static bool _regenerateBaselineFiles = false;
 
         public static void CompareVisualTrees(DeployFilesFixture fixture, VisualTreeObject actual, string fileName) {
-            Action a = () => CompareVisualTreesImplementation(fixture, actual, fileName);
-            a.ShouldNotThrow();
+            CompareVisualTreesImplementation(fixture, actual, fileName);
         }
 
         private static void CompareVisualTreesImplementation(DeployFilesFixture fixture, VisualTreeObject actual, string fileName) {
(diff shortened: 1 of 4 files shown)


src/Package/TestApp/Utility/ViewTreeDump.cs
```diff
@@ -17,8 +17,7 @@ internal class ViewTreeDump {
         private static bool _regenerateBaselineFiles = false;
 
         public static void CompareVisualTrees(DeployFilesFixture fixture, VisualTreeObject actual, string fileName) {
```
int19h: This method has no purpose now, you might as well just rename the private one.

src/Package/TestApp/Utility/ViewTreeDump.cs
```diff
@@ -36,27 +35,36 @@ internal class ViewTreeDump {
         }
 
         private static void CompareVisualTree(VisualTreeObject actual, VisualTreeObject expected, bool compareProperty = true) {
+            bool visible = true;
+            
             // compare
-            actual.Name.ShouldBeEquivalentTo(expected.Name);
+            actual.Name.Should().BeEquivalentTo(expected.Name);
```
int19h: Please don't use `BeEquivalentTo()` for equality comparisons. It's meant for object graphs in very specific circumstances. `Be()` is more appropriate here.

int19h: Actually, since this is a string, so `BeEquivalentTo` in this case is case-insensitive, leading/trailing whitespace ignoring comparison. Unless you need any of that, though, `Be` should be preferred.

src/Package/TestApp/Utility/ViewTreeDump.cs
```diff
@@ -36,27 +35,36 @@ internal class ViewTreeDump {
...
 
-            if (compareProperty) {
+            var visibility = actual.Properties.FirstOrDefault(p => p.Name == "Visibility");
+            if (visibility != null) {
+                visible = (visibility.Value != "Collapsed");
+            }
+
+            if (compareProperty && visible) {
                 var filteredActual = actual.Properties.Where(p => SupportedWpfProperties.IsSupported(p.Name));
                 var filteredExpected = expected.Properties.Where(p => SupportedWpfProperties.IsSupported(p.Name));
```
int19h: This whole thing can be replaced with `filteredActual.Should().BeEquivalentTo(filteredExpected)`

src/Package/TestApp/Utility/ViewTreeDump.cs
```diff
@@ -36,27 +35,36 @@ internal class ViewTreeDump {
...
+                onlyInActual.Count().Should().Be(0);
+                onlyInExpected.Count().Should().Be(0);
             }
 
             actual.Children.Count.ShouldBeEquivalentTo(expected.Children.Count);
 
-            var sortedActualChildren = actual.Children.OrderBy(c => c.Name);
-            var sortedExpectedChildren = expected.Children.OrderBy(c => c.Name);
+            if (visible) {
+                var sortedActualChildren = actual.Children.OrderBy(c => c.Name);
```
int19h: Do `ToArray` here and use indexer below; `ElementAt` is going to be O(n).

PR #1321, opened 2016-03-22
VE Environment UI 

JeongPyoHam:
Combobox is added to show environment list


src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -16,6 +17,10 @@ internal class REnvironmentProvider : IREnvironmentProvider, IDisposable {
         public REnvironmentProvider(IRSession session) {
             _rSession = session;
             _rSession.Mutated += RSession_Mutated;
+
+            IdleTimeAction.Create(() => {
+                GetEnvironmentAsync().SilenceException<Exception>().DoNotWait();
```
MikhailArkhipov: Do we want to hide all of them?

src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -16,6 +17,10 @@ internal class REnvironmentProvider : IREnvironmentProvider, IDisposable {
         public REnvironmentProvider(IRSession session) {
             _rSession = session;
             _rSession.Mutated += RSession_Mutated;
+
+            IdleTimeAction.Create(() => {
```
AlexanderSher: Why do you need `IdleTimeAction` for background thread method?

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -25,24 +36,27 @@ public partial class VariableView : UserControl, IDisposable {
...
             SetRootNode(EvaluationWrapper.Ellipsis);
+            EnvironmentComboBox.ItemsSource = _defaultEnvironments;
+            EnvironmentComboBox.SelectedIndex = 0;
 
-            _globalEnvSubscription = variableProvider.Subscribe(0, VariableProvider.GlobalEnvironmentExpression, OnGlobalEnvironmentEvaluation);
-
+            SortDirection = ListSortDirection.Ascending;
             RootTreeGrid.Sorting += RootTreeGrid_Sorting;
+
+            var sessionProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IRSessionProvider>();
```
AlexanderSher: Please access `IRSession` through `IRInteractiveWorkflow`

src/Package/Impl/DataInspect/REnvironmentCollection.cs
```diff
@@ -6,6 +6,8 @@
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect {
     internal class REnvironmentCollection : List<REnvironment>{
```
int19h: What is the purpose of this class? If you just need to create a collection from JSON, you might as well just make a helper method.

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -57,16 +71,76 @@ public partial class VariableView : UserControl, IDisposable {
...
-            VsAppShell.Current.DispatchOnUIThread(
-                () => {
-                    EnvironmentName.Text = GlobalEnvironmentName;
-                    _rootNode.Model = rootNodeModel;
-                });
+        private void EnvironmentComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            if (!object.Equals(EnvironmentComboBox.ItemsSource, _defaultEnvironments) && e.AddedItems.Count > 0) {
+                var env = e.AddedItems[0] as REnvironment;
+                if (env != null) {
+                    SetRootModelAsync(env).SilenceException<Exception>().DoNotWait();
```
AlexanderSher: I don't like this pattern of silencing any exception during method execution. What exceptions do you expect?

src/Package/Impl/DataInspect/REnvironmentProvider.cs
int19h: `SilenceException<Exception>()` is almost always wrong. Ideally, this should only silence those exceptions that are actually expected during normal operation - for stuff that uses `IRSession`, this is typically `MessageTransportException`, and `OperationCanceledException`, but `DoNoWait` automatically silences the latter. If it really must be a catch-all, it should also log them, otherwise we'll have weird behavior reported by users with no way to diagnose.

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -4,20 +4,31 @@
...
 namespace Microsoft.VisualStudio.R.Package.DataInspect {
     public partial class VariableView : UserControl, IDisposable {
         private readonly IRToolsSettings _settings;
         ObservableTreeNode _rootNode;
-        VariableSubscription _globalEnvSubscription;
+        IREnvironmentProvider _environmentProvider;
+        IRSession _rSession;
+        DebugSession _debugSession;
 
         const string GlobalEnvironmentName = "Global Environment";
```
int19h: is this still needed?

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -4,20 +4,31 @@
...
 
+        private static REnvironmentCollection _defaultEnvironments = new REnvironmentCollection() { new REnvironment(Package.Resources.VariableExplorer_EnvironmentName) };
```
int19h: Where is `VariableExplorer_EnvironmentName` defined? And why does it even need a default collection?

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -25,24 +36,27 @@ public partial class VariableView : UserControl, IDisposable {
...
-
-            var variableProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IVariableDataProvider>();
-
             SetRootNode(EvaluationWrapper.Ellipsis);
+            EnvironmentComboBox.ItemsSource = _defaultEnvironments;
+            EnvironmentComboBox.SelectedIndex = 0;
 
-            _globalEnvSubscription = variableProvider.Subscribe(0, VariableProvider.GlobalEnvironmentExpression, OnGlobalEnvironmentEvaluation);
-
+            SortDirection = ListSortDirection.Ascending;
```
int19h: This can be done in XAML.

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -57,16 +71,76 @@ public partial class VariableView : UserControl, IDisposable {
...
+            EnvironmentComboBox.SelectedIndex = selectedIndex;
+        }
 
-            VsAppShell.Current.DispatchOnUIThread(
-                () => {
-                    EnvironmentName.Text = GlobalEnvironmentName;
-                    _rootNode.Model = rootNodeModel;
-                });
+        private void EnvironmentComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            if (!object.Equals(EnvironmentComboBox.ItemsSource, _defaultEnvironments) && e.AddedItems.Count > 0) {
```
int19h: This looks like it's a reference comparison, so `==` should work.

src/Package/Impl/DataInspect/VariableView.xaml.cs
int19h: Specific exceptions, please.

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -57,16 +71,76 @@ public partial class VariableView : UserControl, IDisposable {
...
+            }
+
+            return _debugSession;
+        }
+
+        private async Task SetRootModelAsync(REnvironment env) {
+            await TaskUtilities.SwitchToBackgroundThread();
+
+            var debugSession = await GetDebugSessionAsync();
+            var frames = await debugSession.GetStackFramesAsync();
```
int19h: Now this makes even less sense than before (as I commented on the first PR). Not only there are two pieces of code retrieving the call stack now, but both are actually used in this environment selector.

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -57,16 +71,76 @@ public partial class VariableView : UserControl, IDisposable {
...
+            var frame = frames.FirstOrDefault(f => f.Index == 0);
+
+            if (frame != null) {
+                const DebugEvaluationResultFields fields = DebugEvaluationResultFields.Classes
+                        | DebugEvaluationResultFields.Expression
+                        | DebugEvaluationResultFields.TypeName
+                        | (DebugEvaluationResultFields.Repr | DebugEvaluationResultFields.ReprStr)
+                        | DebugEvaluationResultFields.Dim
+                        | DebugEvaluationResultFields.Length;
+                DebugEvaluationResult result = await frame.EvaluateAsync(
```
int19h: If the sole purpose of obtaining the stack frames above is to call `EvaluateAsync` here and get `DebugEvaluationResult`, then this isn't actually necessary. `DebugSession` has `EvaluateAsync` that does the same, but lets you specify the environment explicitly (as an expression string, so it can be anything, e.g. `".GlobalEnv"`, and not just a frame).

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -57,16 +71,76 @@ public partial class VariableView : UserControl, IDisposable {
...
+                        _rootNode.Model = rootNodeModel;
+                    });
+            }
+        }
+
+        private string GetExpression(REnvironment env) {
+            if (env.FrameIndex.HasValue) {
+                return Invariant($"sys.frame({env.FrameIndex.Value})");
+            } else {
+                return Invariant($"as.environment('{env.Name}')");
```
int19h: Name can contain symbols that need to be escaped - use `ToRStringLiteral`.

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -76,8 +74,8 @@ public partial class VariableView : UserControl, IDisposable {
 
             var currentItem = EnvironmentComboBox.SelectedItem as REnvironment;
             if (currentItem != null && !e.Environments[0].FrameIndex.HasValue) {
-                int index = e.Environments.FindIndex((env) => env.Name == currentItem.Name);
-                selectedIndex = index == -1 ? selectedIndex : index;
+                var indexes = e.Environments.IndexWhere((env) => env.Name == currentItem.Name);
+                selectedIndex = indexes.Count() > 0 ? indexes.First() : selectedIndex;
```
int19h: Use `Any` rather than `Count` to avoid unnecessarily enumerating the whole thing. Or better yet:
```
selected Index = ...IndexWhere(...).FirstOrDefault() ?? selectedIndex
```

src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -18,9 +20,7 @@ internal class REnvironmentProvider : IREnvironmentProvider, IDisposable {
             _rSession = session;
             _rSession.Mutated += RSession_Mutated;
 
-            IdleTimeAction.Create(() => {
-                GetEnvironmentAsync().SilenceException<Exception>().DoNotWait();
-            }, 10, typeof(REnvironmentProvider));
+            GetEnvironmentAsync().DoNotWait();
```
int19h: You do need to silence `MessageTransportException` (and anything else that `GetEnviromentAsync` can be _expected_ to throw). My earlier comment was about catch-all `SilenceException<Exception>`, not about `SilenceException` in general.

src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -39,8 +39,16 @@ internal class REnvironmentProvider : IREnvironmentProvider, IDisposable {
 
             Debug.Assert(result.JsonResult != null);
 
+            var envs = new List<REnvironment>();
+            var jarray = result.JsonResult as JArray;
+            if (jarray != null) {
```
int19h: This should try to validate and throw if received JSON is not what is expected, rather than silently producing an empty list.

src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -39,8 +39,16 @@ internal class REnvironmentProvider : IREnvironmentProvider, IDisposable {
...
+                foreach (var jsonItem in jarray) {
```
int19h: `jarray.Select(item => new REnvironment(item)).ToList()`

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -88,7 +86,7 @@ public partial class VariableView : UserControl, IDisposable {
             if (!object.Equals(EnvironmentComboBox.ItemsSource, _defaultEnvironments) && e.AddedItems.Count > 0) {
                 var env = e.AddedItems[0] as REnvironment;
                 if (env != null) {
-                    SetRootModelAsync(env).SilenceException<Exception>().DoNotWait();
+                    SetRootModelAsync(env).DoNotWait();
```
int19h: Same thing about exceptions.

src/Package/Test/DataInspect/REnvironmentProviderTest.cs
```diff
@@ -20,63 +20,55 @@ namespace Microsoft.VisualStudio.R.Package.Test.DataInspect {
...
         [Test]
         [Category.Variable.Explorer]
         public async Task GetREnvironmentTest() {
-            await RunEnvironmentTestAsync("1", () => {
-                _environments.Count.Should().BeGreaterOrEqualTo(2);
-                _environments[0].Name.Should().BeEquivalentTo(".GlobalEnv");
-                _environments[_environments.Count - 1].Name.Should().BeEquivalentTo("package:base");
+            await RunEnvironmentTestAsync("1", (environments) => {
+                environments.Count.Should().BeGreaterOrEqualTo(2);
+                environments[0].Name.Should().BeEquivalentTo(".GlobalEnv");
```
int19h: `Be`

int19h: (and the same everywhere below, except where you need case-insensitive, leading/trailing whitespace ignoring comparisons)

src/Package/Test/DataInspect/REnvironmentProviderTest.cs
```diff
@@ -20,63 +20,55 @@ namespace Microsoft.VisualStudio.R.Package.Test.DataInspect {
...
-                _environments.Count.Should().BeGreaterOrEqualTo(3);
-                _environments[0].Name.Should().BeEquivalentTo("foo");
-                _environments[0].FrameIndex.HasValue.Should().BeTrue();
-                _environments[0].FrameIndex.Should().Be(1);
-                _environments[1].Name.Should().BeEquivalentTo(".GlobalEnv");
-                _environments[_environments.Count - 1].Name.Should().BeEquivalentTo("package:base");
+            await RunEnvironmentTestAsync("foo<-function(x){browser();};foo(1);", (environments) => {
+                environments.Count.Should().BeGreaterOrEqualTo(3);
+                environments[0].Name.Should().BeEquivalentTo("foo");
+                environments[0].FrameIndex.HasValue.Should().BeTrue();
```
int19h: Redundant - below will check for that anyway.

PR #1322, opened 2016-03-23
Fix #541, #1191, 547

### LINKED ISSUE #541: Debugger Break command can take a very long time to break
The current implementation tries to invoke `browser()` at the next available prompt. The problem is that if the code is in a long-running computation, there may not even be a prompt for a long time, and meanwhile VS UI is blocked. Need to look at alternative ways to pause execution asynchronously; perhaps via `CallBack`.

### LINKED ISSUE #547: Breakpoints are not instantly updated
When adding or removing breakpoints, the interpreter will only see the updates after it hits the next prompt in the REPL. In other words, if code is actively running and you set a breakpoint, it won't be hit - effectively, breakpoints can only be set before debugging, or when paused on another breakpoint or after stepping, or when R code is expecting keyboard input.

### LINKED ISSUE #1191: R Host process stop: unexpected host response, Escape key after browse(), no debugger
x<-1:10
browser()
<Escape> (maybe several times...)

Trace.Fail in ProtocolError() is called.

int19h:
Fix #541: Debugger Break command can take a very long time to break
Fix #547: Breakpoints are not instantly updated
Fix #1191: R Host process stop: unexpected host response, Escape key after browse(), no debugger

Add support for async background evals in RHost and RSession, such that they do not require queueing requests and waiting for the next prompt to run; and use those new evals to implement debugger break.

Use async background evals to set and remove breakpoints.

Add test for debugger break/continue.

Fix OverlappingBreakpoints test to properly set/remove breakpoints while at prompt (previous behavior inadvertently relied on breakpoint creation/deletion implicitly waiting for next prompt).

### DIFF
--- a/src/Debugger/Test/DebugSessionExtensions.cs
+++ b/src/Debugger/Test/DebugSessionExtensions.cs
@@ -23,7 +23,7 @@ namespace Microsoft.R.Debugger.Test {
             EventHandler<DebugBrowseEventArgs> handler = null;
             handler = (sender, e) => {
                 session.Browse -= handler;
-                browseTcs.SetResult(true);
+                browseTcs.TrySetResult(true);
             };
             session.Browse += handler;
 
--- a/src/Debugger/Test/DebugSessionExtensions.cs
+++ b/src/Debugger/Test/DebugSessionExtensions.cs
@@ -19,13 +19,7 @@ namespace Microsoft.R.Debugger.Test {
             // Now wait until session tells us that it has noticed the prompt and processed it.
             // Note that even if we register the handler after it has already noticed, but
             // before the interaction completed, it will still invoke the handler.
-            var browseTcs = new TaskCompletionSource<bool>();
-            EventHandler<DebugBrowseEventArgs> handler = null;
-            handler = (sender, e) => {
-                session.Browse -= handler;
-                browseTcs.TrySetResult(true);
-            };
-            session.Browse += handler;
+            var browseTask = EventTaskSources.DebugSession.Browse.Create(session);
 
             // Spin until either Browse is raised, or we see a different non-Browse prompt.
             // If the latter happens, we're not going to see Browse raised for that prompt that we've
@@ -34,9 +28,9 @@ namespace Microsoft.R.Debugger.Test {
             // the state is indeterminate.
             while (true) {
                 var interTask = session.RSession.BeginInteractionAsync();
-                var completedTask = await Task.WhenAny(browseTcs.Task, interTask);
+                var completedTask = await Task.WhenAny(browseTask, interTask);
 
-                if (completedTask == browseTcs.Task) {
+                if (completedTask == browseTask) {
                     interTask.ContinueWith(t => t.Result.Dispose(), TaskContinuationOptions.OnlyOnRanToCompletion).DoNotWait();
                     return;
                 }
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -42,9 +43,9 @@ namespace Microsoft.R.Host.Client {
         private readonly FileLogWriter _fileLogWriter;
         private Process _process;
         private volatile Task _runTask;
-        private bool _canEval;
         private int _rLoopDepth;
         private long _nextMessageId = 1;
+        private readonly ConcurrentDictionary<string, EvaluationRequest> _evalRequests = new ConcurrentDictionary<string, EvaluationRequest>();
 
         private TaskCompletionSource<object> _cancelAllTcs;
         private CancellationTokenSource _cancelAllCts = new CancellationTokenSource();
@@ -99,7 +100,7 @@ namespace Microsoft.R.Host.Client {
             return new Message(token);
         }
 
-        private JArray CreateMessage(CancellationToken ct, out string id, string name, params object[] args) {
+        private JArray CreateMessage(out string id, string name, params object[] args) {
             id = "#" + _nextMessageId + "#";
             _nextMessageId += 2;
             return new JArray(id, name, args);
@@ -121,7 +122,7 @@ namespace Microsoft.R.Host.Client {
 
         private async Task<string> SendAsync(string name, CancellationToken ct, params object[] args) {
             string id;
-            var message = CreateMessage(ct, out id, name, args);
+            var message = CreateMessage(out id, name, args);
             await SendAsync(message, ct);
             return id;
         }
@@ -130,7 +131,7 @@ namespace Microsoft.R.Host.Client {
             TaskUtilities.AssertIsOnBackgroundThread();
 
             string id;
-            var message = CreateMessage(ct, out id, ":", request.Id, request.Name, args);
+            var message = CreateMessage(out id, ":", request.Id, request.Name, args);
             await SendAsync(message, ct);
             return id;
         }
@@ -154,21 +155,14 @@ namespace Microsoft.R.Host.Client {
             }
         }
 
-        private async Task ShowDialog(Message request, bool allowEval, MessageButtons buttons, CancellationToken ct) {
+        private async Task ShowDialog(Message request, MessageButtons buttons, CancellationToken ct) {
             TaskUtilities.AssertIsOnBackgroundThread();
 
             request.ExpectArguments(2);
             var contexts = GetContexts(request);
             var s = request.GetString(1, "s", allowNull: true);
 
-            MessageButtons input;
-            try {
-                _canEval = allowEval;
-                input = await _callbacks.ShowDialog(contexts, s, _canEval, buttons, ct);
-            } finally {
-                _canEval = false;
-            }
-
+            MessageButtons input = await _callbacks.ShowDialog(contexts, s, buttons, ct);
             ct.ThrowIfCancellationRequested();
 
             string response;
@@ -192,7 +186,7 @@ namespace Microsoft.R.Host.Client {
             await RespondAsync(request, ct, response);
         }
 
-        private async Task ReadConsole(Message request, bool allowEval, CancellationToken ct) {
+        private async Task ReadConsole(Message request, CancellationToken ct) {
             TaskUtilities.AssertIsOnBackgroundThread();
 
             request.ExpectArguments(5);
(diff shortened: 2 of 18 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -237,7 +251,7 @@ internal sealed class RSession : IRSession, IRCallbacks {
                 if (startupInfo.WorkingDirectory != null) {
                     await evaluation.SetWorkingDirectory(startupInfo.WorkingDirectory);
                 } else {
-                    await evaluation.SetDefaultWorkingDirectory();
+                await evaluation.SetDefaultWorkingDirectory();
```
MikhailArkhipov: Formatting?

src/Host/Client/Impl/Program.cs
```diff
@@ -110,7 +110,7 @@ class Program : IRCallbacks {
 
                     if (s.StartsWith("$$", StringComparison.OrdinalIgnoreCase)) {
                         s = s.Remove(0, 1);
-                    } else if (s.StartsWith("$", StringComparison.OrdinalIgnoreCase) && isEvaluationAllowed) {
+                    } else if (s.StartsWith("$", StringComparison.OrdinalIgnoreCase)) {
```
MikhailArkhipov: StartsWithIgnoreCase ?

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -365,7 +323,12 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
                     if (message == null) {
                         return null;
                     } else if (message.RequestId != null) {
-                        return message;
+                        if (message.Name.StartsWith("=")) {
+                            ProcessEvaluationResult(message);
```
MikhailArkhipov: IgnoreCase for FxCop delight

src/Debugger/Test/DebugSessionExtensions.cs
```diff
@@ -23,7 +23,7 @@ internal static class DebugSessionExtensions {
             EventHandler<DebugBrowseEventArgs> handler = null;
             handler = (sender, e) => {
                 session.Browse -= handler;
-                browseTcs.SetResult(true);
+                browseTcs.TrySetResult(true);
```
AlexanderSher: Do you still need this ad-hoc or you can replace it with `EventTaskSources.DebugSession.Browse`?

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -99,7 +100,7 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
             return new Message(token);
         }
 
-        private JArray CreateMessage(CancellationToken ct, out string id, string name, params object[] args) {
+        private JArray CreateMessage(out string id, string name, params object[] args) {
```
AlexanderSher: Do we need this method to be thread-safe?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -103,24 +105,36 @@ internal sealed class RSession : IRSession, IRCallbacks {
 
         public Task<IRSessionInteraction> BeginInteractionAsync(bool isVisible = true, CancellationToken cancellationToken = default(CancellationToken)) {
             if (!_isHostRunning) {
-                return CanceledInteractionTask;
+                return CanceledBeginInteractionTask;
             }
 
             RSessionRequestSource requestSource = new RSessionRequestSource(isVisible, _contexts, cancellationToken);
```
MikhailArkhipov: Since request can be executed later, can _contexts list be out of date by the time the request gets to processing?

MikhailArkhipov: Also, it looks like Contexts on RSessionRequestSource doesn't need to be public anymore.

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -375,19 +338,27 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
...
                             case "?":
-                                await ShowDialog(message, allowEval, MessageButtons.YesNoCancel, CancellationTokenSource.CreateLinkedTokenSource(ct, _cancelAllCts.Token).Token);
+                                ShowDialog(message, MessageButtons.YesNoCancel, CancellationTokenSource.CreateLinkedTokenSource(ct, _cancelAllCts.Token).Token)
+                                    .SilenceException<MessageTransportException>()
+                                    .DoNotWait();
                                 break;
 
                             case "??":
-                                await ShowDialog(message, allowEval, MessageButtons.YesNo, CancellationTokenSource.CreateLinkedTokenSource(ct, _cancelAllCts.Token).Token);
+                                ShowDialog(message, MessageButtons.YesNo, CancellationTokenSource.CreateLinkedTokenSource(ct, _cancelAllCts.Token).Token)
```
AlexanderSher: `ShowDialog` has 2 awaits - `await _callbacks.ShowDialog(contexts, s, _canEval, buttons, ct);` and `await RespondAsync(request, ct, response);` Do we want to apply `DoNotWait` for both?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -454,14 +461,10 @@ internal sealed class RSession : IRSession, IRCallbacks {
         /// Graph app may call Win32 API directly rather than going via R API callbacks.
         /// </summary>
         /// <returns>Pressed button code</returns>
-        async Task<MessageButtons> IRCallbacks.ShowDialog(IReadOnlyList<IRContext> contexts, string s, bool isEvaluationAllowed, MessageButtons buttons, CancellationToken ct) {
+        async Task<MessageButtons> IRCallbacks.ShowDialog(IReadOnlyList<IRContext> contexts, string s, MessageButtons buttons, CancellationToken ct) {
```
AlexanderSher: So `isEvaluationAllowed` is always `true` now?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -103,24 +105,36 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
             }
 
             var source = new RSessionEvaluationSource(isMutating, cancellationToken);
             _pendingEvaluationSources.Post(source);
 
-            return _isHostRunning ? source.Task : CanceledEvaluationTask;
+            return _isHostRunning ? source.Task : CanceledBeginEvaluationTask;
+        }
+
+        public async Task<REvaluationResult> EvaluateAsync(string expression, REvaluationKind kind, CancellationToken ct = default(CancellationToken)) {
```
AlexanderSher: Do you need `async\await` here?

AlexanderSher: So we allow to call `EvaluateAsync` outside of queue?

src/Host/Client/Impl/Host/RHost.cs
int19h: Good catch, we do now.

src/Host/Client/Impl/Session/RSession.cs
int19h: Good catch. Weird, though, that we haven't seem to run into this before. @AlexanderSher, is this a problem, or are we missing something?

src/Host/Client/Impl/Host/RHost.cs
int19h: We have to, otherwise the loop won't process further incoming messages. Since the R inner loop won't move forward until it gets a response, it would seem okay?

src/Host/Client/Impl/Session/RSession.cs
int19h: Yes, in a sense that you can call `EvaluateAsync` at any time now. There's still a queue on the host side, and if there's an active eval that is not interruptible/reentrant, other evals will not be run in the middle of it. But they will not fail, either - they'll just stay in the queue, and run at the next available opportunity.

src/Host/Client/Impl/Session/RSession.cs
int19h: It's there to write a `try`/`catch` that converts `MessageTransportException` when host suddenly goes down to cancellation.

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: That can be a problem. I think we haven't run into this because we don't have cases when more than one request is scheduled.

src/Debugger/Test/DebugSessionExtensions.cs
int19h: Funnily enough, with this change, it will no longer work. The problem here was that `Browse` was actually raised more than once in some cases, second time being while the first handler is still running, and so `SetResult` was blowing up. So far as I can see, `EventTaskSource` doesn't even try to handle this case. I'm not sure if it should? But I could fix it there and then use it here.

src/Host/Client/Impl/Session/RSession.cs
int19h: Okay. Since this isn't part of this change, can you file a bug? And I'll address it in a separate PR to keep things clean.

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: So evaluation can be requested when host is in one state and get received when it is in the different state?

int19h: Absolutely. That's the whole point of the change (and the corresponding host change) - that evals can be fired at any moment, and will be serviced ASAP in an asynchronous fashion.

It's also why I want to keep `BeginEvaluationAsync` and friends - this lets you do evaluations in a controlled context, and check the context before running the eval. However, we have  evals all around that don't actually need this, and that suffer from the constraints that `BeginEvaluationAsync` model with the queue imposes - setting/removing breakpoints was one example of that, but I'm sure there are more similar cases.

int19h:
I've addressed the feedback that doesn't have any open questions.

src/Debugger/Test/DebugSessionExtensions.cs
AlexanderSher: EventTaskSource waits for only one event notification. Do you mean that here you need to wait for 1 or 2?

int19h: Looks like I've missed the part in `EventTaskSource` where it ensures safety if event is raised again while handler is already running. I'll make it use that.

int19h:
Switched to `EventTaskSource`, and added support for `Mutated` event.

MikhailArkhipov:
:+1:

AlexanderSher:
:+1:

Issue #1319, opened 2016-03-22
Add R Script/ and RMD gives error


MikhailArkhipov:
In regular setup templates are in a different location

PR #1324, opened 2016-03-23
Fix #1320: #1319

### LINKED ISSUE #1320: Show 'Set Directory Here' on the project node

MikhailArkhipov:
#1319: Add R Script/ and RMD gives error
#1320 Show "Set Directory Here..." on project node

### DIFF
--- a/src/ProjectSystem/Impl/Extensions/ProjectTreeExtensions.cs
+++ b/src/ProjectSystem/Impl/Extensions/ProjectTreeExtensions.cs
@@ -6,6 +6,7 @@ using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using Microsoft.VisualStudio.ProjectSystem.Designers;
+using Microsoft.VisualStudio.ProjectSystem.Utilities.Designers;
 
 namespace Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring {
     public static class ProjectTreeExtensions {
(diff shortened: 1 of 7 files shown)


src/ProjectSystem/Impl/Extensions/ProjectTreeExtensions.cs
```diff
@@ -49,5 +49,16 @@ public static class ProjectTreeExtensions {
             }
             return Enumerable.Empty<string>();
         }
+
+        public static string GetSelectedFolderPath(this IImmutableSet<IProjectTree> nodes, UnconfiguredProject unconfiguredProject) {
+            if (nodes.Count == 1) {
+                var n = nodes.First();
+                if (n.Root == n) {
```
AlexanderSher: What does it mean?

MikhailArkhipov: Project node has Root set to itself (in which case I need to use different method for obtaining node path as project node in CPS return empty string from FilePath). Any better way to tell project node from the file node in CPS?

AlexanderSher: Not really, but they have an extension :)
```{csharp}
namespace Microsoft.VisualStudio.ProjectSystem
{
    public static class ProjectTreeExtensions
    {
        /// <summary>
        /// Checks whether the given node is a root node in its tree.
        /// </summary>
        /// <param name="node">The node to test.</param>
        /// <returns><c>true</c> if the node is the root node; otherwise <c>false</c>.</returns>
        [Pure]
        public static bool IsRoot(this IProjectTree node)
        {
            Contract.Requires(node != null);
 
            Requires.NotNull(node, nameof(node));
            return node.Root == node;
        }
 
```

MikhailArkhipov: OK

AlexanderSher:
:+1:

PR #1329, opened 2016-03-23
Open data in Excel (or any registered CVS app) via file writing

MikhailArkhipov:
COM is too slow. Comments are welcome :-)


src/Package/Impl/DataInspect/EvaluationWrapper.cs
```diff
@@ -148,5 +153,16 @@ public sealed class EvaluationWrapper : RSessionDataObject, IIndexedItem {
             return false;
         }
         #endregion
+
+        private bool ComputCsvAvailability(DebugValueEvaluationResult evaluation) {
```
int19h: Spelling: Compute

int19h: Is there any particular reason why this is separate from `ComputeDetailAvailability` (which should probably be renamed to something like `ComputeGridAvailability`)? It would appear that if we can open something in the grid, it should be exportable to Excel, and vice versa.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+            var workflow = workflowProvider.GetOrCreate();
+            var session = workflow.RSession;
+
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
+            folder = Path.Combine(folder, @"RTVS_CSV_Exports\");
+            if(!Directory.Exists(folder)) {
+                Directory.CreateDirectory(folder);
+            }
+
+            // Strip "base::.GlobalEnv$"
```
int19h: If you pass `DebugEvaluationResult` here instead of `string`, you can use its `Name` property to get the tail instead of stripping the prefix (and it'll then work in any frame, not just global - once JP commits packages, you'll have the same problem with them).

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+            string currentStatusText = string.Empty;
+            await VsAppShell.Current.DispatchOnMainThreadAsync(() => {
+                var statusBar = VsAppShell.Current.GetGlobalService<IVsStatusbar>(typeof(SVsStatusbar));
+                statusBar.GetText(out currentStatusText);
+            });
+
+            try {
+                await SetStatusTextAsync(Resources.Status_WritingCSV);
+                using (var e = await session.BeginInteractionAsync()) {
+                    await e.RespondAsync(Invariant($"write.csv({variableName}, file='{rfile}')") + Environment.NewLine);
```
int19h: This should be an eval - this way you can reliably wait for its completion and check for errors before proceeding.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+
+                Task.Run(() => {
+                    try {
+                        Process.Start(file);
+                    } catch(Win32Exception ex) {
```
int19h: Also need to catch `FileNotFoundException`

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+                        VsAppShell.Current.ShowErrorMessage(
+                            string.Format(CultureInfo.InvariantCulture, Resources.Error_CannotOpenCsv, ex.Message));
+                    }
+                }).DoNotWait();
+            } finally {
+                await SetStatusTextAsync(currentStatusText);
```
int19h: It seems like there's a race here if two exports are triggered quickly in parallel - the second one might observe the text set by the first one and stash it away as `currentStatusText`.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
+            folder = Path.Combine(folder, @"RTVS_CSV_Exports\");
+            if(!Directory.Exists(folder)) {
+                Directory.CreateDirectory(folder);
+            }
+
+            // Strip "base::.GlobalEnv$"
+            var index = expression.IndexOf(VariableProvider.GlobalEnvironmentExpression);
+            var variableName = index >= 0 ? expression.Substring(index + VariableProvider.GlobalEnvironmentExpression.Length + 1) : expression;
```
int19h: What if the expression is not a simple variable? There's no such thing in VE today, but there will likely be in the future, and it will definitely be possible once #1335 is implemented (as it will also apply to Watch).

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+                csvFileName += Path.GetFileNameWithoutExtension(projectName);
+                csvFileName += "_";
+            }
+
+            if (!string.IsNullOrEmpty(filePath)) {
+                csvFileName += Path.GetFileNameWithoutExtension(filePath);
+                csvFileName += "_";
+            }
+
+            csvFileName += variableName.Replace('.', '_').Replace('@', '_').Replace('$', '_');
```
int19h: Same comment regarding arbitrary expressions applies here.

However, even considering just the existing implementation, an expression of the form `foo[[1]]$baz` or `foo[['bar']]$baz` is also possible, given something like:
```r
e <- as.environment(list(baz = as.data.frame(0)))
foo <- list(e, bar = e)
```
The latter is particularly problematic, since it has a string literal that can contain pretty much anything.

int19h: Also don't forget about backticks: something like `` `c:\foo` `` is a legitimate variable name, too.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -0,0 +1,108 @@
...
+        public static async Task OpenDataCsvApp(string expression) {
+            if (Interlocked.Exchange(ref _busy, 1) > 0) {
+                return;
+            }
+
+            var workflowProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>();
+            var workflow = workflowProvider.GetOrCreate();
+            var session = workflow.RSession;
+
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
```
int19h: Do we really want these to be dumped to `Documents`? As they are effectively temp files (the command does not imply that it exports anything, so the fact that it does is implementation detail), I'd expect to see them in `%TEMP%`.

src/Package/Impl/DataInspect/EvaluationWrapper.cs
MikhailArkhipov: Open in CSV also handles vectors. Grid wants 2-dimensional and I am not sure it is safe to enable vectors in the grid (although we have user request for that).

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: I made special folder for them so they can be deleted at once. On the other hand, I though about TEMP, just not sure if user ever want to have the file easy to find. I'll check with Shahrokh.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: This is basically for naming of the file. Shahrokh wanted something like project_file_variable naming. There is probably no good universal convention here... I am open to suggestions.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: OK

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: I will do after simple replacement
```
variableName.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
```
and if it is still bad, just substitute name with "expression'" (i.e. literal not the actual expression)

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: PM says under documents is OK, we can make it configurable if we want. Problem with TEMP is that active Excel user will end up with 100s of temp csv files that are hard to find and clean (not everyone ventures under hidden AppData\Local).

MikhailArkhipov:
Merging so Shahrokh can try this out. Anything extra can be added.

int19h:
Note that we will still be creating files like `a[[1]]_b` and even `[[1]]` with this change.

Also, `DebugEvaluationResult.Name` can be `null`, and it will in fact be that when an arbitrary expression is evaluated (e.g. via Watch or Immediate), so the code needs to check for that.

MikhailArkhipov:
```[[ ``` are OK - as long as file name is working - however, I am wondering if there are cases when path chars are legal but overall name is not?. However I need to check for max path length as well. Will add more checks today.

smortaz:
need to add skip rows:

nrows
integer: the maximum number of rows to read in. Negative and other invalid values are ignored.

skip
integer: the number of lines of the data file to skip before beginning to read data.

PR #1339, opened 2016-03-24
Fix #1338: DataTip for indexers is wrong

### LINKED ISSUE #1338: DataTip for indexers is wrong
E.g.:
```r
a$b[c]
```
Hovering over `a` should show the value of the entire expression, but it only shows `a$b`.

int19h:
Since indexer is an unary operator in AST now, use Children.First() instead of LeftOperand to determine which part of it the child node belongs to.


int19h:
Ping.

MikhailArkhipov:
:+1:

Issue #1313, opened 2016-03-22
Hex number has squiggles


huguesv:
Not sure if that's supposed to happen or not, the code works fine.  Discovered it by opening our json.R in the editor, where the hex number on this line has squiggles:
```r
con <- memory_connection(NA, 0x10000);
```

You can also just type the hex number in REPL and it will have squiggles:
```r
> 0x10000
[1] 65536
```

PR #1349, opened 2016-03-24
#1313 Number parsing

MikhailArkhipov:
Made common number parser class
We didn't handle hex numbers properly
Now we can, including hex complex
Added more hex number tests
Fixed one broken project command test


src/R/Core/Impl/Tokens/Number.cs
```diff
@@ -0,0 +1,42 @@
...
+                return false;
+            }
+            if (allowLSuffix && text[text.Length - 1] == 'L') {
+                text = text.Substring(0, text.Length - 1);
+            }
+
+            if (text.Length == 0) {
+                return false;
+            }
+            if (text[0] == '-' || text[0] == '+') {
```
AlexanderSher: Do we handle `-` as sign or as a function?

MikhailArkhipov: if sign is immediate to the number as in -1 it is part of the number. If it is separate as in - 1 then it is unary operator over 1.

int19h: Could this be simplified further by always treating the sign as an operator? Does anything actually rely on `-1` being parsed as a single token?

MikhailArkhipov: The primary reason it is done this way is colorization which goes by tokens. So -1 gets colorized as a number together with minus. It is possible to always treat it as an operator (except in complex types and exponents) but then if one selects different colors for numbers and operators they will see - as operator which is not what is expected. Although C# doesn't seem to care

image

and neither does C++ or JavaScript.

MikhailArkhipov: Although that would be a bigger change than this

MikhailArkhipov: So, I am not changing prefix handling to operator just yet, but will file a separate task

MikhailArkhipov:
https://github.com/Microsoft/RTVS/issues/1352

src/R/Core/Impl/Tokens/Number.cs
int19h: My 2c is that it should colorize in the same way that the actual R parser does things, so if it considers it an operator (which seems to be the case if you look at output of `parse(-1)`), then that should be it.

Also, in the future, when we have "Go to Definition", it will probably make sense to have it for operators as well, in which case treating `-` as a separate token would be good (it would go to `` base::`-` ``).

But sure, let's leave it for later.

int19h:
:+1:

PR #1351, opened 2016-03-24
Excel - address post merge feedback

MikhailArkhipov:
Add extra checks
Clean CSV temp files on exit


src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -71,6 +73,22 @@ internal static class CsvAppFileIO {
...
+                foreach (var f in Directory.EnumerateFiles(folder)) {
+                    try {
+                        File.Delete(f);
+                    } catch(IOException) { } catch (UnauthorizedAccessException) { }
+                }
+            }
+        }
+
+        private static string GetTempCsvFilesFolder() {
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
```
int19h: If we're deleting them on exit now, is it still appropriate?

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -71,6 +73,22 @@ internal static class CsvAppFileIO {
             Interlocked.Exchange(ref _busy, 0);
         }
 
+        public static void Close() {
+            var folder = GetTempCsvFilesFolder();
+            if (Directory.Exists(folder)) {
```
AlexanderSher: Can't you just delete the folder?

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -108,9 +126,10 @@ internal static class CsvAppFileIO {
             if(variableName.StartsWith("$", StringComparison.Ordinal)) {
                 variableName = variableName.Substring(1);
             }
+
             variableName = variableName.Replace('.', '_').Replace('@', '_').Replace('$', '_');
```
int19h: I would suggest also adding `[]` here (or maybe dropping `$` and `@` - are they actually invalid chars? I think dollar at least isn't.).

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: Shahrokh doesn't want them in TEMP, he eventually wants this folder configurable in options.

MikhailArkhipov: That's why I moved it into a function so we can expand on this later

int19h: I don't understand why, though. I understand why you'd want to put these files into Documents when user is doing "Export to CSV". But in this scenario, which is "Open in Excel", these are transient files. Why would the user care where they live? They're going to be gone as soon as he closes VS, anyway.

@mortaz, can you clarify?

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: I think files still need to be deleted though. I can nuke the folder as well.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
int19h: @smortaz, can you clarify the intent?

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
MikhailArkhipov: Actually, some files may still be locked by being open in Excel

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
smortaz: we dont know all the different scenarios that users may have for RTVS-Excel.  for example, someone might want to use this feature to export large DF's or a series or related DF's to Excel to do some Excel centric processing with them post RTVS session.  as such, you dont want them to go hunting around in their TEMP folder digging up .csv files.  i personally think the 1) location and 2) whether they're deleted or not should be options with defaults: 1) ~/RTVS-Excel/ and 2) Delete on Exit.

int19h: I think you're confusing "Export to CSV" with "Open in Excel". They are two different commands, and this particular code only pertains to the latter. If the user wants a file on disk that they want to open later, they should use "Export".

In the case of "Open", writing that data to file is an implementation detail (we didn't even do that before, but loading it directly into Excel sheet via COM APIs proved to be too slow). That file is going to be automatically deleted when VS is closed, so there's no "Excel centric processing post RTVS session" that can happen here.

int19h: (And if they use "Open", and then decide that they want to save it later, the proper workflow is to do "Save As" in Excel, and place it wherever they want it to be, and give it a name that they will remember.)

int19h:
:+1:

PR #1353, opened 2016-03-25
Some package management backend and tests

huguesv:
Implementation and tests for enumeration of available and installed packages.

Many fields are missing when enumerating available packages, because they are missing from the server's PACKAGES index file.  Fetching those missing fields from an alternate location is a separate task.

Optimization of JSON serialization (on R side) is also a separate task.  The AvailablePackagesCranRepo test that fetches 8000 packages takes about 50 secs to run.

The tests include the use of a local test repo so that we can more accurately validate correctness of data (we control the repo and its packages).  I've included the sources of the packages (well there's just 1 to start with) so that we can rebuild them and repopulate the test repos as necessary.

The tests also use a local library to install known packages from the test repo.  Will also be useful when I get to do uninstall/update features.

I've organized the test repo and test packages in a way that will allow me to test installation of multiple packages, using multiple repos, etc later on as I implement the other tasks.  So I expect to later add another test repo and a few more test packages.


src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -0,0 +1,208 @@
...
+using FluentAssertions;
+using Microsoft.R.Components.InteractiveWorkflow;
+using Microsoft.R.Components.Test.Fakes.InteractiveWindow;
+using Microsoft.R.Host.Client;
+using Microsoft.R.Host.Client.Test.Script;
+using Microsoft.R.Support.Settings;
+using Microsoft.UnitTests.Core.XUnit;
+
+namespace Microsoft.R.Components.Test.PackageManager {
+    public class PackageManagerIntegrationTest : IDisposable {
```
AlexanderSher: You can implement `IAsyncLifetime` and move `CreateWorkflowAsync` body into `InitializeAsync`

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -27,6 +34,40 @@ internal class RPackageManager : IRPackageManager {
             return VisualComponent;
         }
 
+        public async Task<IEnumerable<RPackage>> GetInstalledPackagesAsync() {
```
AlexanderSher: I'd like it to be `Task<IList<RPackage>>`

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            rtvslib1Pkg.Depends.Should().Be(TestPackages.RtvsLib1.Depends);
+            rtvslib1Pkg.License.Should().Be(TestPackages.RtvsLib1.License);
+            rtvslib1Pkg.Built.Should().Be(TestPackages.RtvsLib1.Built);
+            rtvslib1Pkg.NeedsCompilation.Should().Be(null);
+            rtvslib1Pkg.Author.Should().Be(TestPackages.RtvsLib1.Author);
+            rtvslib1Pkg.Enhances.Should().Be(null);
+            rtvslib1Pkg.Title.Should().Be(TestPackages.RtvsLib1.Title);
+        }
+
+        private async Task SetLocalRepoAsync(IRSessionEvaluation eval, string localRepoPath) {
```
AlexanderSher: These methods (`SetLocalRepoAsync`, `SetLocalLibAsync` and eval part of `InstallPackageAsync`) will be useful in product itself (I'll need them to set repos at RSession startup). Can you please move them into RPackageManager as well?

AlexanderSher:
I'd like @int19h to take a look at this PR

src/Host/Client/Impl/Extensions/RStringExtensions.cs
```diff
@@ -91,5 +91,13 @@ public static class RStringExtensions {
             }
             return s;
         }
+
+        public static string ToRPath(this string s) {
```
int19h: I'm curious about what happens if packages are installed under the user directory. Normally, R shortens such paths using `~` - if that is also the case here, it needs to be dealt with. We have `MakeRelativeToRUserDirectoryAsync` method which may be of help.

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.R.Host.Client;
 
 namespace Microsoft.R.Host.Client.Session {
     public static class RSessionEvaluationCommands {
```
int19h: This is becoming a dumping ground for all kind of stuff. Perhaps not for this PR (up to you), but I think in the future we should try to split it up. For things that are just straightforward wrappers around `rtvs` members, I think we should have separate extension classes, one for each .R file that we have in the package.

src/R/Components/Impl/PackageManager/IRPackageManager.cs
```diff
@@ -2,11 +2,18 @@
...
+using System.Threading.Tasks;
+using Microsoft.R.Components.PackageManager.Model;
 
 namespace Microsoft.R.Components.PackageManager {
     public interface IRPackageManager : IDisposable {
         IRPackageManagerVisualComponent VisualComponent { get; }
 
         IRPackageManagerVisualComponent GetOrCreateVisualComponent(IRPackageManagerVisualComponentContainerFactory visualComponentContainerFactory, int instanceId = 0);
+
+        Task<IList<RPackage>> GetInstalledPackagesAsync();
```
int19h: If there's no specific meaning to the order of packages, perhaps `ICollection` is more appropriate here.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -27,6 +34,42 @@ internal class RPackageManager : IRPackageManager {
             return VisualComponent;
         }
 
+        public async Task<IList<RPackage>> GetInstalledPackagesAsync() {
+            if (_interactiveWorkflow.RSession == null || !_interactiveWorkflow.RSession.IsHostRunning) {
+                return new List<RPackage>();
+            }
+
+            REvaluationResult result;
+            using (IRSessionEvaluation eval = await _interactiveWorkflow.RSession.BeginEvaluationAsync()) {
```
int19h: This might be one of those cases where it would be better to call `EvaluateAsync` directly on `IRSession` instead of going through `BeginEvaluationAsync`, since you don't care when it happens, or what the context is. Avoiding `BeginEvaluationAsync` means that it'll work even when REPL is running code.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -27,6 +34,42 @@ internal class RPackageManager : IRPackageManager {
...
+            return ParsePackages(result);
+        }
+
+        public async Task<IList<RPackage>> GetAvailablePackagesAsync() {
+            if (_interactiveWorkflow.RSession == null || !_interactiveWorkflow.RSession.IsHostRunning) {
+                return new List<RPackage>();
+            }
+
+            REvaluationResult result;
+            using (IRSessionEvaluation eval = await _interactiveWorkflow.RSession.BeginEvaluationAsync()) {
```
int19h: Same thing.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -27,6 +34,42 @@ internal class RPackageManager : IRPackageManager {
...
+                result = await eval.AvailablePackages();
+            }
+
+            return ParsePackages(result);
+        }
+
+        private static IList<RPackage> ParsePackages(REvaluationResult result) {
+            List<RPackage> packages = new List<RPackage>();
+            foreach (var jsonProp in ((JObject)result.JsonResult).Properties()) {
```
int19h: This looks like it could be a LINQ expression. Something like:
```cs
return (from jsonProp in (JObject)result.JsonResult).Properties()
        let jsonObj = (JObject)jsonProp.Value
        select jsonObj.ToObject<RPackage>()
       ).ToList()
```

src/R/Components/Impl/PackageManager/IRPackageManager.cs
AlexanderSher: For the sake of covariance, I'd prefer to use the most detailed interface types for return values and out parameters

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            using (var eval = await _workflow.RSession.BeginEvaluationAsync()) {
+                await SetLocalRepoAsync(eval, _repo1Path);
+            }
+
+            var result = await _workflow.Packages.GetAvailablePackagesAsync();
+            result.Should().HaveCount(3);
+
+            var rtvslib1Pkg = result.SingleOrDefault(pkg => pkg.Package == TestPackages.RtvsLib1.Package);
+            rtvslib1Pkg.Should().NotBeNull();
+            rtvslib1Pkg.Version.Should().Be(TestPackages.RtvsLib1.Version);
```
int19h: Instead of checking individual fields, I would suggest creating another `RPackage` instance, and using `ShouldBeEquivalentTo`. That will reflect over the fields and compare field by field, and then if the test fails, the error message will be more descriptive (it'll say something like "Version should be X, but was Y", instead of "expected X, got Y").

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
AlexanderSher:
```{csharp}
return ((JObject)result.JsonResult).Properties()
    .Select(p => p.Value.ToObject<RPackage>())
    .ToList();
```

src/R/Components/Impl/PackageManager/IRPackageManager.cs
MikhailArkhipov: IReadOnlyList?

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
int19h: Or maybe even just define `Equals` and friends for `RPackage`, and then you can just use `Be`.

src/R/Components/Impl/PackageManager/IRPackageManager.cs
AlexanderSher: Ok

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
AlexanderSher: Typical service facade.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
AlexanderSher: And also you still need to handle OCE, even if `IsHostRunning` was true

src/Host/Client/Impl/Extensions/RStringExtensions.cs
huguesv: It's still the full path in those cases, it's doesn't return paths with `~` in them.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -27,6 +34,42 @@ internal class RPackageManager : IRPackageManager {
             return VisualComponent;
         }
 
+        public async Task<IList<RPackage>> GetInstalledPackagesAsync() {
+            if (_interactiveWorkflow.RSession == null || !_interactiveWorkflow.RSession.IsHostRunning) {
```
AlexanderSher: `RSession` is always not null

src/Host/Client/Impl/Extensions/RStringExtensions.cs
huguesv:
```
> installed.packages()[[1,'LibPath']]
[1] "C:/Users/huvalo/Documents/R/win-library/3.2"
```

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
huguesv: I don't think the current implementation in the test will be appropriate for the product, let me have a chat with you on what you need.

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            using (var eval = await _workflow.RSession.BeginEvaluationAsync()) {
+                await SetLocalLibAsync(eval, _libPath);
+            }
+
+            var result = await _workflow.Packages.GetInstalledPackagesAsync();
+
+            // Since we can't remove Program Files folder from library paths,
+            // we'll get some results, but nothing from the test folder.
+            result.Should().NotBeEmpty();
+            result.Where(pkg => pkg.LibPath == _libPath.ToRPath()).Should().BeEmpty();
```
AlexanderSher:
```{csharp}
result.Should().NotBeEmpty()
    .And.NotContain(pkg => pkg.LibPath == _libPath.ToRPath();
```

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -0,0 +1,208 @@
...
+                await SetLocalRepoAsync(eval, _repo1Path);
+                await SetLocalLibAsync(eval, _libPath);
+                await InstallPackageAsync(eval, TestPackages.RtvsLib1.Package, _libPath);
+            }
+
+            var result = await _workflow.Packages.GetInstalledPackagesAsync();
+
+            result.Should().NotBeEmpty();
+
+            var rtvslib1Pkg = result.SingleOrDefault(pkg => pkg.Package == TestPackages.RtvsLib1.Package && pkg.LibPath == _libPath.ToRPath());
```
AlexanderSher:
```{csharp}
var rtvslib1Pkg = result.Should().ContainSingle().Which;
```

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
huguesv: If I use RSession.EvaluateAsync to implement GetInstalledPackages and GetAvailablePackages, I'm having trouble in the tests where the test never calls RSession.BeginEvaluationAsync.  The trouble is that it complains that `rtvs` package doesn't exist (it's used by the code I'm evaluating).
The tests that do setup work that still uses BeginEvaluationAsync have no problem when they get to call GetInstalledPackages.

So is this a gotcha of the "you don't care when it happens" ?

int19h: Ahh... It's a gotcha of the way the initialization code is structured currently. Basically when the session creates the host, it queues up the code that loads the package (via `BeginEvaluationAsync`, or rather some internal code that amounts to the same thing and uses the same queue) as a first eval request on the host. So any subsequent eval that goes through `BeginEvaluationAsync` will be next in queue, and initialization code is guaranteed to run by the time it happens. But when you call `EvaluateAsync` directly, it doesn't go in that queue - it just gets fired directly to the host.

@AlexanderSher, I think we'll need to discuss how to properly handle this. But for now let's keep `BeginEvaluationAsync`. It does work after all, it just means that package manager will not be accessible while REPL is running code. And switching to straight `EvaluateAsync` later, once we figure out how to solve this, is cheap and easy.

huguesv: Cool, I'll change it back.

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
huguesv: Some of these may be able to be changed to use EvaluateAsync directly in the future, that will probably be a good time to do some reorganization.  I like the idea of separating them by area.

int19h: We talked about it, and I'm going to make it so that there is a single interface `IRExpressionEvaluator` that is a single point for this, and then they can be extension methods on that. Currently we have that interface already, but only `RHost` implements it - `IRSession` and `IRSessionEvaluation` do not, even though the functionality is very similar. So we'll unify.

huguesv:
I think I've addressed the feedback.

Note that I've added cloning support on RPackage.  I could have done it differently since it's only for tests, let me know if you don't like it.

I've also added evaluation result error checking.

int19h:
:+1:

AlexanderSher:
:+1:

PR #1357, opened 2016-03-25
User function/variable completion + goto/peek definition + roxygen blocks

MikhailArkhipov:
User functions in completion
User variables in completion
Function arguments in completion
User functions signature help
```###``` inserts Roxygen block
Go To Definition (functions, variables, arguments)
Peek Definition (as above)
Tests

### CODE REVIEW

File: src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        /// <summary>
+        /// Enumerates definitions of variables applicable to the given scope.
+        /// Traverses scopes from the provided scope up and enumerates
+        /// variables that appear at the left side of the left-hand assignment
+        /// operator or at the right side of the right-hand assignmend operator.
+        /// Includes both regular variables as well as function definitions.
+        /// Includes assignments that appear in the file up to the specified
+        /// position except in the global scope it enumerates all assignments.
```
int19h: I think there should be the same behavior here regardless of the scope. In global, we want this to look ahead, because functions often call other functions (that are ahead) in their body; and body is usually evaluated after all the assignments are evaluated. But the same can be the case in non-global scopes - you can still have something like:
```r
f <- function() {
  g <- function() h()
  h <- function() ...
  g()
}
```
and it's just as valid. Similarly, something like this is not valid in global scope, but we will still incorrectly list `y` as available:
```r
x <- y
y <- 123
...
So it's a trade-off, and I think it's better if it's done the same way on both global and non-global level.

File: src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+            if (!globalScope) {
+                // See if this is a function scope with arguments
+                var funcDef = scope.Parent as IFunctionDefinition;
+                if (funcDef != null && funcDef.Arguments != null) {
+                    foreach (var arg in funcDef.Arguments) {
+                        var na = arg as IVariable;
+                        if (na != null) {
+                            yield return na;
+                        } else {
+                            var ea = arg as ExpressionArgument;
```
int19h: Can you clarify what an `ExpressionArgument` is?

src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        /// <summary>
+        /// Locates function with a given name inside the scope. Only function
+        /// definition that appear before the given position are analyzed except
+        /// when scope is the global scope.
+        /// </summary>
+        public static IVariable FindFunctionDefinitionByName(this IScope scope, string name, int position) {
+            var variables = scope.GetApplicableVariables(position);
+            var v = variables.FirstOrDefault(x =>
+                x.Name.EqualsOrdinal(name) && (x.Value is RFunction));
```
int19h: Swapping the checks would make this a bit faster.

src/R/Editor/Impl/Completion/Definitions/RCompletionContext.cs
```diff
@@ -118,28 +96,22 @@ public static bool IsCaretInLibraryStatement(ITextView textView)
...
-                    if (s == "library" || s == "require")
-                    {
+                    if (s == "library" || s == "require") {
                         return true;
                     }
                 }
-            }
-            catch (Exception) { }
-        
+            } catch (Exception) { }
```
int19h: While you're here, can you also make this catch more specific?

src/R/Editor/Impl/Completion/Definitions/RCompletionContext.cs
```diff
@@ -118,28 +96,22 @@ public static bool IsCaretInLibraryStatement(ITextView textView)
...
+            SnapshotPoint? pt = REditorDocument.MapCaretPositionFromView(textView);
+            if (pt.HasValue && pt.Value > 0) {
                 int i = pt.Value - 1;
-                for (; i >= 0; i--)
-                {
+                for (; i >= 0; i--) {
                     char ch = snapshot[i];
-                    if (!RTokenizer.IsIdentifierCharacter(ch) && ch != '$' && ch != '@')
-                    {
+                    if (!RTokenizer.IsIdentifierCharacter(ch) && ch != '$' && ch != '@') {
```
int19h: Unrelated to this PR, but it looks like this won't handle `` `foo bar`$baz `` right? Might be a separate bug to file.

src/R/Editor/Impl/Completion/Providers/PackageFunctionCompletionProvider.cs
```diff
@@ -60,8 +60,12 @@ public class FunctionCompletionProvider : IRCompletionListProvider {
                 IEnumerable<INamedItemInfo> functions = pkg.Functions;
                 if (functions != null) {
                     foreach (INamedItemInfo function in functions) {
-                        bool? isSnippet = infoSource?.IsSnippet(function.Name);
-                        if (!isSnippet.HasValue || !isSnippet.Value) {
+                        bool isSnippet = false;
+                        // Snippets are suppressed if user typed namespace
+                        if (!context.IsInNameSpace() && infoSource != null) {
```
int19h: Is this when we're on the right side of `::`? Should it also exclude snippets on the right side of `@` and `$`?

src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs
MikhailArkhipov: It is an interesting question... Originally I did it line this. However, the behavior then changes depending on if file was sourced or not. If it was, then all the functions in global scope become available in the editor. Same happens in RS. So current behavior is identical before and after sourcing file. Also, I have files from R itself (like check.r) where code in the middle of the file references functions declared in the end.

src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs
MikhailArkhipov: ExpressionArgument is a most common type of arguments. i.e. in func(x+1, a) both are expression arguments. 'x = 1' is a named argument. ,, is a missing argument, in func(,, third argument is a 'stub argument'.

src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs
MikhailArkhipov: Sure

src/R/Editor/Impl/Completion/Definitions/RCompletionContext.cs
MikhailArkhipov: OK. This is probably old protection from random editor exceptions which used to happen years ago.

src/R/Editor/Impl/Completion/Providers/PackageFunctionCompletionProvider.cs
MikhailArkhipov: Probably yes.

src/R/Editor/Impl/Completion/Definitions/RCompletionContext.cs
MikhailArkhipov: Nope. Code needs to be generally inspected for backtick handling. I am pretty sure there are quite a few places that may not be doing it right. Opened https://github.com/Microsoft/RTVS/issues/1360

src/R/Editor/Impl/Completion/Providers/PackageFunctionCompletionProvider.cs
MikhailArkhipov: This is [mostly] cosmetic however. By default snippet wins so when typing lm icon with scissors is visible. However, snippet is only inserted if user hits tab-tab. Otherwise there is no difference and typing lm( brings regular function. When in namespace snippets are suppressed and icon looks like function. So yes, code should suppress snippets past selectors as well although there is not much harm at the moment, perhaps occasional wrong icon.

MikhailArkhipov: Also, we don't show anything after $ except variable members (if any) - i.e. only one completion list provider is active, so no snippets, keywords or functions appear anyway. ```::``` is pretty much the only case when we show functions.

src/R/Core/Impl/AST/Extensions/ScopeExtensions.cs
int19h: Right, so it does the right thing (or at least the better thing, given our constraints) in the global scope. What I'm suggesting is that the same logic - list everything, even if it's assigned after the current line - should be applied to all other scopes, because the same case where code in the middle can refer to code at the end, also exists there.

src/R/Editor/Impl/Completion/Definitions/RCompletionContext.cs
```diff
@@ -9,108 +9,84 @@
...
-                if (bufferPosition.HasValue)
-                {
+                if (bufferPosition.HasValue) {
                     ITextSnapshot snapshot = bufferPosition.Value.Snapshot;
                     int caretPosition = bufferPosition.Value.Position;
                     ITextSnapshotLine line = snapshot.GetLineFromPosition(caretPosition);
 
-                    if (line.Length < 8 || caretPosition < line.Start + 8 || snapshot[caretPosition - 1] != '(')
-                    {
+                    if (line.Length < 8 || caretPosition < line.Start + 8 || snapshot[caretPosition - 1] != '(') {
```
int19h: Possibly another one to file for later - looks like this won't handle something like:
```r
library(
  foo
```

src/R/Editor/Impl/Documentation/RoxygenBlock.cs
```diff
@@ -0,0 +1,106 @@
...
+                        sb.Append(Invariant($"#' @param {p.Name}{lineBreak}"));
+                    }
+                }
+
+                if (sb.Length > length) {
+                    sb.Append(Invariant($"#'{lineBreak}"));
+                }
+
+                sb.Append(Invariant($"#' @return{lineBreak}"));
+                sb.Append(Invariant($"#' @export{lineBreak}"));
```
int19h: Should we always be inserting `@export`? My understanding is that it will actually translate to the function being exported from the package (i.e. it's more than documentation); but are all functions with Roxygen comments normally exported?

int19h:
:+1:

src/R/Editor/Impl/Documentation/RoxygenBlock.cs
MikhailArkhipov: Frankly I am fine either way. I modeled the block off what RS does.

PR #1359, opened 2016-03-26
Fix #915: Allow user to specify CRAN mirror via .Rprofile

### LINKED ISSUE #915: Allow user to specify CRAN mirror via .Rprofile
Instead of assuming RStudio mirror

int19h:
Provide a way to not set any CRAN mirror via R Tools | Options at all, using whatever is set by R itself (e.g. via .Rprofile or other startup code) instead.

### DIFF
--- a/src/Package/Impl/Options/R/RToolsOptionsPage.cs
+++ b/src/Package/Impl/Options/R/RToolsOptionsPage.cs
@@ -38,7 +38,7 @@ namespace Microsoft.VisualStudio.R.Package.Options.R {
         [CustomLocDisplayName("Settings_CranMirror")]
         [LocDescription("Settings_CranMirror_Description")]
         [TypeConverter(typeof(CranMirrorTypeConverter))]
-        [DefaultValue("0-Cloud [https]")]
+        [DefaultValue(null)]
         public string CranMirror {
             get { return RToolsSettings.Current.CranMirror; }
             set { RToolsSettings.Current.CranMirror = value; }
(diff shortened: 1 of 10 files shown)


src/Package/Impl/Options/R/RToolsOptionsPage.cs
```diff
@@ -38,7 +38,7 @@ public class RToolsOptionsPage : DialogPage {
         [CustomLocDisplayName("Settings_CranMirror")]
         [LocDescription("Settings_CranMirror_Description")]
         [TypeConverter(typeof(CranMirrorTypeConverter))]
-        [DefaultValue("0-Cloud [https]")]
+        [DefaultValue(null)]
```
AlexanderSher: What will happen if we there is no .RProfile and we don't specify CRAN mirror?

int19h: Then it will use whatever is the default for that R implementation. On vanilla R, if you ask, it says `@CRAN@`, which is a special value that package fetching code expands to "http://CRAN.R-project.org" when it detects it. I think the point of that indirection is so that if you use the default, new versions of R can change it, and that change will be automatically applied.

MikhailArkhipov: If nothing is specified, default was that ugly CRAN selection dialog. That's why we started setting default mirror in the first place. So if nothing is specified we should still set something otherwise the dialog will be back.

int19h: Ah, I see. I'll add a check for `@CRAN@` then.

int19h:
Changed to never allow `@CRAN@`, but otherwise respect settings if users chooses to.

MikhailArkhipov:
:+1:

PR #1366, opened 2016-03-29
one dimensional data.frame doesn't return row names

JeongPyoHam:


src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -20,12 +20,13 @@ grid.data <- function(x, rows, cols) {
...
-  if (is.matrix(x)) {
-    if ((length(rows) == 1) || (length(cols) == 1)) {
-      data <- grid.format(x[rows, cols]);
-    } else {
-      data <- sapply(as.data.frame(x[rows, cols]), grid.format, USE.NAMES=FALSE);
-    }
+  onedim <- ((length(rows) == 1) || (length(cols) == 1));
+  if (onedim) {
+    data <- grid.format(x[rows, cols]);
+    rn <- row.names(x)[rows];
```
int19h: This line, and the one below, seem to be identical between the two branches - can you refactor that out to avoid repeating it?

int19h:
:+1:

PR #1367, opened 2016-03-29
Add functionality to fetch additional package info from a web page.

huguesv:
Add missing fields to RPackage.

### DIFF
--- a/src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
@@ -35,15 +35,12 @@ namespace Microsoft.R.Components.PackageManager.Implementation {
             }
 
             var pkg = result ?? new RPackage();
-            try {
-                var tree = new HtmlTree(new Microsoft.Web.Core.Text.TextStream(content));
-                tree.Build();
 
-                var search = new IndexVisitor(pkg);
-                tree.Accept(search, null);
-            } catch (Exception ex) when(!ex.IsCriticalException()) {
-                throw new RPackageInfoRetrievalException(string.Format("Error parsing index page from {0}", packageUri), ex);
-            }
+            var tree = new HtmlTree(new Microsoft.Web.Core.Text.TextStream(content));
+            tree.Build();
+
+            var search = new IndexVisitor(pkg);
+            tree.Accept(search, null);
 
             return pkg;
         }
--- a/src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
@@ -1,0 +1,217 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Common.Core;
+using Microsoft.Html.Core.Tree;
+using Microsoft.Html.Core.Tree.Nodes;
+using Microsoft.R.Components.PackageManager.Model;
+using Microsoft.R.Components.Test.PackageManager;
+
+namespace Microsoft.R.Components.PackageManager.Implementation {
+    internal class RPackageWebParser {
+        /// <summary>
+        /// Downloads the index html page specified, and extracts a set of known fields from it.
+        /// </summary>
+        /// <param name="packageUri">Location of index.html page.</param>
+        /// <param name="result">RPackage object to fill in. A new one is created if none is passed in.</param>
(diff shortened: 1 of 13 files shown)


src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// won't be overwritten by the ones extracted from the html page.
+        /// </remarks>
+        public static async Task<RPackage> RetrievePackageInfo(Uri packageUri, RPackage result = null) {
+            string content;
+            try {
+                using (var webClient = new WebClient()) {
+                    content = await webClient.DownloadStringTaskAsync(packageUri);
+                }
+            } catch (WebException ex) {
+                throw new RPackageInfoRetrievalException(string.Format("Error reading index page from {0}", packageUri), ex);
```
AlexanderSher: Who is going to catch this exception? RPackageManager or view model?

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
```diff
@@ -0,0 +1,217 @@
...
+            public const string License = "License";
+            public const string Version = "Version";
+            public const string NeedsCompilation = "NeedsCompilation";
+            public const string Author = "Author";
+            public const string URL = "URL";
+            public const string LinkingTo = "LinkingTo";
+            public const string Enhances = "Enhances";
+            public const string Maintainer = "Maintainer";
+            public const string BugReports = "BugReports";
+            public const string Published = "Published";
```
AlexanderSher: Looks like you can use `nameof` for every field

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
huguesv: Intent was that the caller of RPackageManager, so yes probably view model.  The alternative was that RPackageManager catches it, but then view model wouldn't know things went wrong and would have no way of indicating that to the user.

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
```diff
@@ -0,0 +1,217 @@
...
+                if (string.IsNullOrEmpty(_pkg.Description)) {
+                    _pkg.Description = GetElementText(paragraphNode);
+                }
+            }
+
+            private void ParseGenericField(ElementNode tdNode1, ElementNode tdNode2) {
+                string fieldName = GetElementText(tdNode1).Trim(':');
+                string fieldValue = GetElementText(tdNode2);
+
+                switch (fieldName) {
```
AlexanderSher: I don't like this switch, but can't think of anything better :)

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
MikhailArkhipov: Do we need to show some sort of progress and or cancellation? If connection is offline, timeout, I think, is 30 seconds. Or maybe check connection before trying actual download?

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            var pkg = result ?? new RPackage();
+            try {
+                var tree = new HtmlTree(new Microsoft.Web.Core.Text.TextStream(content));
+                tree.Build();
+
+                var search = new IndexVisitor(pkg);
+                tree.Accept(search, null);
+            } catch (Exception ex) when(!ex.IsCriticalException()) {
```
MikhailArkhipov: There is nothing that is supposed to throw there. The parser is the editor-style recoverable parser so it handles broken HTML.

MikhailArkhipov: Although resulting tree may not be what you'd expect. The parser does not do syntax check, in  HTML editor there is a separate validator module.

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
huguesv: I'll try with reflection, and see if performance is not too affected.

MikhailArkhipov: Nah, leave that alone IMO

src/R/Components/Impl/PackageManager/Implementation/RPackageWebParser.cs
huguesv: I'll remove the try/catch then.

MikhailArkhipov:
:+1:

PR #1368, opened 2016-03-29
Fix #1350: Improve toJSON performance for package manager and other uses

### LINKED ISSUE #1350: Improve toJSON performance for package manager and other uses

int19h:
Migrate to the new C++ R-to-JSON serializer implementation that is now baked into R-Host.

Various tweaks to better handle invalid or corner case data from host - null checks, size checks etc.

Rename `setwd` protocol message to avoid confusion with `~+` and `~-`.


int19h:
See https://github.com/Microsoft/R-Host/pull/50 for host-side changes that are utilized here.

int19h:
I'm going to add some more tests specifically for the JSON serializer, but please look at the code in the meantime.

huguesv:
:+1:  once tests are added

int19h:
Here are the tests.

src/Debugger/Test/JsonTest.cs
```diff
@@ -41,12 +44,47 @@ public class JsonTest : IAsyncLifetime {
...
+        [InlineData("list(NA, 1, NA, 2, NA)", "[1,2]")]
+        [InlineData("structure(list(), names = ''[FALSE])", "{}")]
+        [InlineData("list(n = 0, s = 's', u = NULL)", @"{""n"":0,""s"":""s"",""u"":null}")]
+        [InlineData("list(n = 0, na = NA)", @"{""n"":0}")]
+        [InlineData("list(n = 0, na = NA)", @"{""n"":0}")]
+        [InlineData("as.environment(list())", "{}")]
+        [InlineData("as.environment(list(n = 0, s = 's', u = NULL))", @"{""n"":0,""s"":""s"",""u"":null}")]
+        [InlineData("as.environment(list(n = 0, na = NA))", @"{""n"":0}")]
+        [InlineData("as.environment(list(n = 0, na = NA))", @"{""n"":0}")]
+        [InlineData("list(as.environment(list(l = list())))", @"[{""l"":[]}]")]
```
AlexanderSher: Are there any tests for invalid json?

huguesv: It would be good to test some known unsupported scenarios, to make sure an error is raised.  Such as: matrix, data frame, list where some elements don't have name, vector with more than 1 element

int19h: Not yet. It's a bit tricky to write a test for it, because invalid JSON is treated as a fatal error inside R-Host, which means that it immediately terminates the host process. I'm not sure what is the proper way to detect this in the test.

int19h: (Also, fatal process termination means a crash dump is written out to %TEMP%, so even if a test can detect termination, %TEMP% will fill up pretty soon with those crash dumps.)

huguesv: Oh right, I remember this from when I tried to add negative tests for plots.  It's fine for now.

huguesv:
:+1:

AlexanderSher:
:+1:

src/Debugger/Test/JsonTest.cs
int19h: I suppose a proper fix would be to rewrite `to_json` so that it doesn't treat errors as fatal and just reports them. `host.cpp` can still treat an error that bubbles up as fatal at the point where it invokes it, but  invoking `toJSON` in R will then simply report the error as usual, and it can be used to test them.

Problem is, it's very tricky to write C++ code through which R errors can flow through. We have helpers for it that translate R errors to exceptions and vice versa; but because the former requires establishing a fresh R context, it's expensive to do in a tight loop, which is what JSON serializer is. I'll need to look closely at the code to see how this can be done safely.

int19h: I'll go ahead with what is here for now, so that we have the code in the product and can start bashing on it; meanwhile, I will see what can be done about negative tests in another PR.

PR #1383, opened 2016-03-30
R evaluation API unification

int19h:
R evaluation API unification - everything that can eval R code now implements `IRExpressionEvaluator`.

Require `REvaluationKind` to be specified to avoid bugs when not-mutating eval is inappropriate, and force callers to specify it explicitly either way. Fix existing callers accordingly.

Remove `isMutating` parameter from `BeginEvaluationAsync` - all evaluations now begin as non-mutating, and change if `REvaluationKind.Mutating` is used on any `EvaluateAsync` call for that evaluation.

### DIFF
--- a/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
+++ b/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
@@ -8,43 +8,43 @@ using Microsoft.R.Host.Client;
 
 namespace Microsoft.R.Host.Client.Session {
     public static class RSessionEvaluationCommands {
-        public static Task OptionsSetWidth(this IRSessionEvaluation evaluation, int width) {
-            return evaluation.EvaluateNonReentrantAsync($"options(width=as.integer({width}))\n");
+        public static Task OptionsSetWidth(this IRExpressionEvaluator evaluation, int width) {
+            return evaluation.EvaluateAsync($"options(width=as.integer({width}))\n", REvaluationKind.Normal);
         }
 
-        public static async Task<string> GetRUserDirectory(this IRSessionEvaluation evaluation) {
-            var result = await evaluation.EvaluateAsync("Sys.getenv('R_USER')");
+        public static async Task<string> GetRUserDirectory(this IRExpressionEvaluator evaluation) {
+            var result = await evaluation.EvaluateAsync("Sys.getenv('R_USER')", REvaluationKind.Normal);
             return result.StringResult.Replace('/', '\\');
         }
 
-        public static async Task<string> GetWorkingDirectory(this IRSessionEvaluation evaluation) {
-            var result = await evaluation.EvaluateAsync("getwd()");
+        public static async Task<string> GetWorkingDirectory(this IRExpressionEvaluator evaluation) {
+            var result = await evaluation.EvaluateAsync("getwd()", REvaluationKind.Normal);
             return result.StringResult.Replace('/', '\\');
         }
 
-        public static Task SetWorkingDirectory(this IRSessionEvaluation evaluation, string path) {
-            return evaluation.EvaluateNonReentrantAsync($"setwd('{path.Replace('\\', '/')}')\n");
+        public static Task SetWorkingDirectory(this IRExpressionEvaluator evaluation, string path) {
+            return evaluation.EvaluateAsync($"setwd('{path.Replace('\\', '/')}')\n", REvaluationKind.Normal);
         }
 
-        public static Task SetDefaultWorkingDirectory(this IRSessionEvaluation evaluation) {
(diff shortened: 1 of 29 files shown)


int19h:
Note that there are two commits here. First one is the minimal unification of existing implementation - it does not change the semantics of any existing `BeginEvaluationAsync` or `EvaluateAsync` calls. However, it effectively ignores `REvaluationKind.Mutated` for individual evals when `BeginEvaluationAsync` is used.

The second is an attempt to clean that up for a more sensible API - there's no evaluation flag for the entire batch now, but instead it is tracked according to flags specified by individual evals (i.e. if any of them is `REvaluationKind.Mutated`, then the entire batch is marked as mutating, and the event is raised once it's complete).

src/Host/Client/Impl/Session/RSessionEvaluationSource.cs
```diff
@@ -10,19 +10,17 @@ namespace Microsoft.R.Host.Client.Session {
...
-            IsMutating = isMutating;
         }
 
         public Task<IRSessionEvaluation> Task => _tcs.Task;
 
-        public Task BeginEvaluationAsync(IReadOnlyList<IRContext> contexts, IRExpressionEvaluator evaluator, CancellationToken ct) {
+        public async Task<bool> BeginEvaluationAsync(IReadOnlyList<IRContext> contexts, IRExpressionEvaluator evaluator, CancellationToken ct) {
             var evaluation = new RSessionEvaluation(contexts, evaluator, ct);
-            return _tcs.TrySetResult(evaluation) ? evaluation.Task : System.Threading.Tasks.Task.CompletedTask;
+            await (_tcs.TrySetResult(evaluation) ? evaluation.Task : System.Threading.Tasks.Task.CompletedTask);
```
AlexanderSher: You need to `await` only `evaluation.Task`:
```{csharp}
if (_tcs.TrySetResult(evaluation)) {
    await evaluation.Task;
}
```

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -72,66 +72,62 @@ public static class RSessionEvaluationCommands {
...
-        public static Task<REvaluationResult> AvailablePackages(this IRSessionEvaluation evaluation) {
+        public static Task<REvaluationResult> AvailablePackages(this IRExpressionEvaluator evaluation) {
             var script = @"rtvs:::packages.available()";
             return evaluation.EvaluateAsync(script, REvaluationKind.Json);
         }
 
-        public static Task<REvaluationResult> InstallPackage(this IRSessionEvaluation evaluation, string packageName) {
+        public static Task<REvaluationResult> InstallPackage(this IRExpressionEvaluator evaluation, string packageName) {
             var script = string.Format("install.packages(\"{0}\")", packageName);
             return evaluation.EvaluateAsync(script, REvaluationKind.Json);
```
AlexanderSher: Probably mutating

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -8,43 +8,43 @@
 
 namespace Microsoft.R.Host.Client.Session {
     public static class RSessionEvaluationCommands {
-        public static Task OptionsSetWidth(this IRSessionEvaluation evaluation, int width) {
-            return evaluation.EvaluateNonReentrantAsync($"options(width=as.integer({width}))\n");
+        public static Task OptionsSetWidth(this IRExpressionEvaluator evaluation, int width) {
+            return evaluation.EvaluateAsync($"options(width=as.integer({width}))\n", REvaluationKind.Normal);
```
AlexanderSher: Probably mutating

src/Host/Client/Impl/Session/RSessionEvaluationSource.cs
int19h: Indeed, dumb code translation is dumb. Fixed.

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
int19h: And I think it's not `Json`, since it calls a built-in function? No-one calls it, so it's hard to tell.

AlexanderSher: @huguesv, do you need this method?

huguesv: Not needed anymore.  Current plan is to send install.packages code to repl, to get all output there, like source and debug_source do.

AlexanderSher:
:+1:

huguesv:
:+1:

PR #1384, opened 2016-03-30
Fix build.

huguesv:


JeongPyoHam:
Great change. :+1:

JeongPyoHam:
pressed wrong button. :)

Issue #1381, opened 2016-03-30
Snippets should have description tooltips


MikhailArkhipov:
Some don't display any (ggplot for example)

MikhailArkhipov:
In the completion list that is

Issue #1380, opened 2016-03-30
snippets in repl: highlights off by a few chars when done in REPL


smortaz:
try the # section ---- one for example.  you'll see that a + appears in the REPL and as a result the highlighted area is off.

Issue #1379, opened 2016-03-30
Exception in suggested actions (log attached)


MikhailArkhipov:
[ActivityLog.txt](https://github.com/Microsoft/RTVS/files/194632/ActivityLog.txt)

Issue #1365, opened 2016-03-28
Surround With... does not work properly


MikhailArkhipov:
It inserts the construct AFTER the text rather than around it.

MikhailArkhipov:
Actually it works OK, but Surround With filtering was missing

PR #1388, opened 2016-03-30
#1355 #1361 #1279 #1360 #1365 #1381 #1379 #1380 #1308

MikhailArkhipov:
#1355: Ctrl+F3 should be selecting identifier
#1279 Double-clicking in the REPL doesn't highlight words 
- Implemented R-specific TextStructureNavigator for the editor and REPL

#1361 äöü characters breaks formatting
- Changed IsAnsiLetter to IsLetter in tokenizer

#1360 Review variable handling code to make sure it properly handles backticks
- User tokenizer rather than text walk to get identifiers

#1365 Surround With... does not work properly 
- Added filter for the SurroundWith snippet property to ExpansionClient

#1381 Snippets should have description tooltips 
- Propagate snippet descriptions to the completion set

#1379 Exception in suggested actions (log attached) 
- Added checks to brace highlighter and suggested actions

#1380 Snippets in repl: highlights off by a few chars when done in REPL 
- Fix snippet avoiding leading line break

#1308 Review full list of popular shortcuts, see if we can match more 
- Introduced command context for all our tool windows
- Specified key bindings for tool windows, solution explorer and REPL
- Less invasive way to take over Ctrl+Enter and Ctrl+L in global
- Shortcuts now appear in the menus
- Matched default shortcuts to RS
- Commented out RS shortcuts in DS settings import since they are now default

Suppressed automatic completion inside identifiers (used to come up inside backticks)
Added 'Insert Roxygen Block' to context menu
Fixed small issue when smart indent after x <- function() { }ENTER thought it was still inside scope

Optimized AST invalidate code so it invalidates and removes nodes up to nearest well formed scope. This way AST remains usable for smart indenter and similar operations until the next parse pass and we don't need to stash the previous tree snapshot. This will also allow eventually to add incremental parse and AST update since we can just reparse the damaged scope and plant new nodes into the existing AST.

Many new tests for the above


src/R/Editor/Impl/Selection/SelectWordCommand.cs
```diff
@@ -24,60 +22,13 @@ public sealed class SelectWordCommand : ViewCommand {
         public override CommandResult Invoke(Guid group, int id, object inputArg, ref object outputArg) {
```
int19h: So they don't automatically use `ITextStructureNavigator` in this case? Perhaps we should file a bug on VS so that they respect it in selection code.

src/Languages/Core/Impl/Text/CharacterStream.cs
```diff
@@ -207,15 +207,15 @@ public CharacterStream(string text)
         /// <summary>
         /// Determines if current character is an ANSI letter
```
int19h: This comment, and the one below, are wrong now, since it checks for a Unicode letter.

int19h:
Do we really want RS shortcuts by default, even though they take over some VS shortcuts?

src/Package/Impl/Repl/Commands/ReplShortcutSetting.cs
```diff
@@ -2,91 +2,39 @@
...
-using Microsoft.R.Editor.Settings;
 using Microsoft.VisualStudio.R.Package.Shell;
 
 namespace Microsoft.VisualStudio.R.Package.Repl.Commands {
     internal static class ReplShortcutSetting {
-        private const string CommandName = "EditorContextMenus.CodeWindow.ExecuteLineInInteractive";
-
-        private static bool _currentSetting;
+        private static Dictionary<string, string> _editorShortcutRemaps = new Dictionary<string, string>() {
+            { "Edit.LineOpenAbove", "Text Editor::ctrl+k,ctrl+o" }, // Make Ctrl+Enter work in R Editor
```
int19h: What if this command wasn't actually bound to Ctrl+Enter?

src/R/Editor/Impl/Selection/SelectWordCommand.cs
MikhailArkhipov: They do, but default one works for languages like C# and C++ (i.e. where dot is a separator).

src/R/Editor/Application.Test/Documentation/SelectionTest.cs
```diff
@@ -0,0 +1,43 @@
...
+using FluentAssertions;
+using Microsoft.R.Components.ContentTypes;
+using Microsoft.R.Editor.Application.Test.TestShell;
+using Microsoft.UnitTests.Core.XUnit;
+using Xunit;
+
+namespace Microsoft.R.Editor.Application.Test.Selection {
+    [ExcludeFromCodeCoverage]
+    [Collection(CollectionNames.NonParallel)]
+    public class DocumentatonTest {
```
int19h: Nit: "Documentation" (spelling)

src/Package/Impl/Repl/Commands/ReplShortcutSetting.cs
MikhailArkhipov: It is interesting question. I can try and check what is current setting (maybe Vim extensions moves it...)

I have a thread with IDE folks and it looks like core editor ignores narrower definitions and still uses its own :-(. I am going to open a bug on this for them in Dev15.

src/R/Editor/Impl/ContentType/TextViewExtensions.cs
```diff
@@ -74,63 +70,107 @@ public static class TextViewExtensions {
...
 
-            span = Span.FromBounds(0, 0);
-            return string.Empty;
+        private static bool IsVariableNameToken(string lineText, RToken token) {
+            if (token.TokenType == RTokenType.Identifier) {
+                return true;
+            }
+            if (token.TokenType == RTokenType.Operator) {
+                if (token.Length == 1) {
+                    return lineText[token.Start] == '$' || lineText[token.Start] == '@';
```
int19h: `::`?

src/R/Editor/Impl/Navigation/Text/RTextSelector.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Linq;
+using Microsoft.R.Core.Tokens;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.R.Editor.Navigation.Text {
+    internal static class RTextStructure {
+        public static Span? GetWordSpan(ITextSnapshot snapshot, int position) {
```
int19h: This could use some tests.

src/Package/Impl/Repl/Commands/ReplShortcutSetting.cs
MikhailArkhipov: As for RS defaults, it appears that typical users want familiar ones. Also, quite a few didn't find DS profile.

Generally we only assign them to new commands except Ctrl+Enter and Ctrl+L so it doesn't really matter which shortcuts to use. So we leave making custom setting to advanced folks who can create their own profile.

src/R/Editor/Impl/Selection/SelectWordCommand.cs
int19h: Right, but we're providing our own navigator now. So do we still need to override that command, or can we just let the default implementation of the command do its thing (and use our navigator)?

src/Package/Impl/Repl/Commands/ReplShortcutSetting.cs
int19h: It feels like we're removing choice here, especially for those who want to use VS as a true multi-language IDE, and not just for R. It will also make it more complicated to unify with PTVS in one box later, because shortcuts are all over the place (whereas our default shortcuts before this change were more in line with VS and PTVS practice). 

Also, I feel like we might be hearing negative feedback from those who didn't find the shortcuts, but we are not hearing positive feedback from those who did (and who might like/prefer them), simply because they don't see any point in spelling it out. 

Either way, whatever is the default, I think we do need a profile for multi-language VS use that is consistent with VS (generic profile) rather than RS.

src/R/Editor/Impl/ContentType/TextViewExtensions.cs
MikhailArkhipov: Sure, I assume eval is fine with fully qualified

src/R/Editor/Impl/Selection/SelectWordCommand.cs
MikhailArkhipov: We still call it on double click since that thing does not call structure navigator...

int19h: Ah, okay, so I did understand it right. I think they should be doing that (or at least it's worth trying to convince them that they should).

src/R/Editor/Impl/Navigation/Text/RTextSelector.cs
MikhailArkhipov: Will do

src/R/Editor/Impl/Selection/SelectWordCommand.cs
MikhailArkhipov: Yeah, will open a bug on that

src/Package/Impl/Repl/Commands/ReplShortcutSetting.cs
MikhailArkhipov: Apart from Ctrl+Enter and Ctrl+L we only setting shortcuts for our own commands so it doesn't matter what they are - I mean, we are not changing other shortcuts globally. all the shortcuts are set in R Editor, and R Tool windows. Even show solution explorer is out own command so official one in View is unaffected. 4 or 5 combos we specify affect 4-5 obscure commands and only in specific environment. Someone working on C# or C++ won't be affected in any significant way. Also, users familiar with VS can always remap couple of commands in case they are unhappy - they know how to do it. Data folks are not familiar with VS customization on the other hand.

int19h:
:+1:

PR #1389, opened 2016-03-31
Package manager backend

huguesv:
Add load, unload, get loaded packages.
Add install, uninstall package.
Add get library paths.

### DIFF
--- a/src/Debugger/Impl/rtvs/R/packages.R
+++ b/src/Debugger/Impl/rtvs/R/packages.R
@@ -2,18 +2,30 @@
 # Licensed under the MIT License. See LICENSE in the project root for license information.
 
 matrix.as.lists <- function(matrix.data) {
-    res <- list()
-    dn <- dimnames(matrix.data)
+    res <- base::list()
+    dn <- base::dimnames(matrix.data)
     for (r in dn[[1]]) {
-        res[[r]] <- as.list(matrix.data[r,])
+        res[[r]] <- base::as.list(matrix.data[r,])
     }
     res
 }
 
 packages.installed <- function() {
-    matrix.as.lists(installed.packages(fields = c('Title', 'Author')))
+    matrix.as.lists(utils::installed.packages(fields = c('Title', 'Author')))
 }
 
 packages.available <- function() {
-    matrix.as.lists(available.packages())
+    matrix.as.lists(utils::available.packages())
+}
+
+packages.loaded <- function() {
+    prefix <- "package:"
+    prefix.length <- base::nchar(prefix)
+    loaded <- base::search()
+    pkgs <- base::substring(loaded[base::substring(loaded, 1, prefix.length) == prefix], prefix.length + 1)
+    base::as.list(pkgs)
+}
+
+packages.libpaths <- function() {
+    base::as.list(base::.libPaths())
 }
--- a/src/Debugger/Impl/rtvs/R/packages.R
+++ b/src/Debugger/Impl/rtvs/R/packages.R
@@ -19,11 +19,7 @@ packages.available <- function() {
 }
 
 packages.loaded <- function() {
-    prefix <- "package:"
-    prefix.length <- base::nchar(prefix)
-    loaded <- base::search()
-    pkgs <- base::substring(loaded[base::substring(loaded, 1, prefix.length) == prefix], prefix.length + 1)
-    base::as.list(pkgs)
+    base::as.list(base::.packages())
 }
 
 packages.libpaths <- function() {
--- a/src/Debugger/Impl/rtvs/R/packages.R
+++ b/src/Debugger/Impl/rtvs/R/packages.R
@@ -2,26 +2,26 @@
 # Licensed under the MIT License. See LICENSE in the project root for license information.
 
 matrix.as.lists <- function(matrix.data) {
-    res <- base::list()
-    dn <- base::dimnames(matrix.data)
+    res <- list()
+    dn <- dimnames(matrix.data)
     for (r in dn[[1]]) {
-        res[[r]] <- base::as.list(matrix.data[r,])
+        res[[r]] <- as.list(matrix.data[r,])
     }
     res
 }
 
 packages.installed <- function() {
-    matrix.as.lists(utils::installed.packages(fields = c('Title', 'Author')))
+    matrix.as.lists(installed.packages(fields = c('Title', 'Author')))
 }
 
 packages.available <- function() {
-    matrix.as.lists(utils::available.packages())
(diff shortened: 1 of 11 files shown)


src/Debugger/Impl/rtvs/R/packages.R
```diff
@@ -2,18 +2,30 @@
 # Licensed under the MIT License. See LICENSE in the project root for license information.
 
 matrix.as.lists <- function(matrix.data) {
-    res <- list()
-    dn <- dimnames(matrix.data)
+    res <- base::list()
```
int19h: Did it clash with something in the package?

src/Debugger/Impl/rtvs/R/packages.R
```diff
@@ -2,18 +2,30 @@
...
 packages.available <- function() {
-    matrix.as.lists(available.packages())
+    matrix.as.lists(utils::available.packages())
+}
+
+packages.loaded <- function() {
+    prefix <- "package:"
+    prefix.length <- base::nchar(prefix)
+    loaded <- base::search()
+    pkgs <- base::substring(loaded[base::substring(loaded, 1, prefix.length) == prefix], prefix.length + 1)
```
int19h: It looks like this could be replaced with `.packages()`?

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -47,6 +47,87 @@ internal class RPackageManager : IRPackageManager {
             await RPackageWebParser.RetrievePackageInfo(uri, pkg);
         }
 
+        public void InstallPackage(string name, string libraryPath) {
+            string script;
+            if (string.IsNullOrEmpty(libraryPath)) {
+                script = string.Format("install.packages(\"{0}\")", name);
+            } else {
+                script = string.Format("install.packages(\"{0}\", lib=\"{1}\")", name, libraryPath.ToRPath());
```
int19h: All the string literals inserted here and below should use `ToRStringLiteral` to ensure proper quoting for any kinds of weird names.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -47,6 +47,87 @@ internal class RPackageManager : IRPackageManager {
...
+                script = string.Format("library(\"{0}\")", name);
+            } else {
+                script = string.Format("library(\"{0}\", lib.loc=\"{1}\")", name, libraryPath.ToRPath());
+            }
+
+            _interactiveWorkflow.Operations.EnqueueExpression(script, true);
+        }
+
+        public void UnloadPackage(string name) {
+            string script = string.Format("detach(\"package:{0}\", unload=TRUE)", name);
```
int19h: `unloadNamespace`?

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -47,6 +47,87 @@ internal class RPackageManager : IRPackageManager {
...
+
+            _interactiveWorkflow.Operations.EnqueueExpression(script, true);
+        }
+
+        public async Task<string[]> GetLoadedPackagesAsync() {
+            if (!_interactiveWorkflow.RSession.IsHostRunning) {
+                return new string[0];
+            }
+
+            try {
```
int19h: The stuff inside here may also throw `MessageTransportException` - is the intent for the caller to handle that?

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -179,6 +179,145 @@ public class PackageManagerIntegrationTest : IDisposable, IAsyncLifetime {
...
+                _workflow.RSession.IsHostRunning.Should().BeTrue();
+
+                using (var eval = await _workflow.RSession.BeginEvaluationAsync()) {
+                    await SetLocalRepoAsync(eval, _repo1Path);
+                    await SetLocalLibAsync(eval, _libPath);
+                    await InstallPackageAsync(eval, TestPackages.RtvsLib1Description.Package, _libPath);
+                }
+
+                string[] results = await _workflow.Packages.GetLoadedPackagesAsync();
+                results.Should().NotContain("rtvslib1");
```
int19h: There are overloads of `Contain` and `NotContain` that take an enumerable, and check for all elements in it (order doesn't matter, so it's not a subsequence match, just a subset check).

src/Debugger/Impl/rtvs/R/packages.R
huguesv: Just trying to preemptively prevent problems if user masks some of the functions we are using.  Do you think it's overkill?

int19h: You don't need to worry about it in a package, because each package gets its own, completely separate search path, and user's global environment is not on that path. So if they do something like `c <- 1`, they will break themselves, but not us.

It's only a concern for code that goes directly into an eval, because that runs in the global environment by default, and so we've had those issues with `c` (and I started pre-emptively qualifying everything else there to prevent future bugs of the same nature).

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
huguesv: I haven't convinced myself yet as to which exceptions should propagate out of the RPackageManager.  I'm leaning towards catching these low level exceptions and wrapping them in high level package mgr exception(s) that the caller (UI) can handle.  Right now some of the methods are returning empty data/lists in the case of errors, and that prevents the UI from informing the user that something was wrong (it could assume that's the case if it gets empty data, but an exception is clearer).

int19h: @AlexanderSher, what's your preference?

AlexanderSher: I have a message bar in the package manager that can show errors, but I haven't wired it to any error handling yet. Shouldn't be hard though. 
@huguesv, do you have a list of expected exceptions?

huguesv:
All feedback addressed.  Exceptions thrown by pkg mgr are documented in IPackageManager.cs for each function.  It's basically RPackageManagerException, and OperationCanceledException that must be handled by UI.  UI can display ex.Message to the user.

huguesv:
Also I added a test where a package is installed into multiple repositories.  R properly filters duplicates, ie it only lists the package that will actually get loaded by `library`, based on the ordering of the 'repos' option.

int19h:
:+1:

Issue #1377, opened 2016-03-29
Call stack in VE is wrong


int19h:
Run this (no need to attach):
```r
f <- function(x) {
    browser()
    return(x)
}

g <- function(x) f(x)

print(g(x))
```

Once stopped at `browser()`, expand the environment list in VE. It has a frame for `f`, but not for `g`, even though the call chain at this point is global -> `g` -> `f`.

Another, somewhat more convoluted case, which explains the problem better:
```r
f <- function(x) {
    g <- function(y) {
        browser()
        x + y
    }
    return(g)
}

g <- f(1)
print(g(2))
```

Expected behavior is to have two stack frames there, one for global, and one for `g`. But there are three frames - there's an unnamed frame (shows as a blank line) in between global and `g`. When that frame is selected, the following error message appears:

![image](https://cloud.githubusercontent.com/assets/1175142/14124899/0dec8d92-f5bd-11e5-8063-6f47ad190921.png)

int19h:
Basically, the environment selector starts with the current frame, and then walks the `parent.env` chain - but parent of the function environment is not the environment from which the call to that function was made, but rather the environment in which the original `function(...) ...` expression that created the function was evaluated (in other words, it's the lexically enclosing environment).

So in the first case, `parent.env(f)` is `.GlobalEnv`, rather than the environment of `g`, and so `g` is skipped.

And int he second case, `parent.env(g)` is `f`, because `g` was created inside of `f`; but the call is directly from `.GobalEnv` to `g`, so the parent chain is one element longer than the call chain, which accounts for that empty element in the list.

PR #1394, opened 2016-03-31
Do not add environment with empty name

JeongPyoHam:


int19h:
Is this a fix for #1377?

JeongPyoHam:
yes, master

PR #1395, opened 2016-04-01
Package manager UI

AlexanderSher:
This PR doesn't address @huguesv's changes from yesterday, so "avaliable packages" tab and install/uninstall buttons aren't working yet.
packagemanager


src/Host/Process
```diff
@@ -1 +1 @@
-Subproject commit d9b4113a68337fdd03034423d222da3676790837
```
MikhailArkhipov: Probably not needed

src/Languages/Editor/Test/Shell/TestShellBase.cs
```diff
@@ -39,6 +39,10 @@ public class TestShellBase {
             await UIThreadHelper.Instance.InvokeAsync(action);
         }
 
+        public async Task<TResult> DispatchOnMainThreadAsync<TResult>(Func<TResult> action, CancellationToken cancellationToken = new CancellationToken()) {
```
MikhailArkhipov: Technically DispatchOnUIThread was supposed to be async (it used to be Dispatcher.BeginInvoke long time ago). Code we inherited actually expects DispatchOnUIThread  to be non-blocking.

src/Package/Impl/Search/VsSearchControl.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        private void SetIfSpecified(SearchSettingsDataSource vsSettings, GelProperty vsPropertyName, string propertyName) {
+            object value;
+            if (_settings.TryGetValue(propertyName, out value)) {
+                vsSettings.SetValue(vsPropertyName, value);
+            }
+        }
+
+        public bool OnNavigationKeyDown(uint dwNavigationKey, uint dwModifiers) => false;
+        public bool SearchEnabled => true;
```
MikhailArkhipov: OK to search while packages being installed or removed?

src/Package/Impl/Search/VsSearchTask.cs
```diff
@@ -0,0 +1,99 @@
...
+            ErrorCode = 0;
+
+            _searchCallback = pSearchCallback;
+            _handler = handler;
+            _cts = cts;
+            _taskStatus = (long)Created;
+        }
+
+        public void Start() {
+            if (!SetTaskStatus(Started)) {
```
MikhailArkhipov: Nit: Fewer lines if you reverse if

src/Package/Impl/Search/VsSearchTask.cs
```diff
@@ -0,0 +1,99 @@
...
+                return;
+            }
+
+            _handler.Search(SearchQuery.SearchString, _cts.Token).ContinueWith(SearchCompleted);
+        }
+
+        public void Stop() {
+            if (!SetTaskStatus(Stopped)) {
```
MikhailArkhipov: As above

src/Package/Impl/Search/VsSearchTask.cs
```diff
@@ -0,0 +1,99 @@
...
+            switch (task.Status) {
+                case TaskStatus.RanToCompletion:
+                    _searchCallback.ReportComplete(this, (uint)task.Result);
+                    break;
+                case TaskStatus.Faulted:
+                case TaskStatus.Canceled:
+                    _searchCallback.ReportComplete(this, 0);
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
MikhailArkhipov: In private method if this hits it appears to be our own bug... Perhaps assert is more appropriate?

src/Package/Impl/source.extension.vsixmanifest
```diff
@@ -17,6 +17,7 @@
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Core.dll" />
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Actions.dll" />
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Debugger.Engine.dll" />
+    <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Wpf.dll" />
```
MikhailArkhipov: It doesn't seem to reference more than Common.Wpf so do we need it separate?

src/Host/Process
AlexanderSher: facepalm

src/Languages/Editor/Test/Shell/TestShellBase.cs
AlexanderSher: AFAIK it is non-blocking

src/Package/Impl/Search/VsSearchControl.cs
AlexanderSher: Why not? Search results can be updated when package is installed or removed.

src/Package/Impl/Search/VsSearchTask.cs
AlexanderSher: Done

src/Package/Impl/Search/VsSearchTask.cs
AlexanderSher: Done

src/R/Components/Impl/PackageManager/Implementation/View/DesignTime/DesignTimeRPackageViewModel.cs
```diff
@@ -1,24 +1,48 @@
...
 
+using System.Collections.Generic;
 using Microsoft.R.Components.PackageManager.Model;
 using Microsoft.R.Components.PackageManager.ViewModel;
 
 namespace Microsoft.R.Components.PackageManager.Implementation.View.DesignTime {
     internal class DesignTimeRPackageViewModel : IRPackageViewModel {
-        public DesignTimeRPackageViewModel(string name = null
+        public DesignTimeRPackageViewModel() {
+            Name = "ggplot2";
```
MikhailArkhipov: Hm is it a test code of sorts?

src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml
```diff
@@ -2,11 +2,205 @@
...
+                <RowDefinition Height="Auto" />
+                <RowDefinition Height="Auto" />
+                <RowDefinition Height="Auto" />
+                <RowDefinition Height="Auto" />
+            </Grid.RowDefinitions>
+
+            <!-- Icon and Id -->
+            <StackPanel Grid.Row="0" Orientation="Horizontal" MinHeight="32" Margin="0,8">
+
+                <TextBox Background="Transparent" Foreground="{DynamicResource {x:Static wpf:Brushes.UIText}}" BorderThickness="0" Margin="-2,0,-2,0"
```
MikhailArkhipov: I assume these are mapped to VS ones. Does it look OK in dark and high contrast?

src/Package/Impl/Search/VsSearchControl.cs
MikhailArkhipov: OK, just checking

src/R/Components/Impl/PackageManager/Implementation/View/DesignTime/DesignTimeRPackageViewModel.cs
AlexanderSher: It is design time code. Required for WPF designer to render something. Never executed in runtime. I think I can make it debug-only.

src/Package/Impl/source.extension.vsixmanifest
AlexanderSher: Didn't work without it

src/Languages/Editor/Test/Shell/TestShellBase.cs
MikhailArkhipov: So what's the benefit of InvokeAsync vs BeginInvoke?

src/R/Components/Impl/PackageManager/Implementation/View/DesignTime/DesignTimeRPackageViewModel.cs
MikhailArkhipov: I'd replace it with generic Lorem Ipsum data - just in case ;-)

src/Package/Impl/source.extension.vsixmanifest
MikhailArkhipov: I meant can we merge R.Wpf into Common.Wpf

src/R/Components/Impl/PackageManager/Implementation/View/PackageManagerControl.xaml
```diff
@@ -6,10 +6,105 @@
...
+                    <ColumnDefinition Width="auto" />
+                </Grid.ColumnDefinitions>
+
+                <RadioButton Grid.Column="0" x:Name="TabAvailable" GroupName="Tabs" Template="{StaticResource TabToggleButton}" Content="{x:Static components:Resources.PackageManager_Available}" Checked="TabAvailable_Checked" />
+                <RadioButton Grid.Column="2" x:Name="TabInstalled" GroupName="Tabs" Template="{StaticResource TabToggleButton}" Content="{x:Static components:Resources.PackageManager_Installed}" Checked="TabInstalled_Checked" />
+                <RadioButton Grid.Column="4" x:Name="TabLoaded" GroupName="Tabs" Template="{StaticResource TabToggleButton}" Content="{x:Static components:Resources.PackageManager_Loaded}" Checked="TabLoaded_Checked" />
+
+                <!-- container of the search control -->
+                <Border Grid.Column="6" x:Name="SearchControlHost" VerticalAlignment="Center" HorizontalAlignment="Right" Width="320" MinHeight="22" MinWidth="224" />
+
```
MikhailArkhipov: Does it need fixed width? May be too small on large or high DPU monitors.

src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml
AlexanderSher: Two lines need to specify color:
dark

src/R/Components/Impl/PackageManager/Implementation/View/PackageManagerControl.xaml
```diff
@@ -25,7 +120,7 @@
                         <RowDefinition Height="Auto" />
                     </Grid.RowDefinitions>
 
-                    <view:InfiniteScrollList Grid.Row="0" x:Name="ListPackages" Loaded="ListPackages_Loaded" />
+                    <view:PackageList Grid.Row="0" x:Name="ListPackages" Loaded="ListPackages_Loaded" />
                     <Border Grid.Row="1" BorderThickness="0,1,0,0" x:Name="LegalDisclaimer" BorderBrush="{DynamicResource {x:Static SystemColors.ActiveBorderBrushKey}}"
                             Background="{DynamicResource {x:Static rwpf:Brushes.LegalMessageBackground}}">
```
MikhailArkhipov: Should be mapped to VS colors otherwise won't appear right in dark theme

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
```diff
@@ -0,0 +1,227 @@
...
+        private Size _checkBoxSize;
+
+        public PackageSourcesOptionsControl() {
+            InitializeComponent();
+            UpdateDpi();
+        }
+
+        private void UpdateDpi() {
+            var imgs = images16px;
+            if (addButton.Height > 72) {
```
MikhailArkhipov: May need 128 as well for 144+ dpi displays.

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
```diff
@@ -0,0 +1,227 @@
...
+
+            addButton.ImageList = imgs;
+            removeButton.ImageList = imgs;
+            MoveUpButton.ImageList = imgs;
+            MoveDownButton.ImageList = imgs;
+        }
+
+
+        private void PackageSourcesContextMenu_ItemClick(object sender, ToolStripItemClickedEventArgs e) {
+            throw new NotImplementedException();
```
MikhailArkhipov: The UI is hidden right? Otherwise prob should just do nothing.

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
```diff
@@ -0,0 +1,227 @@
...
+            e.DrawBackground();
+
+            if (e.Index < 0 || e.Index >= currentListBox.Items.Count) {
+                return;
+            }
+
+            var currentItem = (RPackageSourceViewModel)currentListBox.Items[e.Index];
+
+            using (var drawFormat = new StringFormat()) {
+                using (var foreBrush = new SolidBrush(currentListBox.SelectionMode == SelectionMode.None ? SystemColors.WindowText : e.ForeColor)) {
```
MikhailArkhipov: VS IDE style brushes

src/Languages/Editor/Test/Shell/TestShellBase.cs
AlexanderSher: Just different syntax. InvokeAsync allows to `await` its task and continue execution.

src/Package/Impl/source.extension.vsixmanifest
AlexanderSher: I would like to have R-specific styles and templates in a separate library. We probably should move controls there as well, but that isn't in this changeset.

src/R/Components/Impl/PackageManager/Implementation/View/DesignTime/DesignTimeRPackageViewModel.cs
AlexanderSher: Sure :)

src/Package/Impl/source.extension.vsixmanifest
MikhailArkhipov: OK, It is just 'Common.Wpf' is actually named Microsoft.R.Common.Wpf.dll ;-)

src/R/Components/Impl/PackageManager/Implementation/View/PackageManagerControl.xaml
AlexanderSher: It looks quite ugly if it is stretched to the whole available length.

src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml
AlexanderSher: Done

src/Common/Core/Test/Threading/BinaryAsyncLockTest.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
```
int19h: Header

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core.Threading {
+    /// <summary>
+    /// BinaryAsyncLock is a helper primitive that can be used instead of SemaphoreSlim.WaitAsync + double-checked locking
+    /// After BinaryAsyncLock is created or reset, the first caller of <see cref="WaitAsync"/> will immediately get <see langword="false" />
+    /// All other callers will wait until <see cref="Release"/> is called and then will get <see langword="true" />
+    /// </summary>
+    public class BinaryAsyncLock {
```
int19h: Would `IDisposable` be inappropriate here?

src/Common/Wpf/Impl/BindableBase.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        /// <summary>
+        /// Notifies listeners that a property value has changed.
+        /// </summary>
+        /// <param name="propertyName">Name of the property used to notify listeners. This
+        /// value is optional and can be provided automatically when invoked from compilers
+        /// that support <see cref="CallerMemberNameAttribute"/>.</param>
+        protected void OnPropertyChanged(string propertyName) {
+            var eventHandler = PropertyChanged;
```
int19h: Why the copy?

src/Package/Impl/Options/PackageManager/PackageSourceOptionsPage.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
```
int19h: Header

src/Package/Impl/Search/VsSearchControl.cs
```diff
@@ -0,0 +1,62 @@
+using System;
```
int19h: Header

MikhailArkhipov:
I can't find the string for the 'Each package is licensed...' in the screenshot above. But it should say 'Microsoft is not responsible...' rather than 'RTVS'

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
AlexanderSher: Since there is no backend, I'll disable this pane for 0.3

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
AlexanderSher: Yes, it will be hidden

src/Common/Core/Test/Threading/BinaryAsyncLockTest.cs
AlexanderSher: Done

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            foreach (var i in collection) {
+                Items.Add(i);
+            }
+
+            OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
+        }
+
+        public void RemoveRange(IEnumerable<T> collection) {
```
MikhailArkhipov: RemoveSet or RemoveMany is probably more appropriate name since it doesn't actually remove from 'start' to 'end' and rather removes disjoint collection of items

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
AlexanderSher: `IDisposable` usually can't be reset to non disposed state

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (collection == null) {
+                throw new ArgumentNullException(nameof(collection));
+            }
+
+            foreach (var i in collection) {
+                Items.Remove(i);
+            }
+
+            OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
```
MikhailArkhipov: Doesn't need to fire if the collection was empty

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (collection == null) {
+                throw new ArgumentNullException(nameof(collection));
+            }
+
+            Items.Clear();
+            foreach (var i in collection) {
+                Items.Add(i);
+            }
+
+            OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
```
MikhailArkhipov: Nothing added = nothing changed

src/Common/Wpf/Impl/BindableBase.cs
AlexanderSher: Not my code, just copy-pasted it. Fixed.

src/Package/Impl/Options/PackageManager/PackageSourceOptionsPage.cs
AlexanderSher: Done

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml
```diff
@@ -0,0 +1,267 @@
...
+                            </TextBlock>
+
+                            <!-- uninstall button -->
+                            <Button x:Name="ButtonUninstall" Style="{StaticResource ButtonStyle}" Grid.Column="1" Click="ButtonUninstall_Click"
+                                    Visibility="{Binding Path=IsInstalled, Converter={x:Static wpf:Converters.FalseIsCollapsed}}"
+                                    ToolTip="{x:Static components:Resources.PackageManager_UninstallButtonToolTip}">
+                                <Rectangle Fill="{StaticResource IconUninstall}" Width="16" Height="16" />
+                            </Button>
+
+                            <!-- install button. It occupies the same position of the install button -->
```
int19h: This comment is recursive :)

src/Package/Impl/Search/VsSearchControl.cs
AlexanderSher: Done

AlexanderSher:
Done

src/Package/Impl/Package.vsct
```diff
@@ -824,7 +824,7 @@
       </Button>
 
       <Button guid="guidRToolsCmdSet" id="icmdShowPackagesWindow" priority="0x0600" type="Button">
-        <!--<Parent guid="guidRToolsCmdSet" id="windowsSubMenuGroup"/>-->
+        <Parent guid="guidRToolsCmdSet" id="windowsSubMenuGroup"/>
```
MikhailArkhipov: Need to bind to Ctrl+7 in 'KeyBindings'

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
AlexanderSher: Renamed

src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml
int19h: Side note regarding screenshot - the disclaimer at the bottom should probably be saying "Microsoft is not responsible", since that's the legal entity.

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
AlexanderSher: Fixed

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -1,9 +1,243 @@
...
+                if (vmAvailablePackages.TryGetValue(installedPackage.Package, out vmPackage)) {
+                    vmPackage.AddDetails(installedPackage, true);
+                    vmInstalledPackages.Add(vmPackage);
+                } else {
+                    vmInstalledPackages.Add(RPackageViewModel.CreateInstalled(installedPackage));
+                }
+            }
+
+            _installedPackages = vmInstalledPackages.OrderBy(p => p.Name).ToList();
+            _availablePackages = vmAvailablePackages.Values.ToList<IRPackageViewModel>();
```
int19h: Not sorted?

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
AlexanderSher: Fixed

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml
AlexanderSher: Fixed

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -0,0 +1,149 @@
...
+                Repository = package.Repository,
+            };
+        }
+
+        public static RPackageViewModel CreateInstalled(RPackage package) {
+            return new RPackageViewModel(package.Package) {
+                Title = package.Title,
+                Authors = package.Author,
+                License = package.License,
+                Urls = package.URL?.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries)
```
int19h: `?.` implies that `URL` can be null; so `Split` would be as well in that case, no? And then `Select` would NRE.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public RPackageViewModel(string name) {
+            Name = name;
+        }
+
+        public void AddDetails(RPackage package, bool isInstalled) {
+            Title = package.Title;
+            Description = package.Description;
+            Authors = package.Author;
+            Urls = package.URL?.Split(new[] {","}, StringSplitOptions.RemoveEmptyEntries)
```
int19h: And same thing here.

int19h: (Maybe the shorter way to write it would be `(package.URL ?? "").Split...`?)

src/Package/Impl/Package.vsct
AlexanderSher: Done

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
AlexanderSher: Fixed

src/Package/Impl/Options/PackageManager/PackageSourceOptionsPage.cs
```diff
@@ -0,0 +1,39 @@
...
+        //        e.ApplyBehavior = ApplyKind.CancelNoNavigate;
+        //    }
+        //}
+
+        //protected override void OnClosed(EventArgs e) {
+        //    PackageSourcesControl.ClearSettings();
+        //    base.OnClosed(e);
+        //}
+
+        protected override IWin32Window Window => PackageSourcesControl;
```
MikhailArkhipov: Where is the PackageSourcesControl sources?

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
AlexanderSher: No, if `package.URL` is `null`, then the rest would not be executed

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
int19h: Indeed. Perhaps make `WaitAsync` return a disposable token that has a property to query whether it was the first issued one or not? Having to use `finally` always makes me uneasy...

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -1,9 +1,243 @@
...
+        }
+
+        private void LoadLoadedPackages() {
```
MikhailArkhipov: Is it needed?

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+
+namespace Microsoft.Common.Wpf.Collections {
+    /// <summary>
+    /// ObservableCollection that supports bulk changes
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class RangeObservableCollection<T> : ObservableCollection<T> {
```
int19h: In light of the renaming below, it feels like this could use a clearer name, too. `BatchObservableCollection`?

src/Package/Impl/Options/PackageManager/PackageSourceOptionsPage.cs
AlexanderSher: `Microsoft.R.Components.PackageManager.Implementation.View.PackageSourcesOptionsControl`

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
MikhailArkhipov: Do we then need to remove attribute from the package so Tools | Options page will be completely out?

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
int19h: You're right. Weird that they did this. I would have expected to have to keep chaining `?.`, unless I wanted one of the calls to actually happen for `null` (e.g. what if I have an extension method that handles nulls in a special manner?).

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
AlexanderSher: So you suggest to define some kind of `BinaryAsyncLockWaitResult` which will be comparable to `bool`:
```{csharp}
using (var areLoaded = await _availableAndInstalledLock.WaitAsync()) {
    if (!areLoaded) {
        IsLoading = true;
        await LoadInstalledAndAvailablePackagesAsync();
        IsLoading = false;
    }
}
```

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
AlexanderSher: ok

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
int19h: I was thinking of a property on the returned object instead of comparing it to bool directly, but either one would work; and since you'd always want to check when using this object, I guess the latter would make more sense. That would take `operator true` and `operator false`, right?

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
AlexanderSher: Not yet.

src/R/Components/Impl/PackageManager/Implementation/View/PackageSourcesOptionsControl.cs
AlexanderSher: Yes

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public RangeObservableCollection(IEnumerable<T> collection) : base(collection) { }
+
+        public void AddRange(IEnumerable<T> collection) {
+            if (collection == null) {
+                throw new ArgumentNullException(nameof(collection));
+            }
+
+            foreach (var i in collection) {
+                Items.Add(i);
```
MikhailArkhipov: So this collection appears to be accessede from async tasks but it doesn't seem to make any attempts for locking... Collection<T> is not thread safe.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
AlexanderSher: Why? If you have found that value is null on the previous step, why do you want to check for null on the next step?

src/Common/Wpf/Impl/Collections/RangeObservableCollection.cs
MikhailArkhipov: nvm, it seems that all calls are via

```
async Task SwitchToAvailablePackagesAsync() {
```

which confuses me since it a s async method and yet has to be called on UI thread...

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
int19h: Because I might not want a `null` to `null` mapping on the next step. For example, my extension method might be something like:
```
string Stringify(this Foo foo) {
  if (foo == null) return "<null>"
  ...
}
```
And then if I do `bar?.GetFoo().Stringify()`, I would expect to get `<null>` if `bar` was null; but I  get actual null instead, because `Stringify()` isn't called at all. I'd prefer to have full control on every step by using `.` or `?.` as necessary.

(I recall now that there was actually a discussion about this back when they were still designing the feature.)

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
AlexanderSher:
```{csharp}
public class BinaryAsyncLockWaitResult : IDisposable {
    private TaskCompletionSource<bool> _tcs;
    public BinaryAsyncLockWaitResult (TaskCompletionSource<bool> tcs) {
        _tcs = tcs;
    }
    public static implicit operator double(BinaryAsyncLockWaitResult r) {
        return r.result;
    }
    public void Dispose() {
        _tcs.TrySetResult(true);
    }
}
```

MikhailArkhipov:
:+1: pending Pavel's CR. Also, may need to consider adding some telemetry in 0.4

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -0,0 +1,149 @@
...
+                .Select(s => s.Trim())
+                .ToArray() ?? new string[0];
+            LibraryPath = package.LibPath;
+            Built = package.Built;
+
+            if (isInstalled) {
+                InstalledVersion = package.Version;
+                IsInstalled = true;
+            }
+            IsUpdateAvailable = InstalledVersion != LatestVersion;
```
huguesv: You should put a TODO for proper version comparison.

AlexanderSher: Done

huguesv:
You'll have to handle RPackageManagerException at some level, right now I don't see it being caught anywhere.

AlexanderSher:
As i said in the description, your changes from yesterday aren't addressed :)

int19h:
:+1:, let's get it in for the bash

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
AlexanderSher: Wouldn't include this change in current PR - `finally` block is still needed in `EnsureInstalledAndAvailablePackagesLoadedAsync` and change isn't trivial.

Issue #1275, opened 2016-03-14
Smart indent should be aligning arguments


MikhailArkhipov:
```
x <- function(a<ENTER>
```

should position caret under 'a'

MikhailArkhipov:
#1285

huguesv:
While it does achieve what was described in the example above, I feel it's not quite correct and need to allow more flexibility, like in RS.

- It should be aligning based on the column of the previous argument.  Right now it's always aligning with column+1 of the open paren.
- When there's no previous argument and you press ENTER, it should indent one level relative to the first char of the function name.

In RS it is easy to do this:
```
res <- this.function.name.is.somewhat.long(
  'this long message would go over 80 cols if aligned with open paren',
  'other parameter'
)

```
But trying to do the same in RTVS is infuriating.


Examples (where indentation is 2 spaces):

action:
```
res <- myfunc(<ENTER>
```

expected:
```
res <- myfunc(
  <caret>
```

actual:
```
res <- myfunc(
              <caret>
```


action:
```
res <- myfunc(1,<ENTER>
```

expected & actual (already correct):
```
res <- myfunc(1,
              <caret>



action:
```
res <- myfunc(
  1,<ENTER>
```

expected:
```
res <- myfunc(
  1,<ENTER>
  <caret>
```

actual:
```
myfunc(
       1,
       <caret>
)
```

**Comment by MikhailArkhipov** [comment]
We used to handle user indent, but user found it improper and rather have arguments aligned. However, we can technically combine smart indent and block indent (i.e. that is what RS does based on the previous line).

**Comment by int19h** [comment]
If they're typing a new call, user indent will equate to having the arguments aligned (in a way that user explicitly asked for by the placement of the first argument), no?

**Comment by MikhailArkhipov** [comment]
RS:

```
x <- function(a,<ENTER>
                     <caret>
```

PR #1396, opened 2016-04-01
#1275 + more shortcut work

MikhailArkhipov:
#1275: Smart indent should be aligning arguments 
- Pass original user indentation to the range formatter so it can apply it in function arguments
- Format on Enter and ; before the typing is applied
- Format on } after } is applied to the buffer
- Tests

Shortcuts:
- Remove DTE shortcut take over hack
- Return R-specific command context from factory CreateEditorInstace instead of core editor context
- Explicitly inherit core editor bindings when view is created
- Add 'clear repl' to RMD editor


src/Package/Impl/Editors/BaseEditorFactory.cs
```diff
@@ -318,5 +312,30 @@ public class BaseEditorFactory : IVsEditorFactory, IDisposable {
 
         protected virtual void Dispose(bool disposing) { }
         #endregion
+
+        public static void InitKeyBindings(ITextView textView) {
```
AlexanderSher: Looks more like a hepler or extension method. Maybe move it to the TextViewExtensions?

MikhailArkhipov: It only applies to views that host main editor files. I.e. doesn't apply to REPL or History.

src/R/Editor/Impl/Formatting/FormatOperations.cs
```diff
@@ -23,8 +23,8 @@ internal static class FormatOperations {
         /// <summary>
         /// Formats statement that the caret is at
         /// </summary>
-        public static void FormatCurrentStatement(ITextView textView, ITextBuffer textBuffer) {
-            SnapshotPoint? caretPoint = MapCaretToBuffer(textView, textBuffer);
+        public static void FormatCurrentStatement(ITextView textView, ITextBuffer textBuffer, int caretOffset = 0) {
```
AlexanderSher: Previously default was `-1`

MikhailArkhipov: Yes. This was because we were always formatting AFTER text buffer changed so after ENTER caret was on the next line and that led to -1 offset. Now we format BEFORE the change so default is 0. The only case we format after the change is on } (because of the way brace completion and provisional text work in the core editor)

AlexanderSher:
:+1: 
Probably @int19h should also take a look

int19h:
:+1:

PR #1399, opened 2016-04-01
Stop update grid after closed

JeongPyoHam:
- needless update grid after closing the window pane

### DIFF
--- a/src/Package/Impl/DataInspect/VariableGridWindowPane.cs
+++ b/src/Package/Impl/DataInspect/VariableGridWindowPane.cs
@@ -1,14 +1,16 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
 using System.Runtime.InteropServices;
 using Microsoft.VisualStudio.Imaging;
 using Microsoft.VisualStudio.R.Package.Windows;
+using Microsoft.VisualStudio.Shell.Interop;
 using static System.FormattableString;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect {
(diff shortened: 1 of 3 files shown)


src/Package/Impl/DataInspect/VariableGridWindowPane.cs
```diff
@@ -25,5 +27,31 @@ internal sealed class VariableGridWindowPane : RToolWindowPane {
...
+        public int OnDockableChange(int fDockable, int x, int y, int w, int h) {
+            return VSConstants.S_OK;
+        }
+
+        public int OnMove(int x, int y, int w, int h) {
+            return VSConstants.S_OK;
+        }
+
+        public int OnShow(int fShow) {
+            return VSConstants.S_OK;
```
huguesv: Does the reverse need to happen in OnShow?  When you re-open the tool window after closing it, does scrolling work again?

JeongPyoHam: The only way of showing grid pane is from VE, which resets new evaluation. So, OnShow is not required. SetEvaluation() does the work.

int19h: There are ways to have it show later that would not go through VE. Windows | Save Window Layout while it's visible, followed by Apply Window Layout later, for example.

MikhailArkhipov: I.e. try Import/Export settings, save window layout, restart with no VE visible, then go via Import/Export again and apply saved settings. Or hide VE, restart then apply DS profile.

MikhailArkhipov: Also: We need to override Dispose() from WindowPane. Also, WindowPane exposes OnClose which you may be able to use instead of FrameNotify.

PR #1433, opened 2016-04-02
Fix #1318, #1432

### LINKED ISSUE #1318: cxxfunction: Error in system(cmd, intern = !verbose) : 'C:/Program' not found
from user:
```r
setwd("e:/hmad")

emad_dat <- list(mad = 50)

library(rstan)
fit <- stan(file = 'e:/hmad/EMAD1.stan', data = emad_dat,
            iter = 10000, chains = 4)
```
I get the error: 
> Error in system(cmd, intern = !verbose) : 'C:/Program' not found

Easier repro:
```r
fx <- cxxfunction(signature(x = "integer", y = "numeric"), 'return ScalarReal( INTEGER(x)[0] * REAL(y)[0] )')
```

### LINKED ISSUE #1432: If R\bin is in PATH, wrong R.dll is loaded
When we start the host process, we append the path to the designated R interpreter bin folder to PATH, so that the DLL gets picked up. But because we append it at the end, if it already has a path to bin somewhere on it, that path will be preferred, and we end up picking the wrong DLL. If it happens to be a 32-bit DLL, or if it's an old version of R that we don't support, the host process cannot start at all.

(This is likely to be the root cause of various user-reported issues where REPL is completely inoperable on startup.)

int19h:
Fix #1318: cxxfunction: Error in system(cmd, intern = !verbose) : 'C:/Program' not found

Shorten path when setting R_HOME.

Fix #1432: If R\bin is in PATH, wrong R.dll is loaded

When adding R\bin to PATH when launching host, insert it at the beginning rather than at the end.

### DIFF
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -527,8 +527,11 @@ namespace Microsoft.R.Host.Client {
                 UseShellExecute = false
             };
 
-            psi.EnvironmentVariables["R_HOME"] = rHome;
-            psi.EnvironmentVariables["PATH"] = Environment.GetEnvironmentVariable("PATH") + ";" + rBinPath;
+            var shortHome = new StringBuilder(NativeMethods.MAX_PATH);
+            NativeMethods.GetShortPathName(rHome, shortHome, shortHome.Capacity);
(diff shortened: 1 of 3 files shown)


MikhailArkhipov:
:+1:

huguesv:
:+1:

Issue #1421, opened 2016-04-01
PM should be normalizing whitespace in package details


MikhailArkhipov:
Otherwise line breaks look odd

![image](https://cloud.githubusercontent.com/assets/12820357/14220934/1ef7e12e-f818-11e5-97be-9bd0c898b8d5.png)

PR #1434, opened 2016-04-02
#1421 PM should be normalizing whitespace in package details

MikhailArkhipov:


src/Languages/Core/Impl/Formatting/Whitespace.cs
```diff
@@ -110,5 +111,28 @@ public static class Whitespace {
...
+            var sb = new StringBuilder();
+
+            while (!cs.IsEndOfStream()) {
+                var current = cs.Position;
+                cs.SkipWhitespace();
+                if (cs.Position - current > 0) {
+                    sb.Append(' ');
+                }
+
+                while (!cs.IsEndOfStream() && !cs.IsWhiteSpace()) {
```
AlexanderSher: There can be one loop instead of two, but I'm not sure that it will be easier to read, so up to you

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -35,16 +36,16 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
...
                 Urls = package.URL?.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries)
                     .Select(s => s.Trim())
                     .ToArray() ?? new string[0],
                 NeedsCompilation = package.NeedsCompilation != null && !package.NeedsCompilation.EqualsIgnoreCase("no"),
                 LibraryPath = package.LibPath,
                 Repository = package.Repository,
-                Description = package.Description,
+                Description = package.Description.NormalizeWhitespace(),
                 Built = package.Built,
                 Depends = package.Depends,
```
AlexanderSher: `Depends`, `Imports` and `Suggests` also need normalization

MikhailArkhipov: OK

src/Languages/Core/Impl/Formatting/Whitespace.cs
MikhailArkhipov: Just borrowed old code

AlexanderSher:
:+1:

PR #1438, opened 2016-04-02
Retrieve environment names

### LINKED ISSUE #702: Remove unnecessary stack frames from the call stack
in stack view there are stuff a typical user wont care about & it clutters the view.

image

int19h:
When retrieving call stack, include formatted environment names for those environments that have them; and use those names in the Call Stack window when available.

Fix #702: Remove unnecessary stack frames from the call stack

### DIFF
--- a/src/Debugger/Engine/Impl/AD7StackFrame.cs
+++ b/src/Debugger/Engine/Impl/AD7StackFrame.cs
@@ -109,7 +109,7 @@ namespace Microsoft.R.Debugger.Engine {
             }
 
             if (dwFieldSpec.HasFlag(enum_FRAMEINFO_FLAGS.FIF_FUNCNAME)) {
-                fi.m_bstrFuncName = StackFrame.CallingFrame?.Call ?? (StackFrame.IsGlobal ? "<global>" : "<unknown>");
+                fi.m_bstrFuncName = StackFrame.EnvironmentName ?? StackFrame.CallingFrame?.Call ?? "<unknown>";
                 fi.m_dwValidFields |= enum_FRAMEINFO_FLAGS.FIF_FUNCNAME;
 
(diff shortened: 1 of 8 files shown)


int19h:
The first part is to assist Variable Explorer call stack implementation.

#702 is a low-hanging fruit (more so than I originally thought), so I figured I might as well do that while i'm there; since VE is more visible than Call Stack tool window, there might be more people paying attention to it now.

int19h:
Also, regarding tests - I intend to add more tests for `IsGlobal`, `EnvironmentName` etc, but i want to use the better test helpers that I've been working on in a private branch for that. So I'll go back to the tests here once the helpers are in.

int19h:
@JeongPyoHam, please take a look at `EnvironmentName` to see if it is suitable for the purpose of showing call stack in VE as we've discussed today.

int19h:
Ping.

JeongPyoHam:
go ahead, sir

src/Debugger/Engine/Impl/AD7StackFrame.cs
```diff
@@ -109,7 +109,7 @@ internal sealed class AD7StackFrame : IDebugStackFrame2, IDebugExpressionContext
             }
 
             if (dwFieldSpec.HasFlag(enum_FRAMEINFO_FLAGS.FIF_FUNCNAME)) {
-                fi.m_bstrFuncName = StackFrame.CallingFrame?.Call ?? (StackFrame.IsGlobal ? "<global>" : "<unknown>");
+                fi.m_bstrFuncName = StackFrame.EnvironmentName ?? StackFrame.CallingFrame?.Call ?? "<unknown>";
```
int19h: Test

Issue #1430, opened 2016-04-01
VS not responding after some editing


huguesv:
Start with the file attached.
[script.zip](https://github.com/Microsoft/RTVS/files/200564/script.zip)

Copy/paste the last line on the line immediately next to it (put cursor on first column and hit CTRL-C, CTRL-V)
Move the cursor to location in picture (before closing paren on the first call to merge)

Consistently repros.  Code isn't stuck in an obvious loop as far as I can tell.

![image](https://cloud.githubusercontent.com/assets/1696845/14222630/9d88312e-f827-11e5-928d-2623ad67b6b2.png)

MikhailArkhipov:
Doesn't repro for me. Don't mind if you show it :-) Also, may be some setup issue or clean build issue since I don't see those task canceled exceptions either. This one may be related.

huguesv:
Oops, turns out I forgot the last step.  When the cursor is in the position of the picture, press CTRL-SHIFT-LEFTARROW (which selects the preceding word).  I got that to hang on both my dev machine and in the VM where I installed today's build.

MikhailArkhipov:
Facepalm

MikhailArkhipov:
Looks like VS doesn't like us returning 0-length span and keeps calling same method.

PR #1440, opened 2016-04-04
#1430 VS not responding after some editing

MikhailArkhipov:
- VS doesn't like us returning empty spans in text structure navigator - it keeps calling. So when span we find is empty, pass control down to default.
- Improve span selection with non-character items and empty lines
- Add more tests


int19h:
:+1:

Issue #1441, opened 2016-04-04
Package manager: Move list of authors to the bottom


MikhailArkhipov:
It is sometimes very long and one has to scroll down to see library/repo or dependencies important details

![image](https://cloud.githubusercontent.com/assets/12820357/14258687/5fed18a6-fa58-11e5-8937-9210700a2875.png)

Issue #1425, opened 2016-04-01
Link to repo in package details should be clickable


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/14220990/9a083f30-f818-11e5-9d6a-7176a3b8c1ed.png)

MikhailArkhipov:
#1422

Issue #1422, opened 2016-04-01
Package Manager UI: Library in details page should be clickable


huguesv:
It should open the folder in windows file explorer.

PR #1442, opened 2016-04-04
#1425 #1422 #1441 Link to repo in package details should be clickable

MikhailArkhipov:
#1422 Package Manager UI: Library in details page should be clickable
#1441 Packaga manager: Move list of authors to the bottom
#1425 Link to repo in package details should be clickable

### DIFF
--- a/src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml.cs
@@ -1,8 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
+using System.Diagnostics;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Navigation;
 
 namespace Microsoft.R.Components.PackageManager.Implementation.View {
     /// <summary>
@@ -14,15 +17,23 @@ namespace Microsoft.R.Components.PackageManager.Implementation.View {
         }
 
         private void HyperlinkUrl_Click(object sender, RoutedEventArgs e) {
-            throw new System.NotImplementedException();
+            throw new NotImplementedException();
         }
 
         private void ButtonInstall_Click(object sender, RoutedEventArgs e) {
-            throw new System.NotImplementedException();
+            throw new NotImplementedException();
         }
         
         private void ButtonUninstall_Click(object sender, RoutedEventArgs e) {
-            throw new System.NotImplementedException();
+            throw new NotImplementedException();
+        }
+
+        private void RepositoryUri_RequestNavigate(object sender, RequestNavigateEventArgs e) {
+            Process.Start(e.Uri.AbsoluteUri);
+        }
+
+        private void LibraryPath_RequestNavigate(object sender, RequestNavigateEventArgs e) {
+            Process.Start(e.Uri.ToString().Replace('/', '\\'));
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -23,6 +23,9 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
         private bool _isUpdateAvailable;
 
         public static RPackageViewModel CreateAvailable(RPackage package) {
+            Uri uri = null;
```
AlexanderSher: Let's call it repositoryUri

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -55,6 +63,8 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
                 IsInstalled = true,
                 HasDetails = true
             };
+
```
AlexanderSher: Formatting?

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -30,11 +33,15 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
...
                 NeedsCompilation = package.NeedsCompilation != null && !package.NeedsCompilation.EqualsIgnoreCase("no"),
-                Repository = package.Repository,
+                RepositoryUri = uri,
+                RepositoryText = uri != null ? null : package.Repository,
             };
         }
 
         public static RPackageViewModel CreateInstalled(RPackage package) {
+            Uri uri = null;
+            Uri.TryCreate(package.Repository, UriKind.Absolute, out uri);
```
AlexanderSher: What will happen if `package.Repository` is null?

src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml.cs
```diff
@@ -14,15 +17,23 @@ public partial class PackageDetails : UserControl {
...
-            throw new System.NotImplementedException();
+            throw new NotImplementedException();
+        }
+
+        private void RepositoryUri_RequestNavigate(object sender, RequestNavigateEventArgs e) {
+            Process.Start(e.Uri.AbsoluteUri);
+        }
+
+        private void LibraryPath_RequestNavigate(object sender, RequestNavigateEventArgs e) {
+            Process.Start(e.Uri.ToString().Replace('/', '\\'));
```
AlexanderSher: @huguesv, is it guaranteed that `Package.LibPath` is URL?

huguesv: It's a path on disk in the following format:
```
> i1[1,'LibPath']
[1] "C:/Users/huvalo/Documents/R/win-library/3.2"
> 
```

Also, I've never seen it return a path with a ` in it, it's always expanded.  So the code above should work AFAIK.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
MikhailArkhipov: You get null URI so text will be used which is also null. You get Repositiry: and then empty

src/R/Components/Impl/PackageManager/Implementation/View/PackageDetails.xaml.cs
MikhailArkhipov: In the worst case it won't open anything in explorer

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
MikhailArkhipov: Fixed

AlexanderSher:
:+1:

Issue #1410, opened 2016-04-01
Package Manager UI: Details should show package title


huguesv:
Title is a short sentence, which is different from the description, which is a long paragraph.

AlexanderSher:
Fixed in #1447

Issue #1439, opened 2016-04-04
Package Manager: GetInstalledPackagesAsync doesn't contain package description


AlexanderSher:
`IRPackageManager.GetInstalledPackagesAsync` returns `Description == null`.
Close it if `by design`

MikhailArkhipov:
Can't be by design

PR #1443, opened 2016-04-04
#1410: Package Manager UI: Details should show package title

MikhailArkhipov:


huguesv:
I was thinking we show both title and description.

MikhailArkhipov:
They may be duplicate text... Description is often missing however.

int19h:
Don't we already show title in XAML?

int19h:
Ah, I see now, title != package name. 

In details, I think it would be most appropriate to show both title and description.

huguesv:
We show it in the tooltip in the left pane.  I'd like us to show it at the top in the details pane.

Description shouldn't be missing.  It's a mandatory field:
https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-DESCRIPTION-file

Right now Description is missing from installed packages, but that's a bug and will be fixed.
https://github.com/Microsoft/RTVS/issues/1439

MikhailArkhipov:
Nope.

image

We could do NuGet style

image

MikhailArkhipov:
Maybe I misunderstood the bug but Description comes empty in 

            return new RPackageViewModel(package.Package) {

AlexanderSher:
Guys, I've already added a row with title to the Details
So you can leave this bug on me

MikhailArkhipov:
Maybe the bug is actually in fetching descriptions then...

MikhailArkhipov:
OK

AlexanderSher:
> Maybe the bug is actually in fetching descriptions then...

https://github.com/Microsoft/RTVS/issues/1439

Issue #1382, opened 2016-03-30
BreakContinue test is unstable


int19h:
Occasionally fails with:
```
Test Name:	SteppingTest.BreakContinue
Test FullName:	Microsoft.R.Debugger.Test.SteppingTest.BreakContinue
Test Source:	C:\git\RTVS\src\Debugger\Test\SteppingTest.cs : line 44
Test Outcome:	Failed
Test Duration:	0:00:01.166

Result StackTrace:	
at FluentAssertions.Execution.FallbackTestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\Shared\Execution\FallbackTestFramework.cs:line 21
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions.Net40\Execution\TestFrameworkProvider.cs:line 42
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\DefaultAssertionStrategy.cs:line 25
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\AssertionScope.cs:line 197
   at FluentAssertions.Execution.GivenSelector`1.FailWith(String message, Object[] args) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\GivenSelector.cs:line 122
   at FluentAssertions.Execution.GivenSelector`1.FailWith(String message, Func`2[] args) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\GivenSelector.cs:line 96
   at Microsoft.UnitTests.Core.FluentAssertions.CollectionAssertionsExtensions.HaveTail[TActual,TExpected](GenericCollectionAssertions`1 assertions, IEnumerable`1 expectation, Func`3 predicate, String because, Object[] reasonArgs) in C:\git\RTVS\src\UnitTests\Core\Impl\FluentAssertions\CollectionAssertionsExtensions.cs:line 71
[...]

PR #1445, opened 2016-04-04
Package manager backend

### LINKED ISSUE #1431: Package Manager: REPL (and all dependent code) is frozen while available packages are loaded
Probably we should use another instance of RSession to get available packages.
Another option is to call CancelAll

### LINKED ISSUE #1439: Package Manager: GetInstalledPackagesAsync doesn't contain package description

huguesv:
Fix #1439 Package Manager: GetInstalledPackagesAsync doesn't contain package description
Fix #1431 Package Manager: REPL (and all dependent code) is frozen while available packages are loaded

When dealing with the REPL session, EvaluateAsync is used so that code can execute even if the REPL is busy.

To avoid making the REPL busy while querying list of available and installed packages, the query is done in a separate session, which is initialized with the same (repo+libpath) options as the REPL session.  This is based on similar code in FunctionRdDataProvider, which uses its own host session.

Make EvaluateAsync wait for the AfterHostStarted initialization to be completed.  This is the code that loads rtvs package, among other things, which is where i was running into issues.

### DIFF
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -41,6 +41,7 @@ namespace Microsoft.R.Host.Client.Session {
         private IReadOnlyList<IRContext> _contexts;
         private RHost _host;
         private Task _hostRunTask;
+        private TaskCompletionSource<object> _afterHostStartedTcs = new TaskCompletionSource<object>();
         private TaskCompletionSource<object> _initializationTcs;
         private RSessionRequestSource _currentRequestSource;
         private readonly IRHostClientApp _hostClientApp;
@@ -130,6 +131,8 @@ namespace Microsoft.R.Host.Client.Session {
                 return await CanceledEvaluateTask;
             }
 
+            await _afterHostStartedTcs.Task;
+
             try {
                 var result = await _host.EvaluateAsync(expression, kind, ct);
                 if (kind.HasFlag(REvaluationKind.Mutating)) {
@@ -176,6 +179,9 @@ namespace Microsoft.R.Host.Client.Session {
 
             var initializationTask = _initializationTcs.Task;
             _hostRunTask = CreateAndRunHost(startupInfo, timeout);
+
+            Interlocked.Exchange(ref _afterHostStartedTcs, new TaskCompletionSource<object>());
+
             ScheduleAfterHostStarted(startupInfo);
 
             await initializationTask;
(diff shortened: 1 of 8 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -176,6 +179,9 @@ internal sealed class RSession : IRSession, IRCallbacks {
 
             var initializationTask = _initializationTcs.Task;
             _hostRunTask = CreateAndRunHost(startupInfo, timeout);
+
+            Interlocked.Exchange(ref _afterHostStartedTcs, new TaskCompletionSource<object>());
```
int19h: There is a race between this and `await` on the task - if you call `EvaluateAsync` before this line runs, it'll be waiting on the old task (which, if this session has never been started before, will never be completed).

I think this rather needs to follow the same pattern as `EnsureHostStartedAsync` does for `_initializationTcs`.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -180,8 +229,40 @@ internal class RPackageManager : IRPackageManager {
             }
         }
 
+        private async Task CreatePackageQuerySessionAsync() {
+            await _pkgQuerySessionSemaphore.WaitAsync();
```
AlexanderSher: Session can't dispose itself, you don't need to subscribe for `_pkgQuerySession.Disposed` if you don't share its instance

huguesv: Removed.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -36,11 +52,11 @@ internal class RPackageManager : IRPackageManager {
         }
 
         public async Task<IReadOnlyList<RPackage>> GetInstalledPackagesAsync() {
-            return await GetPackages(async (eval) => await eval.InstalledPackages());
+            return await GetPackagesAsync(async (eval) => await eval.InstalledPackages());
```
AlexanderSher: I'd like to be able to execute `GetInstalledPackagesAsync` and `GetAvailablePackagesAsync` concurrently. `GetAvailablePackagesAsync` is much slower, while `GetInstalledPackagesAsync` will be executed every time user switch the tab

int19h: If the eval uses direct `EvaluateAsync` rather than `BeginEvaluateAsync`, and specifies `REvaluationKind.Reentrant`, it can be interrupted by other evals, allowing to interleave them.

huguesv: Done

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -79,12 +79,12 @@ public static class RSessionEvaluationCommands {
 
         public static Task<REvaluationResult> InstalledPackages(this IRExpressionEvaluator evaluation) {
             var script = @"rtvs:::packages.installed()";
-            return evaluation.EvaluateAsync(script, REvaluationKind.Json);
+            return evaluation.EvaluateAsync(script, REvaluationKind.Json | REvaluationKind.Reentrant);
```
int19h: This means that this eval can be interrupted by the (much longer) eval for available packages, in which case it won't complete until the other one does. I would suggest using that flag only for available packages.

int19h:
:+1:

huguesv:
I moved the initialization of the task continuation source to happen earlier in time, next to where the initialization continuation source is.
With my original code, SteppingTest.BreakContinue was failing, but now it's passing again.

int19h:
There is a known issue with `BreakContinue`, so ignore it for now:
https://github.com/Microsoft/RTVS/issues/1382

PR #1448, opened 2016-04-05
Package manager: check Installed tab by default

JeongPyoHam:


huguesv:
I just tried with the latest code, and it looks like this has already been taken care of.  The installed page is selected by default.

Issue #1370, opened 2016-03-29
Review package functions display in VE


MikhailArkhipov:
Not sure if 'Value' column is necessary for a typical user, it just shows lazyLoadDBFetch(...). Also, list of functions appears to be different in VE, intellisense and RS. We probably want to have a look why. 

![image](https://cloud.githubusercontent.com/assets/12820357/14123800/143e77ec-f5b7-11e5-8b7a-96f9d50fe7ad.png)

PR #1464, opened 2016-04-05
Show predefined string for promise's value

JeongPyoHam:


src/R/Editor/Impl/Data/RSessionDataObject.cs
```diff
@@ -57,7 +57,7 @@ public class RSessionDataObject : IRSessionDataObject {
                 const string PromiseVaue = "<promise>";
```
huguesv: Might as well fix that typo while you're here.

huguesv:
:+1:

int19h:
Why not show the promise expression? It's more informative, and it is also what RS does.

huguesv:
It creates an unreadable wall of text.  Is the lost information useful in any way?

int19h:
It depends on what it is. In many cases, it can tell you what the value is, before it becomes evaluated. E.g. suppose you have this:
```r
f <- function(x) {
 ...
}
f(123)
```
and you have a breakpoint set at the first line of the function. Without this change, you'll see `x` show `123` at the breakpoint, but with it, you'll see `<promise>`.

Can you give a specific scenario where it results in a wall of text? We probably just need to do some trimming.

huguesv:
image

Maybe only special case lazyLoadDBfetch?

src/R/Editor/Impl/Data/RSessionDataObject.cs
JeongPyoHam: Ah! fixed.

int19h:
It appears that RS changes behavior depending on whether you're viewing a function environment, or a package environment. In function environment, promise code is displayed. In package environment, it just says `<Promise>`. Perhaps we should be doing the same.

int19h:
But yes, since `lazyLoadDBfetch` appears to be something that the package loader uses to initialize a lot of those, special-casing it would appear to be a viable approach, as well. Basically if it's something else, it's probably interesting.

JeongPyoHam:
In the tiny cell, function body, which is cut off, doesn't show more than just 'function', and hard to be useful. When code is required in that cell, new issue will be opened.

int19h:
This is not just about the function bodies, though. This is about promises for function arguments, which are more often than not simple expressions that are helpful, as in my example above.

JeongPyoHam:
As long as it shows more than just <promise>, you can use the tiny part for leverage to your justification. The claimed usefulness is limited to trivial cases where representation fits into the tiny cell.
This is for addressing the specific issue of clearing up package's list, and one change can't contain everything that appears in mind randomly. The concern of showing really useful information in the mentioned area should be addressed in bigger scope such as triggering View with double click or hovering mosue shows code or expanding the cell to bigger one at click.
If even such small snippet required, there will be feedback.

int19h:
My concern is that this change is addressing the specific issue of clearing up package list (#1370) by changing the behavior globally, such that it affects way more than packages; and the resulting change of behavior actually makes it different from RS when it was identical before (for function arguments).

PR #1465, opened 2016-04-06
Fix  #1408: Evaluation children with names that are not identifiers do not work correctly

### LINKED ISSUE #1408: Evaluation children with names that are not identifiers do not work correctly
```r
`123` <- 'hello'
```

123 appears in VE, and shows empty value, class, and type

int19h:
Backtick-quote names as necessary.


MikhailArkhipov:
:+1:

PR #1466, opened 2016-04-06
Clean up references, update component versions

MikhailArkhipov:
In preparation for Dev15:
- Make sure we are getting from nuget what is available rather than referencing VS SDK
- Update external components to recent versions
- Consolidate versions so all projects use same consistent versions of components
- Update MicroBuild to 0.2
- Add basic versioning to setup

Actual Dev15 work is going on in a separate branch. This change will make it easier to compare project files and make conditional reference groups.


src/R.sln
```diff
@@ -61,8 +61,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.VisualStudio.Shel
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.R.Actions", "R\Actions\Impl\Microsoft.R.Actions.csproj", "{B68D4AD2-2DC2-473E-AB06-408172C4FB92}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.R.Actions.Test", "R\Actions\Test\Microsoft.R.Actions.Test.csproj", "{E4528324-C399-48D2-AB38-66D951158F12}"
```
MikhailArkhipov: Need to double check this, in my solution project is still there

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -55,7 +55,7 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
                 LibraryPath = package.LibPath,
                 RepositoryUri = repositoryUri,
                 RepositoryText = repositoryUri != null ? null : package.Repository,
-                Description = package.Description.NormalizeWhitespace(),
+                Description = MakeDescription(package),
```
MikhailArkhipov: Will revert, stale change

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -194,5 +194,9 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
 
             HasDetails = true;
         }
+
```
MikhailArkhipov: Will revert, stale change

int19h:
Since this affects SetupBundle, can you also update http://github.com/Microsoft/RTVS-pr ?

MikhailArkhipov:
sure. BTW should we update R Host to boost 1.60?

int19h:
Ah, yes. If there's a new version of any packages that it uses, might as well update it, too. If there are any issues, we can always back that one package out.

int19h:
:+1:

Issue #1475, opened 2016-04-07
Attempt to uninstall from Loaded tab replaces loaded list with installed list


MikhailArkhipov:
Try uninstallng base from Loaded tab
List replaced by installed packages list

Issue #1472, opened 2016-04-06
Tooltips are not visible on Update and Install buttons


MikhailArkhipov:

MikhailArkhipov:
Needs textblock around multi-binding

```
                                    <TextBlock>
                                        <TextBlock.Text>
                                            <MultiBinding StringFormat="{x:Static components:Resources.PackageManager_InstallButtonToolTip}">
                                                <Binding Path="Name" />
                                                <Binding Path="LatestVersion" />
                                            </MultiBinding>
                                        </TextBlock.Text>
                                    </TextBlock>

```

MikhailArkhipov:
Ditto on Load and Unload

Issue #1471, opened 2016-04-06
Package manager: Load library doesn't work


huguesv:
Looks like UI may be passing in a repo path instead of a library path:
```
Error in library("ggplot2", lib.loc = "/src/contrib") : 
  no library trees found in 'lib.loc'
> 
```

MikhailArkhipov:
It is passing    repo indeed
```
 await _packageManager.LoadPackageAsync(package.Name, package.RepositoryUri?.AbsolutePath.ToRPath());
```

MikhailArkhipov:
Just needs this

```
            var libPath = await GetLibPath();
            await _packageManager.LoadPackageAsync(package.Name, libPath);
```

Issue #1458, opened 2016-04-05
pkg manager: says name twice, wastes space


smortaz:
on the window title it says:

R Package Manager
Package Manager                  <Looooots of wasted white space>

i think users will understand it's the pkg manager, from the title and the look of the form...  removing the 2nd Package Manager also enables more actual pkgs to be displayed on the form.

huguesv:
:+1:

PR #1477, opened 2016-04-07
#1471 #1458 	#1475 #1472 #1455

MikhailArkhipov:


src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -211,7 +211,8 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             _coreShell.AssertIsOnMainThread();
             BeforeLoadUnload();
 
-            await _packageManager.LoadPackageAsync(package.Name, package.RepositoryUri?.AbsolutePath.ToRPath());
+            var libPath = await GetLibPath();
+            await _packageManager.LoadPackageAsync(package.Name, libPath);
```
AlexanderSher: No, it should be `await _packageManager.LoadPackageAsync(package.Name, package.LibraryPath.ToRPath())`, otherwise it will fail for system packages.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -184,17 +184,17 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
 
         private async Task UninstallAsync(IRPackageViewModel package) {
             _coreShell.AssertIsOnMainThread();
-            if (_selectedTab == SelectedTab.InstalledPackages) {
-                IsLoading = true;
-            }
+            IsLoading = true;
```
AlexanderSher: Hm... So what will happen if user clicks `uninstall` and then switches to the `available` tab?

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml.cs
```diff
@@ -51,7 +51,10 @@ public partial class PackageList : UserControl {
         }
 
         private void List_PreviewKeyUp(object sender, KeyEventArgs e) {
-            
+            HandleDefaultAction();
```
AlexanderSher: I think it should be 
```{csharp}
if (e.KeyCode == Keys.Enter) {
    HandleDefaultAction();
}
```

MikhailArkhipov: Totally

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: Makes sense

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
AlexanderSher: Or, user can call this method directly from the `available` tab. What bug are you fixing with this change?

MikhailArkhipov: When user uninstalls from loaded packages the tab was never populated again (IsLoading was not set) and the list was not updated.

MikhailArkhipov: Lemme check these schenarios

MikhailArkhipov: Switching to available quickly seems to work since I think, available requests state and since uninstall process already started, enumeration package state happens well after uninstallation started. Although yes, theoretically, depending on timing, state may not be correct. I think maybe model update should fire event on main thread when collections change so active tab can update itself. Parser does this guaranteeing that when tree change event fires it will be processed in UI thread i.e by the UI element that is currently active.

MikhailArkhipov: Or maybe make a method similar to AfterLoadUnload that will update active tab like AfterUninstall since uninstall affects every tab

MikhailArkhipov: So code gets confused when something didnt happen. For example, a dependent package that user tries to uninstall won't get unloaded or uninstalled. So similarly to loading where sequence is always available -> installed -> loaded, uninstall should go in reverse. Such as, first try and unload the package and then check if it was really unloaded. If it was, then proceed with uninstall otherwise bail. I changed code to this extent and the results appears to be more predictable.

AlexanderSher: No, it affects only `installed` and `loaded` tabs - `available` tab isn't changed, only some of its items will change `IsInstalled` and `IsLoaded` values, but the list will be the same.
So yes, instead of 
```{csharp}
if (_selectedTab == SelectedTab.InstalledPackages) {
    IsLoading = true;
}
```
it should be 
```{csharp}
if (_selectedTab == SelectedTab.InstalledPackages || _selectedTab == SelectedTab.LoadedPackages) {
    IsLoading = true;
}
```

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -186,15 +186,23 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
+            }
 
-            IsLoading = false;
-            if (_selectedTab == SelectedTab.InstalledPackages) {
-                ReplaceItems(_installedPackages);
-            } else if(_selectedTab == SelectedTab.LoadedPackages) {
-                ReplaceItems(_loadedPackages);
+            if (!package.IsLoaded) {
+                var libPath = await GetLibPath();
+                await _packageManager.UninstallPackageAsync(package.Name, libPath);
```
AlexanderSher: I think here it also should be `await _packageManager.UninstallPackageAsync(package.Name, package.LibraryPath.ToRPath())`

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: Available tab also shows loaded state. So it is affected when previously loaded package is uninstalled. I've seen when uninstall from Loaded and switching to Available still displays package as loaded.

MikhailArkhipov: OK, if state will change then you are right

AlexanderSher:
:+1:

PR #1480, opened 2016-04-07
Fix #1416: After loading PM should set selection to the first item

### LINKED ISSUE #1416: After loading PM should set selection to the first item
Currently it doesn't so right panel stays empty

AlexanderSher:


MikhailArkhipov:
:+1:

PR #1481, opened 2016-04-07
 #1449: R Toolbar is not visible by default

MikhailArkhipov:


src/Package/Impl/Packages/BasePackage.cs
```diff
@@ -58,6 +61,14 @@ public abstract class BasePackage<TLanguageService> : VisualStudio.Shell.Package
             foreach (var commmand in CreateMenuCommands()) {
                 menuCommandService.AddCommand(commmand);
             }
+
+            var dte = VsAppShell.Current.GetGlobalService<DTE2>(typeof(DTE));
+            var cbs = dte?.CommandBars;
+            var cb = dte?.CommandBars["R Toolbar"] as CommandBar;
```
AlexanderSher: Do we have this string defined anywhere as a constant?

src/Package/Impl/Utilities/ViewUtilities.cs
```diff
@@ -80,9 +80,11 @@ public static class ViewUtilities {
         }
 
         public static void SaveFile(this ITextView textView) {
-            RunningDocumentTable rdt = new RunningDocumentTable(RPackage.Current);
-            string filePath = textView.TextBuffer.GetFilePath();
-            rdt.SaveFileIfDirty(filePath);
+            if (textView != null) {
+                RunningDocumentTable rdt = new RunningDocumentTable(RPackage.Current);
+                string filePath = textView.TextBuffer.GetFilePath();
```
AlexanderSher: With the change above, `filePath` can be `null`. Will `SaveFileIfDirty` handle it?

src/Package/Impl/Packages/BasePackage.cs
MikhailArkhipov: in VSCT only so it is not directly importable. VSCT can handle C++ .h files, but that's not for C#.

src/Package/Impl/Utilities/ViewUtilities.cs
MikhailArkhipov: Yes

MikhailArkhipov: Actually, let me add another check

AlexanderSher:
:+1:

PR #1484, opened 2016-04-07
Fix #1136: Create debugger data inspection tests 

### LINKED ISSUE #737: Variable Explorer doesn't show name of element in 1-element vector
Start withL
```r
 x <- 123
```
At this point, VE shows `x` with value `123`, not expandable - as expected. Now do:
```r
names(x) <- 'abc'
```
VE still doesn't allow to expand `x` after this, and so there's no way to see the name of the element. On the other hand, in the debugger, the same variable will now show a single child item, exposing the name.

### LINKED ISSUE #1136: Create debugger data inspection tests

int19h:
(and fixes to various minor code issues exposed by the tests)

Add value tests for promises and active bindings.

Add value tests for children of evaluation results.

Remove duplicate JSON test data.

Use ShouldBeOfType instead of As where appropriate, because the latter doesn't assert the type.

Fix expression for slots to enable assignment.

Always sort children of environments - not just for frames.

Add synthetic parent environment nodes to environments.

Disable BreakContinue test.

Fix #737: Variable Explorer doesn't show name of element in 1-element vector

### DIFF
--- a/src/Common/Core/Test/Match/MatchAny.cs
+++ b/src/Common/Core/Test/Match/MatchAny.cs
@@ -2,25 +2,58 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using FluentAssertions.Formatting;
 
 namespace Microsoft.Common.Core.Test.Match {
     public class MatchAny<T> : IEquatable<T> {
         public static readonly MatchAny<T> Instance = new MatchAny<T>();
 
-        public bool Equals(T other) {
-            return true;
+        private readonly Func<T, bool> _condition;
+        private readonly Func<string> _toString;
+        private readonly object _wrapped;
+
+        static MatchAny() {
+            FluentAssertions.Formatting.Formatter.AddFormatter(new Formatter());
         }
 
-        public override bool Equals(object obj) {
-            return true;
+        private MatchAny(Func<T, bool> condition, object wrapped) {
+            _condition = condition;
+            _toString = wrapped.ToString;
+            _wrapped = wrapped;
         }
 
-        public override int GetHashCode() {
-            return 0;
+        public MatchAny() {
+            _condition = _ => true;
+            _toString = () => "<any>";
         }
 
-        public override string ToString() {
-            return "<any>";
+        public MatchAny(Expression<Func<T, bool>> condition) {
+            _condition = condition.Compile();
+            _toString = () => $"<satisfies {condition.ToString()}>";
+        }
+
+        public static MatchAny<T> ThatMatches<TOther>(IEquatable<TOther> match) {
+            return new MatchAny<T>(x => x is TOther && match.Equals((TOther)(object)x), match);
+        }
+
+        public bool Equals(T other) =>
+            _condition(other);
+
+        public override bool Equals(object obj) =>
+            (obj == null || obj is T) && Equals((T)obj);
+
+        public override int GetHashCode() => 0;
+
+        public override string ToString() => _toString();
+
+        private class Formatter : IValueFormatter {
+            public bool CanHandle(object value) =>
+                (value as MatchAny<T>)?._wrapped != null;
+
+            public string ToString(object value, bool useLineBreaks, IList<object> processedObjects = null, int nestedPropertyLevel = 0) =>
+                FluentAssertions.Formatting.Formatter.ToString(((MatchAny<T>)value)._wrapped, useLineBreaks, processedObjects, nestedPropertyLevel);
         }
     }
 }
--- a/src/Common/Core/Test/Match/MatchAny.cs
+++ b/src/Common/Core/Test/Match/MatchAny.cs
@@ -34,8 +34,9 @@ namespace Microsoft.Common.Core.Test.Match {
             _toString = () => $"<satisfies {condition.ToString()}>";
         }
 
-        public static MatchAny<T> ThatMatches<TOther>(IEquatable<TOther> match) {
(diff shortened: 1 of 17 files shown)


src/Common/Core/Test/Match/MatchAny.cs
```diff
@@ -2,25 +2,58 @@
...
-            return true;
+        public MatchAny(Expression<Func<T, bool>> condition) {
+            _condition = condition.Compile();
+            _toString = () => $"<satisfies {condition.ToString()}>";
         }
 
-        public override int GetHashCode() {
-            return 0;
+        public static MatchAny<T> ThatMatches<TOther>(IEquatable<TOther> match) {
+            return new MatchAny<T>(x => x is TOther && match.Equals((TOther)(object)x), match);
```
AlexanderSher: I don't like this conversion. If you add constraint `where TOther : T` you wouldn't need it.

src/Common/Core/Test/Match/MatchAny.cs
```diff
@@ -2,25 +2,58 @@
...
+        public override int GetHashCode() => 0;
+
+        public override string ToString() => _toString();
+
+        private class Formatter : IValueFormatter {
+            public bool CanHandle(object value) =>
+                (value as MatchAny<T>)?._wrapped != null;
+
+            public string ToString(object value, bool useLineBreaks, IList<object> processedObjects = null, int nestedPropertyLevel = 0) =>
+                FluentAssertions.Formatting.Formatter.ToString(((MatchAny<T>)value)._wrapped, useLineBreaks, processedObjects, nestedPropertyLevel);
```
AlexanderSher: Why do you use `_wrapped` instead of `_toString`?

int19h: Because it needs to delegate to the formatter of the wrapped object, which is not necessarily the same as its `ToString`.

src/Common/Core/Test/Match/MatchAny.cs
int19h: Indeed. This was a leftover from a different experiment that didn't quite play out.

src/Common/Core/Test/Match/MatchElements.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public override bool Equals(object obj) =>
+            Equals(obj as IEnumerable<T>);
+
+        public override int GetHashCode() {
+            return 0;
+        }
+
+        public override string ToString() {
+            return "<any>";
```
AlexanderSher: ?

src/Common/Core/Test/Match/MatchMembers.cs
```diff
@@ -23,12 +23,20 @@ public class MatchMembers<T> : IEquatable<T> {
...
+            // If selector is of the form x => x.Member, which is the most common kind, then shorten it to Member.
+            // Otherwise, just use the whole body of the lambda.
+            var member = memberSelector.Body as MemberExpression;
+            if (member != null && member.Expression is ParameterExpression) {
+                _memberName = member.Member.Name;
+            } else {
+                _memberName = memberSelector.ToString();
+            }
+
+            _memberName = (memberSelector.Body as MemberExpression)?.Member.Name ?? memberSelector.ToString();
```
AlexanderSher: This line overrides what was set by the `if/else` above

src/Debugger/Engine/Impl/AD7Property.cs
```diff
@@ -55,10 +55,21 @@ public AD7Property(AD7Property parent, DebugEvaluationResult result, bool isSynt
...
         }
 
-        private IReadOnlyList<DebugEvaluationResult> CreateChildren() {
-            return TaskExtensions.RunSynchronouslyOnUIThread(ct => (EvaluationResult as DebugValueEvaluationResult)?.GetChildrenAsync(PrefetchedFields, ChildrenMaxLength, ReprMaxLength, ct))
-                ?? new DebugEvaluationResult[0];
-        }
+        private IReadOnlyList<DebugEvaluationResult> CreateChildren() =>
+            TaskExtensions.RunSynchronouslyOnUIThread(async ct => {
+                var valueResult = EvaluationResult as DebugValueEvaluationResult;
+                if (valueResult != null) {
```
AlexanderSher: You can invert this `if` to reduce nesting

src/Debugger/Test/ValuesTest.cs
```diff
@@ -144,5 +213,213 @@ public class ValuesTest : IAsyncLifetime {
...
+            new ChildrenDataRow("setClass(\"C\", contains = \"list\", slots = list(y = \"numeric\"))(list(x = 1), y = 2)", 1, nameCount: 1, slotCount: 2, attrCount: 3) {
+                { "@.Data", "list(1)" },
+                { "@y", "2" },
+                { "$x", "1" },
+            },
+        };
+
+        [CompositeTest]
+        [Category.R.Debugger]
+        [MemberData("ChildrenData")]
```
AlexanderSher: `nameof(ChildrenData)`

src/Common/Core/Test/Match/MatchElements.cs
int19h: Totally forgot about it. Will fix.

src/Common/Core/Test/Match/MatchMembers.cs
int19h: *facepalm*

int19h: Looks like this won't need a full fledged formatter since it's enumerable. I just need to not override `ToString`, then the default formatter should kick in.

AlexanderSher:
:+1:

Issue #1390, opened 2016-03-31
Use packages.rds for additional package data instead of parsing web page


huguesv:
packages.rds has the full metadata for packages in the repository, including stuff that is missing from PACKAGES.gz.

It's 2254kb, and has the advantage of giving us all the data at once, rather than the parsing one page at a time when needed.

It's located at `/web/packages/packages.rds`

Examples:
```
https://cran.revolutionanalytics.com/web/packages/packages.rds
http://cloud.r-project.org/web/packages/packages.rds
```

For mran specifically, there's a json version of that file:
```
https://mran.microsoft.com/packagedata/allpackages.json
```

```
url <- "https://cran.revolutionanalytics.com/web/packages/packages.rds"
tf <- file.path(tempdir(), "packages.rds")
download.file(url, destfile = tf, mode = "wb")
x <- as.data.frame(readRDS(tf))
head(x[, c("Package", "Title")])
```

huguesv:
Note that fetching all of this data in advance, rather than one page at a time, is what we need in order to make our search be able to search within title and description of packages.

PR #1500, opened 2016-04-08
Package details

huguesv:
Retrieve details for all packages within the initial call to get available packages.

Fixes https://github.com/Microsoft/RTVS/issues/1390


huguesv:
The .xaml change to display the title is placeholder, I don't have access to the nuget xaml.
image

MikhailArkhipov:
Would be nice to align description line with the package name

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml
```diff
@@ -83,18 +83,21 @@
...
                         <!-- title & summary-->
-                        <TextBlock Grid.Column="2" FontWeight="Bold" TextTrimming="CharacterEllipsis" TextWrapping="NoWrap" VerticalAlignment="Center" Text="{Binding Path=Name, Mode=OneWay}"
+                        <StackPanel Grid.Column="2">
+                            <TextBlock FontWeight="Bold" TextTrimming="CharacterEllipsis" TextWrapping="NoWrap" VerticalAlignment="Center" Text="{Binding Path=Name, Mode=OneWay}"
                                    FontSize="{Binding ElementName=_self, Path=FontSize, Converter={x:Static wpf:Converters.FontScale122}}" Margin="4,0,0,0">
-                            <TextBlock.ToolTip>
-                                <ToolTip Visibility="{Binding Path=Title, Mode=OneWay, Converter={x:Static wpf:Converters.NullOrEmptyIsCollapsed}}">
-                                    <TextBlock Style="{StaticResource TooltipStyle}">
+                                <TextBlock.ToolTip>
+                                    <ToolTip Visibility="{Binding Path=Title, Mode=OneWay, Converter={x:Static wpf:Converters.NullOrEmptyIsCollapsed}}">
```
MikhailArkhipov: Since title is already displayed in the list, I'd bind title to description. Although for the tooltip to look OK we would need to add line breaks every ~100 characters.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -59,24 +59,6 @@ internal class RPackageManager : IRPackageManager {
             return await GetPackagesAsync(async (eval) => await eval.AvailablePackages());
```
MikhailArkhipov: Should we perhaps prefetch data when package loads? Or is list population is now fast enough?

src/R/Components/Impl/PackageManager/IRPackageManager.cs
```diff
@@ -41,30 +41,6 @@ public interface IRPackageManager : IDisposable {
         Task<IReadOnlyList<RPackage>> GetAvailablePackagesAsync();
```
MikhailArkhipov: BTW RPackage is the same name as our VS package. I guess at some point we may want to rename VS package to something else.

src/Debugger/Impl/rtvs/R/packages.R
```diff
@@ -1,27 +1,86 @@
...
+  as.list(.libPaths())
 }
 
-packages.loaded <- function() {
-    as.list(.packages())
+download.packages.rds <- function(repo.url) {
+  # download the repository's packages.rds and return path to the temp file
+  pkg.rds.url <- gsub("src/contrib", "web/packages/packages.rds", repo.url)
+  pkg.rds.path <- tempfile('pkg', fileext = '.rds')
+  suppressWarnings(download.file(pkg.rds.url, destfile = pkg.rds.path, quiet = TRUE, mode = "wb"))
```
MikhailArkhipov: Can it throw or otherwise fail badly if the file is locked or TEMP is not set?

MikhailArkhipov: what happens offline?

MikhailArkhipov: I mean, if we are offline and then go online, can we refresh automatically or can user somehow do it?

huguesv: Yes it can throw, and that's what the tryCatch is for.  It will return as much data as it can, if it can't get the additional details, then it will return the basic info that R's available.packages gives us.
I'm not concerned with usage of temp, as R's available packages makes us of it too.
There was a JSON deserializing problem when returning empty list of package (ie offline) which I just added a fix for.

huguesv: The offline then later online scenario may be a good reason to give the user the ability to refresh.  Maybe if they close pkg mgr window then later re-open it, we could refresh at that time.  Of course we can make it refresh when the user resets the session, but that's not always an option.  I just don't want it to be too in your face big refresh button, I don't want the user to think they get stale data unless they press that.

int19h:
:+1:

huguesv:
Adjusted tooltip with description.
image

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
huguesv: That's something to consider if people feel it's not fast enough.  If I have package manager opened in my last VS session, when I start VS it opens in the "installed" tab, I immediately click on "available" and it's about 3 secs before all available show up.

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml
huguesv: I changed it, see my new screenshot below.  I didn't do any fancy line breaks, I think it looks fine.

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml
```diff
@@ -83,18 +83,21 @@
...
-                                    </TextBlock>
-                                </ToolTip>
-                            </TextBlock.ToolTip>
-                        </TextBlock>
+                                        <Run Text="{Binding Path=Description, Mode=OneWay}" />
+                                        </TextBlock>
+                                    </ToolTip>
+                                </TextBlock.ToolTip>
+                            </TextBlock>
+                            <TextBlock Text="{Binding Path=Title, Mode=OneWay}" Margin="4,0,0,0"></TextBlock>
```
AlexanderSher: Empty tag

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageViewModel.cs
```diff
@@ -38,6 +38,14 @@ internal class RPackageViewModel : BindableBase, IRPackageViewModel {
...
                 RepositoryUri = repositoryUri,
                 RepositoryText = repositoryUri != null ? null : package.Repository,
+                Title = package.Title.NormalizeWhitespace(),
+                Description = package.Description.NormalizeWhitespace(),
+                Built = package.Built,
+                Urls = package.URL?.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries)
+                    .Select(s => s.Trim())
+                    .ToArray() ?? new string[0],
+                Authors = package.Author.NormalizeWhitespace(),
+                HasDetails = true,
```
AlexanderSher: Looks like we don't need this property anymore

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
MikhailArkhipov: 3 sec is fine

MikhailArkhipov:
:+1:

smortaz:
that looks great!!!

is there room to bump the pkg name one size up?  the name should ideally be
in a slightly bigger font than the description below it.  not a hugues deal
tho ;)

Issue #1413, opened 2016-04-01
Package Manager UI: Update button update logic needs to properly compare versions


huguesv:
Right now it's not checking that the version available is greater than the installed version, it's just checking that it's different (so we'll enable update if it's a downgrade).

huguesv:
Version strings sometimes have weird `-`, which appear to be treated as if they were `.` for purpose of comparison.

PR #1504, opened 2016-04-08
Add proper package version comparison.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/1413

I left the data type on the view model as a string, it could be changed to RPackageVersion if we like that better.  I remember it used to be a RPackageVersion instance and got changed to a string at some point, but I'm not sure why.


AlexanderSher:
:+1:

Issue #1511, opened 2016-04-08
Encoding list in the import dialog may appear empty


MikhailArkhipov:
Looks like population is async and may not complete timely

JeongPyoHam:
When encoding is empty, read.csv uses default option. UI should show default value such as 'Automatic' rather than leaving the combobox empty.

MikhailArkhipov:
Have the fix. Automatic is already in the list, it is DoNotWait on the encoding list population was trouble. Selected encoding is used to read the file.

Issue #1507, opened 2016-04-08
VS crash: ObjectDisposedException for MarshalingWindowFrame


int19h:
Not clear if all of these are necessary, but the sequence was as follows:

1. Start VS; no RTVS toolwindows are visible (not even REPL).
2. Open Package Manager.
3. Open "Installed" tab, and search for "rtf" (package was already installed prior).
4. Click "Uninstall".
5. Open REPL window.

(I believe the key part here is trying to use Package Manager without having the REPL)

ObjectDisposedException for object name 'MarshalingWindowFrame' happens at this point. Stack doesn't seem to be particularly useful:
```
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at Microsoft.VisualStudio.ErrorHandler.ThrowOnFailure(Int32 hr, Int32[] expectedHRFailure)
   at Microsoft.VisualStudio.ErrorHandler.ThrowOnFailure(Int32 hr)
   at Microsoft.VisualStudio.R.Package.Shell.VisualComponentToolWindowAdapter`1.<>c__DisplayClass13_0.<Show>b__0()
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Threading.DispatcherOperation.Invoke()
   at System.Windows.Threading.Dispatcher.ProcessQueue()
[...]

PR #1517, opened 2016-04-08
#1511 #1507

MikhailArkhipov:
#1511 Encoding list in the import dialog may appear empty
#1507 VS crash: ObjectDisposedException for MarshalingWindowFrame

### DIFF
--- a/src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
+++ b/src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
@@ -59,9 +59,11 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.DataImport {
             [Package.Resources.ImportData_UseNumber] = "NULL"
         };
 
+        private Task _encodingsInitTask;
+
         public ImportDataWindow() {
             InitializeComponent();
-            SetEncodingComboBoxAsync().DoNotWait();
+            _encodingsInitTask = SetEncodingComboBoxAsync();
         }
 
         public ImportDataWindow(string filePath, string name)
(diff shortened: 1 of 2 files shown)


src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -162,6 +164,12 @@ public ImportDataWindow(string filePath, string name)
         }
 
         private async Task PreviewAsync() {
+            if(_encodingsInitTask == null) {
```
int19h: It looks like it can never be null?

int19h:
:+1:

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
MikhailArkhipov: It is null when InitializeComponent creates combobox and it fires selection change

src/Package/Impl/PackageManager/VsRPackageManagerVisualComponent.cs
```diff
@@ -27,13 +30,9 @@ internal class VsRPackageManagerVisualComponentContainerFactory : ToolWindowPane
...
-            var workflow = _workflowProvider.GetOrCreate();
-            if (workflow.ActiveWindow == null) {
-                workflow
-                    .GetOrCreateVisualComponent(_componentContainerFactory)
-                    .ContinueOnRanToCompletion(w => w.Container.Show(false));
-            }
-
+            var o = new object();
+            var shell = VsAppShell.Current.GetGlobalService<IVsUIShell>(typeof(SVsUIShell));
+            shell.PostExecCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdShowReplWindow, 0, ref o);
```
AlexanderSher:
```{csharp}
var workflow = _workflowProvider.GetOrCreate();
if (workflow.ActiveWindow == null) {
    VsAppShell.Current.DispatchOnMainThreadAsync(() => workflow.GetOrCreateVisualComponent(_componentContainerFactory));
}
```

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
AlexanderSher: yep, it is safe to just call `await _encodingsInitTask;`

PR #1518, opened 2016-04-08
Fix XAML bindings

AlexanderSher:
+ Adjust fix for #1507: VS crash: ObjectDisposedException for MarshalingWindowFrame (we don't actually need to show REPL when PM is loaded - we just need to start it).


MikhailArkhipov:
👍

Issue #1509, opened 2016-04-08
After installing, Uninstall button in details is visible but greyed out


huguesv:
![image](https://cloud.githubusercontent.com/assets/1696845/14398120/9b126b70-fd96-11e5-9f16-e54eb68fecd1.png)

In this case, it also failed to appear in the list on the left.

MikhailArkhipov:
Specifically reproes in Available after install:

Uninstall button in details is disabled
Uninstall button in list is visible but does nothing

Switching to Installed tab and back recovers

MikhailArkhipov:
Sounds like 'IsChanging' property remails true

PR #1525, opened 2016-04-11
#1509 After installing, Uninstall button in details is visible but gr…

MikhailArkhipov:
…eyed out

### DIFF
--- a/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
@@ -143,7 +143,10 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
             }
 
             package.IsChanging = true;
-            DispatchOnMainThread(() => InstallAsync(package));
+            DispatchOnMainThread(async () => {
+                await InstallAsync(package);
+                package.IsChanging = false;
+            });
         }
 
         public void Update(IRPackageViewModel package) {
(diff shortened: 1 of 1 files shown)


src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -157,7 +160,10 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             }
 
             package.IsChanging = true;
-            DispatchOnMainThread(() => InstallAsync(package));
+            DispatchOnMainThread(async () => {
+                await InstallAsync(package);
+                package.IsChanging = false;
```
AlexanderSher: Let's move this directly inside `InstallAsync`

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -182,7 +188,10 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             }
 
             package.IsChanging = true;
-            DispatchOnMainThread(() => UninstallAsync(package));
+            DispatchOnMainThread(async () => {
+                await UninstallAsync(package);
+                package.IsChanging = false;
```
AlexanderSher: Same here - let's move it to the end of `UninstallAsync`

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: OK. Install/Uninstall have multiple exit points though due to the prompt

MikhailArkhipov: Done

AlexanderSher:
:+1:

PR #1527, opened 2016-04-11
Fix #1519 Crash in PlotContentProvider

### LINKED ISSUE #1519: Crash in PlotContentProvider
```
System events:
Time: 2016-04-08T14:40:45
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.Threading.Tasks.TaskCanceledException
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.VisualStudio.R.Package.Plots.PlotContentProvider+<ApplyNewSize>d__25.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.VisualStudio.R.Package.Plots.PlotContentProvider+<RSession_Connected>d__4.MoveNext()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__6_1(System.Object)
   at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()
```

huguesv:
This is one case where we didn't use DoNotWait, so I added the equivalent catch handlers.

### DIFF
--- a/src/Package/Impl/Plots/PlotContentProvider.cs
+++ b/src/Package/Impl/Plots/PlotContentProvider.cs
@@ -33,7 +33,7 @@ namespace Microsoft.VisualStudio.R.Package.Plots {
             if (_lastPixelWidth >= 0 && _lastPixelHeight >= 0) {
                 try {
                     await ApplyNewSize();
-                } catch (OperationCanceledException) { } catch (MessageTransportException) { } catch (RException) { }
+                } catch (OperationCanceledException) { } catch (MessageTransportException) { }
             }
 
             VsAppShell.Current.DispatchOnUIThread(() => {
@@ -277,7 +277,6 @@ namespace Microsoft.VisualStudio.R.Package.Plots {
             // in an Microsoft.R.Host.Client.RException, and we don't need to do anything with them,
             // as the user can see them in the REPL.
             task.SilenceException<MessageTransportException>()
-                .SilenceException<Microsoft.R.Host.Client.RException>()
                 .DoNotWait();
         }
     }
--- a/src/Package/Impl/Plots/PlotContentProvider.cs
+++ b/src/Package/Impl/Plots/PlotContentProvider.cs
@@ -31,7 +31,9 @@ namespace Microsoft.VisualStudio.R.Package.Plots {
         private async void RSession_Connected(object sender, EventArgs e) {
             // Let the host know the size of plot window
             if (_lastPixelWidth >= 0 && _lastPixelHeight >= 0) {
-                await ApplyNewSize();
+                try {
+                    await ApplyNewSize();
+                } catch (OperationCanceledException) { } catch (MessageTransportException) { } catch (RException) { }
             }
 
             VsAppShell.Current.DispatchOnUIThread(() => {
--- a/src/Debugger/Engine/Impl/AD7BoundBreakpoint.cs
+++ b/src/Debugger/Engine/Impl/AD7BoundBreakpoint.cs
@@ -82,7 +82,6 @@ namespace Microsoft.R.Debugger.Engine {
                                 // later, but as soon as it's actually possible.
                                 DebugBreakpoint.DeleteAsync()
                                     .SilenceException<MessageTransportException>()
-                                    .SilenceException<RException>()
                                     .DoNotWait();
                             } else {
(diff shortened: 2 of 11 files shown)


src/Package/Impl/Plots/PlotContentProvider.cs
```diff
@@ -31,7 +31,9 @@ internal sealed class PlotContentProvider : IPlotContentProvider {
         private async void RSession_Connected(object sender, EventArgs e) {
             // Let the host know the size of plot window
             if (_lastPixelWidth >= 0 && _lastPixelHeight >= 0) {
-                await ApplyNewSize();
+                try {
+                    await ApplyNewSize();
+                } catch (OperationCanceledException) { } catch (MessageTransportException) { } catch (RException) { }
```
int19h: Do you really want to catch `RException` here? It would indicate an actual eval failure, which is usually a bug and not expected. Or is there some known case where eval can fail here and it is expected?

huguesv: It's for R library errors like "invalid graphics state" and such.

From DoNotWait:
```
        public static void DoNotWait(System.Threading.Tasks.Task task) {
            // Errors like invalid graphics state which go to the REPL stderr will come back
            // in an Microsoft.R.Host.Client.RException, and we don't need to do anything with them,
            // as the user can see them in the REPL.
            task.SilenceException<MessageTransportException>()
                .SilenceException<Microsoft.R.Host.Client.RException>()
                .DoNotWait();
        }
```

int19h: I don't think we're still doing it, actually, if you're doing interactions - i.e. stderr output should no longer translate to `RException` from `RespondAsync`. @AlexanderSher, is that right?

AlexanderSher: Yes, error output just printed out in different color.

int19h: Amusingly, nothing is actually throwing `RException` anywhere in the code now.

huguesv: Shall I remove `RException` class and all the catches for it?

int19h: I would like to appropriate that exception for failing evals sometime later. Right now `EvaluateAsync` never throws but rather just returns `REvaluationResult`, which you then have to check... and every place that uses it ends up doing that same check over and over again (or sometimes forget to do so, resulting in errors being silently swallowed). So I'd like to add some helper for `EvaluateAsync` that would just throw, and this would be the type that it'll throw.

But all places that currently catch `RException` are no-ops, and are a leftover from the time when interactions did throw. So it should be safe to remove all the catches.

huguesv: Okay, will do that.

huguesv: Done

huguesv:
Removed all references to RException, but left that exception class in as requested.

int19h:
:+1:

Issue #1513, opened 2016-04-08
Variable names with non-ASCII characters produce errors


huguesv:
I imported bikes.csv (set separator to semicolon, keep everything else default), and tried to expand the node in VE, go this message:
![image](https://cloud.githubusercontent.com/assets/1696845/14398230/6b42e554-fd97-11e5-9b23-f817b40f2e6a.png)

\pytools\r\TestData\bikes.csv

int19h:
The issue is with variable/children names with Latin-1 characters in general. A very basic repro is:
```r
ü <- 1
```

PR #1528, opened 2016-04-11
Add unit tests for #1513.

int19h:
Note that this is just the test - the actual fix is https://github.com/Microsoft/R-Host/pull/54


AlexanderSher:
:+1:

Issue #1489, opened 2016-04-07
Package Manager will send library() and detach() commands to REPL


int19h:
When the green circle next to a package is clicked, the corresponding package is loaded or unloaded. However, the fact that it does so is not reflected in any obvious way in the UI, because loading & unloading is done directly via `RSession`, and so the corresponding expression is not visible in the REPL.

It would be better if loading/unloading would submit the code to the REPL window, such that it (and any output that it produces) is immediately visible.

int19h:
RS does submit it to the REPL - e.g. when you click on "abc" in the package manager, you see something like this:
```
> library("abc", lib.loc="~/R/win-library/3.2")
Loading required package: abc.data
Loading required package: nnet
Loading required package: quantreg
Loading required package: SparseM

Attaching package: ‘SparseM’

The following object is masked from ‘package:base’:

    backsolve

Loading required package: MASS
Loading required package: locfit
locfit 1.5-9.1 	 2013-03-22
```
while in RTVS, you don't see the first line. If the package doesn't produce any warnings on load, you don't see any output at all - it just silently updates `search()`.

smortaz:
this is important because i want to see my record of load/unload/etc. in the History window.

AlexanderSher:
Depends on #1469: Allow BeginInteractionAsync to notify RInteractiveEvaluator that response should be added to REPL 

The problem we have here is that Interactive Window doesn't allow to update its text buffer when its executing, so I can add command to the history, but I don't know if I can correctly add it to the Interactive Window.

MikhailArkhipov:
Not entirely convinced this is worth the risk.

MikhailArkhipov:
There are more places line this, opened Epic https://github.com/Microsoft/RTVS/issues/1534

PR #1529, opened 2016-04-11
Fix for #1489: Loading/unloading packages is not obvious in UI

AlexanderSher:
The main problem with this bug was in IInteractiveWindow, which has no waiting or blocking version of ExecuteInput.
To track the end of interaction, TaskCompletionSource instances are passed to the RInteractiveEvaluator. Since ProcessQueuedInput is always executed in UI thread, it is safe to assume that IInteractiveEvaluator.ExecuteCodeAsync will be called right after SetInputCompletionSources

### DIFF
--- a/src/Debugger/Engine/Impl/AD7BoundBreakpoint.cs
+++ b/src/Debugger/Engine/Impl/AD7BoundBreakpoint.cs
@@ -82,7 +82,6 @@ namespace Microsoft.R.Debugger.Engine {
                                 // later, but as soon as it's actually possible.
                                 DebugBreakpoint.DeleteAsync()
                                     .SilenceException<MessageTransportException>()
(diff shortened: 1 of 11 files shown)


MikhailArkhipov:
How do you access risks

AlexanderSher:
All tests except `VariableExplorer` have passed (probably we have broken them with recent changes). `RInteractiveEvaluator` doesn't wait for the any of the `TaskCompletionSource` instances, so the change should only affect PackageManager. Solution itself though becomes less controlled - instead of sending commands to the RHost directly, we rely on the `InteractiveWindow` which may fail for some reason before it calls `IInteractiveEvaluator.ExecuteCodeAsync`. In this case, `PackageManager` will be blocked until user types something inside interactive window or calls `SourceFile`.

Issue #1340, opened 2016-03-24
Non-ASCII output is garbled when R locale is not Latin-1


Marchen:
The content of the attached file is

"Language"
"English"
"日本語"

If I read the file using read.csv(), it returns

read.csv("test_cp932.txt")
  Language
1  English
2   “ú{ê

Adding fileEncoding = "CP932" had no effect.

[test_cp932.txt](https://github.com/Microsoft/RTVS/files/187791/test_cp932.txt)

int19h:
Is it the data that is garbled, or is it the output in the REPL? If you use `write.csv` on the data, is the output correct?

Marchen:
This problem only occurred in the REPL. If I use write.csv to file, the output data was correct. But if I print the result to the  R interactive, the result was garbled like

> write.csv(data)
"","Language"
"1","English"
"2","“ú{ê"

And I found this is not a problem of read.csv but a problem of showing data in data.frame because following code caused same problem.

> data.frame(lang="日本語")
    lang
1 “ú{ê

int19h:
Thank you. So it seems that this is the same problem as in #1211?

Marchen:
I'm not sure, actually. When I print a element of data.frame, it works correctly.

> df <- data.frame(lang = "日本語")
> df[1, 1]
[1] 日本語
Levels: 日本語

but when I print entire data.frame, it garbles Japanese characters.
> df
    lang
1 “ú{ê

Following part of print.data.frame() seems to cause garbling.
> m <- as.matrix(format.data.frame(x, digits = digits, 
            na.encode = FALSE))

but again, I'm not sure this is the same problem as in #1211.

int19h:
I wonder if it escapes things in a different way, and we handle one correctly, but not the other...

Could you try these two things. and give me the output for each:
```r
lapply(capture.output(df), utf8ToInt)

lapply(capture.output(df[1, 1]), utf8ToInt)
```

Marchen:
Your codes returned following.

lapply(capture.output(df), utf8ToInt)
[[1]]
[1]  32  32  32  32 108  97 110 103

[[2]]
[1] NA

lapply(capture.output(df[1, 1]), utf8ToInt)
[[1]]
[1] NA

[[2]]
[1] NA

int19h:
Tentative estimate - needs investigation to root cause.

int19h:
[...]

PR #1539, opened 2016-04-12
Unit test for #1340 fix

int19h:
and fixes to various minor issues caused by the change

PR for the actual fix is https://github.com/Microsoft/R-Host/pull/55

### DIFF
--- a/src/Debugger/Engine/Impl/AD7Property.cs
+++ b/src/Debugger/Engine/Impl/AD7Property.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Common.Core;
+using Microsoft.R.Host.Client;
 using Microsoft.R.Support.Settings;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Debugger.Interop;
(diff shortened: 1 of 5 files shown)


src/Debugger/Engine/Impl/AD7Property.cs
```diff
@@ -313,7 +314,7 @@ public AD7Property(AD7Property parent, DebugEvaluationResult result, bool isSynt
             if (fields.HasFlag(enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_VALUE)) {
                 if (valueResult != null) {
                     // TODO: handle radix
-                    dpi.bstrValue = valueResult.GetRepresentation().Deparse;
+                    dpi.bstrValue = valueResult.GetRepresentation().Deparse.ToUnicodeQuotes();
```
int19h: This one is not directly related to the fix, but it's another place where this call was missing where it should have been present.

src/Host/Process
```diff
@@ -1 +1 @@
-Subproject commit 9f82bae74ecfdff91681bf8569dd4c49d0063397
+Subproject commit 7d4a4112debaefb566860a21981e9a562e28db7b
```
int19h: I'll update it to the actual commit once the host part goes in.

src/Host/Client/Test/Session/RSessionTest.Output.cs
```diff
@@ -0,0 +1,58 @@
...
+
+                using (var inter = await _session.BeginInteractionAsync()) {
+                    await inter.RespondAsync("'日本語'\n");
+                }
+
+                output.ToString().Should().Be("[1] \"日本語\"\n");
+            }
+        }
+    }
+}
```
huguesv: Can you add an example of an inexact conversion that causes it to be escaped?

int19h: It already exists in JsonTest - it started failing, that's how I noticed :) it's this data:

https://github.com/Microsoft/RTVS/blob/131fdf77f4a9786d2d55755ff5b8e83c73ede4dc/src/Debugger/Test/JsonTest.cs#L63

"δ" is lowercase delta, which got mistranslated as "d".

huguesv:
:+1:

PR #1543, opened 2016-04-12
Fix #1530: Package Manager UI: Remove checkmark when package is loaded

### LINKED ISSUE #1530: Package Manager UI: Remove checkmark when package is loaded
Should be only filled green circle

AlexanderSher:
nocheckmark


huguesv:
:+1:

int19h:
:+1:

Issue #1532, opened 2016-04-11
Encodings in Data Import dialog don't work correctly


MikhailArkhipov:
We need map of R encoding names to .NET encoding names and code pages

MikhailArkhipov:
So general importing of non-ASCII data in R is pretty miserable... read.csv works with encoding="UTF-8" but that's pretty much it. Reading in encoding like KOI-8R (encoding="866") doesnt work. RS has same results. So - we can make this quite a bit better:

a. Showing list of encoding .NET handles with descriptive names and code pages. No need to show 'shift_jis", "shift-jis" and all their spelling variations. Japanese (Shift_JIS) - 932 is better.

b. Convert file to temp file in UTF-8 before calling read.csv on the temp file and import always in UTF-8. This appears to produce proper results without loss of data.

int19h:
Can you try this?
```r
read.csv(..., fileEncoding = 'cp866')
```

MikhailArkhipov:
Same. cp866 is the same as 866

![image](https://cloud.githubusercontent.com/assets/12820357/14451486/616a0038-003c-11e6-8bcb-c4dba0beb0a3.png)

MikhailArkhipov:
with UTF-8 conversion I properly get

![image](https://cloud.githubusercontent.com/assets/12820357/14451523/ab86da7e-003c-11e6-9913-d320a597d9d4.png)

And VE grid shows actual Russian

![image](https://cloud.githubusercontent.com/assets/12820357/14451543/c6327e5a-003c-11e6-97ec-62d73a7541e9.png)

See also http://stackoverflow.com/questions/18789330/r-on-windows-character-encoding-hell

MikhailArkhipov:
Now need to figure out why VE grid shows proper characters while data frame in preview shows ```<U+...>```

MikhailArkhipov:
Now, ```fileEncoding``` appears to work BUT produces odd error message

![image](https://cloud.githubusercontent.com/assets/12820357/14451604/460fddac-003d-11e6-8d79-708013d2799f.png)

MikhailArkhipov:
Also, default R reader doesn't like UTF-8 with BOM while with conversion everything works fine

MikhailArkhipov:
There

![image](https://cloud.githubusercontent.com/assets/12820357/14451943/d003844e-003f-11e6-98ef-886c7a5ab61d.png)

int19h:
[...]

PR #1544, opened 2016-04-12
#1532: Encodings in Data Import dialog don't work correctly

MikhailArkhipov:
Display list of .NET encodings
Convert files to UTF-8 and import via UTF-8
Fixed display of UTF-8 in Data Frame preview
Fixed threading/WPF control access issues
Added *.txt to import list
Added progress bar
Added R Encodings table - not used atm, but may come useful at some point

### DIFF
--- a/src/Host/Client/Impl/Encodings/REncodingsMap.cs
+++ b/src/Host/Client/Impl/Encodings/REncodingsMap.cs
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
-
+#if _NOT_USED_YET_
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -1080,3 +1080,4 @@ namespace Microsoft.R.Host.Client.Encodings {
     };
     }
 }
+#endif
--- a/src/Host/Client/Impl/Encodings/REncodingsMap.cs
+++ b/src/Host/Client/Impl/Encodings/REncodingsMap.cs
@@ -1,0 +1,1082 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.R.Host.Client.Encodings {
+    public static class REncodingsMap {
+        // https://msdn.microsoft.com/en-us/library/system.text.encoding.getencodings(v=vs.110).aspx
+        // https://www.ibm.com/support/knowledgecenter/SSMKHH_9.0.0/com.ibm.etools.mft.doc/ac00408_.htm
+        public static Encoding GetEncoding(string rEncodingName) {
+            var encSet = _encodingsMap.Keys.FirstOrDefault(x => x.Contains(rEncodingName));
+            if(encSet != null) {
+                try {
+                    return _encodingsMap[encSet].Invoke();
+                } catch(ArgumentException) { }
+            }
+            return Encoding.Default;
+        }
+
+        private static Dictionary<HashSet<string>, Func<Encoding>> _encodingsMap = new Dictionary<HashSet<string>, Func<Encoding>>() {
+        { new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {"UTF-8", "ibm-1208", "ibm-1209", "ibm-5304", "ibm-5305", "ibm-13496", "ibm-13497", "ibm-17592", "ibm-17593",
+            "windows-65001", "cp1208", "x-UTF_8J", "unicode-1-1-utf-8", "unicode-2-0-utf-8"},
+            () => { return Encoding.UTF8; }
+        },
+
+        { new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {"UTF-16", "ISO-10646-UCS-2", "ibm-1204", "ibm-1205", "unicode", "csUnicode", "ucs-2"},
+            () => { return Encoding.Unicode; }
+        },
+
+        { new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {"UTF-16BE", "x-utf-16be", "UnicodeBigUnmarked", "ibm-1200", "ibm-1201", "ibm-13488", "ibm-13489", "ibm-17584", "ibm-17585",
+             "ibm-21680", "ibm-21681", "ibm-25776", "ibm-25777", "ibm-29872", "ibm-29873", "ibm-61955", "ibm-61956",
+             "windows-1201", "cp1200", "cp1201", "UTF16_BigEndian"},
+            () => { return Encoding.BigEndianUnicode; }
+        },
+
+        { new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {"UTF-16LE", "x-utf-16le", "UnicodeLittleUnmarked", "ibm-1202", "ibm-1203", "ibm-13490", "ibm-13491", "ibm-17586", "ibm-17587",
+             "ibm-21682", "ibm-21683", "ibm-25778", "ibm-25779", "ibm-29874", "ibm-29875", "UTF16_LittleEndian", "windows-1200"},
+            () => { return Encoding.Unicode; }
+        },
+
+        { new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {"UTF-32", "ISO-10646-UCS-4", "ibm-1236", "ibm-1237", "csUCS4", "ucs-4"},
+            () => { return Encoding.UTF32; }
+        },
+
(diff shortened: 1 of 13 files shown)


src/Host/Client/Impl/Encodings/REncodingsMap.cs
```diff
@@ -0,0 +1,1082 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.R.Host.Client.Encodings {
+    public static class REncodingsMap {
```
AlexanderSher: Do you use this class anywhere?

MikhailArkhipov: Nope, as I mentioned in comments. May come useful at some point so - ifdef it out? Exclude?

src/Host/Client/Impl/Encodings/REncodingsMap.cs
```diff
@@ -0,0 +1,1082 @@
...
+            var encSet = _encodingsMap.Keys.FirstOrDefault(x => x.Contains(rEncodingName));
+            if(encSet != null) {
+                try {
+                    return _encodingsMap[encSet].Invoke();
+                } catch(ArgumentException) { }
+            }
+            return Encoding.Default;
+        }
+
+        private static Dictionary<HashSet<string>, Func<Encoding>> _encodingsMap = new Dictionary<HashSet<string>, Func<Encoding>>() {
```
int19h: All encodings in .NET also have names. This will be a bit cleaner and easier to add/remove things if it just does name-to-name mapping, and then the code that uses it can call `Encoding.GetEncoding` to get the actual thing.

src/Host/Client/Impl/Encodings/REncodingsMap.cs
AlexanderSher: ifdef probably

MikhailArkhipov: OK

src/Host/Client/Impl/Encodings/REncodingsMap.cs
MikhailArkhipov: Since it is unused atm, I'll ifdef it out

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -309,11 +256,131 @@ class DataFramePreviewRowItem {
...
+                                    var newProgressValue = 90 * (double)read / (double)sr.BaseStream.Length;
+                                    if (newProgressValue - progressValue >= 10) {
+                                        progressValue = newProgressValue;
+                                        await ReportProgress(progressValue);
+                                    }
+                                }
+                            }
+                        }
+                    }
+                } catch (IOException ex) {
```
int19h: Also `UnauthorizedAccessException`.

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -102,89 +105,63 @@ public ImportDataWindow(string filePath, string name)
...
-                    var grid = await GridDataSource.GetGridDataAsync(expression, null);
-
-                    PopulateDataFramePreview(grid);
-                    ErrorBlock.Visibility = Visibility.Collapsed;
-                    DataFramePreview.Visibility = Visibility.Visible;
-                } catch (Exception ex) {
-                    OnError(ex.Message);
+            int cp = GetSelectedValueAsInt(EncodingComboBox);
+            PreviewFileContent(FilePathBox.Text, cp);
+            await ConvertToUtf8(FilePathBox.Text, cp, MaxPreviewLines);
```
AlexanderSher: Bad idea awaiting inside void method. If you want to insure that async method is executed in UI thread, use `ICoreShell.DispatchOnMainThreadAsync`

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -309,11 +256,131 @@ class DataFramePreviewRowItem {
...
+
+        private void DeleteTempFile() {
+            try {
+                if (!string.IsNullOrEmpty(_tempFilePath) && File.Exists(_tempFilePath)) {
+                    File.Delete(_tempFilePath);
+                }
+            } catch (UnauthorizedAccessException) { }
+        }
+
+        private async void DoDefaultAction() {
```
int19h: Since this is `async void` and will crash if any exceptions leak, can you check that nothing in the code below can throw something potentially expected?

AlexanderSher: I'd rather use `ICoreShell.DispatchOnMainThreadAsync` and avoid dealing with `async void` at all

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
MikhailArkhipov: This stuff comes from WPF event handlers which are usually void... I can try ```DispatchOnMainThreadAsync(async () => ...).DoNotWait()``` I guess

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
MikhailArkhipov: Done

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -309,11 +256,135 @@ class DataFramePreviewRowItem {
...
+            InputFilePreview.Text = text;
+        }
+
+        private Task ConvertToUtf8(string file, int codePage, int maxLines = Int32.MaxValue) {
+            DeleteTempFile();
+            bool reportProgress = maxLines == Int32.MaxValue;
+
+            // Convert file to UTF-8 so it can be properly imported
+            return Task.Run(async () => {
+                try {
```
AlexanderSher: If you move the body of the `try` block into separated method, you can simply call `await TaskUtilities.SwitchToBackgroundThread();` at its beginning, and also you can simply call `OnError(ex.Message))` instead of `VsAppShell.Current.DispatchOnUIThread(() => OnError(ex.Message));`

AlexanderSher: And you wouldn't need `Task.Run`

MikhailArkhipov: Yep, good point. Fixed

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -309,11 +256,135 @@ class DataFramePreviewRowItem {
...
+
+        private void DeleteTempFile() {
+            try {
+                if (!string.IsNullOrEmpty(_tempFilePath) && File.Exists(_tempFilePath)) {
+                    File.Delete(_tempFilePath);
+                }
+            } catch (UnauthorizedAccessException) { }
+        }
+
+        private async Task DoDefaultAction() {
```
AlexanderSher: You can make this method sync and don't await `DispatchOnMainThreadAsync`

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -309,11 +256,138 @@ class DataFramePreviewRowItem {
...
+        }
+
+        private async Task ConvertToUtf8(string file, int codePage, int maxLines = Int32.MaxValue) {
+            try {
+                DeleteTempFile();
+                await ConvertToUtf8Worker(file, codePage, maxLines);
+            } catch (IOException ex) {
+                OnError(ex.Message);
+            } catch (UnauthorizedAccessException ex) {
+                OnError(ex.Message);
```
AlexanderSher: Does `OnError` hide the progress bar?

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
MikhailArkhipov: sure but we stull need to ```await RunAsync(expression);```

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
MikhailArkhipov: Nope, but it should

MikhailArkhipov:
Added NRows, fixed styling in Dark theme

PR #1546, opened 2016-04-12
Fix #1545: CRAN mirror setting allows arbitrary strings

### LINKED ISSUE #1545: CRAN mirror setting allows arbitrary strings
The field in the property grid is editable, and arbitrary input is accepted. However, if that input doesn't actually match the name of any known mirror, it is silently ignored, as if nothing was set.

int19h:
Reject unknown CRAN mirror names.


MikhailArkhipov:
👍

PR #1552, opened 2016-04-12
Json for grid data

JeongPyoHam:


src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
```diff
@@ -26,28 +27,13 @@ public class GridDataSource {
 
             REvaluationResult? result = null;
             using (var evaluator = await rSession.BeginEvaluationAsync()) {
-                result = await evaluator.EvaluateAsync($"rtvs:::grid.dput(rtvs:::grid.data({expression}, {rows}, {columns}))", REvaluationKind.Normal);
+                result = await evaluator.EvaluateAsync($"rtvs:::toJSON(rtvs:::grid.data({expression}, {rows}, {columns}))", REvaluationKind.Normal);
```
int19h: You don't need to call `toJSON` anymore. Instead, pass `REvaluationKind.Json` as the evaluation flag, and then use `REvaluationResult.JsonResult`.

JeongPyoHam: parser requires it as NRE

int19h: You mean JSON parser? When you do the above, you will get the value already parsed (as a `JToken`). You can then call `ToObject` method on that to get `GridData` out of it.

The difference between doing that, and the line of code above, is that with the above setup, the JSON for the grid will be returned inside a string literal in the protocol message JSON - so any quotation marks will be escaped, making the logs that much harder to read.

Issue #1328, opened 2016-03-23
Package manager backend: Profile and optimize serialization as needed


huguesv:
Separate task because the initial implementation will go in unoptimized so that the UI can start using it early.

huguesv:
Serialization on the R side is definitely causing perf issues.  Opened https://github.com/Microsoft/RTVS/issues/1350 to track that performance improvement work.

huguesv:
The most time consuming part is now the conversion from matrix to lists.
```
> a <- available.packages()
> al <- rtvs:::matrix.as.lists(a)
> system.time(rtvs:::matrix.as.lists(a))
   user  system elapsed 
   8.16    0.00    8.52 
> system.time(available.packages())
   user  system elapsed 
   1.20    0.01    1.23 
> 
```

PR #1554, opened 2016-04-12
Remove names from the package lists returned to VS.

huguesv:
The only optimization that I've been aware of that hadn't been done.  I'll close https://github.com/Microsoft/RTVS/issues/1328 after this.

This reduces the length of the JSON by 100kb, saving on serialization/transport/deserialization, and removes the special case of empty lists on deserialization.

We'll also now be able to have the same package be listed twice for "installed packages", which was not possible before with named lists.  I'm still preventing duplicates from happening right now, as I'm not sure the UI would expect it, but we can stop removing duplicates later.


int19h:
:+1:

MikhailArkhipov:
👍

PR #1555, opened 2016-04-13
Fix NullReferenceException in RHost when RSession does not have a _hostClientApp

int19h:


int19h:
This happens e.g. when the package manager host process crashes, since it will try to do a `SendMessage` with the error. The bug here is that the task-returning method returns null, which will cause `await` on the receiving side to throw.

AlexanderSher:
Funny
:+1:

Issue #1548, opened 2016-04-12
Same-file IntelliSense does not work when using "=" as an assignment operator


jflam:
"=" is a valid assignment operator (albeit not idimoatic R), but IntelliSense does not recognize it in scenarios where the variable is assigned to in the editor, but not evaluated (works fine if evaluated):

![image](https://cloud.githubusercontent.com/assets/8877/14474008/b09a2912-00af-11e6-9166-3ba09bd310a6.png)

MikhailArkhipov:
Trivial fix, just added another operator type check.

Issue #1541, opened 2016-04-12
Assertion failed: children collection is sorted and yet no position to insert can be found


int19h:
Happened while trying to reproduce #1268, when pasting the following code into REPL:
```r
id <- 1:5
url <- paste0( "http://faxian.smzdm.com/mall/%E4%BA%AC%E4%B8%9C/p", id)
read_html(url[1], encoding = "UTF-8") %>% html_nodes("h2.itemName") %>% html_node("a") %>% html_node("span.black") %>% html_text() %>% repair_encoding(., from = "UTF_8")
```

![image](https://cloud.githubusercontent.com/assets/1175142/14469881/850ad5bc-009b-11e6-9c44-2aeb3c40a4b5.png)

Children collection is sorted and yet no position to insert can be found

   at Microsoft.Languages.Core.Text.TextRangeCollection`1.AddSorted(T item) in C:\git\RTVS\src\Languages\Core\Impl\Text\TextRangeCollection.cs:line 134

   at Microsoft.R.Core.AST.AstNode.AppendChild(IAstNode child) in C:\git\RTVS\src\R\Core\Impl\AST\AstNode.cs:line 64

   at Microsoft.R.Core.AST.AstNode.set_Parent(IAstNode value) in C:\git\RTVS\src\R\Core\Impl\AST\AstNode.cs:line 45

   at Microsoft.R.Core.AST.AstNode.AppendChild(IAstNode child) in C:\git\RTVS\src\R\Core\Impl\AST\AstNode.cs:line 56

   at Microsoft.R.Core.AST.Expressions.Expression.MakeNode(ParseContext context) in C:\git\RTVS\src\R\Core\Impl\AST\Expressions\ExpressionParser.cs:line 541

   at Microsoft.R.Core.AST.Expressions.Expression.ProcessHigherPrecendenceOperators(ParseContext context, IOperator currentOperator) in C:\git\RTVS\src\R\Core\Impl\AST\Expressions\ExpressionParser.cs:line 499

   at Microsoft.R.Core.AST.Expressions.Expression.HandleOperatorPrecede......

int19h:
Full stack:
```
 	[Managed to Native Transition]	
 	System.dll!Microsoft.Win32.SafeNativeMethods.MessageBox(System.IntPtr hWnd, string text, string caption, int type)	Unknown
[...]

Issue #1537, opened 2016-04-11
IntelliSense in REPL proposes completions that don't exist


huguesv:
In REPL:
Define some variables ex; `i1 <- installed.packages()`
Type `graphics::i` and TAB
You'll see `graphics::i1` in the completion list
But it's not really there:
```
> graphics::i1
Error: 'i1' is not an exported object from 'namespace:graphics'
> 
``

PR #1556, opened 2016-04-13
#1537 #1541 #1548

MikhailArkhipov:
#1541 Assertion failed: children collection is sorted...
- Remove checking for inner part of ```%...%``` custom operators

#1537 IntelliSense in REPL proposes completions that don't exist
- Suppress variable intellisense in namespaces

#1548 Same-file IntelliSense does not work when using "="
- Add ```=``` and ```->``` to operators when looking up variable definitions


AlexanderSher:
:+1:

Issue #1557, opened 2016-04-13
Data Import: make file field read only and switch to message box error reporting


MikhailArkhipov:

MikhailArkhipov:
Also needs more styling since read-only data grid is gray on white window

PR #1561, opened 2016-04-13
#1557

MikhailArkhipov:
Data Import: make file field read only and switch to message box error reporting 
And proper grid styling


AlexanderSher:
:+1:

PR #1564, opened 2016-04-14
Add helpers and doc comments for IRExpressionEvaluator to clarify and streamline common usage scenarios

int19h:


int19h:
The purpose of this all is to provide an eval interface that is easy to use _right_, by making the best approach also the most concise one. So, new overloads have implicit error checking w/exceptions  (so it's impossible forget about / ignore errors if you use them), and focus on JSON over plain text - ideally, I would like to move all evals to JSON in the future, as it does a few things better even for the simplest cases (e.g. it handles Unicode in a lossless fashion, avoiding conversion to native encoding and back).

Changes to debugger here are targeted for minimal effect (I do intend to untangle the mess around various eval helpers, and `InvokeDebugHelperAsync` in particular, in the future - but that would be a far more risky change), and are there mainly to showcase the use and make sure that the code is actually exercised. But if there's risk concern, I can back them out, and leave just the new helpers.

src/Debugger/Impl/DebugSession.cs
```diff
@@ -143,8 +143,9 @@ public sealed class DebugSession : IDisposable {
             using (var eval = await RSession.BeginEvaluationAsync(cancellationToken)) {
                 res = await eval.EvaluateAsync(expression, json ? REvaluationKind.Json : REvaluationKind.Normal);
                 if (res.ParseStatus != RParseStatus.OK || res.Error != null || (json && res.JsonResult == null)) {
-                    Trace.Fail(Invariant($"Internal debugger evaluation {expression} failed: {res}"));
-                    throw new REvaluationException(res);
+                    string message = Invariant($"Internal debugger evaluation {expression} failed: {res}");
+                    Trace.Fail(message);
```
AlexanderSher: Do we need Trace.Fail if we throw exception after that?

src/Host/Client/Impl/Definitions/IRExpressionEvaluator.cs
```diff
@@ -13,62 +14,146 @@ namespace Microsoft.R.Host.Client {
...
+        /// </remarks>
         Cancelable = 1 << 5,
+        /// <summary>
+        /// Indicates that this expression should be evaluated in a fresh blank environment, that has the
+        /// otherwise designated target environment set as its parent, and which is discarded immediately
+        /// after evaluation.
+        /// </summary>
+        /// <remarks>
+        /// Similar to <c>local()</c>, in that it will contain and discard any assignments to local variables
+        /// that are performed as part of evaluation.
```
AlexanderSher: Is it compatible with `BaseEnv` and `EmptyEnv`?

src/Host/Client/Impl/Definitions/IRExpressionEvaluator.cs
```diff
@@ -13,62 +14,146 @@ namespace Microsoft.R.Host.Client {
     [Flags]
     public enum REvaluationKind {
```
AlexanderSher: Let's put this enum into its own file

MikhailArkhipov:
Not directly related to this but it would be nice to provide simple way to get value of variable or function return - to let easily eval and get result back in C#.

src/Host/Client/Impl/Definitions/IRExpressionEvaluator.cs
int19h: Yes - that's what "otherwise designated target environment" refers to.

src/Debugger/Impl/DebugSession.cs
int19h: In the debugger, kinda, because it's a COM API, so all exceptions get translated to HRESULTs at the AD7 boundary (and more often than not, swallowed silently by VS). 

Ideally, these asserts should rather be in AD7 code. When I get the time to go do some cleanup there, I'm going to do a generic wrapper, and just use it in all methods. But that's a much bigger change.

src/Host/Client/Impl/Definitions/IRExpressionEvaluator.cs
int19h: Actually, on a second thought, I think I'll get rid of this altogether. It doesn't do anything that `local()` won't do, and we don't use it anywhere - it's really just a leftover from back when I tried to use this as a workaround for the issue when evals were breaking into debugger because of debug bits (which has long since been fixed properly). I won't touch the host for now, but get rid of this here so that no-one tries to use it.

int19h:
You mean, get the value already in .NET representation? That's what `EvaluateAsync<T>` will do in most cases - i.e. you can use `EvaluateAsync<bool>` etc. So long as the variable value / function return value is JSON-serializable - which it is for most common primitive types. See the mapping rules here:

https://github.com/Microsoft/R-Host/blob/master/src/json.h#L30

And on .NET side, `ToObject` will take care of mapping the corresponding JSON type to most primitive .NET types, enums (with appropriate attributes), `[DataContract]` etc. Take a look:

https://github.com/JamesNK/Newtonsoft.Json/blob/93ba0c4b849934b0ce3b5dee58290ddebd600a03/Src/Newtonsoft.Json/Linq/JToken.cs#L1932

Ideally, I'd love to move all result-producing evals to JSON mode eventually, and retire the current "normal" mode completely, as it should be redundant. If it's something simple enough that you can convert it to string and parse it on .NET side, it is probably also JSON-serializable. If it's not simple enough, then it's better to convert it to a JSON-serializable data structure (as we do for e.g. package list currently), then to marshal it as string and then try to parse that string manually.

int19h:
(and, of course, you can always explicitly wrap your expression into `toString()` or something similar, and then the result of that will definitely be JSON-serializable)

MikhailArkhipov:
I meant more of what VE does, i.e. getting data frames back in a way that is less involved with the debug session. We can chat offline, that's offtopic here :-)

int19h:
Addressed feedback and removed various unused stuff as discussed in the comments.

AlexanderSher:
:+1:

PR #1566, opened 2016-04-14
Several Package Manager fixes

### LINKED ISSUE #1565: Package Manager: "Update" button doesn't work
Clicking the Update button in the package description pane doesn't do anything. However, clicking on the update icon (up-pointing arrow) in the package list does work.

AlexanderSher:
- Add error notification to package manager
- Fix #1565: Package Manager: "Update" button doesn't work
- Adjust package list items height


src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -68,6 +70,16 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             private set { SetProperty(ref _isLoading, value); }
         }
 
+        public string FirstError {
+            get { return _firstError; }
+            private set { SetProperty(ref _firstError, value); }
+        }
+
+        public bool HasMultipleErrors {
```
int19h: Would it be easier to make this a computed property (based on `_errorMessages.Count`)?

AlexanderSher: I still have to raise `PropertyChanged` event, so I'd rather change the property

int19h:
:+1:

huguesv:
:+1:

PR #1569, opened 2016-04-15
Package manager backend: detect package dll lock

huguesv:
I haven't gotten the UI part to work successfully in the case where the user wants us to reset the repl for them before install/uninstall (Task returned BeginInteractionAsync is awaited forever), so it's just the backend + tests in this PR.

FYI RestartManager impl comes from VS.

### DIFF
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -50,6 +50,8 @@ namespace Microsoft.R.Host.Client {
         private TaskCompletionSource<object> _cancelAllTcs;
         private CancellationTokenSource _cancelAllCts = new CancellationTokenSource();
 
+        public int ProcessId { get { return _process != null ? _process.Id : 0; } }
+
         public RHost(string name, IRCallbacks callbacks) {
             Check.ArgumentStringNullOrEmpty(nameof(name), name);
 
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -50,7 +50,7 @@ namespace Microsoft.R.Host.Client {
         private TaskCompletionSource<object> _cancelAllTcs;
         private CancellationTokenSource _cancelAllCts = new CancellationTokenSource();
 
-        public int ProcessId { get { return _process != null ? _process.Id : 0; } }
+        public int? ProcessId => _process?.Id;
 
         public RHost(string name, IRCallbacks callbacks) {
             Check.ArgumentStringNullOrEmpty(nameof(name), name);
--- a/src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
@@ -1,0 +1,143 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace Microsoft.R.Components.PackageManager.Implementation {
+    internal class RestartManager {
+        /// <summary>
+        /// Given a list of filenames with aboslute path, returns list of process currently locking those files.
+        /// </summary>
+        /// <param name="filePaths">Filenames with absolute path.</param>
+        /// <returns>List of processes locking the files. null if it encounters any error.</returns>
+        public static IList<Process> GetProcessesUsingFiles(string[] filePaths) {
+            uint sessionHandle;
+            List<Process> processes = new List<Process>();
+
+            int restartManager = NativeMethods.RmStartSession(out sessionHandle, 0, Guid.NewGuid().ToString("N"));
+            if (restartManager != 0) {
+                return null;
+            }
+
+            try {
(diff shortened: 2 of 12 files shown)


src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -50,6 +50,8 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
         private TaskCompletionSource<object> _cancelAllTcs;
         private CancellationTokenSource _cancelAllCts = new CancellationTokenSource();
 
+        public int ProcessId { get { return _process != null ? _process.Id : 0; } }
```
int19h: You can use `=>` here to make the one-liner slightly shorter:
```cs
public int ProcessId => ...
```

int19h: Also, perhaps it's better to make it `int?`, and return `null` if there's no host.

int19h: Or maybe even just expose the process directly... @AlexanderSher, I recall we talked about something where having the process accessible could be advantageous. Should we just do that?

src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
```diff
@@ -0,0 +1,143 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace Microsoft.R.Components.PackageManager.Implementation {
+    internal class RestartManager {
```
int19h: Looks like this could be `static`.

int19h: Also, this is very generic, and would probably be a better fit in `Microsoft.Common.Core`.

src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <summary>
+        /// Given a list of filenames with aboslute path, returns list of process currently locking those files.
+        /// </summary>
+        /// <param name="filePaths">Filenames with absolute path.</param>
+        /// <returns>List of processes locking the files. null if it encounters any error.</returns>
+        public static IList<Process> GetProcessesUsingFiles(string[] filePaths) {
+            uint sessionHandle;
+            List<Process> processes = new List<Process>();
+
+            int restartManager = NativeMethods.RmStartSession(out sessionHandle, 0, Guid.NewGuid().ToString("N"));
```
int19h: This is really a status/error code, right? Can you rename it to `error` or something similarly descriptive?

src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
```diff
@@ -0,0 +1,143 @@
...
+            try {
+                restartManager = NativeMethods.RmRegisterResources(sessionHandle, (uint)filePaths.Length, filePaths, 0, null, 0, null);
+                if (restartManager != 0) {
+                    return null;
+                }
+
+                uint pnProcInfoNeeded = 0, pnProcInfo = 0, lpdwRebootReasons = RmRebootReasonNone;
+                restartManager = NativeMethods.RmGetList(sessionHandle, out pnProcInfoNeeded, ref pnProcInfo, null, ref lpdwRebootReasons);
+
+                if (restartManager == ERROR_MORE_DATA) {
```
int19h: This likely needs to be a loop, if the list can change between the calls. Basically keep trying until it fits.

src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
```diff
@@ -0,0 +1,143 @@
...
+                    RM_PROCESS_INFO[] processInfo = new RM_PROCESS_INFO[pnProcInfoNeeded];
+                    pnProcInfo = (uint)processInfo.Length;
+
+                    restartManager = NativeMethods.RmGetList(sessionHandle, out pnProcInfoNeeded, ref pnProcInfo, processInfo, ref lpdwRebootReasons);
```
int19h: Again, assuming the list can change between consecutive `RmGetList` calls, this might end up filling fewer than `pnProcInfoNeeded` items in the array - I'm assuming that `pnProcInfo` is `ref` precisely so that the function can communicate the actual number of items it has filled. The loop below should take that into account.

src/R/Components/Impl/PackageManager/Implementation/RestartManager.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Security;
+
+namespace Microsoft.R.Components.PackageManager.Implementation {
+    internal class RestartManager {
+        /// <summary>
+        /// Given a list of filenames with aboslute path, returns list of process currently locking those files.
+        /// </summary>
+        /// <param name="filePaths">Filenames with absolute path.</param>
+        /// <returns>List of processes locking the files. null if it encounters any error.</returns>
+        public static IList<Process> GetProcessesUsingFiles(string[] filePaths) {
```
int19h: I would make this `IEnumerable<Process>`, and use `yield return` inside to fetch them on demand. The caller can then do `ToList` or not, as needed.

huguesv:
I think I've addressed all feedback.  Kept the process id as (nullable) int rather than returning Process object.  My thinking is that if we don't need it right now, we'll avoid potential misuse (someone accessing stuff through the Process rather than going through our existing host APIs).

int19h:
:+1:

Issue #1573, opened 2016-04-15
Default encoding of R Markdown file created by RTVS cause garbling of non-ascii characters


Marchen:
Default encoding of R Markdown seems to be system encoding (for my case, it was Japanese CP932), but it causes garbling of Japanese text in HTML.
If I change the encoding of the R Markdown file to UTF-8, Japanese characters are shown correctly.
So I think for the default encoding of R Markdown file, UTF-8 is more preferable.

![rmd](https://cloud.githubusercontent.com/assets/3863577/14566497/c6436716-036a-11e6-9255-0e0d177e273a.png)

MikhailArkhipov:
We use the same template as RS and generally pass same options to ```rmarkdown::render```. However, when saving file we should be setting encoding via ITextDocument.Encoding
https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.text.itextdocument.encoding.aspx

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/06dd7fe7f461afe30972f30b903ec51ae9337950

Marchen:
Thanks for the fix!

PR #1575, opened 2016-04-15
 #1573: Default encoding of R Markdown file created by RTVS cause gar…

MikhailArkhipov:
…bling of non-ascii characters

### DIFF
--- a/src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
+++ b/src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
@@ -5,6 +5,7 @@ using System.ComponentModel.Composition;
 using System.IO;
 using Microsoft.Markdown.Editor.Flavor;
 using Microsoft.VisualStudio.R.Package.Publishing.Definitions;
+using static System.FormattableString;
 
 namespace Microsoft.VisualStudio.R.Package.Publishing {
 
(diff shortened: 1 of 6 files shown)


src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -28,7 +29,7 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
             string outputFile = Path.GetFileName(outputFilePath);
             string outputFolder = Path.GetDirectoryName(outputFilePath).Replace('\\', '/');
             // Run rmarkdown::render
-            return $"\"rmarkdown::render(\'{inputFile}\', output_format=\'{format}\', output_file='{outputFile}',  output_dir='{outputFolder}')\"";
```
int19h: What if they try to publish an existing, non-UTF .Rmd?

MikhailArkhipov: This is generated command that we invoke so we specify encoding that we save the file with. However, it is a good question if someone tries to do the same manually. In this case they won't be going via this path and rather would need to a) save file in VS with desired encoding and b) specify that encoding in the command line. Although I guess no one would ever publish RMD manually.

MikhailArkhipov: Alternatively, we can try and see if file is convertible to Encoding.Default and leave it as it as if it is convertable.

int19h: If we control the encoding here, and this is not intended for direct user consumption, I wouldn't bother.

int19h:
:+1:

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
MikhailArkhipov: Well, there is a side effect since if file is modified, we do save it forcibly in UTF-8 which may not be what the user wants. Let me try and add the detection.

int19h: But if we save it using native encoding (when it can be thus saved), won't the original reported issue re-manifest?

Also, note that OS locale and R locale do not necessarily match (which was a root cause of many of our bugs regarding Unicode).

MikhailArkhipov:
Added detection if file can be converted. It appears that rmarkdown always produces UTF-8 page. However, original encoding has to be specified. Added the warning that file is not convertible and if user is OK with re-saving file in UTF-8

int19h:
:+1:

PR #1577, opened 2016-04-15
Add REvaluationKind.NoResult corresponding to "0" evaluation flag

int19h:
and reimplement ExecuteAsync to use that flag instead of the ";NULL" hack.


int19h:
This is the host change that adds the flag: https://github.com/Microsoft/R-Host/pull/59

huguesv:
:+1:

PR #1580, opened 2016-04-15
Change implementation of TaskExtensions.DoNotWait 

AlexanderSher:
To ensure that exception is returned back to the synchronization context, or shown with Debug.Fail if there is no synchronization context that can be used

### DIFF
--- a/src/Common/Core/Impl/Extensions/TaskExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/TaskExtensions.cs
@@ -71,26 +71,28 @@ namespace Microsoft.Common.Core {
 
             var synchronizationContext = SynchronizationContext.Current;
             if (synchronizationContext != null && synchronizationContext.GetType() != typeof(SynchronizationContext)) {
-                task.ContinueWith(ReThrowTaskExceptionIntoSynchronizationContext, synchronizationContext, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+                task.ContinueWith(DoNotWaitSynchronizationContextContinuation, synchronizationContext, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
             }
 
-            task.ContinueWith(DebugFailTaskException, TaskContinuationOptions.ExecuteSynchronously);
+            task.ContinueWith(DoNotWaitThreadContinuation, TaskContinuationOptions.ExecuteSynchronously);
         }
 
         private static void ReThrowTaskException(object state) {
-            var task = (Task) state;
+            var task = (Task)state;
             if (task.IsFaulted && task.Exception != null) {
-                ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();
+                var exception = task.Exception.InnerException;
+                ExceptionDispatchInfo.Capture(exception).Throw();
             }
         }
 
-        private static void DebugFailTaskException(Task task) {
+        private static void DoNotWaitThreadContinuation(Task task) {
             if (task.IsFaulted && task.Exception != null) {
-                Debug.Fail(task.Exception.InnerException.Message, task.Exception.InnerException.StackTrace);
+                var exception = task.Exception.InnerException;
+                ThreadPool.QueueUserWorkItem(s => ((ExceptionDispatchInfo)s).Throw(), ExceptionDispatchInfo.Capture(exception));
             }
         }
 
-        private static void ReThrowTaskExceptionIntoSynchronizationContext(Task task, object state) {
+        private static void DoNotWaitSynchronizationContextContinuation(Task task, object state) {
             var context = (SynchronizationContext) state;
             context.Post(ReThrowTaskException, task);
         }
--- a/src/Common/Core/Impl/Extensions/TaskExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/TaskExtensions.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Diagnostics;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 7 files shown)


src/Common/Core/Impl/Extensions/TaskExtensions.cs
```diff
@@ -56,14 +58,41 @@ public static class TaskExtensions {
...
-        /// Suppresses warnings about unawaited tasks and ensures that unhandled
-        /// errors will cause the process to terminate.
+        /// Rethrows task exceptions back to the callers synchronization context if it is possible
         /// </summary>
         /// <remarks>
         /// <see cref="OperationCanceledException"/> is always ignored.
         /// </remarks>
-        public static async void DoNotWait(this Task task) {
-            await task.SilenceException<OperationCanceledException>();
+        public static void DoNotWait(this Task task) {
```
int19h: @zooba, you might want to take a look, since this was originally a piece of code borrowed from PTVS, and it might be good to merge the changes back at some point.

src/Common/Core/Impl/Extensions/TaskExtensions.cs
```diff
@@ -56,14 +58,41 @@ public static class TaskExtensions {
         }
 
         /// <summary>
-        /// Suppresses warnings about unawaited tasks and ensures that unhandled
-        /// errors will cause the process to terminate.
+        /// Rethrows task exceptions back to the callers synchronization context if it is possible
         /// </summary>
         /// <remarks>
         /// <see cref="OperationCanceledException"/> is always ignored.
```
int19h: Is the updated version still ignoring OCE as documented? I don't see it in the code.

AlexanderSher: When task failed with OCE or `taskCompletionSource.SetCanceled()` is called, `Task.IsFaulted` will be `false`.
The only case when it is faulted is when you explicitly set `OperationCanceledException`:
```{csharp}
taskCompletionSource.SetException(new OperationCanceledException())
```
but that is a wrong way to cancel `TaskCompletionSource` and I'd prefer to fail for such code.

int19h:
:+1:

src/Common/Core/Impl/Extensions/TaskExtensions.cs
zooba: What does it achieve, other than randomly inserting exceptions into another thread?

int19h: I think the point here is to make it so that exceptions get propagated to the main event loop in case UI thread calls `DoNotWait` (which is by far the most common use), so that any top-level error handling/reporting/logging facilities can kick in.

AlexanderSher: Original implementation
```{csharp}
public static async void DoNotWait(this Task task) {
    await task
}
```
and the new one have several differences (both insert exceptions into another thread):
- Compiler independent implementation. `async`/`await` generates different code for Async CTP, 4.5 and 4.6.2. New implementation doesn't rely on this implementation.
- Performance. No async state machine, execution context capture and other side effects that are required for the general `async void` case.
- Selected `SynchronizationContext`. `AsyncVoidMethodBuilder` always uses `SynchronizationContextNoFlow`. New implementation doesn't care if context was captured, but rather checks if context isn't default.
- Continuation. Original implementation uses `MoveNextRunner.Run` which calls `AsyncMethodBuilderCore.ThrowAsync`. The difference is that when `DoNotWait` is called from background thread, exception is scheduled into `ThreadPool`. It crashes VS, but doesn't give us any idea about the issue. New implementation will not crash VS but will show the Trace.Fail dialog

src/Common/Core/Impl/Extensions/TaskExtensions.cs
```diff
@@ -56,14 +58,41 @@ public static class TaskExtensions {
...
+        private static void ReThrowTaskException(object state) {
+            var task = (Task) state;
+            if (task.IsFaulted && task.Exception != null) {
+                ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();
+            }
+        }
+
+        private static void DebugFailTaskException(Task task) {
+            if (task.IsFaulted && task.Exception != null) {
+                Debug.Fail(task.Exception.InnerException.Message, task.Exception.InnerException.StackTrace);
```
AlexanderSher: @int19h , do you think we should use `Trace.Fail` here?

AlexanderSher:
Ok, @int19h asked me for a detailed explanation.
Original implementation:
```{csharp}
public static async void DoNotWait(this Task task) {
    await task
}
```
is converted by the current version of compiler into:
```{csharp}
public static void DoNotWait(this Task task) {
    DoNotWaitStateMachine stateMachine;
    stateMachine.task = task;
    stateMachine._builder = AsyncVoidMethodBuilder.Create();
    stateMachine._state = -1;
    stateMachine._builder.Start(ref stateMachine);
}
```
with the following IAsyncStateMachine implementation (I've removed parts that aren't valuable):
```{csharp}
private struct DoNotWaitStateMachine : IAsyncStateMachine {
    public int _state;
    public AsyncVoidMethodBuilder _builder;
    public Task task;
    private TaskAwaiter _awaiter;

    void IAsyncStateMachine.MoveNext() {
        try {
            TaskAwaiter awaiter;
            if (this._state != 0) {
                awaiter = this.task.GetAwaiter();
                if (!awaiter.IsCompleted) {
                    this._state = 0;
                    this._awaiter = awaiter;
                    this._builder.AwaitUnsafeOnCompleted(ref awaiter, ref this);
                    return;
                }
            } else {
                awaiter = this._awaiter;
            }
            awaiter.GetResult();
        } catch (Exception ex) {
            this._state = -2;
            this._builder.SetException(ex);
            return;
        }
        this._state = -2;
        this._builder.SetResult();
    }

    void IAsyncStateMachine.SetStateMachine(IAsyncStateMachine stateMachine) {
        this._builder.SetStateMachine(stateMachine);
    }
}
```
This can be simplified even more for our case, cause this code works for any `async void` implementation with any amount of `await` calls in it.
So here is how it works:

1. `AsyncVoidMethodBuilder.Create` saves reference to `ExecutionContext.SynchronizationContextNoFlow` context (it doesn't "capture" context) and calls `SynchronizationContext.OperationStarted`
1. `AsyncVoidMethodBuilder.Start` opens `copy-on-write` scope and calls `IAsyncStateMachine.MoveNext`. All this happens in a caller's thread.
1. Since `_state == -1`, we get its awaiter
1.1. If task is completed, we just call `awaiter.GetResult()`
1.1. If it is not completed, awaiter is passed to the `AsyncVoidMethodBuilder.AwaitUnsafeOnCompleted`, which will create a continuation and pass it to `awaiter.UnsafeOnCompleted`. Continuation will call `IAsyncStateMachine.MoveNext` with `_state == 0` which will get previously saved `awaiter` and call `awaiter.GetResult()`
1. If `awaiter.GetResult()` throws exception, it is passed to the AsyncVoidMethodBuilder.SetException`, which calls `SynchronizationContext.OperationCompleted` and throws exception inside previously saved `SynchronizationContext`. If it is null, it throws it into `ThreadPool`.

New implementation does pretty much the same but it does it explicitly, without useless execution context manipulations, duplicated `OperationStarted/OperationCompleted` notification and so on.
The difference is between `ExecutionContext.SynchronizationContextNoFlow` and `ExecutionContext.SynchronizationContext` that is used in new implementation. `SynchronizationContext` isn't preserved when action is executed with `ExecutionContext.Run`, which we can fix using `ExecutionContext.SetSynchronizationContext` if needed. `SynchronizationContextNoFlow` isn't preserved when context is captured and there is no way we can fix it cause `SynchronizationContextNoFlow` is internal.
Another difference is in the way exceptions are handled, but I will probably return throwing exceptions to the `ThreadPool` back for release builds cause `Trace.Fail` isn't blocking UI thread and user may not notice the popup.

zooba:
So, essentially you took the autogenerated code and stripped out the bits that are unused? And now when they change the internal behavior we need to reimplement our wrapper (and do it differently depending on .NET version)?

If you can show significant perf benefit (either speed or size) in the context of RTVS (not a microbenchmark), then I'll consider it. Otherwise, I prefer the maintainability of letting the compiler generate the code.

(In PTVS, we have an unhandled exception filter on the current AppDomain when debugging, and use a different Debug.Fail handler in unit tests to avoid the dialog popup. This handles many more exceptions than just what the `DoNotWait` version would. We also have an unobserved task exception handler, but that doesn't see as many crashes.)

int19h:
My understanding is that those "unused" bits actually do things that can potentially be affecting semantics, it's just that they don't do so with all default settings. So stripping them out actually makes the implementation more predictable. Especially if the defaults change subtly in a future .NET version (like they did when unobserved task exceptions stopped killing the process in 4.5).

AlexanderSher:
> So, essentially you took the autogenerated code and stripped out the bits that are unused?

No, I haven't used compiler-generated code as basis. New version more or less matches code generated by C# 6 compiler, but the goal is actually the opposite - to have predictable behavior that doesn't depend on compiler and don't have any side effects that are provided by compiler-generated version.
I've shown the stripped version of compiler-generated code just to compare.

Issue #1579, opened 2016-04-15
Crash viewing in Excel in specific scenario


MikhailArkhipov:
Start VS
Open REPL
```x <- mtcars```
File | New | R Script
Try viewing x in Excel

Crash: (COM Exception).

```dte.Solution.Projects.Count > 0``` (Misc project was created with new file) but accessing
```dte.ActiveSolutionProjects``` throws since there is no solution technically

Issue #1574, opened 2016-04-15
It is possible to get duplicate signature sessions


MikhailArkhipov:
I’ve just installed RTVS 0.3 and I noticed something that could be improved. This is something I do at the beginning of almost all scripts I author.

Observed: 2 floating help panes
Expected: 1 floating help pane

Observed: Inside string literal when I hit Ctrl+Space I get a list of functions I cannot use because I’m in string literal.
Expected: A list of files under “D:\” path

![sig](https://cloud.githubusercontent.com/assets/12820357/14570709/89c8249a-02fa-11e6-955b-add669effdf8.png)

MikhailArkhipov:
BTW, to get list of files string has to be in R format (i.e. forward slashes) like "D:/"

MikhailArkhipov:
Also, string must be closed. The issue here is that literal is now closed and caret technically does not belong to it (i.e. caret position is beyond the end of the open literal which by default ends at the line break and end of the line is not inclusive. This is why Ctrl+J or Ctrl+Space brings list of functions.

PR #1590, opened 2016-04-16
#1574 #1579

MikhailArkhipov:
#1579 Crash viewing in Excel in specific scenario
- check for misc files solution
- add catching com exceptions just in case

#1574 It is possible to get duplicate signature sessions
- minor issue when caret is at the end of an unterminated string
- fix tokenizer so it actually creates token when opening quote is at the very end of the file

### DIFF
--- a/src/Package/Impl/Utilities/ProjectUtilities.cs
+++ b/src/Package/Impl/Utilities/ProjectUtilities.cs
@@ -56,10 +56,20 @@ namespace Microsoft.VisualStudio.R.Package.Utilities {
         public static EnvDTE.Project GetActiveProject() {
             DTE dte = VsAppShell.Current.GetGlobalService<DTE>();
             if (dte.Solution.Projects.Count > 0) {
-                var projects = (Array)dte.ActiveSolutionProjects;
-                if (projects != null && projects.Length == 1) {
-                    return (EnvDTE.Project)projects.GetValue(0);
-                }
+                try {
+                    if (dte.Solution != null) {
+                        Array projects;
+                        var solutionFile = dte.Solution.FullName;
+                        if (!string.IsNullOrEmpty(solutionFile)) {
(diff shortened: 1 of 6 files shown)


src/Package/Impl/Utilities/ProjectUtilities.cs
```diff
@@ -56,10 +56,20 @@ public static class ProjectUtilities {
         public static EnvDTE.Project GetActiveProject() {
             DTE dte = VsAppShell.Current.GetGlobalService<DTE>();
             if (dte.Solution.Projects.Count > 0) {
-                var projects = (Array)dte.ActiveSolutionProjects;
-                if (projects != null && projects.Length == 1) {
-                    return (EnvDTE.Project)projects.GetValue(0);
-                }
+                try {
+                    if (dte.Solution != null) {
```
int19h: Given that we have accessed `dte.Solution` three lines above, this seems redundant (or else that line should also be inside `if`).

src/Package/Impl/Utilities/ProjectUtilities.cs
```diff
@@ -56,10 +56,20 @@ public static class ProjectUtilities {
...
+                        Array projects;
+                        var solutionFile = dte.Solution.FullName;
+                        if (!string.IsNullOrEmpty(solutionFile)) {
+                            projects = (Array)dte.ActiveSolutionProjects;
+                            return (EnvDTE.Project)projects.GetValue(0);
+                        } else {
+                            var e = dte.Solution.Projects.GetEnumerator();
```
int19h: Enumerator needs to be disposed.

Better yet, you can do `Projects.Cast<EnvDTE.Project>().First()`, to avoid having to muck around with enumerators directly.

int19h:
:+1:

Issue #931, opened 2016-02-19
CPS hangs on exit when a lot of plots were executed in bulk


MikhailArkhipov:
- Open/Create R project
- Select all code below, send to REPL
- Exit VS or close solution or unload project
  VS hangs
# R Script

``` r
plot(AirPassengers)
plot(BJsales)
plot(BJsales.lead)
plot(BOD)
plot(CO2)
plot(ChickWeight)
plot(DNase)
plot(EuStockMarkets)
plot(Formaldehyde)
#plot(HairEyeColor)
#plot(Harman23.cor)
#plot(Harman74.cor)
plot(Indometh)
plot(InsectSprays)
plot(JohnsonJohnson)
plot(LakeHuron)
plot(LifeCycleSavings)
plot(Loblolly)
plot(Nile)
plot(Orange)
plot(OrchardSprays)
plot(PlantGrowth)
plot(Puromycin)
plot(Seatbelts)
plot(Theoph)
plot(Titanic)
plot(ToothGrowth)
plot(UCBAdmissions)
plot(UKDriverDeaths)
plot(UKgas)
plot(USAccDeaths)
plot(USArrests)
plot(USJudgeRatings)
plot(USPersonalExpenditure)
plot(UScitiesD)
plot(VADeaths)
plot(WWWusage)
plot(WorldPhones)
#plot(ability.cov)
plot(airmiles)
plot(airquality)
plot(anscombe)
plot(attenu)
plot(attitude)
plot(austres)
plot(beaver1)
plot(beaver2)
plot(cars)
plot(chickwts)
plot(co2)
plot(crimtab)
plot(discoveries)
plot(esoph)
plot(euro)
plot(euro.cross)
plot(eurodist)
plot(faithful)
plot(fdeaths)
plot(freeny)
plot(freeny.x)
plot(freeny.y)
plot(infert)
plot(iris)
plot(iris3)
plot(islands)
plot(ldeaths)
plot(lh)
plot(longley)
plot(lynx)
plot(mdeaths)
plot(morley)
plot(mtcars)
plot(nhtemp)
plot(nottem)
plot(npk)
plot(occupationalStatus)
plot(precip)
plot(presidents)
plot(pressure)
plot(quakes)
plot(randu)
plot(rivers)
plot(rock)
plot(sleep)
plot(stack.loss)
plot(stack.x)
plot(stackloss)
#plot(state.abb)
plot(state.area)
plot(state.center)
plot(state.division)
#plot(state.name)
[...]

PR #1608, opened 2016-04-18
Fix #1607: History isn't saved when user exits VS

### LINKED ISSUE #1607: History isn't saved when user exits VS
Regression because of https://github.com/Microsoft/RTVS/issues/931

AlexanderSher:
Verified that #931 isn't regressed


int19h:
:+1:

PR #1609, opened 2016-04-18
Fix #1538: Package manager could warn user when uninstall didn't remove all files

### LINKED ISSUE #1538: Package manager could warn user when uninstall didn't remove all files
If you install `ffbase`, load it, then try to uninstall, uninstall will not be able to delete `ffbase/libs/x64/ffbase.dll`.  You have to reset the REPL, then you can manually delete the leftover file.

The user experience is that after uninstall, `ffbase` still shows up in the list of installed packages, which can be confusing.

After uninstall is done, we could check if the folder still exist and warn the user that some files could not be deleted and will have to be removed manually.  Right now there are no warning/errors being displayed in REPL.

We could also fix up the result of installed.packages to remove packages that don't have a NAMESPACE file, so it doesn't show up in the UI anymore. Though if we do that, we open ourselves up for another error later on, if they try to re-install `ffbase` after they've installed, loaded, uninstalled it.  It fill fail to re-install:
```
Warning: cannot remove prior installation of package ffbase
```

AlexanderSher:
User will be notified prior to uninstalling/updating package that it is locked and can't be uninstalled.

### DIFF
--- a/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
@@ -14,6 +14,7 @@ using Microsoft.Common.Core.Shell;
 using Microsoft.Common.Core.Threading;
 using Microsoft.Common.Wpf;
 using Microsoft.Common.Wpf.Collections;
+using Microsoft.R.Components.PackageManager.Model;
 using Microsoft.R.Components.PackageManager.ViewModel;
 using Microsoft.R.Components.Settings;
 using Microsoft.R.Host.Client;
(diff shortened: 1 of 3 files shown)


src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -158,6 +159,27 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
+            } catch (RPackageManagerException ex) {
+                AddErrorMessage(ex.Message);
+            }
+
+            await ReloadInstalledAndLoadedPackagesAsync();
+            if (_selectedTab == SelectedTab.InstalledPackages) {
+                IsLoading = false;
+                ReplaceItems(_installedPackages);
+            }
+            package.IsChanging = false;
```
MikhailArkhipov: Should it be set to true here? Or is it set elsewhere? I think we may consider something like using(var x = package.BeginChange()) so we won't have to track changing state set/clear

huguesv:
:+1:

MikhailArkhipov:
👍

Issue #1613, opened 2016-04-18
User operator definition is incorrectly squiggled


MikhailArkhipov:
%a% <- function(x, y) x + y

PR #1617, opened 2016-04-18
#1613 User operator definition is incorrectly squiggled

MikhailArkhipov:
+ Tests
+ make % completion trigger

### DIFF
--- a/src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
+++ b/src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using Microsoft.Common.Core;
 using Microsoft.Languages.Core.Text;
 using Microsoft.Languages.Core.Tokens;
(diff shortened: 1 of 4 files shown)


src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
```diff
@@ -383,6 +384,27 @@ enum OperationType {
         }
 
         private OperationType HandleOperator(ParseContext context, out ParseErrorType errorType) {
+            // We must distinguish between operator definition and the operator use.
+            // in case of %a% <- left side behaves as identifier rather than the operator
```
int19h: `%a% <- ...` doesn't seem to be valid R. You have to write `` `%a%` <- ...`` for it to parse.

Similarly, ``1 `%a%` 2`` is actually not valid - you need to write ``1 %a% 2`` for it to parse as an operator.

So it should be sufficient to look at presence/absence of backticks to distinguish these two cases.

MikhailArkhipov: Hm I guess you are right...

PR #1628, opened 2016-04-18
Fix #1618: Variable explorer, package manager double click on scrollbar arrow opens grid view, installs packages

### LINKED ISSUE #1618: Variable explorer, package manager double click on scrollbar arrow opens grid view, installs packages
We should only do that when double clicking an item in the list.

image

Same issue in package manager, where in the available list, it will install a package when you double-click the scrollbar.

AlexanderSher:


MikhailArkhipov:
👍

int19h:
:+1:

Issue #1626, opened 2016-04-18
Plot tests fail image comparison on certain machines


huguesv:
We compare all the pixels in the image that is produced with a baseline, and in some cases there is a tiny variance in black levels which makes tests fail.  This is happening only to Alex's machine for some reason.

We can change the tests so that they produce the expected images as part of the test, using `png()` calls.

PR #1641, opened 2016-04-19
Generate expected plot images as part of the test instead of using pr…

huguesv:
…e-rendered images. This is to avoid comparison errors where colors may be slightly different depending on machine where it is rendered.

Fixes https://github.com/Microsoft/RTVS/issues/1626


src/Host/Client/Test/IdeGraphicsDeviceTest.cs
```diff
@@ -478,46 +501,22 @@ public class IdeGraphicsDeviceTest {
...
-            var expectedPlotPaths = GetTestExpectedFiles();
-            actualPlotPaths.Length.Should().Be(expectedPlotPaths.Length);
+        private void CompareImages(string[] actualPlotPaths, string[] expectedPlotPaths) {
+            actualPlotPaths.Should().HaveSameCount(expectedPlotPaths);
 
-            foreach (string path in expectedPlotPaths) {
-                var actualPlotPath = actualPlotPaths.First(p => Path.GetFileName(p) == Path.GetFileName(path));
-                File.ReadAllBytes(actualPlotPath).Should().Equal(File.ReadAllBytes(path));
+            for (int i = 0; i < actualPlotPaths.Length; i++) {
+                File.ReadAllBytes(actualPlotPaths[i]).Should().Equal(File.ReadAllBytes(expectedPlotPaths[i]));
```
int19h: If you do `actualPlotPaths,Select(p => File>ReadAllBytes(p))` etc, then you can do `Equal` on the two resulting collections, and get somewhat better diagnostics from asserts - e.g. index of the first mismatching image.

int19h: (although it might have to be `ShouldBeEquivalentTo` for it to compare data rather than references; `Be` will just use `Object.Equals`, I think, which doesn't compare data for arrays)

PR #1648, opened 2016-04-19
FxCop cleanup

MikhailArkhipov:
Change targets file so FxCop can read its results back into VS
FxCop cleanup (test code was left alone)
Fix 2 more VE tests


src/Debugger/Engine/Impl/AD7StackFrame.cs
```diff
@@ -149,7 +150,7 @@ internal sealed class AD7StackFrame : IDebugStackFrame2, IDebugExpressionContext
         }
 
         int IDebugExpressionContext2.GetName(out string pbstrName) {
-            pbstrName = string.Format("{0}:{1}|{2}", StackFrame.FileName, StackFrame.LineNumber, StackFrame.CallingFrame?.Call);
+            pbstrName = string.Format(CultureInfo.InvariantCulture, "{0}:{1}|{2}", StackFrame.FileName, StackFrame.LineNumber, StackFrame.CallingFrame?.Call);
```
AlexanderSher: `Invariant($"{StackFrame.FileName}:{StackFrame.LineNumber}|{StackFrame.CallingFrame?.Call}")`

src/R.Settings.targets
```diff
@@ -3,10 +3,10 @@
   <!-- Common Build Settings and Configurations -->
   <PropertyGroup>
     <DefineConstants Condition="'$(VisualStudioVersion)'=='14.0'">$(DefineConstants);VS14</DefineConstants>
-    <RootDirectory Condition="'$(RootDirectory)' == ''">$(MSBuildThisFileDirectory)\..\</RootDirectory>
-    <LibDirectory Condition="'$(LibDirectory)' == ''">$(RootDirectory)\lib\</LibDirectory>
-    <BinDirectory Condition="'$(BinDirectory)' == ''">$(RootDirectory)\bin\</BinDirectory>
-    <ObjDirectory Condition="'$(ObjDirectory)' == ''">$(RootDirectory)\obj\</ObjDirectory>
+    <RootDirectory Condition="'$(RootDirectory)' == ''">$(MSBuildThisFileDirectory)..</RootDirectory>
```
AlexanderSher: Directory name should always end with `\`. So 
`<RootDirectory Condition="'$(RootDirectory)' == ''">$(MSBuildThisFileDirectory)..\</RootDirectory>`

AlexanderSher: `<LibDirectory Condition="'$(LibDirectory)' == ''">$(RootDirectory)lib\</LibDirectory>`

MikhailArkhipov: That was what broke code analysis. They can't handle ``\\..\\bin```

src/R/Actions/Impl/Utility/RInstallation.cs
```diff
@@ -259,7 +259,7 @@ public static class RInstallation {
         }
 
         public static Version GetRVersionFromFolderName(string folderName) {
-            if (folderName.StartsWith("R-")) {
+            if (folderName.StartsWith("R-", StringComparison.OrdinalIgnoreCase)) {
```
AlexanderSher: `folderName.StartsWithIgnoreCase`

src/R.Settings.targets
MikhailArkhipov: So we need to guarantee no double backslashes. ```$(Foo)\bar``` looks more natural than ```$(Foo)bar```

src/R/Core/Impl/AST/DataTypes/RNull.cs
```diff
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.R.Core.AST.DataTypes.Definitions;
 
 namespace Microsoft.R.Core.AST.DataTypes {
     public sealed class RNull : RObject, IRVector {
-        public static readonly RNull Null = new RNull();
+        private static readonly RNull _null = new RNull();
```
AlexanderSher: public static RNull Null { get; } = new RNull();

src/R/Actions/Impl/Utility/RInstallation.cs
MikhailArkhipov: This code is used in setup custom actions via linked items and custom actions can't reference assemblies that haven't been installed yet :-)

src/R.Settings.targets
AlexanderSher: maybe, but all built-in variables include trailing backslashes, so I'd like to be consistent with them.

---
In reply to: [60325204](https://github.com/Microsoft/RTVS/pull/1648#discussion_r60325204) [](ancestors = 60325204)

src/R/Actions/Impl/Utility/RInstallation.cs
AlexanderSher: :)

---
In reply to: [60325363](https://github.com/Microsoft/RTVS/pull/1648#discussion_r60325363) [](ancestors = 60325363)

src/R.Settings.targets
MikhailArkhipov: OK should be better now

src/Common/Core/Impl/Extensions/CharExtensions.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.Common.Core {
     public static class CharExtensions {
-        public static readonly char[] LineBreakChars = new char[] { '\n', '\r' };
+        public static char[] LineBreakChars => new char[] { '\n', '\r' };
```
AlexanderSher: `public static char[] LineBreakChars { get; } = new char[] { '\n', '\r' };`
Otherwise it will create new array every time you access the property

MikhailArkhipov: Ah I see

Issue #1647, opened 2016-04-19
Duplicate access keys in R Tools menu


MikhailArkhipov:
![1](https://cloud.githubusercontent.com/assets/12820357/14657192/1c2243e0-0641-11e6-8e81-feb21059de05.png)

Also R Documentation vs Data.

Issue #1645, opened 2016-04-19
GoTo/Peek Definition sometimes doesn't work when file was just opened


MikhailArkhipov:
[eval.R.txt](https://github.com/Microsoft/RTVS/files/226878/eval.R.txt)

Open attached file
Goto line 268
Try getting definition for process_slots()
Nothing
Now edit file
Try again - it works.

Not 100% repro, need to investigate deeper.

MikhailArkhipov:
This is because right click does not reposition caret and Peek/GoTo are based on the caret position. Specific scenario: 

find function name
Ctrl+F3 to find its use
Note that found item is *selected* and caret is at its end
Right click on the item and hit ESC
Note that caret was not repositioned on right click
Only repros when selection exists

Fix is to base Peek on selection when it exists, otherwise use caret position

Issue #836, opened 2016-02-05
Completion does not show package methods in function arguments


MikhailArkhipov:
jsonlite::f <- flatten is in the list
str(jsonlite:: <- no flatten in the list

![image](https://cloud.githubusercontent.com/assets/12820357/12860590/92eb1632-cc12-11e5-8907-89169c3a6baa.png)

PR #1649, opened 2016-04-20
#836 #1645 #1647

MikhailArkhipov:
#1647 Duplicate access keys in R Tools menu

#1645 GoTo/Peek Definition sometimes doesn't work
- when there is a selection, right click doesn't move caret

#836 Completion does not show package methods in function arguments
- package name lookup didn't stop at ( or ,


int19h:
:+1:

huguesv:
More duplicates: In Windows menu, `h` access key is used twice.  In Plots menu, `v` is used twice.

MikhailArkhipov:
I fixed H in the Windows menu. Will fix v next PR

Issue #1551, opened 2016-04-12
read.csv behind Data Import dialog should appear in REPL


MikhailArkhipov:
```After running the tool, it would be helpful if it printed out the associated R code in History; reproducibility of code is important to data scientists because of the prevalence of peer review.```

PR #1651, opened 2016-04-20
#1551 read.csv behind Data Import dialog should appear in REPL

MikhailArkhipov:
- Make temp file less 'temporary' so re-invoking command from history would work
- Reuse file name rather than generating it every time
- Don't delete converted file so command from history won't fail
- Add asserts to workflow operations that require UI thread calls

### DIFF
--- a/src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
+++ b/src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
@@ -12,6 +12,7 @@ using System.Windows.Controls;
 using System.Windows.Data;
 using System.Windows.Input;
 using Microsoft.Common.Core;
+using Microsoft.Languages.Editor.Tasks;
 using Microsoft.R.Components.InteractiveWorkflow;
 using Microsoft.R.Host.Client;
 using Microsoft.VisualStudio.PlatformUI;
@@ -25,7 +26,7 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.DataImport {
     /// </summary>
     public partial class ImportDataWindow : DialogWindow {
         private const int MaxPreviewLines = 20;
-        private string _tempFilePath;
+        private string _utf8FilePath;
 
         public IDictionary<string, string> Separators { get; } = new Dictionary<string, string> {
             [Package.Resources.ImportData_Whitespace] = "",
@@ -62,6 +63,7 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.DataImport {
             [Package.Resources.ImportData_UseFirstColumn] = "1",
             [Package.Resources.ImportData_UseNumber] = null
         };
+        public object IdleTimeTask { get; private set; }
 
         public ImportDataWindow() {
             InitializeComponent();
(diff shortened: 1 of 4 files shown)


src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -62,6 +63,7 @@ public partial class ImportDataWindow : DialogWindow {
             [Package.Resources.ImportData_UseFirstColumn] = "1",
             [Package.Resources.ImportData_UseNumber] = null
         };
+        public object IdleTimeTask { get; private set; }
```
MikhailArkhipov: Whoops that is accidental

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -164,35 +164,23 @@ public ImportDataWindow(string filePath, string name)
...
             try {
-                REvaluationResult result = await EvaluateAsync(expression);
-                if (result.ParseStatus == RParseStatus.OK && result.Error == null) {
-                    Close();
-                } else {
-                    OnError(result.ToString());
-                }
+                var workflow = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>().GetOrCreate();
+                workflow.Operations.ExecuteExpression(expression);
+                return true;
```
AlexanderSher: You will get true even if `readcsv` fails. Is it ok?

MikhailArkhipov: I expect user to see error in REPL. Since Execute() is void, there is nothing we can report here.

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -287,7 +270,7 @@ class DataFramePreviewRowItem {
             await TaskUtilities.SwitchToBackgroundThread();
 
             Encoding encoding = Encoding.GetEncoding(codePage);
-            _tempFilePath = Path.GetTempFileName();
+            _utf8FilePath = Path.Combine(Path.GetTempPath(), Path.GetFileName(file) + ".utf8");
```
AlexanderSher: What happens if there is such file already? Or `Path.GetTempPath` guarantees that it is unique?

MikhailArkhipov: We'll reuse it then - this prevents us from producing many large oddly named files each time user imports. If they import same file again, we'll reuse the name. Also, readable name looks nicer in REPL. We also can't delete file on close since REPL is still executing (I tried 10 sec timeout, didn't work, it depends on the file size and the CPU speed)

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
AlexanderSher: Ok

---
In reply to: [60455182](https://github.com/Microsoft/RTVS/pull/1651#discussion_r60455182) [](ancestors = 60455182)

src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
AlexanderSher: Ok

---
In reply to: [60455799](https://github.com/Microsoft/RTVS/pull/1651#discussion_r60455799) [](ancestors = 60455799)

AlexanderSher:
:+1:

PR #1668, opened 2016-04-21
Fix #1605: Intellisense should be disabled for readline() in interactive window

### LINKED ISSUE #1294: Lines of code treated as response to readline() when running multiline snippet in REPL
from user:

In RGUI this script works fine:

h <- 'Hallo'
name <- readline('Name: ')

paste(h, name)

In R-Interactive readline is not doing what it should do.

--- 

in line at a time it works.  if sent to repl as a chunk, the readline doesnt work.

### LINKED ISSUE #1605: Intellisense should be disabled for readline() in interactive window
readline() is of course used for user-generated text. It should therefore not trigger IntelliSense when used in the interactive window. See, for example, the way that swirl() prompts me with inappropriate content (in this case, one of the sample datasets). 

image

This is R Tools for Visual Studio 0.2, running with Microsoft R Open on Visual Studio 2015 Update 2.

AlexanderSher:
Fix #1294: Lines of code treated as response to readline() when running multiline snippet in REPL


MikhailArkhipov:
👍

Issue #1535, opened 2016-04-11
Import from csv: Row Names "Use first column" error


huguesv:
Import bikes.csv
Separator: semicolon
Row names: Use first column
```
Error in data[[rowvar]] : attempt to select less than one element
```

huguesv:
I'll take this.  Problem is that we are passing `row.names="1"`.  We should be passing `row.names=1`.

huguesv:
Fixed

PR #1669, opened 2016-04-21
When row name is set to first column, set row.names to 1, not "1".

huguesv:
Fix for https://github.com/Microsoft/RTVS/issues/1535

### DIFF
--- a/src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
+++ b/src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
@@ -57,9 +57,9 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.DataImport {
 
         public IDictionary<string, int> Encodings { get; } = new Dictionary<string, int>();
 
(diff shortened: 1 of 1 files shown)


src/Package/Impl/DataInspect/DataImport/ImportDataWindow.xaml.cs
```diff
@@ -57,9 +57,9 @@ public partial class ImportDataWindow : DialogWindow {
 
         public IDictionary<string, int> Encodings { get; } = new Dictionary<string, int>();
 
-        public IDictionary<string, string> RowNames { get; } = new Dictionary<string, string> {
+        public IDictionary<string, int?> RowNames { get; } = new Dictionary<string, int?> {
```
AlexanderSher: Then it is `RowNumbers` I think

huguesv: This is still controlling what we are passing to `row.names` parameter of `read.csv`.

AlexanderSher:
:+1:

PR #1670, opened 2016-04-22
Fix SilenceException so it doesn't convert task cancellation into task failure

AlexanderSher:
Added tests
Unblock `RSession` thread for `ReadUserInput` implementation


int19h:
:+1:

PR #1674, opened 2016-04-22
Dev 14/15 compat

MikhailArkhipov:
- Changes are around project files, references and ifdefs
- Project structure didn't change
- Builds in 14 and 15
- Runs in 14 and 15
- No 15 setup yet
- Still issues with R project in Dev15, but other parts appear to be working


src/Common/Core/Impl/Versions/VsVersion.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core {
+    public static class VsVersion {
```
AlexanderSher: Do we need Vs-specific code in this assembly?

huguesv:
Does this build on a machine that doesn't have Dev 15?

src/Mocks/VisualStudio/ProjectServiceMock.cs
```diff
@@ -53,4 +55,49 @@ public sealed class ProjectServiceMock : ProjectService {
             throw new NotImplementedException();
         }
     }
+#else
+    [ExcludeFromCodeCoverage]
+    public sealed class ProjectServiceMock : IProjectService {
```
AlexanderSher: Can we split this into separate files?

src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
```diff
@@ -861,11 +867,11 @@
       <HintPath>$(LibDirectory)\Microsoft.VisualStudio.InteractiveWindow.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.VisualStudio.Language.Intellisense, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\..\NugetPackages\Microsoft.VisualStudio.Language.Intellisense.14.1.24720\lib\net45\Microsoft.VisualStudio.Language.Intellisense.dll</HintPath>
+      <HintPath>..\..\..\NugetPackages\Microsoft.VisualStudio.Language.Intellisense.$(Vs14Version)\lib\net45\Microsoft.VisualStudio.Language.Intellisense.dll</HintPath>
```
AlexanderSher: How NuGet restore handles this?

MikhailArkhipov:
Yes. I have pure Dev14 machine and laptop with Dev15 and Dev14 SxS. Haven't tried on Dev15 clean box, that is TBD when we try and build for Dev15 in the lab. but I need to figure out setup build in Dev15 still since officially there is no Wixtools for Dev15 yet.

src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
MikhailArkhipov: It doesn't really. I think I may just put back version number back. Generally versioning issue happen around CPS related code that is in the project assembly, package and 3 test modules. So working with nuget in these component has to be careful and project files may need to be fixed if nuget references are added/removed/updated.

src/Common/Core/Impl/Versions/VsVersion.cs
MikhailArkhipov: Where can we place it so it is available across?

MikhailArkhipov: Ex 

```Microsoft.UnitTests.Core\XUnit\AssemblyMefCatalogFixture.cs - (112, 114) : return (string)Registry.GetValue(Invariant($"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\VisualStudio\\{VsVersion.Version}"), "InstallDir", string.Empty);```

File: src/Mocks/VisualStudio/ProjectServiceMock.cs (hunk 2 above)
Comment (MikhailArkhipov) [inline src/Mocks/VisualStudio/ProjectServiceMock.cs]: Sure

File: src/Common/Core/Impl/Versions/VsVersion.cs (hunk 1 above)
Comment (AlexanderSher) [inline src/Common/Core/Impl/Versions/VsVersion.cs]: We don't have general assembly for VS-specific things. Maybe we need to add one. Or use `#ifdef` in this particular case.

File: src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs [hunk 4]
```diff
@@ -83,14 +95,21 @@ internal sealed class RProjectLoadHooks {
         }
 
         [AppliesTo("RTools")]
+#if VS14
         [UnconfiguredProjectAutoLoad2(completeBy: UnconfiguredProjectLoadCheckpoint.CapabilitiesEstablished)]
+#else
```
AlexanderSher: I think we should use `#if VS15` rather than `#else`

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
```diff
@@ -207,12 +243,18 @@ internal sealed class RProjectLoadHooks {
...
+#if VS14
         private async Task<bool> GetSaveDefaultWorkspace(string rdataPath) {
+#else
+        private bool GetSaveDefaultWorkspace(string rdataPath) {
+#endif
             switch (_toolsSettings.SaveRDataOnProjectUnload) {
                 case YesNoAsk.Yes:
                     return true;
                 case YesNoAsk.Ask:
+#if VS14
```
AlexanderSher: So in VS15 this method is called in UI thread?

src/ProjectSystem/Impl/IO/MsBuildFileSystemWatcher.cs
```diff
@@ -149,7 +150,7 @@ public sealed partial class MsBuildFileSystemWatcher : IDisposable {
                 } catch (Exception e) {
                     _hasErrors = 1;
                     _log.WatcherApplyChangeFailed(change.ToString(), e);
-                    Debug.Fail(Invariant($"Failed to apply change {change}:\n{e}"));
+                    System.Diagnostics.Debug.Fail(Invariant($"Failed to apply change {change}:\n{e}"));
```
AlexanderSher: Why?

MikhailArkhipov: b/c there is new 'Debug' namespace in CPS :-(

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
MikhailArkhipov: since most of the code is UI thread (starting with the package load) I added 'requires UI thread' to the AutoLoad attribute. If you think it is worth keeping first 2 lines on bg thread, we can keep code as is

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
MikhailArkhipov: I hope new code is going to be same going forward and not just VS15, but on the other hand, who knows

src/Common/Core/Impl/Versions/VsVersion.cs
MikhailArkhipov: On the other hand, this solution is VS solution so one can perceive this as version of the toolset.

src/ProjectSystem/Impl/IO/MsBuildFileSystemWatcher.cs
AlexanderSher: picard-facepalm

AlexanderSher:
:+1:

int19h:
:+1:

PR #1688, opened 2016-04-26
Debugger refactoring and cleanup

int19h:
Clean up the use of evaluation - ditch unnecessary helpers, don't use `BeginEvaluationAsync` where it's not necessary, do not tie `DebugEvaluationResult` to a stack frame.

Rename things so that they're more descriptive and/or not misleading.

Split types into separate source files for better readability.

Remove unused stuff.

Comments, comments everywhere!

### DIFF
--- a/src/Debugger/Impl/DebugEvaluationResult.cs
+++ b/src/Debugger/Impl/DebugEvaluationResult.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.R.Host.Client;
 using Newtonsoft.Json.Linq;
 using static System.FormattableString;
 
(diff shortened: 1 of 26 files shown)


huguesv:
:+1:

src/Debugger/Impl/DebugEvaluationResult.cs
```diff
@@ -2,113 +2,114 @@
...
 
-        public static string ToRVector(this DebugEvaluationResultFields fields) {
-            var fieldNames = _mapping.Where(kv => fields.HasFlag(kv.Key)).Select(kv => "'" + kv.Value + "'");
-            return Invariant($"base::c({string.Join(", ", fieldNames)})");
-        }
-    }
+        /// <summary>
+        /// R expression designating the environment in which the evaluation that produced this result took place.
+        /// </summary>
+        public string EnvironmentExpression { get; }
```
AlexanderSher: I can't find where this property is used

int19h: No-one uses it outside of the class (but I saw no reason to make it private). Inside the class, the first overload of `EvaluateAsync` uses it, as do `GetChildrenAsync` in `DebugValueEvaluationResult`.

AlexanderSher:
:+1:

PR #1692, opened 2016-04-26
1691

MikhailArkhipov:
If function was not found, RD info fetcher kept calling. We need to fill entry with a stub. This is limited fix. Technically we should allow return() to be a function, but that would be a bigger change + needed multiple test updates, I will do it in 0.4


AlexanderSher:
:+1:

int19h:
:+1:

PR #1707, opened 2016-04-27
Fix #1694: Hang after reset REPL during readline() call

### LINKED ISSUE #1694: Hang after reset REPL during readline() call
Call `readline`
Hit Reset

In the following code, `reader` that is returned is null and causes a null ref exception.
```
        public Task<string> ReadUserInput(string prompt, int maximumLength, CancellationToken ct) {
            _coreShell.DispatchOnUIThread(() => _interactiveWindow.Write(prompt));
            return Task.Run(() => {
                using (var reader = _interactiveWindow.ReadStandardInput()) {
                    return Task.FromResult(reader.ReadToEnd());
                }
            }, ct);
        }

```

AlexanderSher:


MikhailArkhipov:
:+1:

huguesv:
:+1:

PR #1708, opened 2016-04-27
VE and viewers

MikhailArkhipov:
Tests to come when PR is agreed upon :-)

- Allows multiple grid instances
- View details on lists, functions, vectors, time series
- View(x) from REPL opens grid
- View(func) opens editor with function code
- library() opens pakage manager
- Tooltips in VE for functions etc


src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -401,6 +401,15 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
                                 _callbacks.DirectoryChanged();
                                 break;
 
+                            case "library":
+                                _callbacks.ViewLibrary();
```
AlexanderSher: Would've been great if we generalize this somehow on the level of the protocol an `RHost`, providing general `Override` command to have a switch inside `RSessionCallback`

AlexanderSher: Or, since we need to explicitly call `OverrideFunction`, create a list of callbacks in RSession, and add the callback to the OverrideFunction as a parameter

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -153,19 +153,23 @@ public static class RSessionEvaluationCommands {
...
 
-        public static Task<REvaluationResult> SetChangeDirectoryRedirection(this IRExpressionEvaluator evaluation) {
-            var script =
-@"utils::assignInNamespace('setwd', function(dir) {
-    old <- .Internal(setwd(dir))
-    rtvs:::send_message('setwd', dir)
-    invisible(old)
-  }, 'base')";
+        public static Task<REvaluationResult> OverrideFunction(this IRExpressionEvaluator evaluation, string name, string ns) {
+            var script = Invariant($"utils::assignInNamespace('{name}', rtvs:::{name}, '{ns}')");
```
AlexanderSher: I'd prefer to have wrapper in `rtvs` package, even if it is just a single-line call

src/Package/Impl/DataInspect/VariableViewModel.cs
```diff
@@ -4,41 +4,53 @@
...
             Index = index;
+            var result = DebugEvaluation as DebugValueEvaluationResult;
+            if (result != null) {
+                SetViewButtonStatus(result);
+            }
+        }
 
-            CanShowDetail = ComputeDetailAvailability(DebugEvaluation as DebugValueEvaluationResult);
+        private void SetViewButtonStatus(DebugValueEvaluationResult result) {
+            var aggregator = VsAppShell.Current.ExportProvider.GetExportedValue<IObjectDetailsViewerAggregator>();
```
AlexanderSher: `IObjectDetailsViewerAggregator` is accessed often. Can you pass it to ctor as a parameter?

src/Package/Impl/DataInspect/VariableViewModel.cs
```diff
@@ -4,41 +4,53 @@
...
+            _detailsViewer = aggregator.GetViewer(result);
+            _title = result.Name;
+
+            CanShowDetail = _detailsViewer != null;
             if (CanShowDetail) {
-                ShowDetailCommand = new DelegateCommand(ShowVariableGridWindowPane, (o) => CanShowDetail);
+                ShowDetailCommand = new DelegateCommand(async (o) => await _detailsViewer.ViewAsync(result.Expression, _title), (o) => CanShowDetail);
```
AlexanderSher: `DelegateCommand` doesn't have proper ctor for async methods. You should either call `ViewAsync` with `DoNotWait`, or just provide a plain method and use it directly instead of command. Also, for `DelegateCommand` you need to call `RaiseCanExecuteChanged` manually :(

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+            return result;
+        }
+
+        private async Task<DebugSession> GetDebugSessionAsync() {
+            if (_debugSession == null) {
+                if (_rSession == null) {
+                    _rSession = _rSessionProvider.GetInteractiveWindowRSession();
+                }
+                _debugSession = await _debugSessionProvider.GetDebugSessionAsync(_rSession);
```
AlexanderSher: @int19h, do you think we should attach DebugSession to Workflow instance?

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+                    if (File.Exists(tempFile)) {
+                        File.Delete(tempFile);
+                    }
+
+                    using (var sw = new StreamWriter(tempFile)) {
+                        sw.Write(functionCode);
+                    }
+
+                    VsAppShell.Current.DispatchOnUIThread(() => {
+                        var dte = VsAppShell.Current.GetGlobalService<DTE>();
```
AlexanderSher: Do we need UI thread only for `GetGlobalService`?

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
MikhailArkhipov: Ideally we should probably separate pure data-related things from debugging things. I.e. VE code should be talking to some data fetching code rather than to code that looks like debugging

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
MikhailArkhipov: DTE itself is also single thread. Also, need to delete file only when DTE is done with it.

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+        [ImportingConstructor]
+        public FunctionViewer(IRSessionProvider sessionProvider, IDataObjectEvaluator evaluator) {
+            _sessionProvider = sessionProvider;
+            _evaluator = evaluator;
+        }
+
+        #region IObjectDetailsViewer
+        public ViewerCapabilities Capabilities => ViewerCapabilities.Function;
+
+        public bool CanView(DebugValueEvaluationResult evaluation) {
```
AlexanderSher: Is it possible that `CanView` returns `true` when `evaluation == null`? If not, we can verify it inside aggregator and call `CanView` only when `evaluation != null`.

MikhailArkhipov: Since it is a MEF export it can also be imported somewhere else. So this is basically like command GetStatus and Invoke - no matter if you return 'no supported' Invoke can still be called.

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
int19h: Yes, I have thought about it. `DebugSession` currently handles three unrelated things - breakpoints/stepping, stack tracing, and data/eval. The latter part doesn't really need anything from the former, and can be separated. Stack tracing is kinda related, but it is a one-way dependency (i.e. stack needs eval, but eval doesn't need stack). With my recent changes decoupling `DebugEvaluationResult` from `DebugStackFrame`, this one-way dependency is reflected in the code, as well.

The only real reason why it was all in the same place was because it used helpers from the same helper package, and so `DebugSession.InitializeAsync` was loading that package. But now the package is used everywhere else and is loaded elsewhere, so that reason is gone.

So, yeah. I have some more debugger refactoring work queued up, so timing is good too. Just please file an issue for it and assign to me so that I don't forget.

src/Package/Impl/DataInspect/Viewers/GridViewer.cs
```diff
@@ -0,0 +1,65 @@
...
+                } else if (evaluation.Classes.Any(t => _listClasses.Contains(t))) {
+                    return evaluation.Dim == null && evaluation.Classes.Count == 1;
+                }
+                return evaluation.Length > 1;
+            }
+            return false;
+        }
+
+        public async Task ViewAsync(string expression, string title) {
+            var evaluation = await _evaluator.EvaluateAsync(expression, _fields) as DebugValueEvaluationResult;
```
AlexanderSher: It feels like we call `EvaluateAsync` for the same expression. If so, can we reuse `DebugValueEvaluationResult` that was used in `CanView`?

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
MikhailArkhipov: #1709

src/Package/Impl/DataInspect/Viewers/GridViewer.cs
MikhailArkhipov: CanView only fetches class, length and dim. View may require a lot more so this is to reduce traffic.

src/Debugger/Impl/rtvs/R/overrides.R
```diff
@@ -0,0 +1,36 @@
...
+    if (missing(title)) {
+      title <- ""
+    }
+    invisible(rtvs:::send_message('View', deparse(substitute(x)), title))
+  } else {
+    print(x)
+  }
+}
+
+browser <- function(url) {
```
int19h: We really should rename it, because now it's clashing with `base::browser` (only inside this package, but still). Let's call it `open_url` or something like that, and rename the message as well.

src/Package/Impl/DataInspect/Viewers/GridViewer.cs
MikhailArkhipov: So aggregator does it once and passes over to all viewers it imports via Linq. Each viewer typically needs specific fields to fetch but that is one time deal.

src/Debugger/Impl/rtvs/R/overrides.R
```diff
@@ -0,0 +1,36 @@
...
+  rtvs:::send_message('setwd', dir)
+  invisible(old)
+}
+
+library <- function(package, ...) {
+  args <- list(...)
+  if (missing(package) && length(args) == 0) {
+    invisible(rtvs:::send_message('library'))
+  } else {
+    pkgname <- as.character(substitute(package))
```
int19h: Why not just pass `...` directly to `library()`? The fewer things we do here before invoking the real thing, the fewer chances to subtly change its semantics.

src/Debugger/Impl/rtvs/R/overrides.R
MikhailArkhipov: OK

src/Debugger/Impl/rtvs/R/overrides.R
```diff
@@ -0,0 +1,36 @@
...
+
+setwd <- function(dir) {
+  old <- .Internal(setwd(dir))
+  rtvs:::send_message('setwd', dir)
+  invisible(old)
+}
+
+library <- function(package, ...) {
+  args <- list(...)
+  if (missing(package) && length(args) == 0) {
```
int19h: `if (nargs() == 0)` will avoid the need to force all args into the list.

src/Host/Client/Impl/Host/RHost.cs
int19h: I have long wanted to get rid of all this hardcoded stuff in the host and in `IRCallbacks`, and instead offer a generic facility to register custom message callbacks by name.

src/Debugger/Impl/rtvs/R/overrides.R
MikhailArkhipov: Ah, that took me a while to figure out :-) Problem is when user types in library(abc). 'abc' object does not really exist so passing func(...) library(...) produces 'abc' object does not exist. So basically this snippet is borrowed from the library.r itself, it flattens object to string effectively making library('abc')

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -153,19 +153,23 @@ public static class RSessionEvaluationCommands {
...
+            return evaluation.EvaluateAsync(script, REvaluationKind.Mutating);
+        }
+
+
+        public static Task<REvaluationResult> SetViewRedirection(this IRExpressionEvaluator evaluation) {
+            var script = @"
```
int19h: This will leave a variable named `.rtvs` in global scope. It should either use `local()` to avoid polluting, but better yet, it can be rewritten to avoid the need for a temp variable altogether:
```r
attach(as.environment(list(View = rtvs:::view)), warn.conflicts = FALSE)
```

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
AlexanderSher: Ok

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
MikhailArkhipov: Cool. I won't have to filter out .rtvs then

src/Host/Client/Impl/Host/RHost.cs
int19h: Also, I'm not sure we really need it (yet, anyway). A more reliable way to override `setwd` would be to do the same thing we do for `parse` - i.e. override the C implementation in `R_FunTab` (the place where `.Internal` looks things up).

Although for this particular thing, I wouldn't even bother overriding; instead, make the host code check for current directory every time it hits a new prompt, compare it with the old value, and if it's different, send `setwd`. It's common enough and low level enough that I'm not concerned with it being part of core host functionality.

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public async Task<DebugEvaluationResult> EvaluateAsync(string expression, DebugEvaluationResultFields fields) {
+            await TaskUtilities.SwitchToBackgroundThread();
+            DebugEvaluationResult result = null;
+
+            var debugSession = await GetDebugSessionAsync();
+            var frames = await debugSession.GetStackFramesAsync();
+
+            for (int i = frames.Count - 1; i >= 0; i--) {
```
int19h: What is this loop trying to achieve? Find the correct frame to evaluate it in?

If so, that's not the right approach - the same expression can be legitimate in a different frame, but produce a totally different result (e.g. if it has a variable with the same name, but different value). It's why `DebugSession.EvaluateAsync` takes environment as one of the arguments. We should be doing the same thing here - capture the environment context, and pass it in.

int19h: Although for `View()` specifically, by definition, when it's invoked, it's from the top of the stack (since that's where the code is currently running). So it appears that you can just use `frames.Last()`.

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
MikhailArkhipov: Well, its unfortunate but

image

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            string functionName = evaluation.Expression as string;
+            if (string.IsNullOrEmpty(functionName)) {
+                return;
+            }
+
+            var session = _sessionProvider.GetInteractiveWindowRSession();
+            string functionCode = null;
+            using (var e = await session.BeginEvaluationAsync()) {
```
int19h: It probably doesn't need this here, since you're not checking anything on `e` - so just invoke `EvaluateAsync` directly on the session.

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+            return evaluation != null && evaluation.Classes.Count == 1 && evaluation.Classes[0].EqualsOrdinal("function");
+        }
+
+        public async Task ViewAsync(string expression, string title) {
+            var evaluation = await _evaluator.EvaluateAsync(expression, DebugEvaluationResultFields.Expression) as DebugValueEvaluationResult;
+            if(evaluation == null) {
+                return;
+            }
+
+            string functionName = evaluation.Expression as string;
```
int19h: The cast is unnecessary - it's already a string.

src/Package/Impl/Packages/R/RPackageToolWindowProvider.cs
```diff
@@ -56,6 +61,14 @@ internal class RPackageToolWindowProvider {
             return CreateContainer(toolWindowType.GUID, id) as ToolWindowPane;
         }
 
+        #region IPackageLibraryViewer
+        public void ViewLibrary() {
```
AlexanderSher: `CreatePackageManagerToolWindow(0).Container.Show(true)`

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+        #region IObjectDetailsViewer
+        public ViewerCapabilities Capabilities => ViewerCapabilities.Function;
+
+        public bool CanView(DebugValueEvaluationResult evaluation) {
+            return evaluation != null && evaluation.Classes.Count == 1 && evaluation.Classes[0].EqualsOrdinal("function");
+        }
+
+        public async Task ViewAsync(string expression, string title) {
+            var evaluation = await _evaluator.EvaluateAsync(expression, DebugEvaluationResultFields.Expression) as DebugValueEvaluationResult;
+            if(evaluation == null) {
```
int19h: Do we want to silently ignore eval errors? It might be worth checking for `DebugErrorEvaluationResult` here, and displaying the error to the user.

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
MikhailArkhipov: Yes, it is to allow View(x) when sitting on a breakpoint. However, it can be defined in any of the outer frames, no?

MikhailArkhipov: When x is not global that is

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            string functionName = evaluation.Expression as string;
+            if (string.IsNullOrEmpty(functionName)) {
+                return;
+            }
+
+            var session = _sessionProvider.GetInteractiveWindowRSession();
+            string functionCode = null;
+            using (var e = await session.BeginEvaluationAsync()) {
+                var result = await e.EvaluateAsync(Invariant($"paste0(deparse({functionName}), collapse='\n')"), REvaluationKind.Normal);
```
int19h: Use `EvaluateAsync<string>()` - you get correct Unicode handling that way, and errors will be thrown as `RException`.

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -0,0 +1,86 @@
...
+                        File.Delete(tempFile);
+                    }
+
+                    using (var sw = new StreamWriter(tempFile)) {
+                        sw.Write(functionCode);
+                    }
+
+                    VsAppShell.Current.DispatchOnUIThread(() => {
+                        var dte = VsAppShell.Current.GetGlobalService<DTE>();
+                        dte.ItemOperations.OpenFile(tempFile);
```
int19h: Does this block until the editor is closed? If not, won't `Delete` conflict with the editor?

Also, as I recall, VS has the notion of documents that are not backed by files (it uses it to e.g. show client JS code when attached to IE). It might be possible to use that here, then we don't have to deal with temp files at all, and we can control the title of the document, too (i.e. it could be something like "R function: foo").

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -21,6 +21,8 @@ internal sealed class RSessionCallback : IRSessionCallback {
         private readonly IRSettings _settings;
         private readonly ICoreShell _coreShell;
 
+        public object VsAppShell { get; private set; }
```
AlexanderSher: How it compiles?

src/Package/Impl/Packages/R/RPackageToolWindowProvider.cs
MikhailArkhipov: For some reason it didn't work for me... I can try again

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -77,5 +79,15 @@ internal sealed class RSessionCallback : IRSessionCallback {
...
             return CranMirrorList.UrlFromName(mirrorName);
         }
+
+        public void ViewObject(string expression, string title) {
+            var viewer = _coreShell.ExportProvider.GetExportedValue<IObjectViewer>();
+            viewer?.ViewObjectDetails(expression, title);
+        }
+
+        public void ViewLibrary() {
+            var viewer = _coreShell.ExportProvider.GetExportedValue<IPackageLibraryViewer>();
```
AlexanderSher: I'm not sure if we need this interface. Package manager can be shown by 
```{csharp}
var containerFactory = _coreShell.ExportProvider.GetExportedValue<IRPackageManagerVisualComponentContainerFactory>();
workflow.Packages.GetOrCreateVisualComponent(containerFactory, id).Show(true);
```

src/Debugger/Impl/rtvs/R/overrides.R
int19h: That's really weird - passing an argument shouldn't cause it to be evaluated (since arguments are promises). 

I'm more concerned with the fact that we're calling `list(...)`. That will cause evaluation of all arguments in that list. If `library` adds some argument in the future that they expect to evaluate lazily, we will break that.

int19h: This works for me:
```r
f <- function(...) library(...)
f(utils)
```
even though:
```
> utils
Error: object 'utils' not found
```

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
int19h: `attach()` takes a `name` argument - you should pass that.

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
int19h: Ah, I guess this is true - you could pause on a breakpoint and then run `View()` in the REPL for any active frame.

Okay, then it should capture the current frame number on the stack inside `View()` and pass that, and then use it to obtain the corresponding frame by index (via `sys.frame`).

int19h: Actually, I'm wrong. When you're paused on a breakpoint, anything you type in REPL is always evaluated at the top of the call stack. It doesn't matter if you switch frames from Call Stack or VE.

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
MikhailArkhipov: No, editor does not lock files. Yeah, we can look into transients at some point, not sure if they are supported in all scenarios.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
MikhailArkhipov: Damn right click menu. It actually does compile go figure

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
MikhailArkhipov: Makes sense

src/Debugger/Impl/rtvs/R/grid.R
```diff
@@ -30,7 +26,11 @@ grid.data <- function(x, rows, cols) {
   onedim <- ((length(rows) == 1) || (length(cols) == 1));
   if (onedim || is.matrix(x)) {
     if (onedim) {
-      data <- grid.format(x[rows, cols]);
+      if(is.list(x) || is.vector(x) || is.ts(x)) {
```
int19h: This excludes vectors that have named elements. A better way to check would be `is(x, 'vector')`. Because lists are also vectors, this would make `is.list` check redundant, as well.

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
MikhailArkhipov: I did basically this
```x <- mtcars```
inside deeper frame and View(x) from command line didn't work since it needed deeper frame rather than .GlobalEnv

**Comment by MikhailArkhipov** [comment]
Updated per feedback

File: src/Debugger/Impl/rtvs/R/overrides.R [hunk 23]
```diff
@@ -0,0 +1,38 @@
...
+      title <- ""
+    }
+    invisible(rtvs:::send_message('View', deparse(substitute(x)), title))
+  } else {
+    print(x)
+  }
+}
+
+open_url <- function(url) {
+  rtvs:::send_message('Browser', url)
```
int19h: Can you rename the message as well? It's always confusing to me whenever I see it in RHost :)

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
int19h: `ToRStringLiteral` (unlikely that we'll ever have anything that needs to be escaped here, but you never know).

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
```diff
@@ -0,0 +1,45 @@
...
+            _debugSessionProvider = debugSessionProvider;
+        }
+
+        public async Task<DebugEvaluationResult> EvaluateAsync(string expression, DebugEvaluationResultFields fields) {
+            await TaskUtilities.SwitchToBackgroundThread();
+
+            var debugSession = await GetDebugSessionAsync();
+            var frames = await debugSession.GetStackFramesAsync();
+
+            return await frames.Last().EvaluateAsync(expression, fields) as DebugValueEvaluationResult;
```
int19h: On the oft chance that stack frame list is empty (I think this should never happen, but I'm not 100% sure), can you check and throw (or return null, whichever is more appropriate) here?

int19h:
:+1:

int19h:
Note that with another change that just went in, you'll need to remove `DebugValueEvaluationResult.Repr` everywhere, and start passing the appropriate repr function to `EvaluateAsync` / `GetChildrenAsync`.

To make it easier, feel free to change the `repr` argument of all these functions to not have a default. It's currently set to `null` (meaning no representation) by default, but in practice most locations do need it, and forgetting to provide it means that `Representation` is set to null, so maybe that wasn't the best idea.

src/Package/Impl/DataInspect/Viewers/GridViewer.cs
```diff
@@ -0,0 +1,61 @@
...
+                } else if (evaluation.Classes.Any(t => _listClasses.Contains(t))) {
+                    return evaluation.Dim == null && evaluation.Classes.Count == 1;
+                }
+                return evaluation.Length > 1;
+            }
+            return false;
+        }
+
+        public async Task ViewAsync(string expression, string title) {
+            var evaluation = await EvaluateAsync(expression, _fields, "rtvs:::make_repr_str()") as DebugValueEvaluationResult;
```
AlexanderSher: Can you move it to the base class?

MikhailArkhipov: You mean repr thing? Function viewer doesnt need it

AlexanderSher: Ok, got it.

PR #1710, opened 2016-04-27
Refactor representation handling in DebugEvaluationResult

int19h:
Instead of hardcoded flags selecting one of the three implementations, DebugSession.EvaluateAsync and all its wrappers now take repr function as an expression. Factories for such functions are provided for all the previous options.


MikhailArkhipov:
:+1:

PR #1711, opened 2016-04-27
Implement graphics device locator functionality (rtvs)

huguesv:
Before calling locator
image

After calling locator
image

You can't see from the screenshot but the mouse cursor in plot window changes to a cross.
I've disabled the other plot commands that don't work during locator mode.
I've considered making them invisible, like in RS.  Let me know what you think.
Status bar and tool window text are also changed during locator mode.  I think it's sufficient feedback for user.
The one thing that I don't know how to do is bind ESC to the End Locator command.  This would match RGui and RS key mapping.

Click on End Locator and it returns to normal mode
image


smortaz:
nice. once you have a private build, pls ping the user that requested the
feature & have them try it out.

src/Package/Impl/Plots/PlotWindowPane.cs
```diff
@@ -38,10 +43,13 @@ internal class PlotWindowPane : RToolWindowPane, IVsWindowFrameNotify3 {
             var historyProvider = VsAppShell.Current.ExportProvider.GetExportedValue<IPlotHistoryProvider>();
 
             _plotHistory = historyProvider.GetPlotHistory(sessionProvider.GetInteractiveWindowRSession());
+            _plotHistory.PlotContentProvider.Locator = this;
             _plotHistory.HistoryChanged += OnPlotHistoryHistoryChanged;
 
             var presenter = new XamlPresenter(_plotHistory.PlotContentProvider);
             presenter.SizeChanged += PlotWindowPane_SizeChanged;
+            presenter.RootContainer.MouseLeftButtonUp += RootContainer_MouseLeftButtonUp;
```
AlexanderSher: Do you unsubscribe it anywhere?

src/Package/Impl/Plots/PlotWindowPane.cs
```diff
@@ -99,5 +128,45 @@ internal class PlotWindowPane : RToolWindowPane, IVsWindowFrameNotify3 {
...
+        public bool IsInLocatorMode {
+            get { return _locatorTcs != null; }
+        }
+
+        public void StartLocatorMode(CancellationToken ct, TaskCompletionSource<LocatorResult> tcs) {
+            _locatorTcs = tcs;
+            ct.Register(() => EndLocatorMode());
+
+            VsAppShell.Current.DispatchOnUIThread(() => {
+                var rootContainer = ((XamlPresenter)Content).RootContainer;
```
AlexanderSher: Can you extract the body of this callback into separate method?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -62,6 +62,22 @@ internal sealed class RSessionCallback : IRSessionCallback {
             history.PlotContentProvider.LoadFile(filePath);
         }
 
+        public async Task<LocatorResult> Locator(CancellationToken ct) {
+            var historyProvider = _coreShell.ExportProvider.GetExportedValue<IPlotHistoryProvider>();
+            var history = historyProvider.GetPlotHistory(_session);
+            var tcs = new TaskCompletionSource<LocatorResult>();
+            _coreShell.DispatchOnUIThread(() => {
+                if (history.PlotContentProvider.Locator != null) {
+                    history.PlotContentProvider.Locator.StartLocatorMode(ct, tcs);
```
AlexanderSher: It is usually a design flaw if we have to pass `TaskCompletionSource` as a parameter.
Here it would be enough to make `StartLocatorMode` async:
```{csharp}
public async Task<LocatorResult> Locator(CancellationToken ct) {
    var historyProvider = _coreShell.ExportProvider.GetExportedValue<IPlotHistoryProvider>();
    var history = historyProvider.GetPlotHistory(_session);
    return history.PlotContentProvider.Locator != null 
        ? history.PlotContentProvider.Locator.StartLocatorModeAsync(ct, tcs)
        : Task.FromResult(new LocatorResult());
}
```
and add call of `ICoreShell.SwitchToMainThreadAsync()` to the beginning of `StartLocatorModeAsync` implementation

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -405,6 +405,12 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
                                 await _callbacks.Plot(message.GetString(0, "xaml_file_path"), ct);
                                 break;
 
+                            case "Locator":
```
AlexanderSher: I don't see R code that overrides this function

int19h: It's not an R function - it comes from the graphics device.

https://github.com/Microsoft/R-Host/pull/63

src/Host/Client/Impl/Definitions/LocatorResult.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Client {
+    public class LocatorResult {
```
int19h: It sounds like this could be a struct.

int19h: Preferably with readonly fields, and a ctor, as usual for structs.

src/Package/Impl/Plots/PlotWindowPane.cs
```diff
@@ -99,5 +123,49 @@ internal class PlotWindowPane : RToolWindowPane, IVsWindowFrameNotify3 {
...
+
+        public async Task<LocatorResult> StartLocatorModeAsync(CancellationToken ct) {
+            _locatorTcs = new TaskCompletionSource<LocatorResult>();
+            ct.Register(EndLocatorMode);
+
+            await VsAppShell.Current.SwitchToMainThreadAsync();
+            SetLocatorModeUI(true);
+
+            var task = _locatorTcs.Task;
+            await task;
```
int19h: You can simplify it to `return await task` here.

int19h:
:+1:

PR #1716, opened 2016-04-29
Viewer tests

MikhailArkhipov:
- Split classes and add extra checks
- Fix whatever was discovered during test writing
- Add interfaces to couple of DebugEval classes so they can be mocked/substituted. We can add more as needed.


src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
```diff
@@ -13,14 +14,8 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.DataSource {
...
             await TaskUtilities.SwitchToBackgroundThread();
-
-            if (rSession == null) {
-                rSession = VsAppShell.Current.ExportProvider.GetExportedValue<IRSessionProvider>().GetInteractiveWindowRSession();
-                if (rSession == null) {
-                    throw new InvalidOperationException(Invariant($"{nameof(IRSessionProvider)} failed to return RSession for {nameof(VariableViewModel)}"));
-                }
-            }
-
+            rSession = rSession ?? VsAppShell.Current.ExportProvider.GetExportedValue<IRSessionProvider>().GetInteractiveWindowRSession();
```
AlexanderSher: What is the case when `rSession` is null?

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
```diff
@@ -28,22 +26,16 @@ public sealed class DebugObjectEvaluator : IDataObjectEvaluator {
...
 
-            var debugSession = await GetDebugSessionAsync();
+            // Don't cache sessions since they can be disposed, expecially the debug session
+            // when host is restarts or gets re-created in tests
+            var rSession = _rSessionProvider.GetInteractiveWindowRSession();
+            var debugSession = await _debugSessionProvider.GetDebugSessionAsync(rSession);
+
             var frames = await debugSession.GetStackFramesAsync();
             if(frames == null || frames.Count == 0) {
                 throw new InvalidOperationException("Debugger frames stack is empty");
```
AlexanderSher: Should we fail in this case?

src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
MikhailArkhipov: It is passed as null in data import dialog. Not entirely sure why, we can probably change this and have it always non-null

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
MikhailArkhipov: Pavel wanted throw. I don't mind either way.

src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
MikhailArkhipov: Actually no one passes session, it is always null

src/Package/Impl/DataInspect/Viewers/ObjectDetailsViewerAggregator.cs
```diff
@@ -10,9 +10,9 @@
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
     [Export(typeof(IObjectDetailsViewerAggregator))]
-    public sealed class ObjectDetailsViewerAggregator : IObjectDetailsViewerAggregator {
+    internal sealed class ObjectDetailsViewerAggregator : IObjectDetailsViewerAggregator {
         [ImportMany]
-        private IEnumerable<Lazy<IObjectDetailsViewer>> Viewers { get; set; }
+        internal IEnumerable<Lazy<IObjectDetailsViewer>> Viewers { get; set; }
```
AlexanderSher: Is it for tests?

MikhailArkhipov: Yes

src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public async Task ViewAsync(string expression, string title) {
+            var evaluation = await EvaluateAsync(expression, _fields, "rtvs:::make_repr_str()") as DebugValueEvaluationResult;
+            if (evaluation != null) {
+                await VsAppShell.Current.SwitchToMainThreadAsync();
+
+                var id = _toolWindowIdBase + evaluation.GetHashCode() % (Int32.MaxValue - _toolWindowIdBase);
+                VariableGridWindowPane pane = ToolWindowUtilities.ShowWindowPane<VariableGridWindowPane>(id, true);
+                title = !string.IsNullOrEmpty(title) ? title : evaluation.Expression;
+                pane.SetEvaluation(new VariableViewModel(evaluation, _aggregator), title);
```
AlexanderSher: Is it possible that `VariableViewModel` will use a viewer that is different from the current one?

src/Package/Impl/uur1ketl.tmp_proj
```diff
@@ -0,0 +1,1322 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
AlexanderSher: What is it?

src/Package/Test/DataInspect/ReplCommandsTest.cs
```diff
@@ -0,0 +1,232 @@
...
+                gridViewer = await _aggregator.GetViewer("c(1:10)");
+                gridViewer.Should().NotBeNull().And.BeOfType<VectorViewer>();
+
+                gridViewer.Capabilities.Should().HaveFlag(ViewerCapabilities.List | ViewerCapabilities.Table);
+            }
+        }
+
+        [Test]
+        [Category.Variable.Explorer]
+        public void ViewerTest02() {
```
AlexanderSher: I don't understand what this test covers that the `ViewerExportTest` doesn't cover

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
AlexanderSher: Me too. Let's leave it as it is.

src/Package/Test/DataInspect/ReplCommandsTest.cs
```diff
@@ -0,0 +1,232 @@
...
+                gridViewer.CanView(eval).Should().BeFalse();
+            }
+        }
+
+        [Test]
+        [Category.Variable.Explorer]
+        public async Task Viewer1DTest() {
+            using (var hostScript = new RHostScript(_sessionProvider)) {
+                var viewer = await _aggregator.GetViewer("airmiles") as Viewer1D;
+                viewer.Should().NotBeNull();
```
AlexanderSher: You have tested it in `ViewerExportTest`, so here you can just create an instance of Viewer1D directly, so you won't need `RHostScript`. You can also make it a composite test by providing json strings and using real `DebugValueEvaluationResult` (first 4 parameters in ctor can be null)

src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
MikhailArkhipov: Not really. Viewer is determined by specific data type. Whenit changes, new model is created.

src/Package/Impl/uur1ketl.tmp_proj
MikhailArkhipov: Hmmm VS keeps generating these or whatever. Will delete

src/Package/Test/DataInspect/ReplCommandsTest.cs
MikhailArkhipov: I guess that's true.

src/Package/Test/DataInspect/ReplCommandsTest.cs
MikhailArkhipov: Yep. I will look into composite next round - there are many tests in the viewer and VE space that can be composite.

src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
AlexanderSher: Hm... In this case we probably can add another ctor to `VariableViewModel` that will accept `IObjectDetailsViewer` directly instead of retrieving it from `IObjectDetailsViewerAggregator` and remove this export from `GridViewerBase` completely.

MikhailArkhipov: Well only model really knows which viewer(s) that particular model needs. It may decide and fetch specific one rather than just base it on expression. So I think the selection process should be encapsulated.

MikhailArkhipov: In the original question I thought you meant if viewer can change over lifetime of the model.That is no. But model does can decide differently on which viewer(s) it wants.

AlexanderSher: Got it

AlexanderSher:
👍

Issue #1726, opened 2016-05-03
Xaml binding error in variable explorer


huguesv:
```
System.Windows.Data Error: 40 : BindingExpression path error: 'OpenInCsvAppCommandTooltip' property not found on 'object' ''VariableViewModel' (HashCode=3957848)'. BindingExpression:Path=Model.Content.OpenInCsvAppCommandTooltip; DataItem='ObservableTreeNode' (HashCode=40992173); target element is 'Button' (Name=''); target property is 'ToolTip' (type 'Object')

```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/9b267f51472a0d5807eb8e475992ad8350b44624

Issue #1591, opened 2016-04-16
Peek/Goto definition for library functions should work like View(function)


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/9b267f51472a0d5807eb8e475992ad8350b44624

PR #1728, opened 2016-05-04
#1591 #1726

MikhailArkhipov:
#1591 Peek/Goto definition for library functions should work like View(func)
#1726 Xaml binding error in variable explorer

+ tests


src/Host/Client/Impl/Session/InteractiveWindowRSessionHelper.cs
```diff
@@ -1,9 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using Microsoft.R.Host.Client;
-
-namespace Microsoft.VisualStudio.R.Package.Repl {
+namespace Microsoft.R.Host.Client.Session {
     public static class InteractiveWindowRSessionHelper {
```
AlexanderSher: I don't like it. Host.Client has no idea about interactive window.

MikhailArkhipov:
Will move around and use interactive workflow provider as discussed offline

PR #1731, opened 2016-05-05
Debugger refactoring

int19h:
* Split evaluation, stack tracing, and execution tracing (stepping and breakpoints) into separate components, and rename types and members to be more descriptive of the functions they perform.

* Use interfaces for the entirety of the resulting public API surface.

* Rewrite unit tests to use NSubstitute instead of hand-coded mocks and match objects.

* Various small fixes.


int19h:
You really want to use CodeFlow for this one, by the way, since it's much better at detecting when file has been renamed and then edited than GitHub diff.

src/Debugger/Impl/DataInspection/IRValueInfo.cs
```diff
@@ -0,0 +1,102 @@
...
+        RValueFlags Flags { get; }
+
+        /// <summary>
+        /// <see langword="true"/> if <see cref="IREvaluationInfo.DescribeChildrenAsync"/> is known to return any items,
+        /// otherwise <see langword="false"/>.
+        /// </summary>
+        bool HasChildren { get; }
+    }
+
+    public static class RValueInfoExtensions {
```
AlexanderSher: Please move extensions into separate file

src/Debugger/Impl/DataInspection/RActiveBindingInfo.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.R.Host.Client;
+
+namespace Microsoft.R.DataInspection {
+    internal sealed class RActiveBindingInfo : REvaluationInfo, IRActiveBindingInfo {
```
AlexanderSher: `IRActiveBindingInfo` is already derived from `REvaluationInfo`

src/Debugger/Impl/DataInspection/REvaluationInfo.cs
```diff
@@ -0,0 +1,104 @@
...
+                    var jEvalResult = (JObject)kv.Value;
+                    var evalResult = Parse(Session, EnvironmentExpression, name, jEvalResult);
+                    children.Add(evalResult);
+                }
+            }
+
+            return children;
+        }
+
+        public override bool Equals(object obj) =>
```
AlexanderSher: Where is it needed?

src/Debugger/Impl/DataInspection/IRValueInfo.cs
int19h: Is there any particular point in it? They really are just interface methods with default implementations, and are tightly tied to the interface in question. I don't want any outside code to even be aware of the fact that they're extension methods.

---
In reply to: [62135595](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62135595) [](ancestors = 62135595)

src/Debugger/Impl/DataInspection/REvaluationInfo.cs
int19h: This is leftover from when the tests were using match objects. I'll get rid of it.

---
In reply to: [62135945](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62135945) [](ancestors = 62135945)

src/Debugger/Impl/DataInspection/IRValueInfo.cs
AlexanderSher: So these methods are used only inside `IRValueInfo` implementation?

---
In reply to: [62136058](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136058) [](ancestors = 62136058,62135595)

src/Package/Impl/DataInspect/Viewers/ViewerBase.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Threading.Tasks;
 using Microsoft.R.Components.Extensions;
+using Microsoft.R.DataInspection;
 using Microsoft.R.Debugger;
```
MikhailArkhipov: This probably no longer needed?

src/Debugger/Impl/DataInspection/IRValueInfo.cs
int19h: They are used inside, but are also intended to be used by API clients. My point is rather that they are really a part of the interface as far as clients should be concerned - the only reason why it's an extension class is because C# doesn't support default implementations for interface methods. 

---
In reply to: [62136331](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136331) [](ancestors = 62136331,62136058,62135595)

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
```diff
@@ -33,12 +34,12 @@ internal sealed class FunctionViewer : ViewerBase, IObjectDetailsViewer {
         #region IObjectDetailsViewer
         public ViewerCapabilities Capabilities => ViewerCapabilities.Function;
 
-        public bool CanView(IDebugValueEvaluationResult evaluation) {
-            return evaluation != null && evaluation.Classes.Count == 1 && evaluation.Classes.Any(t => _classes.Contains(t));
+        public bool CanView(IRValueInfo evaluation) {
+            return evaluation != null && evaluation.Classes.Count == 1 && evaluation.Classes[0].EqualsOrdinal("function");
```
MikhailArkhipov: what about "formula" above

src/Debugger/Impl/DataInspection/IRValueInfo.cs
AlexanderSher: Why default implementation matters? You expect these interfaces to be implemented by users?

---
In reply to: [62136453](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136453) [](ancestors = 62136453,62136331,62136058,62135595)

src/Package/Impl/DataInspect/Viewers/ViewerBase.cs
int19h: Yep. I'll do a sweep - there shouldn't be anything using it anymore. In fact, it shouldn't even be declared.

---
In reply to: [62136426](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136426) [](ancestors = 62136426)

int19h: Ah, it's because of Microsoft.R.Debugger.Engine. I'll probably rename that to just R.Debugger later on, but I'll remove this (and elsewhere) for now.

---
In reply to: [62136654](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136654) [](ancestors = 62136654,62136426)

src/Debugger/Impl/ExecutionTracing/IRExecutionTracer.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        /// <summary>
+        /// If a step operation is currently in progress, cancel it.
+        /// </summary>
+        bool CancelStep();
+
+        /// <summary>
+        /// Enables or disables breakpoints. When enabled, <see cref="IRBreakpoint.BreakpointHit"/> events are raised.
+        /// </summary>
+        Task EnableBreakpointsAsync(bool enable, CancellationToken cancellationToken = default(CancellationToken));
```
AlexanderSher: Probably should be `ToggleBreakpointsAsync`

src/Debugger/Impl/ExecutionTracing/RBreakpoint.cs
```diff
@@ -10,19 +10,20 @@
...
-    /// A breakpoint in R code.
-    /// </summary>
-    public sealed class DebugBreakpoint {
-        public DebugSession Session { get; }
-        public DebugSourceLocation Location { get; }
+namespace Microsoft.R.ExecutionTracing {
+    internal sealed class RBreakpoint : IRBreakpoint {
+        private readonly RExecutionTracer _tracer;
+
+        public IRExecutionTracer Tracer => _tracer;
```
AlexanderSher: `public IRExecutionTracer Tracer { get; }`

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
int19h: Bad merge - will fix. Looks like we don't have a test that would catch this, either - should I add one?

---
In reply to: [62136631](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136631) [](ancestors = 62136631)

MikhailArkhipov:
:+1:

src/Package/Impl/DataInspect/Viewers/FunctionViewer.cs
MikhailArkhipov: Yes, we miss it. FunctionViewerTest checks function but not formula

---
In reply to: [62137005](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62137005) [](ancestors = 62137005,62136631)

src/Debugger/Impl/ExecutionTracing/IRExecutionTracer.cs
int19h: Toggle implies changing state to the reverse of what it's currently is (and so requires you to be aware of what the current state is). This one actually lets you do EnableBreakpointsAsync(true) regardless of what the current state is - and if they're already enabled, it'll do nothing.

---
In reply to: [62136887](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136887) [](ancestors = 62136887)

src/Debugger/Impl/ExecutionTracing/RBreakpoint.cs
int19h: That doesn't work, because return types don't match - and _tracer has to be RExecutionTracer rather than IRExecutionTracer, so that RemoveBreakpoint can be called on it.

---
In reply to: [62136948](https://github.com/Microsoft/RTVS/pull/1731#discussion_r62136948) [](ancestors = 62136948)

src/Debugger/Impl/ExecutionTracing/IRExecutionTracer.cs
```diff
@@ -0,0 +1,115 @@
...
+
+namespace Microsoft.R.ExecutionTracing {
+    /// <summary>
+    /// Wraps <see cref="IRSession"/> and provides functionality to step through R code and set breakpoints in it.
+    /// </summary>
+    /// <remarks>
+    /// Use <see cref="RSessionExtensions.TraceExecutionAsync"/> to obtain an object implementing this interface
+    /// for a given <see cref="IRSession"/>.
+    /// </remarks>
+    public interface IRExecutionTracer {
```
int19h: Suggestions for a better name are welcome, by the way :)

src/Package/Impl/DataInspect/Viewers/DebugObjectEvaluator.cs
```diff
@@ -1,41 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See LICENSE in the project root for license information.
```
int19h: Seems to be another merge issue, I'll bring this back.

src/Debugger/Impl/DataInspection/RValueProperties.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.R.DataInspection {
+    /// <summary>
+    /// Used to specify properties of <see cref="IRValueInfo"/> to fill when evaluating expressions using
+    /// <see cref="RSessionExtensions.EvaluateAndDescribeAsync"/>,
+    /// <see cref="RStackFrameExtensions.EvaluateAndDescribeAsync"/>,
+    /// or <see cref="REvaluationInfoExtensions.DescribeChildrenAsync"/>.
+    /// </summary>
+    [Flags]
+    public enum RValueProperties : ulong {
```
int19h: I really don't like this name. It's technically correct in a sense that all members here _currently_ correspond to properties of `IRValueInfo`. But 1) this may not necessarily be true in the future, and 2) it is actually passed to various methods that return `IREvaluationInfo`, which may or may not be `IRValueInfo`, which is confusing.

I considered renaming it to `REvaluationProperties`, but that also feels weird, because it reads as "properties of evaluation", not "properties of what is produced by evaluation".

I would really like to get the names right in this refactoring, to avoid sweeping changes across all the code that uses it in the future, and this is a particularly high-traffic type. So, I would really appreciate any suggestions on the naming.

int19h: (ping to show it on the most recent commit)

src/Debugger/Impl/DataInspection/RValueProperties.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.R.StackTracing;
+using static System.FormattableString;
+
+namespace Microsoft.R.DataInspection {
+    /// <summary>
+    /// Used to specify properties of <see cref="IRValueInfo"/> to fill when evaluating expressions using
+    /// <see cref="RSessionExtensions.TryEvaluateAndDescribeAsync"/> and its various wrappers.
+    /// </summary>
+    [Flags]
+    public enum RValueProperties : ulong {
```
int19h: I really don't like this name. It's technically correct in a sense that all members here _currently_ correspond to properties of `IRValueInfo`. But 1) this may not necessarily be true in the future, and 2) it is actually passed to various methods that return `IREvaluationInfo`, which may or may not be `IRValueInfo`, which is confusing.

I considered renaming it to `REvaluationProperties`, but that also feels weird, because it reads as "properties of evaluation", not "properties of what is produced by evaluation".

I would really like to get the names right in this refactoring, to avoid sweeping changes across all the code that uses it in the future, and this is a particularly high-traffic type. So, I would really appreciate any suggestions on the naming.

MikhailArkhipov: Well, it is not ideal but name at least looks better than other options. Perhaps RValueAttributes or RValueInfoAttributes? sort of like FileInfo I guess.

int19h: (pokes @AlexanderSher)

The "Properties" part is right I think, since it does directly map to properties of the resulting objects. It's the "Value" part that I'd like to get rid of.

Since it's used with methods, all of which begin with `Describe`..., maybe `RDescribeProperties`?

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -131,35 +132,29 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
...
+                        const REvaluationResultProperties fields =
+                            REvaluationResultProperties.ExpressionProperty |
+                            REvaluationResultProperties.AccessorKindProperty |
+                            REvaluationResultProperties.TypeNameProperty |
+                            REvaluationResultProperties.ClassesProperty |
+                            REvaluationResultProperties.LengthProperty |
+                            REvaluationResultProperties.SlotCountProperty |
+                            REvaluationResultProperties.AttributeCountProperty |
+                            REvaluationResultProperties.DimProperty |
+                            REvaluationResultProperties.FlagsProperty;
```
AlexanderSher: So you don't use `using static Microsoft.R.DataInspection.REvaluationResultProperties`

int19h: Nope, just didn't finish going over all these before committing by mistake. I'll do `using static` in all places where there's more than a couple.

int19h: Here we go.

AlexanderSher:
:+1:

Issue #1235, opened 2016-03-09
library, data, ... pop up external window


smortaz:
from user:

Examples of functions that do this:
```r
library()
data()
```

int19h:
Okay, so there are some alternatives here. The default and current behavior is to open a separate window (R offers that):

![image](https://cloud.githubusercontent.com/assets/1175142/13864953/4a2d4d9a-ec63-11e5-9375-7c93cefcc99e.png)

We can set `options('pager')` to `'console'`, in which case it will just print out to the REPL - that's by far the easiest solution (1h to implement), and it works in a sense that there are no windows, and REPL is scrollable, so functionally it's adequate. But clearly the intent is to provide a separate viewing window...

RS opens the text as a document in the editor well instead - this can also be done, but it requires more effort, as we'll need to implement it as a virtual document to give R control over the title, and also support automatic deletion of the displayed temporary file once the document is closed. I'd estimate it as at least one full day of work to wire everything up and test.

@jflam, your take?

MikhailArkhipov:
I remember JavaScript debugger in VS opens runtine JS as 'transient' document. Typically comes to the editor factory with name like ^something.js. Basically, no file on disk, just transient text buffer with contents.

int19h:
Yes, VS supports that. it's a bit easier for debuggers though, because there you just need to implement [IDebugDocument2](https://msdn.microsoft.com/en-us/library/bb145153.aspx) to retrieve the source code, and VS debugger service wires it up to the editor. Since this scenario doesn't actually involve the debugger, I'll need to figure out how it all works under the hood first.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/c15056c2fa91884f38d5b672002d84598d3be5e1 for library and View(). data TBD.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/2ac3848c8cd29014a9aeca12f0d3855e26f04bd8

PR #1734, opened 2016-05-05
#1235 library, data, ... pop up external window

MikhailArkhipov:
Implement override via 'pager' option


MikhailArkhipov:
Yeah I know, yet another method on callback interface... We should get together and discuss perhaps calling by name or something. Doesn't want to get into reflection though.

src/Common/Core/Impl/IO/Exports.cs
```diff
@@ -2,12 +2,8 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.ComponentModel.Composition;
-using Microsoft.Common.Core.IO;
-using Microsoft.VisualStudio.ProjectSystem.Utilities;
 
-namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
-
-    [AppliesTo("RTools")]
+namespace Microsoft.Common.Core.IO {
```
AlexanderSher: `Export` was separated from `FileSystem` implementation so that test assemblies (e.g. `Components.Tests`) could either use real `FileSystem` or replace it with fake. If you move it to `Common.Core`, this won't work anymore.

MikhailArkhipov: I'll undo this one but there is a bunch of missing MEF imports getting thrown in interactive tests (including this one), need to clean them up eventually. Luckily interactive tests don't rely on those, but that is just so far...

PR #1743, opened 2016-05-06
Object tools

MarcoPortmann:
Adds four simple R commands (summary, dim, names, head) to the toolbar
and the editor context menu.


MarcoPortmann:
These are my first lines of C# code I've ever written. I guess there could be a more efficient way to implement these four commands since all of them are of the form "commandName(<Editor.SelectedText>)" where only the commandName differs.

MikhailArkhipov:
Hi https://github.com/MarcoPortmann . Pretty impressive for the first C# code :-) However, commands like these are better added as code snippets. For example, look at the completion list for lm - you can see scissors icon.

image

 That means that you can type lm[TAB] then ( and get regular completion or type lm[TAB][TAB] and get 
image

VS then allows you to use TAB to navigate between fields to fill values. There is actually a snippet for summary:

image

You can add more snippets either to the product in Package\Impl\Snippets or to custom snippets just for you in Tools | Code Snippet Manager

MikhailArkhipov:
[summary.snippet.txt](https://github.com/Microsoft/RTVS/files/252688/summary.snippet.txt)

MarcoPortmann:
Thanks @MikhailArkhipov, I agree that the code snippets are very handy. However, the idea of the commands I added are slightly different, I think. They are not added to the R script, they are just executed to check the current state of the variables. I find that very handy when I'm debugging a script. These commands are rather supposed to complement the variable explorer.

smortaz:
i think marco has a point.  sometime i just want to right click, click and
get my answer, rather than actually type something, get a popup, tab tab
tab for something simple like a summary().  it's a function of click count
& context.

MarcoPortmann:
Although, I see the potential problem the the menus are getting cluttered quite quickly even with a small number of such commands. But at least for me there is a list of about 4-5 commands which would be sufficient.

MikhailArkhipov:
Question is how many commands should be included in the menu so it does not turn into this

image

Everyone has their own preferences but if there IS a list of commands used frequently by most people, then sure, we can discuss that. Also, it is hard to find suitable keyboard shortcuts in VS that are not taken by something else...

smortaz:
good point.  can we get rid of the codemap, using, etc. when in R mode?

MarcoPortmann:
quickaccess

I agree, I don't know whether other users have the same needs. I use a mini toolbar in my editor. Therefore, commands use less space than in the context menu. More elegant solution would be to allow users to customize these commands. Or to use code snippets with the option to execute them once.

MikhailArkhipov:
@smortaz : this is C# menu, not R but it is just an example of uncontrolled menu growth :-)

smortaz:
ah, yeah, it looks ridiculous.  i like the idea of pulling often used stuff
& putting them elsewhere where they're a single click away.

int19h:
How about making this a hover thing instead of a context menu? Kinda like DataTip in debugging. In fact, that scenario alone is worth it IMO, as right now you have to have debugger attached in order to get variable values when hovering over them in the editor, even though VE is always accessible.

If we implement our own DataTips that work outside of debug mode and _show_ the value, then that popup will also be the logical place to put commands that _describe_ it.

To remind, this is what the standard VS DataTip UI looks currently when you hover over variable:

image

So I'm thinking of something very similar, but with a toolbar or a menu right below.

MarcoPortmann:
@int19h, I like the idea of a DataTip with a menu below.

tip

The tip could contain a value, dim, object type, similar to the picture above. This solution would also solve the menu cluttering problem. Would that work only for well-defined variables, i.e. when you place the cursor on "mydata" in  "mydata <- subset(whatever)" or could it even work when you select an expression like "subset(whatever)" and hover over the selection?

int19h:
It could be implemented either way. The tricky part with arbitrary selections is that evaluating them may result in side effects (e.g. consider if you accidentally select and hover over something like `rm(x)`...). That may be mitigated by blocking mouse hover if an expression has any function calls, assignment operators etc, and providing a command / keyboard shortcut to activate the same popup manually.

int19h:
For debugger DataTips (which let you hover over selection) we currently do not show the hover if expression has any function calls (including nested subexpressions to catch things like `x[f()]`). You can see the way it does that here:

https://github.com/Microsoft/RTVS/blob/master/src/R/Core/Impl/DataTips/RDataTip.cs

MarcoPortmann:
Yes, exactly. I was about to editing and adding that issue to my comment above. But I think the hover tip would still work with arbitrary selections. As long as only the "empty tip" with the menu is showed and the expression is not evaluated unless the user selects one of the additional commands from the menu.

MikhailArkhipov:
There is support for hovering items in VS, it is called "Suggestions" (aka Light Bulb or Smart Tags). We show them for library(). Now, if you want something custom, then it is pretty big undertaking - I can imagine configurable WPF custom presenter for command items etc.

MikhailArkhipov:
@MarcoPortmann - could you attach screenshot with the menu that includes added commands? Thanks

MarcoPortmann:
objcommands
@MikhailArkhipov, here comes the screenshot.
Hover tip with a menu, showing basic info for well-defined variables and only the menu for complex expressions sounds great. But I can imagine that it is a slightly larger undertaking. :)

MikhailArkhipov:
```Ctrl+S``` will override File Save (putting Ctrl+S in a double-keystroke shortcut will remove single-keystroke shortcut). Similarly, ```Ctrl+N``` will override File New. Not sure if users will expect this. Typically double keystrokes start with Ctrl+K or M

MikhailArkhipov:
```Ctrl+H``` = Replace and ```Ctrl+D``` is 'Go To Find Combo'

MikhailArkhipov:
VS context specific suggestions (aka light bulb):

image

MarcoPortmann:
Oh, :( didn't know that it affects single-keystroke shortcut. Just assumed it distinguishes double-keystroke shortcuts from single with a time delay. Now I understand what you said earlier: not easy to find suitable keyboard shortcuts in VS.

MikhailArkhipov:
No worries. Keep PR open we are still discussing.

int19h:
Since there hasn't been any activity on this PR for a long time now, should we close it? The code has changed a lot since then, so this will need a lot of merging at a minimum - probably to the point where it's easier to redo it from scratch.

MikhailArkhipov:
👍

MarcoPortmann:
Unfortunately, I haven't had the time to work on it. I still think that shortcuts (or an office-style mini toolbar or something similar) that allow to quickly check code and the content of R objects by showing the head/summary/dim/view/names of the word at the caret position or for the current text selection are extremely handy. I have this feature in my (otherwise buggy) version of Tinn-R and I use them on a daily base.
But anyway, closing the PR and starting from scratch is certainly wiser than revise the few lines of code.

int19h:
Could you create a new issue for this, so that we can track and triage it appropriately? And reference this PR from there, so that all the discussion, screenshots etc above, as well as the code, is preserved for a future take on it.

Issue #1174, opened 2016-03-02
Remove dependency on Web Developer Tools installation in RTVS


jflam:
We use types in one assembly - for HTML parsing, correct? If so, we need to remove the dependency in installing all of Web Developer Tools just for this dependency.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/aea56bc8f9187c1ff097609daacdd28aef06e488

PR #1744, opened 2016-05-06
#1174 Remove dependency on Web Developer Tools installation in RTVS

MikhailArkhipov:
Include HTML parser sources and tests. ASP.NET, PHP and Razor related code as well as HTML editor-specific code (such as incremental update support or tree cloning) was removed.


MikhailArkhipov:
I converted test to xUnit, but left regular asserts. We can convert to fluent in the future.

PR #1745, opened 2016-05-06
Debugger assembly & namespace refactoring

int19h:
Remove the existing Microsoft.R.Debugger assembly, and move classes in it to Microsoft.R.Host.Client + same thing for the tests.

Rename Microsoft.R.Debugger.Engine -> Microsoft.R.Debugger.


huguesv:
:+1:

Issue #1679, opened 2016-04-25
Publishing markdown in folders named in Unicode fails


MikhailArkhipov:
```
In addition: Warning messages:
1: In dir.create(output_dir, recursive = TRUE) :
  cannot create dir 'C:\Users\mikhaila\documents\visual studio 2015\Projects\rproject15\rproject15\????', reason 'Invalid argument'
2: In normalizePath(path.expand(path), winslash, mustWork) :
  path[1]="C:/Users/mikhaila/documents/visual studio 2015/Projects/rproject15/rproject15/????": The filename, directory name, or volume label syntax is incorrect
3: In normalizePath(path.expand(path), winslash, mustWork) :
  path[1]="C:/Users/mikhaila/documents/visual studio 2015/Projects/rproject15/rproject15/????/markdown.rmd": The filename, directory name, or volume label syntax is incorrect
Execution halted
Markdown conversion failed
```

will try setting cp for non-Unicode

MikhailArkhipov:
Setting command line CP makes display OK but publishing still fails

```
Error in tools::file_path_as_absolute(input) : 
  file 'C:/Users/mikhaila/Documents/Visual Studio 2015/Projects/rproject15/rproject15/фыва/markdown.rmd' does not exist
Calls: <Anonymous> -> setwd -> dirname -> <Anonymous>
Execution halted
Markdown conversion failed
```

also, display of the temp folder is wrong. 

![image](https://cloud.githubusercontent.com/assets/12820357/14773575/5c82543e-0a60-11e6-9c1c-606db957764e.png)


Most probably same as #1661

int19h:
Can we avoid Rscript for publish, and do direct eval in the current R session (or maybe a freshly spawned session)? I suspect the problem here is that it has to go through command line, which means that it has to undergo codepage/Unicode translation back and forth; whereas with eval, it's Unicode end-to-end now (if you use `REvaluationKind.Json` for the result).

MikhailArkhipov:
Yeah #1156

MikhailArkhipov:
After switching to R Host per #1156 both evaluation and interaction produce
```
Error in find_external_resources(original_input, encoding) : 
  The input file file 'C:/Users/mikhaila/Documents/Visual Studio 2015/Projects/rproject6/rproject6/<U+0444><U+044B><U+0432><U+0430>/mark down.rmd' does not exist.
In addition: Warning messages:
[...]

PR #1748, opened 2016-05-09
 #1679 Publishing markdown in folders named in Unicode fails

MikhailArkhipov:
Transfer system non-Unicode locale to R. Provide list of locales in settings in care user prefers to set it explicitly just for R.

image

image


src/Common/Core/Impl/Extensions/CultureExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Globalization;
+using System.Linq;
+
+namespace Microsoft.Common.Core {
+    public static class CultureExtensions {
+        private static readonly char[] _separators = new char[] { ',', ' ' };
+
+        public static string LanguageNameFromLCID(int lcid) {
+            var culture = CultureInfo.GetCultures(CultureTypes.InstalledWin32Cultures).FirstOrDefault(x => x.LCID == lcid);
+            return culture.GetLanguageName();
```
AlexanderSher: culture can be null

src/Common/Core/Impl/Extensions/CultureExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        public static IEnumerable<string> GetLanguageNames() {
+            IEnumerable<CultureInfo> cultures = CultureInfo.GetCultures(CultureTypes.InstalledWin32Cultures);
+            return cultures.Select(c => c.GetLanguageName()).Distinct().OrderBy(x => x);
```
int19h: Are we sure that for the cases where there will be duplicates (I'm assuming this is after stripping things after separators), after eliminating them, the remaining entry will be the one that the user wants?

E.g. for Chinese, there are two common encodings, Big5 and GB. If we show only one entry for "China", and always pick one and never the other, it will present a problem.

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -64,7 +66,15 @@ internal sealed class RToolsSettingsImplementation : IRToolsSettings {
...
-                IdleTimeAction.Create(async () => await SetMirrorToSession(), 20, typeof(RToolsSettingsImplementation));
+                IdleTimeAction.Create(async () => await SetMirrorToSession(), 20, _cranMirror ?? (object)this.GetType());
+            }
+        }
+
+        public string RLocale {
+            get { return _locale; }
+            set {
+                _locale = value;
+                IdleTimeAction.Create(async () => await SetSessionLocale(), 20, _locale ?? (object)typeof(CultureInfo));
```
AlexanderSher: `SetSessionLocale` runs on background thread, so I don't think we need to wait for idle.

src/Common/Core/Impl/Extensions/CultureExtensions.cs
int19h: Perhaps it is better to not deal with culture names at all here, and just show encoding names? Those are guaranteed to be unique; and given that we only set LC_CTYPE (and not all locale categories), showing culture name is somewhat misleading anyway...

And it appears that R's (actually msvcrt.dll) `setlocale` lets you specify only the encoding name like this: `Sys.setlocale(LC_CTYPE, ".1251")`

(see here for syntax: https://msdn.microsoft.com/en-us/library/aa246449(v=vs.60).aspx)

MikhailArkhipov: It is indeed an interesting problem. There is no really a good list of what particular C runtime accepts in LC_CTYPE. I can ping MSVCRT folks and get their list. .NET list does not map 1:1 and there is no 'country' name there. RS only provides ~10 encodings in their list although Sys.setlocale accepts quite a few more. I can try and make list like Language.LCID if the helps?

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
MikhailArkhipov: Probably. I need to check why the other piece of similar code does this.

src/Common/Core/Impl/Extensions/CultureExtensions.cs
int19h: My understanding is that they don't use their own database for msvcrt.dll, but rather call into Win32 API - so whatever locales Windows supports, msvcrt also does.

int19h: Also, it should support any codepage that Windows does (save for 65001). So if we set LC_CTYPE using codepage number only, that should make it that much more reliable. The trick will be getting the list of all encoding names.

MikhailArkhipov: Although despite of what MSDN says setlocale(".OCP") didn't work for me

MikhailArkhipov: Language name seemed to be friendlier than .1251. I am not sure how many are familiar with Windows LCID numbers. Actual code page is 866 but it is not unique since several other cultures use 866 as OEM. So maybe do have both?

int19h: Sounds like maybe `Encoding.GetEncodings` could be used to generate a list, with `EncodingName` as display label, and the actual codepage number used when setting (and filter out all encodings that don't have a codepage).

MikhailArkhipov: Basically something like Russian (866) or something like this?

int19h:
```
> Encoding.GetEncoding(866).EncodingName
"Cyrillic (DOS)"
> Encoding.GetEncoding(1251).EncodingName
"Cyrillic (Windows)"
> Encoding.GetEncoding(950).EncodingName
"Chinese Traditional (Big5)"
> Encoding.GetEncoding(936).EncodingName
"Chinese Simplified (GB2312)"
```

MikhailArkhipov: Also, R does not support all Win32 locales. Here is 65001

image

int19h: It's an msvcrt limitation (and also present in the current VC++ runtime). Basically 65001 is "too special" or something.

MikhailArkhipov: Let me try encodings, but we somehow need to filter out what R rejects. It seem to be OK with LCID and languages though and finds the right CP.

int19h: LCID is fine too - I'm pretty sure it maps one to the other internally via Win32, so either one works. So long as we get all the LCIDs for a given language, for that case of multiple popular encodings.

MikhailArkhipov: setting just cp looks a bit odd

image

int19h: But it's accurate, since the only part of the locale name that matters when you set `LC_CTYPE` is the codepage (which uniquely specifies the encoding).

PR #1750, opened 2016-05-09
#1679 Publishing markdown in folders named in Unicode fails

MikhailArkhipov:
Changed to encodings

image


src/Common/Core/Impl/Extensions/CultureExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace Microsoft.Common.Core {
+    public static class CultureExtensions {
```
int19h: Is this still used anywhere?

MikhailArkhipov: No, I'll remove it.

src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+        public override bool GetStandardValuesSupported(ITypeDescriptorContext context) {
+            return true;
+        }
+
+        public override StandardValuesCollection GetStandardValues(ITypeDescriptorContext context) {
+            var codePageList = new List<int>() { 0 };
+            var codePages = Encoding.GetEncodings()
+                                    .OrderBy(e => e.DisplayName)
+                                    .Select(e => e.CodePage)
+                                    .Where(cp => cp != Encoding.UTF7.CodePage &&
```
int19h: This would be more readable as a static readonly array, and you can then do `.Except` instead of `.Where`

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -159,6 +161,14 @@ public static class RSessionEvaluationCommands {
             return evaluation.EvaluateAsync(script, REvaluationKind.Mutating);
         }
 
+        public static Task SetCodePage(this IRExpressionEvaluator evaluation, int codePage) {
+            if (codePage == 0) {
+                codePage = NativeMethods.GetOEMCP();
+            }
+            var script = Invariant($"Sys.setlocale('LC_CTYPE', '.{codePage}')");
+            return evaluation.EvaluateAsync(script, REvaluationKind.Mutating);
```
int19h: `ExecuteAsync` (and you can then drop `Mutating`).

int19h:
:+1:

PR #1762, opened 2016-05-12
Project property pages for run settings

huguesv:
Screenshots:
image

image


src/Package/Impl/ProjectSystem/IVsHierarchyExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.VisualStudio.ProjectSystem.Designers;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
+    internal static class IVsHierarchyExtensions {
+        /// <summary>
+        /// Returns EnvDTE.Project object for the hierarchy
+        /// </summary>
+        public static EnvDTE.Project GetDTEProject(this IVsHierarchy hierarchy) {
+            //UIThreadHelper.VerifyOnUIThread();
```
huguesv: What's our equivalent for this?

src/Package/Impl/Rules/Run.cs
```diff
@@ -0,0 +1,270 @@
...
+// <auto-generated>
+//     This code was generated by a tool.
+//     Runtime Version:4.0.30319.42000
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+
+// WARNING: this file was auto-generated then modified.
```
huguesv: I sent out an email to cps mailing list regarding my troubles with XamlPropertyRule build action.  I haven't heard back yet.

src/Package/Impl/ProjectSystem/IVsHierarchyExtensions.cs
AlexanderSher: CoreShellExtensions.AssertIsOnMainThread

---
In reply to: [63072611](https://github.com/Microsoft/RTVS/pull/1762#discussion_r63072611) [](ancestors = 63072611)

src/Package/Impl/ProjectSystem/PropertyPages/WpfBasedPropertyPage.cs
```diff
@@ -0,0 +1,85 @@
...
+        protected abstract PropertyPageViewModel CreatePropertyPageViewModel();
+
+        protected abstract PropertyPageControl CreatePropertyPageControl();
+
+        protected async override Task OnSetObjects(bool isClosing) {
+            if (isClosing) {
+                _control.DetachViewModel();
+                return;
+            } else {
+                //viewModel can be non-null when the configuration is chaged.
```
AlexanderSher: typo

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageViewModel.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// </summary>
+        internal static readonly string DifferentStringOptions = Resources.PropertyPageDifferentOptions;
+
+        /// <summary>
+        /// Special value that indicates that the bool property values for the selected
+        /// configurations are not identical.
+        /// </summary>
+        internal static readonly bool? DifferentBoolOptions = null;
+
+        public UnconfiguredProject UnconfiguredProject { get; set; }
```
AlexanderSher: Why it is in the base class?

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        [ImportingConstructor]
+        public PropertyPageProvider(UnconfiguredProject unconfiguredProject) {
+            PropertyPages = new OrderPrecedenceImportCollection<IPageMetadata>(projectCapabilityCheckProvider: unconfiguredProject);
+        }
+
+        public Task<IReadOnlyCollection<IPageMetadata>> GetPagesAsync() {
+            return Task.Run<IReadOnlyCollection<IPageMetadata>>(
+                () => PropertyPages.Select(p => p.Value).ToList().AsReadOnly()
+            );
```
AlexanderSher: Why do you need to run task here? Task.FromResult looks good enough

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageControl.cs
```diff
@@ -0,0 +1,82 @@
...
+                    OnStatusChanged(new EventArgs());
+                }
+            }
+        }
+
+        public virtual void InitializePropertyPage(PropertyPageViewModel viewModel) {
+            _ignoreEvents = true;
+            IsDirty = false;
+            ViewModel = viewModel;
+            ViewModel.PropertyChanged += ViewModel_PropertyChanged;
```
AlexanderSher: This is an antipattern. No user code should explicitly subscribe to PropertyChanged event. Chains of PropertyChanged calls are permanent bug origins. What are you trying to achieve with it?

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageViewModel.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void PopIgnoreEvents() {
+            Debug.Assert(_ignoreEventsNestingCount > 0);
+            if (_ignoreEventsNestingCount > 0) {
+                _ignoreEventsNestingCount--;
+            }
+        }
+
+        public abstract Task Initialize();
+
+        public abstract Task<int> Save();
```
AlexanderSher: This method never used by this class or any other `PropertyPage*` classes. You can move it to the derived class.

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageViewModel.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// The project properties for the currently selected project configurations.
+        /// </summary>
+        public IRProjectProperties[] ConfiguredProjectsProperties { get; set; }
+
+        /// <summary>
+        /// Since calls to ignore events can be nested, a downstream call could change the outer 
+        /// value.  To guard against this, IgnoreEvents returns true if the count is > 0 and there is no setter. 
+        /// PushIgnoreEvents\PopIgnoreEvents  are used instead to control the count.
+        /// </summary>
+        private int _ignoreEventsNestingCount = 0;
```
AlexanderSher: We have `CountdownDisposable` for this type of things, but I still don't understand the purpose. We save properties only when Save is called - why do we care about events here?

src/Package/Impl/ProjectSystem/PropertyPages/WpfBasedPropertyPage.cs
```diff
@@ -0,0 +1,85 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using System.Windows.Controls;
+using System.Windows.Forms;
+
+namespace Microsoft.VisualStudio.R.Package.ProjectSystem.PropertyPages {
+    internal abstract partial class WpfBasedPropertyPage : PropertyPage {
```
AlexanderSher: Is it reusable for non-project property pages?

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageProvider.cs
huguesv: You're right.  Not sure why dotnet project system was doing that.

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageControl.cs
huguesv: It's all about propagating 'dirty' bit flag from view model to control then to page.  Page calls IPropertyPageSite.OnStatusChange which updates the UI with a '*' and causes an Apply on the page.

I find it all very confusing, but it works.  I took it as-is from the dotnet project system.

I can try to simplify it if you think it's worth it.

src/Package/Impl/ProjectSystem/PropertyPages/WpfBasedPropertyPage.cs
huguesv: No, it's all for project properties only.

src/Package/Impl/ProjectSystem/PropertyPages/PropertyPageViewModel.cs
huguesv: `Save` is used by `PropertyPageControl`, which doesn't know about the derived class.

int19h:
:+1: , but let @AlexanderSher comment on the stuff that he pointed out.

AlexanderSher:
Ok, let's check it in. I'll take a look later how this code can be decoupled (or maybe we don't even need it).

huguesv:
This now builds and runs properly on Dev14 and Dev15.  I filed a few follow up bugs:
https://github.com/Microsoft/RTVS/issues/1817
https://github.com/Microsoft/RTVS/issues/1816

huguesv:
I've included the fix for #1816

Issue #1759, opened 2016-05-12
Indent in function arguments in formatting is broken


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/15202814/7ddd4be8-17b2-11e6-81fa-fdec7ad77e81.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/468732e509cab0e9ff2a8d509e9afc2059d2174d

Issue #1751, opened 2016-05-09
Autoformat needs to handle multi-line statements better


MikhailArkhipov:
```
x %>% y %>%
        z |
```

typing ```%>%``` at the caret unindents the line. The reason is that formatter does not see previous line since it is not using AST to determine if given line is a continuation of a statement. Also, the said statement is not in the AST since expression is not syntactically correct.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/468732e509cab0e9ff2a8d509e9afc2059d2174d

Issue #1739, opened 2016-05-05
Auto-formatting brings inconsistent result when Expand braces is True


Marchen:
Format Selection command brings different results for same code when Expand braces is True.
Editor settings: #1738.

![autoformat](https://cloud.githubusercontent.com/assets/3863577/15038290/6d868464-12dd-11e6-8e79-fad8251727f0.PNG)

![step1](https://cloud.githubusercontent.com/assets/3863577/15038295/7a17b39c-12dd-11e6-8ca8-3fe2f17b8c9d.PNG)

Ctrl + K, Ctrl + F

![step2](https://cloud.githubusercontent.com/assets/3863577/15038300/827e9e06-12dd-11e6-9805-a27f9a0244b7.PNG)

Ctrl + K, Ctrl + F

![step3](https://cloud.githubusercontent.com/assets/3863577/15038302/87c1749c-12dd-11e6-87a7-63d504f61faf.PNG)

Ctrl + K, Ctrl + F

![step4](https://cloud.githubusercontent.com/assets/3863577/15038307/8dbd6bf8-12dd-11e6-9931-990536fb8037.PNG)

Ctrl + K, Ctrl + F

![step5](https://cloud.githubusercontent.com/assets/3863577/15038308/92bfec66-12dd-11e6-933c-564afd24090a.PNG)

Ctrl + K, Ctrl + F

![step6](https://cloud.githubusercontent.com/assets/3863577/15038320/a8bc5126-12dd-11e6-9e5e-834d78ce5989.PNG)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/468732e509cab0e9ff2a8d509e9afc2059d2174d

Issue #1738, opened 2016-05-05
Auto formatting uses Space when R editor setting for indentation is Tab


Marchen:
Though VS Options > Text Editor > R > Tab > Tab is set to Keep tabs, auto formatting insert spaces rather than tabs.
 

![options](https://cloud.githubusercontent.com/assets/3863577/15038098/df0876da-12db-11e6-8475-61e22cb48289.PNG)

See the line 430.

![before](https://cloud.githubusercontent.com/assets/3863577/15037997/f9f56dc8-12da-11e6-91d5-dbd32add1bcd.PNG)
![after](https://cloud.githubusercontent.com/assets/3863577/15037998/fce3870e-12da-11e6-9225-6975b9387677.PNG)

MikhailArkhipov:
Spaces may still be inserted if indentation length is not a multiply of tab size. Ex indent of 5 will consist of tab and 1 space.

Marchen:
Sure.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/468732e509cab0e9ff2a8d509e9afc2059d2174d

PR #1765, opened 2016-05-12
#1738 #1689 #1759 #1751 #1717 #1690 #1739

MikhailArkhipov:
Amounts to the following changes to the formatter and smart indent:
- respect single-line scopes like ```if(TRUE) {1}``` and keep them single line
- indent in function arguments at a regular level rather than align by first argument
- keep user indentation in arguments even if not aligned
- improve smart indent in argument lists with function definitions in them
- respect line breaks and user indent in multi-line arguments such as positioning of )
- update tests and add new ones

Fix VariableExplorer test instability by switching to JSON serializer
Update test baselines reflecting changes


MikhailArkhipov:
Let me look at couple changes in tests one more time

MikhailArkhipov:
Verified that original file matches so where if looked like

```
if(...) {
}
else ...
```

'else' positioning is respected and not pulled to ```} else```

MikhailArkhipov:
#1771 #1767 #1761 #1113

src/Package/Impl/DataInspect/VariableSearchTask.cs
```diff
@@ -0,0 +1,41 @@
...
+        protected override void OnStartSearch() {
+            base.OnStartSearch();
+
+            var searchString = SearchQuery.SearchString;
+            if (_grid?.Items != null && !string.IsNullOrWhiteSpace(searchString)) {
+                foreach(var itemControl in _grid.Items) {
+                    var tn = itemControl as ObservableTreeNode;
+                    var model = tn?.Model?.Content as VariableViewModel;
+                    if(model != null) {
+                        if (model.Name.StartsWithOrdinal(searchString)) {
```
int19h: I think we should do two passes here, first one using `StartsWith`, and then another one using `Contains` (so that variables that start with it are listed first, but you can also find stuff in the middle).

MikhailArkhipov: It only finds one at the moment. No multiple selection

---
In reply to: [63251922](https://github.com/Microsoft/RTVS/pull/1765#discussion_r63251922) [](ancestors = 63251922)

MikhailArkhipov: I.e. it moves actual selection not highlight

---
In reply to: [63252079](https://github.com/Microsoft/RTVS/pull/1765#discussion_r63252079) [](ancestors = 63252079,63251922)

int19h: Ah. Then do the second pass with `Contains` if we didn't find anything with `StartsWith`.

src/Package/Impl/DataInspect/VariableWindowPane.cs
```diff
@@ -21,5 +23,28 @@ internal sealed class VariableWindowPane : RToolWindowPane {
 
             ToolBar = new CommandID(RGuidList.RCmdSetGuid, RPackageCommandId.variableWindowToolBarId);
         }
+
+        public override bool SearchEnabled
```
int19h: Formatting.

src/Package/Impl/DataInspect/VariableSearchTask.cs
MikhailArkhipov: OK

---
In reply to: [63252176](https://github.com/Microsoft/RTVS/pull/1765#discussion_r63252176) [](ancestors = 63252176)

src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
```diff
@@ -32,7 +33,12 @@ internal abstract class GridViewerBase : ViewerBase, IObjectDetailsViewer {
             if (evaluation != null) {
                 await VsAppShell.Current.SwitchToMainThreadAsync();
 
-                var id = _toolWindowIdBase + evaluation.GetHashCode() % (Int32.MaxValue - _toolWindowIdBase);
+                var id = Math.Abs(_toolWindowIdBase + expression.GetHashCode() % (Int32.MaxValue - _toolWindowIdBase));
+
+                var existingPane = ToolWindowUtilities.FindWindowPane<VariableGridWindowPane>(id);
+                var frame = existingPane?.Frame as IVsWindowFrame;
+                frame?.CloseFrame(0);
```
int19h: If it exists, shouldn't we just focus on it?

src/Package/Impl/Help/HelpWindowPane.cs
```diff
@@ -28,5 +37,64 @@ internal class HelpWindowPane : VisualComponentToolWindow<IHelpVisualComponent>
...
+
+            protected override async void OnStartSearch() {
+                base.OnStartSearch();
+                _callback.ReportComplete(this, 1);
+
+                var searchString = SearchQuery.SearchString;
+                if (!string.IsNullOrWhiteSpace(searchString)) {
+                    var session = _workflowProvider.GetOrCreate().RSession;
+                    using (var inter = await session.BeginInteractionAsync()) {
+                        await inter.RespondAsync("?" + searchString + Environment.NewLine);
```
int19h: Use `ExecuteAsync` and `rtvs:::show_help`, like `ShowHelpOnCurrentCommand` (so that it doesn't require an active prompt, and falls back to `??` if `?` didn't work).

src/R/Editor/Impl/Completion/Providers/PackageFunctionCompletionProvider.cs
```diff
@@ -79,6 +80,16 @@ public class PackageFunctionCompletionProvider : IRCompletionListProvider {
         }
         #endregion
 
+        #region IRHelpSearchTermProvider
+        public IReadOnlyCollection<string> GetEntries() {
+            var list = new List<string>();
+            foreach (IPackageInfo pkg in PackageIndex.Packages) {
```
int19h: Looks like it's `SelectMany`?

src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
MikhailArkhipov: Perhaps. I need to check if it actually updates the data or if the data is detached.

---
In reply to: [63252280](https://github.com/Microsoft/RTVS/pull/1765#discussion_r63252280) [](ancestors = 63252280)

int19h: If it doesn't refresh, I'd just treat it as a separate bug to fix in another PR.

int19h:
:+1:

src/Package/Impl/Help/HelpWindowPane.cs
MikhailArkhipov: It didn't work for me. Test also does Interact. 

---
In reply to: [63252584](https://github.com/Microsoft/RTVS/pull/1765#discussion_r63252584) [](ancestors = 63252584)

MikhailArkhipov: OK, will change then test and command to "rtvs:::show_help

---
In reply to: [63253119](https://github.com/Microsoft/RTVS/pull/1765#discussion_r63253119) [](ancestors = 63253119,63252584)

int19h: The tricky part with `?` is that it doesn't cause the help to open when it is evaluated. Instead, it returns an object (actually just a string URL with a custom class) that causes help to open when it is printed. So if you just evaluate `?x`, you get that string back, and nothing opens. `show_help` wraps all the stuff that's needed to actually make it open.

On the other hand, if you want to test whether search is successful or not, this means that you can eval `?x` first and check the result. If it didn't find anything, it'll be a 0-length vector. On the other hand, `??` never fails.

PR #1770, opened 2016-05-13
Fix #1377 #1760 #908 #1733

### LINKED ISSUE #908: Data in grid launched from debugger's Local doesn't update to following command's result
1. launch data grid viewer from debugger Local window
2. run any command to change the data such as deleting the variable

actual: the grid window doesn't change
expected: it should update to command's result

### LINKED ISSUE #1377: Call stack in VE is wrong

### LINKED ISSUE #1733: View(x) should work for all language objects, not just formulas
Formulas are just language objects with class "formula" slapped on them. `View` should work for other stuff as well, like e.g. `quote(1+2)`.

### LINKED ISSUE #1760: VE exception when inside a call that is multiline
Run in REPL with VE visible:
```r
(function() { browser(); })()
```
Error message produced:
```
Error: JSON serialization failed for input:

c("(function() {", "    browser()", "})")

Vector must have 0 or 1 elements.
```
Caused by `sys.call` for that function being a character vector with more than one element, due to function definition being a part of the call, and presence of `{}` making `deparse` of that call span several lines:
```r
> (function() { sys.call(1); })()
(function() {
    sys.call(1)
})()
```r
> (function() { deparse(sys.call(1)); })()
[1] "(function() {"            "    deparse(sys.call(1))"
[3] "})()"

### PR DESCRIPTION
**Pull request by int19h**
Fix #1377: Call stack in VE is wrong
Fix #1760: VE exception when inside a call that is multiline
Fix #908: Data in grid launched from debugger's Local doesn't update to following command's result
Fix #1733: View(x) should work for all language objects, not just formulas

Harden describe_traceback() to better deal with failure to retrieve source information.

Use TracebackAsync and search() to obtain environment list in REnvironmentProvider, and use data
binding to display that data in Variable Explorer environment selector. Add icons for different
environment types.

Save the environment expression when opening data viewers, and use it consistently when fetching
data, so that it is never retrieved from the wrong environment.

Tweak Grid View and Open CSV button appearance in variable values for more consistency with Locals.

Rename FunctionViewer -> CodeViewer, and enable it to handle all language objects.

Fix various issues with value representation, add new tests.

Rewrite REnvironmentProvider tests.

### CODE REVIEW

**Comment by int19h** [comment]
Also appears to fix #1757

File: src/Package/Impl/DataInspect/Definitions/IObjectDetailsViewerAggregator.cs [hunk 1]
```diff
@@ -3,10 +3,18 @@
...
+using static Microsoft.R.DataInspection.REvaluationResultProperties;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect {
     public interface IObjectDetailsViewerAggregator {
         IObjectDetailsViewer GetViewer(IRValueInfo result);
-        Task<IObjectDetailsViewer> GetViewer(string expression);
+    }
+
+    public static class ObjectDetailsViewerAggregatorExtensions {
+        public static async Task<IObjectDetailsViewer> GetViewer(this IObjectDetailsViewerAggregator aggregator, IRSession session, string environmentExpression, string expression) {
```
MikhailArkhipov: Can we get code into a separate file from the interface

src/Package/Impl/DataInspect/Viewers/CodeViewer.cs
```diff
@@ -2,28 +2,26 @@
...
-using Microsoft.VisualStudio.R.Package.Repl;
 using Microsoft.VisualStudio.R.Package.Shell;
-using static System.FormattableString;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
     [Export(typeof(IObjectDetailsViewer))]
-    internal sealed class FunctionViewer : ViewerBase, IObjectDetailsViewer {
-        private readonly static string[] _classes = new string[] { "function", "formula" };
+    internal sealed class CodeViewer : ViewerBase, IObjectDetailsViewer {
+        private readonly static HashSet<string> _types = new HashSet<string> { "closure", "language" };
```
MikhailArkhipov: Hashset may be slower on small collections and also has expense to build it at startup but OK

MikhailArkhipov:
:+1:

src/Host/Client/Impl/DataInspection/REnvironments.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.R.DataInspection {
+    /// <summary>
+    /// Expressions for various common R environments, suitable for use as <c>environmentExpression</c> argument
+    /// when calling <see cref="RSessionExtensions.EvaluateAndDescribeAsync"/>.
+    /// </summary>
+    public static class REnvironments {
```
AlexanderSher: Is it valuable for DataInspection only?

src/Host/Client/Impl/Session/RSessionRequestSource.cs
```diff
@@ -13,19 +13,17 @@ internal sealed class RSessionRequestSource {
...
             _responseTcs = new TaskCompletionSource<object>();
             ct.Register(() => _createRequestTcs.TrySetCanceled(ct));
 
             IsVisible = isVisible;
-            Contexts = contexts ?? new[] { RHost.TopLevelContext };
         }
 
-        public void Request(string prompt, int maxLength, TaskCompletionSource<string> requestTcs) {
-            var request = new RSessionInteraction(requestTcs, _responseTcs, prompt, maxLength, Contexts);
+        public void Request(IReadOnlyList<IRContext> contexts, string prompt, int maxLength, TaskCompletionSource<string> requestTcs) {
```
AlexanderSher: Do we have any tests that explicitly cover this?

src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -3,66 +3,96 @@
...
 using System.Threading.Tasks;
 using Microsoft.Common.Core;
 using Microsoft.R.Host.Client;
+using Microsoft.R.StackTracing;
 using Microsoft.VisualStudio.R.Package.Shell;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect {
-    internal class REnvironmentProvider : IREnvironmentProvider, IDisposable {
+    internal class REnvironmentProvider : IREnvironmentProvider {
```
AlexanderSher: Is it a ViewModel object?

src/Package/Impl/DataInspect/VariableView.xaml
```diff
@@ -88,55 +89,48 @@
...
+                                        <Setter Property="BorderBrush"
+                                Value="{DynamicResource {x:Static vsui:EnvironmentColors.ToolWindowButtonHoverActiveBorderBrushKey}}" />
+                                        <Setter Property="Background"
+                                Value="{DynamicResource {x:Static vsui:EnvironmentColors.ToolWindowButtonHoverActiveBrushKey}}" />
+                                    </Trigger>
+                                </Style.Triggers>
+                            </Style>
+                        </StackPanel.Resources>
+                        <StackPanel.Visibility>
+                            <MultiBinding Converter="{wpf:Lambda '(x, y) => x || y ? Visibility.Visible : Visibility.Collapsed'}">
```
AlexanderSher: `{x:Static wpf:Converters.AnyIsNotCollapsed}`
Unfortunately, lambdas can't run in designer, so if we have converter for a common case, I'd prefer to use it instead of lambda

AlexanderSher:
>             <local:BooleanToVisibilityConverter x:Key="HidingBoolToVisibilityConverter" ValueForTrue="Visible" ValueForFalse="Hidden" />

Since you're here, can you please get rid of these two and use Microsoft.R.Wpf.Converters instead?

---
Refers to: src/Package/Impl/DataInspect/VariableView.xaml:30 in 07fe341. [](commit_id = 07fe3414344a8030930f4c2d1e9523fb3a577f85, deletion_comment = False)

src/Package/Impl/DataInspect/VariableView.xaml
```diff
@@ -160,12 +154,55 @@
...
+                  SelectedItem="{Binding SelectedEnvironment}"
+                  SelectionChanged="EnvironmentComboBox_SelectionChanged">
             <ComboBox.ItemTemplate>
                 <DataTemplate>
-                    <TextBlock Text="{Binding Name}" />
+                    <StackPanel Orientation="Horizontal">
+                        <imaging:CrispImage Width="16" Height="16">
+                            <imaging:CrispImage.Style>
+                                <Style TargetType="{x:Type imaging:CrispImage}">
+                                    <Setter Property="Moniker" Value="{x:Static imagecatalog:KnownMonikers.ObjectPublic}"/>
```
AlexanderSher: I'd rather make a converter for it. Probably put in Microsoft.R.Wpf

src/Package/Impl/DataInspect/REnvironmentProvider.cs
AlexanderSher: You can derive it from BindableBase

src/Package/Impl/DataInspect/REnvironmentProvider.cs
```diff
@@ -3,66 +3,96 @@
...
+        public void Dispose() {
+            if (_rSession != null) {
+                _rSession.Mutated -= RSession_Mutated;
+                _rSession = null;
             }
         }
 
-        private async Task GetEnvironmentAsync() {
-            await TaskUtilities.SwitchToBackgroundThread();
+        public IReadOnlyList<IREnvironment> Environments {
```
AlexanderSher: I'd suggest to use `BatchObservableCollection` here

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -101,20 +83,17 @@ public partial class VariableView : UserControl, IDisposable {
...
-            var result = await _evaluator.EvaluateAsync(GetExpression(env), properties, RValueRepresentations.Str());
-            if (result != null) {
-                var wrapper = new VariableViewModel(result, _aggregator);
-                var rootNodeModel = new VariableNode(_settings, wrapper);
-                VsAppShell.Current.DispatchOnUIThread(() => _rootNode.Model = rootNodeModel);
+            IRValueInfo result;
+            try {
+                result = await _session.EvaluateAndDescribeAsync(env.EnvironmentExpression, properties, null);
+            } catch (RException ex) {
+                SetRootNode(VariableViewModel.Error(ex.Message));
```
AlexanderSher: So this method is called from both UI and background thread?

src/Package/Impl/DataInspect/Viewers/CodeViewer.cs
AlexanderSher: I also don't see the value of this change - you don't have any unions or intersections, array has `Exists` and `Find` methods

---
In reply to: [63196108](https://github.com/Microsoft/RTVS/pull/1770#discussion_r63196108) [](ancestors = 63196108)

src/Package/Test/DataInspect/REnvironmentProviderTest.cs
```diff
@@ -2,127 +2,162 @@
...
-    public class REnvironmentProviderTest {
-        [Test]
-        [Category.Variable.Explorer]
-        public async Task GetREnvironmentTest() {
-            await RunEnvironmentTestAsync("1", (environments) => {
-                environments.Count.Should().BeGreaterOrEqualTo(2);
-                environments[0].Name.Should().Be(".GlobalEnv");
-                environments[environments.Count - 1].Name.Should().Be("package:base");
-            });
+    public class REnvironmentProviderTest : IAsyncLifetime {
```
AlexanderSher: Can we move these tests to the `Microsoft.R.Host.Client.Test` assembly?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -90,7 +90,7 @@ internal sealed class RSessionCallback : IRSessionCallback {
 
         public void ViewObject(string expression, string title) {
             var viewer = _coreShell.ExportProvider.GetExportedValue<IObjectViewer>();
-            viewer?.ViewObjectDetails(expression, title);
+            viewer?.ViewObjectDetails(_session, "base::.GlobalEnv", expression, title);
```
AlexanderSher: `REnvironments.GlobalEnv`

src/Host/Client/Impl/Session/RSessionRequestSource.cs
int19h: Not directly. I can certainly add one.

src/Host/Client/Impl/DataInspection/REnvironments.cs
int19h: Probably. When you're using regular `EvaluateAsync`, you don't specify the target environment as an expression, but rather use `REvaluationKind.BaseEnv` etc. I guess it may occasionally be useful for when your evaluated expression needs to incorporate a reference to an environment somewhere. 

Where would you prefer it to be?

AlexanderSher: `Definitions` maybe?

---
In reply to: [63222890](https://github.com/Microsoft/RTVS/pull/1770#discussion_r63222890) [](ancestors = 63222890)

src/Package/Impl/DataInspect/REnvironmentProvider.cs
int19h: Does it make a difference, given than the entire collection is refreshed completely every time?

src/Package/Impl/DataInspect/Definitions/IObjectDetailsViewerAggregator.cs
int19h: Sure.

src/Package/Impl/DataInspect/VariableView.xaml
int19h: Lambda, or you'd prefer a full fledged class? This is unlikely to be reused.

src/Package/Impl/DataInspect/VariableView.xaml.cs
int19h: Yeah, this is wrong. I'll add a dispatch.

src/Package/Impl/DataInspect/Viewers/CodeViewer.cs
int19h: It's an artifact of an earlier squashed change where I was using set intersects to match against classes. It's definitely better off with an array now, I'll change it.

src/Package/Test/DataInspect/REnvironmentProviderTest.cs
int19h: But they are testing a class from `Microsoft.VisualStudio.R.Package`. Do we want a reference to it from that test assembly?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
int19h: I'll do a code sweep for this. I added it late in the work, and probably didn't find all places where it needs to go - and there are also some pre-existing ones that will also be better off with an indirection.

src/Package/Impl/DataInspect/REnvironmentProvider.cs
AlexanderSher: Changing the content of collection (even if the whole collection is replaced) is less destructive than replacing the datasource. The latter forces ItemsControl to pretty much rebuild itself.

---
In reply to: [63223439](https://github.com/Microsoft/RTVS/pull/1770#discussion_r63223439) [](ancestors = 63223439)

src/Package/Test/DataInspect/REnvironmentProviderTest.cs
AlexanderSher: I see, they are for `REnvironmentProvider`. No, in this case, let's leave them where they are.

---
In reply to: [63224242](https://github.com/Microsoft/RTVS/pull/1770#discussion_r63224242) [](ancestors = 63224242)

src/Package/Impl/DataInspect/VariableView.xaml
AlexanderSher: If lambda can handle `switch` - sure

---
In reply to: [63223696](https://github.com/Microsoft/RTVS/pull/1770#discussion_r63223696) [](ancestors = 63223696)

int19h:
Addressed review feedback.

AlexanderSher:
:shipit:

Issue #984, opened 2016-02-24
Support echo=true for Source R Script command


int19h:
User-reported as follows:

> I would like to be able to change the RTools.SourceRScript command (that is currently invoked by Ctrl+Shift+S) a bit. Currently it submits the following to the REPL window: `source(“/Full/Path/To/Active/Script.R”)`, whereas I would love it to run `source(“/Full/Path/To/Active/Script.R”, echo = TRUE)` instead. 

Basically, we need to add a second command that is the same, but with echo.

huguesv:
I'm assuming we want an echo version of Source Selected File(s), which is a different command for project context menu?

int19h:
We do. The tricky part here is what to do when debugging - we need to use `rtvs::debug_source` in that case, which doesn't support echo currently.

And adding that support is non-trivial - the reason why regular `source` can do it easily is because it evaluates the file one line at a time, but that also breaks stepping (and is why we needed `debug_source` in the first place). The only way I can think of implementing that for `debug_source` is injecting print statements into the parsed expression.

Or we could say that source w/echo simply does not support stepping, and then always use `source`. @jflam?

huguesv:
Implemented the simple way ie. no stepping support.

PR #1774, opened 2016-05-13
source with echo

huguesv:
#984

For now, I've made it use source() instead of debug_source() in the case where we are debugging and user wants echo.


int19h:
:+1:

PR #1775, opened 2016-05-13
Use more detailed names for import data commands and dialog titles.

### LINKED ISSUE #1659: Command names for data import are confusing out of context
The names / tooltips of these commands are simply "Import from URL" and "Import from file". When they appear outside of VE / Data Grid context, it's unclear what they actually do, especially on the top-level toolbar (which is disjoint from the REPL), because it's not clear what is being imported, or where it will end up. First you hover over them, and see this:

image

Then you click on it, and get the equally vague dialog:

image

To make this less confusing, the names need to be changed to something lengthier that is sufficiently descriptive regardless of the context, and that specifically references R - e.g. "Import data into R session from URL" - and the dialog title should also be adjusted accordingly.

huguesv:
Fixes #1659


src/Package/Impl/DataInspect/Commands/ImportDataSetTextFileCommand.cs
```diff
@@ -17,7 +17,10 @@ internal sealed class ImportDataSetTextFileCommand : SessionCommand {
         protected override void Handle() {
             base.Handle();
 
-            string filePath = VsAppShell.Current.BrowseForFileOpen(IntPtr.Zero, Package.Resources.CsvFileFilter);
+            string filePath = VsAppShell.Current.BrowseForFileOpen(
```
AlexanderSher: Please pass `ICoreShell` through ctor

huguesv: None of the other commands that need VsAppShell have it passed in that I can see, so I guess I'll change all of them.

MikhailArkhipov: Some code does but if it can accept in ctor it is better for testability. Sometimes it cant though.

huguesv: It's actually IApplicationShell that it needs.

huguesv:
Commands that were using VsAppShell.Current are now getting IApplicationShell passed in instead.

int19h:
:+1:

PR #1777, opened 2016-05-14
Fix #1725: Support non-system encoding for Source R Script command

### LINKED ISSUE #1725: Support non-system encoding for Source R Script command
When editing a script of UTF-8 encoding, I can't run the script from Source R Script because my system encoding is CP932.

int19h:
Add support for encoding to SourceFile helper methods, and fix encoding handling in rtvs:::debug_source.

Use ITextDocument encoding when available in SourceRScript.

Enable SourceRScript command whenever there's an active REPL window, regardless of whether it is visible or not; and make it visible when command is invoked.

Add tests for SourceRScript.

### DIFF
--- a/src/Package/Impl/Repl/Commands/SourceRScriptCommand.cs
+++ b/src/Package/Impl/Repl/Commands/SourceRScriptCommand.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System;
+using Microsoft.Languages.Editor.EditorHelpers;
 using Microsoft.R.Components.ContentTypes;
 using Microsoft.R.Components.InteractiveWorkflow;
 using Microsoft.VisualStudio.R.Package.Commands;
@@ -32,8 +32,8 @@ namespace Microsoft.VisualStudio.R.Package.Repl.Commands {
         }
 
         protected override void SetStatus() {
-            Visible = _interactiveWorkflow.ActiveWindow != null && _interactiveWorkflow.ActiveWindow.Container.IsOnScreen;
-            Enabled = GetFilePath() != null;
+            Visible = _interactiveWorkflow.ActiveWindow != null;
+            Enabled = Visible && !string.IsNullOrEmpty(GetFilePath());
         }
 
         protected override void Handle() {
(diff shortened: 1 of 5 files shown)


src/Package/Impl/Repl/Commands/SourceRScriptCommand.cs
```diff
@@ -42,8 +42,14 @@ public SourceRScriptCommand(IRInteractiveWorkflow interactiveWorkflow, IActiveWp
...
                 if (textView != null) {
-                    textView.SaveFile();
-                    _interactiveWorkflow.Operations.SourceFile(filePath, _echo);
+                    if (RPackage.Current != null) {
+                        textView.SaveFile();
+                    }
+
+                    var encoding = textView.TextBuffer.GetTextDocument()?.Encoding;
+
+                    _interactiveWorkflow.ActiveWindow?.Container.Show(false);
```
AlexanderSher: There is no reason to call `SourceFile` if `ActiveWindow` became null between `SetStatus` and `Handle` calls

src/Package/Test/Repl/ReplCommandTest.cs
```diff
@@ -114,5 +117,39 @@ public class ReplCommandTest: IDisposable {
...
+
+
+        [CompositeTest]
+        [Category.Repl]
+        [InlineData(false, "utf-8")]
+        [InlineData(true, "Windows-1252")]
+        public void SourceRScriptTest(bool echo, string encoding) {
+            var textBuffer = new TextBufferMock("", RContentTypeDefinition.ContentType);
+            var textView = new WpfTextViewMock(textBuffer);
+            var tracker = Substitute.For<IActiveWpfTextViewTracker>();
```
AlexanderSher: Why you don't want to use real `RInteractiveWorkflow`?

src/Package/Impl/Repl/Commands/SourceRScriptCommand.cs
int19h: I'll make it a local and combine the check with `textView`.

src/Package/Test/Repl/ReplCommandTest.cs
int19h: How can I make it have an `ActiveWindow` without creating an actual GUI window?

int19h: Ah, I see, I can mock the factory for that when calling `GetOrCreateVisualComponent`

AlexanderSher:
👍

PR #1780, opened 2016-05-16
Switch to Nuget 3 and make RTVS compilable for both VS14 and VS15

AlexanderSher:
rtvs for vs15

We still can't produce signed builds because microbuild lab has no machines that have VS15 installed

Changes to `CONTRIBUTING.md` will be submitted as a separate PR.


MikhailArkhipov:
:shipit:

Issue #1672, opened 2016-04-22
Parameter alignment auto-formatting breaks Code Completion


jflam:
xyplot<TAB><TAB>

The nice reformatting of parameters puts all of the highlighted tab completions in the wrong areas:

![image](https://cloud.githubusercontent.com/assets/8877/14746775/12b6ee4c-0866-11e6-924b-a5b1bf5e7749.png)

MikhailArkhipov:
#1640. Multiline snippets are not particularly useful in REPL due to its stitched buffer nature. Also, We have no control over highlights, it is very old IDE code (circa 2003-2005). It is not known up front if particular snippet is going to be multiline or not.

MikhailArkhipov:
Need to look into if we can suppress some snippets in REPL

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0e1f37a27c39fe0e6694ac171b5e6eebe648bd1f

PR #1781, opened 2016-05-16
#1672 Parameter alignment auto-formatting breaks Code Completion

MikhailArkhipov:
Fix incorrect mapping when snipped is getting inserted as it needs to account for projections in REPL.

### DIFF
--- a/src/Package/Impl/Expansions/ExpansionClient.cs
+++ b/src/Package/Impl/Expansions/ExpansionClient.cs
@@ -4,6 +4,7 @@
 using System.Diagnostics;
 using Microsoft.Languages.Core.Text;
 using Microsoft.R.Editor;
+using Microsoft.R.Editor.Document;
 using Microsoft.R.Editor.Formatting;
 using Microsoft.R.Editor.Settings;
 using Microsoft.VisualStudio.R.Package.Utilities;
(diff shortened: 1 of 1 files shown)


src/Package/Impl/Expansions/ExpansionClient.cs
```diff
@@ -138,15 +139,25 @@ public sealed class ExpansionClient : IVsExpansionClient {
...
+                var expansion = document.TextBuffer.GetBufferAdapter<IVsExpansion>();
                 _earlyEndExpansionHappened = false;
+
                 Span span;
                 _shortcut = TextView.GetItemBeforeCaret(out span, x => true);
-
                 VsExpansion? exp = _cache.GetExpansion(_shortcut);
+
+                // Get view span
                 var ts = span.Length > 0 ? TextSpanFromSpan(TextView, span) : TextSpanFromPoint(caretPoint);
```
AlexanderSher: Do we care about virtual space here?

MikhailArkhipov: There is ```if (!TextView.Caret.InVirtualSpace) { ``` up above

int19h:
:+1:

PR #1784, opened 2016-05-17
Fix debug_parse to not lose source information for the parsed file.

int19h:


int19h:
This fixes the various debugger unit test failures.

PR #1785, opened 2016-05-17
Command refactoring

int19h:
Add a generic interface for commands that has no references to VS APIs, and a helper wrapper to produce PackageCommand, to improve testability.

Convert SourceRScript to the new interface + wrapper, and move it to the R.Components assembly.

Change SourceRScript test to use non-mocked IInteractiveWindow, and verify side effect of the code being executed, instead of the code itself.


src/Package/Impl/Commands/MenuCommandToOleMenuCommandShim.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.ComponentModel.Design;
+using Microsoft.R.Components.Controller;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.R.Package.Commands {
+    internal class MenuCommandToOleMenuCommandShim : PackageCommand {
+        private readonly IMenuCommand _command;
+
+        public MenuCommandToOleMenuCommandShim(IMenuCommand command, Guid group, int id)
```
AlexanderSher: I think it's better to have `group`/`id` before `command`.

src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs
```diff
@@ -0,0 +1,63 @@
...
+                if (_interactiveWorkflow.ActiveWindow == null) {
+                    status |= CommandStatus.Invisible;
+                } else if (!string.IsNullOrEmpty(GetFilePath())) {
+                    status |= CommandStatus.Enabled;
+                }
+                return status;
+            }
+        }
+
+        public CommandResult Invoke(object inputArg, ref object outputArg) {
```
AlexanderSher: Why `ref` instead of `out`?

src/R/Components/Test/Repl/ReplCommandTest.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.VisualStudio.Editor.Mocks;
+using Microsoft.VisualStudio.R.Package.Repl.Commands;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.R.Components.Test.Repl {
+    [ExcludeFromCodeCoverage]
+    [Collection(CollectionNames.NonParallel)]
```
AlexanderSher: You can remove this - `RComponentsMefCatalogFixture` creates unique MEF container for every test, so they can run in parallel

src/R/Components/Impl/Controller/IMenuCommand.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.R.Components.Controller {
+    /// <summary>
+    /// An object that implements a single command. 
+    /// </summary>
+    public interface IMenuCommand {
```
AlexanderSher: I would've call it `IUserCommand`

src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs
AlexanderSher: Also, it looks like `CommandResult` is never used, so we probably can just return value instead of using outputArg

---
In reply to: [63465651](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63465651) [](ancestors = 63465651)

int19h: It's `ref` because that's what `ICommandTarget` uses. I'm assuming this is originally to make the implementations shorter, since there's no reason why it would need to read from it.

`CommandResult` is indeed currently unused, but if we ever want to support command chaining for these, it may come in handy.

src/R/Components/Impl/Controller/IMenuCommand.cs
int19h: I'd call it `ICommand`, but then it clashes with the existing `ICommand` that is under `Microsoft.Languages.Editor.Controller`. Perhaps the latter should really be `IEditorCommand`? Its members - specifically, `NeedCheckout` - seem to imply that it's meant for that specifically.

MikhailArkhipov: NeedCheckout is generally obsolete since VS no longer requires manual SCC interactions. There is a task to reshuffle dependencies. Perhaps we could still make existing interface work? The difference is in missing ```Guid group, int id``` which in ICommand allow command implementation to register for multiple IDs. However, the command doesn't have to check for them as controller does the work.

MikhailArkhipov: Something like
```
_command.Invoke(Guid.Empty, 0, inArg, ref outArg);
```

src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs
MikhailArkhipov: And ICommandTarget is using ref b/c of IOleCommandTarget in/out last parameter
https://msdn.microsoft.com/en-us/library/windows/desktop/ms690300(v=vs.85).aspx

AlexanderSher: Then we can probably create `CommandResult<T>`. I want to have async version of this interface, and async methods can't have `out` parameters

---
In reply to: [63471471](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63471471) [](ancestors = 63471471)

MikhailArkhipov: We have to be careful with async here since call may go down OLE chain in VS depending on the particular command implementation.

AlexanderSher: Currently we do something like this (from `ShowHelpOnCurrentCommand`):
```{csharp}
            try {
                if (!_workflow.RSession.IsHostRunning) {
                    return;
                }

                // Fetch identifier under the cursor
                string item = GetItemUnderCaret();
                if (item == null || item.Length >= MaxHelpItemLength) {
                    return;
                }

                _workflow.RSession.ExecuteAsync($"rtvs:::show_help({item.ToRStringLiteral()})")
                    .SilenceException<RException>()
                    .SilenceException<MessageTransportException>()
                    .DoNotWait();
            } catch (Exception ex) {
                Debug.Assert(false, string.Format(CultureInfo.InvariantCulture, "Help on current item failed. Exception: {0}", ex.Message));
                // Catch everything so exceptions don't leave the async void method
                if (ex.IsCriticalException()) {
                    throw;
                }
            }
        }
```

With such implementation, there is no good way to wait until command is executed. We can call `DoNotWait` inside CommandShim and make `Handle` async`. This will not change implementation, but will help in test.

---
In reply to: [63559342](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63559342) [](ancestors = 63559342)

src/R/Components/Impl/Controller/IMenuCommand.cs
int19h: The nice thing about having a new interface that doesn't take IDs at all is that it is then not tied to the OLE/VS command system in any way - and so reusable in VSCode etc in theory (obviously there are dependencies on VS editor components etc that make it infeasible for many commands - but those that only use `IRSession` and/or `IRInteractiveWorkflow` should be readily portable).

src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs
int19h: Should I just make the interface async outright then?

src/R/Components/Impl/Controller/IMenuCommand.cs
int19h: My suggestion is to do this gradually, specifically:

- Rename existing interface to `IEditorCommand` or whatever is most appropriate for its actual use.
- Define a new `ICommand` interface with the desired properties (async etc), and start using it.
- Eventually, migrate all command implementations to that new interface, and retire the old one (or shim it).

src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs
AlexanderSher: We may want to have both sync and async versions. For the SourceRScriptCommand, it is probably good to have `InvokeAsync` that will call `SourceFileAsync` 

---
In reply to: [63588197](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63588197) [](ancestors = 63588197)

src/R/Components/Impl/Controller/IMenuCommand.cs
AlexanderSher: Agreed

---
In reply to: [63588679](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63588679) [](ancestors = 63588679)

src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs
int19h: Do we really gain anything from having a sync version if we have async? If the implementation of a particular command is synchronous, it can always just do what it needs to do directly, and then return CompletedTask.

---
In reply to: [63589496](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63589496) [](ancestors = 63589496,63588197)

AlexanderSher: > Do we really gain anything from having a sync version if we have async?
Understanding that this command is sync. Nothing else. 

---
In reply to: [63590758](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63590758) [](ancestors = 63590758,63589496,63588197)

MikhailArkhipov: I wouldn't spend too much time on renaming at the moment. There is bunch of thing to review in commands, controllers and associated OLE shims. Probably can be done as part of dependencies rework task

On May 17, 2016, at 12:51, Alexander Sher <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/R/Components/Impl/Repl/Commands/SourceRScriptCommand.cs<https://github.com/Microsoft/RTVS/pull/1785#discussion_r63591395>:

> +            _echo = echo;
> +        }
> +
> +        public CommandStatus Status {
> +            get {
> +                var status = CommandStatus.Supported;
> +                if (_interactiveWorkflow.ActiveWindow == null) {
> +                    status |= CommandStatus.Invisible;
> +                } else if (!string.IsNullOrEmpty(GetFilePath())) {
> +                    status |= CommandStatus.Enabled;
> +                }
> +                return status;
> +            }
> +        }
> +
> +        public CommandResult Invoke(object inputArg, ref object outputArg) {


Do we really gain anything from having a sync version if we have async?
Understanding that this command is sync. Nothing else.

________________________________

In reply to: 63590758<https://github.com/Microsoft/RTVS/pull/1785#discussion_r63590758> <ancestors%20=%2063590758,63589496,63588197>

-

int19h: Renaming is suggested because the existing interface occupies the "ICommand" name that would really be more appropriate for the new thing (and even though they're in different namespaces, a lot of code that uses the interface needs both of them, causing conflicts). The proposed approach is a simple mechanical renaming of existing ICommand to IEditorCommand (or something along these lines), so it would be a simple Refactor -> Rename, not requiring any particular effort.

---
In reply to: [63592100](https://github.com/Microsoft/RTVS/pull/1785#discussion_r63592100) [](ancestors = 63592100)

int19h:
Addressed review feedback.

AlexanderSher:
:+1:

PR #1788, opened 2016-05-17
Set options(width=) to match the size of interactive window

### LINKED ISSUE #1224: REPL should automatically update options('width') when it is resized
when width param is specified

huguesv:
The value is stored so that it can be set again if the session is reset.

Fixes #1224


int19h:
:+1:

src/R/Components/Impl/InteractiveWorkflow/TerminalWidthChangedEventArgs.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+
+namespace Microsoft.R.Components.InteractiveWorkflow {
+    public class TerminalWidthChangedEventArgs : EventArgs {
+        public TerminalWidthChangedEventArgs(int newWidth) {
+            NewWidth = newWidth;
+        }
+
+        public int NewWidth { get; private set; }
```
AlexanderSher: Just `public int NewWidth { get; }`

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -43,14 +43,21 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
...
                         RBasePath = _settings.RBasePath,
                         RHostCommandLineArguments = _settings.RCommandLineArguments,
                         CranMirrorName = _settings.CranMirror,
                         CodePage = _settings.RCodePage,
-                        WorkingDirectory = _settings.WorkingDirectory
-                    }, new RSessionCallback(CurrentWindow, Session, _settings, _coreShell));
+                        WorkingDirectory = _settings.WorkingDirectory,
+                    };
+
+                    IInteractiveWindowVisualComponent visualComponent;
```
AlexanderSher: `RInteractiveEvaluator` already has access to the `InteractiveWindow` through `CurrentWindow` property, so the whole `VisualElement_SizeChanged` can be implemented directly inside `RInteractiveEvaluator` itself. I know that this class pretends to be on the lower level than ViewModel, but `CurrentWindow` is defined by `IInteractiveEvaluator` itself, I don't think there is a reason to spread this logic between 3 classes.

AlexanderSher:
Will you write a test for it?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
huguesv: Sure, if we don't care about a bit of UI specific code there, that will simplify things quite a bit.

AlexanderSher: We already have UI specific code there. Everything that works with CurrentWindow is technically a UI specific code.

huguesv:
Moved the code to the evaluator.  I still need to fix RInteractiveEvaluatorTest, which is now failing because WpfTextViewMock returns null VisualElement.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -35,6 +37,8 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
         }
 
         public void Dispose() {
+            CurrentWindow.TextView.VisualElement.SizeChanged -= VisualElement_SizeChanged;
```
AlexanderSher: Let's check `CurrentWindow

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -207,5 +218,13 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
                 _coreShell.DispatchOnUIThread(() => CurrentWindow.WriteLine(message));
             }
         }
+
+        private void VisualElement_SizeChanged(object sender, System.Windows.SizeChangedEventArgs e) {
+            int width = (int)(e.NewSize.Width / CurrentWindow.TextView.FormattedLineSource.ColumnWidth);
+            // From R docs:  Valid values are 10...10000 with default normally 80.
+            _terminalWidth = Math.Max(10, Math.Min(10000, width));
+
+            Session.OptionsSetWidth(_terminalWidth);
```
AlexanderSher: `.DoNotWait()`

int19h: Can we rename the method to `...Async` too, so it's more obvious that it needs to be awaited or otherwise observed?

huguesv:
I'll rename all those evaluation extension methods to Async suffix.

huguesv:
I've added the Async suffix, and added SilenceException when setting the width on the session.  That should take care of the review comments.

AlexanderSher:
:+1:

PR #1789, opened 2016-05-17
Fix 1764: Package manager tests fail on R 3.4

AlexanderSher:
Some of the tests will still fail because of the external bug: https://bugs.r-project.org/bugzilla/show_bug.cgi?id=16914


huguesv:
👍

int19h:
:+1:

PR #1794, opened 2016-05-18
Fix conditional reference to Microsoft.UnitTests.References.14.0 / Microsoft.UnitTests.References.15.0 in test assemblies

AlexanderSher:


huguesv:
👍

PR #1798, opened 2016-05-20
Add simple helper class that allows to run background thread action on delay

AlexanderSher:


huguesv:
So looks like I would be using it this way?
```
private DelayedAsyncAction _resizePlotAction = new DelayedAction(ResizePlot, 250);
private int _lastWidth;
private int _lastHeight;
private int _lastResolution;

private async Task ResizePlot() {
   await _session.ResizePlot(_lastWidth, _lastHeight, _lastResolution);
}


private void OnControlResize(object sender, mouseeventargs args) {
// figure out width, height, res from event args
   _lastWidth = ...
   _lastHeight = ...
   _lastResolution = ...
   _resizePlotAction.Invoke();
}

```

Since the values for size will potentially change during the timeout period, if I was able to set the callback as part of `Invoke()` that would avoid having to cache the values.

```
private DelayedAsyncAction _resizePlotAction = new DelayedAction(250);

private void OnControlResize(object sender, mouseeventargs args) {
  // figure out width, height, resolution from event args
   _resizePlotAction.Invoke(async() => { await _session.ResizePlot(width, height, resolution); });
}

```

AlexanderSher:
There you are. And you don't need curly braces:
```
_resizePlotAction.Invoke(() => _session.ResizePlotAsync(width, height, resolution));
```

huguesv:
👍

PR #1807, opened 2016-05-21
Fix #1071: Up/down arrow in History doesn't update viewport

### LINKED ISSUE #1071: Up/down arrow in History doesn't update viewport
When history window is filled with entries such that it requires vertical scrolling to see them all, using up/down arrow keys to change the current selected entry will not scroll the window if selection goes outside of viewport.

E.g. size the window so that it can only fit 5 entries, and fill it with 10. Scroll all the way up, click on the first entry to highlight it, then press down arrow. Highlighting will move down with each press, and will eventually go off the screen at item 6, because the window doesn't scroll down.

AlexanderSher:
+ Fix Microsoft.R.Components.Test dependencies

### DIFF
--- a/src/R/Components/Impl/History/Implementation/RHistory.cs
+++ b/src/R/Components/Impl/History/Implementation/RHistory.cs
@@ -330,6 +330,7 @@ namespace Microsoft.R.Components.History.Implementation {
                 return;
             }
 
+            IRHistoryEntry entryToSelect;
             if (HasSelectedEntries) {
                 var lastSelectedEntry = _entries.GetSelectedEntries().Last();
                 if (lastSelectedEntry.Next == null) {
@@ -337,14 +338,14 @@ namespace Microsoft.R.Components.History.Implementation {
                 }
 
                 ClearHistoryEntrySelection();
-                lastSelectedEntry.Next.IsSelected = true;
+                entryToSelect = lastSelectedEntry.Next;
             } else {
                 if (VisualComponent == null) {
                     return;
                 }
 
                 if (VisualComponent.TextView.Selection.IsEmpty) {
-                    _entries.FirstOrDefault().IsSelected = true;
+                    entryToSelect = _entries.FirstOrDefault();
                 } else {
                     var entry = GetHistoryEntryFromPosition(VisualComponent.TextView.Selection.End.Position);
                     if (entry.Next == null) {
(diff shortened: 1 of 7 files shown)


src/R/Components/Impl/History/Implementation/RHistory.cs
```diff
@@ -330,32 +330,41 @@ internal sealed class RHistory : IRHistory {
...
+                    entryToSelect = _entries.FirstOrDefault();
                 } else {
                     var entry = GetHistoryEntryFromPosition(VisualComponent.TextView.Selection.End.Position);
                     if (entry.Next == null) {
                         return;
                     }
 
                     VisualComponent.TextView.Selection.Clear();
-                    entry.Next.IsSelected = true;
+                    entryToSelect = entry.Next.Next;
```
int19h: Can `Next.Next` be null?

src/R/Components/Impl/History/Implementation/RHistory.cs
```diff
@@ -364,32 +373,41 @@ internal sealed class RHistory : IRHistory {
...
                         return;
                     }
 
                     VisualComponent.TextView.Selection.Clear();
-                    entry.Previous.IsSelected = true;
+                    entryToSelect = entry.Previous;
                 }
             }
 
+            entryToSelect.IsSelected = true;
```
int19h: This assumes `entryToSelect` is never null, but it is set to `LastOrDefault()` in one of the branches above, implying that it could be null.

AlexanderSher: `_entries.Count` shouldn't be `0` because `HasEntries` is true. I'll replace it with `Last()`

src/R/Components/Impl/History/Implementation/RHistory.cs
AlexanderSher: That's an error - it should be just `entry.Next`

src/R/Components/Impl/History/Implementation/RHistory.cs
```diff
@@ -364,32 +373,41 @@ internal sealed class RHistory : IRHistory {
...
+
+            var snapshotPoint = entryToSelect.Span.GetStartPoint(_historyTextBuffer.CurrentSnapshot);
+            var line = VisualComponent.TextView.GetTextViewLineContainingBufferPosition(snapshotPoint);
+            if (line.VisibilityState != VisibilityState.FullyVisible) {
```
MikhailArkhipov: Can be made into a small function, use is almost same as above.

MikhailArkhipov:
:+1:

PR #1808, opened 2016-05-21
Fix #1806, #1796, #1805

### LINKED ISSUE #1796: Do not advertise false capabilities for the debug engine
When we register our debug engine, we're currently setting a bunch of capabilities flags that do not correspond to what is actually implemented. In particular, we incorrectly claim to support:

- Conditional breakpoints
- Set Next Statement
- Just My Code
- exceptions

These need to be declared unsupported until we actually implement them, to prevent VS from showing inapplicable UI (as in e.g. for #1779).

### LINKED ISSUE #1805: debug_source produces warning message
```
> rtvs::debug_source("//pytools/r/TestData/dbg.R", encoding = "Windows-1252")
Warning message:
In parse(text = text, encoding = encoding, srcfile = srcfile) :
  argument 'encoding = "Windows-1252"' will be ignored
> 
````

### LINKED ISSUE #1806: No error message for conditional and hit count breakpoints
This is partly related to #1796, but there's some more wiring that needs to be done for VS to provide coherent error messages when user is trying to do something unsupported with a breakpoint.

int19h:
Fix #1806: No error message for conditional and hit count breakpoints

Implement IDebugErrorBreakpoint2 and IDebugErrorBreakpointResolution2, and return them when binding a breakpoint of an unsupported kind (conditional, hit count, function etc).

Fix #1796: Do not advertise false capabilities for the debug engine

Expose all registry entries on the registration attribute, and default them all to false. Only set the supported features to true on the package.

Fix #1805: debug_source produces warning message

Do not specify encoding in call to parse() - it's ignored when parsing text rather than file.

### DIFF
--- a/src/Debugger/Impl/AD7ErrorBreakpoint.cs
+++ b/src/Debugger/Impl/AD7ErrorBreakpoint.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Debugger.Interop;
+
+namespace Microsoft.R.Debugger {
+    internal sealed class AD7ErrorBreakpoint : IDebugErrorBreakpoint2 {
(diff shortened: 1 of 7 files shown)


huguesv:
👍

PR #1811, opened 2016-05-23
Add icons to variables in Variable Explorer:

int19h:
- active bindings
- promises
- functions
- data frames
- matrices
- errors
- normal variables (everything else)

### DIFF
--- a/src/Package/Impl/DataInspect/VariableView.xaml
+++ b/src/Package/Impl/DataInspect/VariableView.xaml
@@ -29,48 +29,6 @@
 
             <local:MultiplyConverter x:Key="MultiplyConverter" />
 
-            <wpf:Lambda x:Key="REnvironmentKindConverter">
-                (REnvironmentKind kind) => {
-                    switch (kind) {
-                        case REnvironmentKind.Global:
-                            return KnownMonikers.GlobalVariable;
-                        case REnvironmentKind.Function:
-                            return KnownMonikers.MethodPublic;
-                        case REnvironmentKind.Package:
-                            return KnownMonikers.Package;
-                        case REnvironmentKind.Error:
-                            return KnownMonikers.StatusInvalid;
-                        default:
-                            return KnownMonikers.ObjectPublic;
-                    }
-                }
-            </wpf:Lambda>
-
-            <wpf:Lambda x:Key="VariableIconConverter">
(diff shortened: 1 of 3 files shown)


int19h:
Here are all the icons that are readily available:

http://glyphlist.azurewebsites.net/knownmonikers/

If you can think of anything else from there that we could use for some category of variables, let me know.

int19h:
image

int19h:
Alternatively, we could abandon standard VS mappings (which are kinda weird - like, why is method a transparent box?), and do something like this:

 Scalar values (1-element vectors, NULL etc)

 Vectors (except 1-element), lists

 Environments

 S4 objects (including ref class instances)

 Ref class definitions / constructors

 Functions

 Language objects (symbols and calls)

 Matrices

 Data frames

 Formulas

 Active bindings (e.g. ref class properties)

 Promises

smortaz:
+1

huguesv:
I like your alternative icons proposal.

int19h:
image

int19h:
One annoyance that I see now is that, because most of these icons are grey, they blend with the variable name, whereas the colored debugger icons did not. However, a uniform color could be added easily, so long as we can decide what it should be. Bikeshedding time!

MikhailArkhipov:
Looks crowded... I'd probably drop icons on ```[[x]]``` they are self-explanatory

int19h:
The icons are assigned based on the type/class. Note that it is possible for a list element to be an S4 object or an environment, for example (look at `envFoo@foo` to see one such case).

Maybe drop the icons for scalars?

int19h:
It seems that the distance between icon and name was very small, significantly contributing to the cluttered appearance. Here it is with the spacing increased 4x (and dark theme, but that doesn't really make a difference):

image

int19h:
Another option is to make the icons more subdued by blending them into the background. This is 50% alpha:

image

huguesv:
I don't like the 50% alpha very much.
Margin is 4 in the combo box, vs. 8 in the list.
Solution explorer looks like it may be using 6, should we use that?

huguesv:
Otherwise :+1: on the code changes.

MikhailArkhipov:
Actually faded looks better. Still not sure about icons on simple values but we'll see what users say.

:+1:

PR #1820, opened 2016-05-25
Sorting in data grid

MikhailArkhipov:


src/Package/Impl/DataInspect/VisualGrid/VisualGridScroller.cs
```diff
@@ -269,5 +244,11 @@ internal class VisualGridScroller {
                 Points.ViewportWidth = DataGrid.RenderSize.Width;
             }
         }
+        private void OnSortOrderChanged(object sender, EventArgs e) {
+            if(_owner?.Data == null || _owner.ColumnHeader == null) {
```
MikhailArkhipov: Should be doing _owner?.ColumnHeader here

src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
```diff
@@ -11,17 +11,29 @@
...
             string columns = gridRange?.Columns.ToRString();
 
             using (var evaluator = await rSession.BeginEvaluationAsync()) {
                 try {
-                    return await evaluator.EvaluateAsync<GridData>($"rtvs:::grid.data({expression}, {rows}, {columns})", REvaluationKind.Normal);
+                    string exp;
+                    if (sortOrder != null && !sortOrder.IsEmpty) {
+                        if (gridRange.Value.Columns.Count > 1) {
+                            string dataFrameSortExpression = sortOrder.GetDataFrameSortExpression();
+                            exp = Invariant($"rtvs:::grid.data({expression}, {rows}, {columns}, \"{dataFrameSortExpression}\")");
```
int19h: `ToRStringLiteral`

src/Package/Impl/DataInspect/VisualGrid/SortOrderType.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.VisualStudio.R.Package.DataInspect {
+    public enum SortOrderType {
+        None,
```
AlexanderSher: Do we use `None` anywhere?

src/Package/Impl/DataInspect/VisualGrid/SortOrder.cs
```diff
@@ -0,0 +1,82 @@
...
+                if (!first) {
+                    sb.Append(", ");
+                } else {
+                    first = false;
+                }
+                if (s.Descending) {
+                    sb.Append('-');
+                }
+                // NOTE: name must match name of the data frame in grid.r
+                sb.Append(Invariant($"x.df['{s.ColumnName}']"));
```
int19h: `ToRStringLiteral`

src/Package/Impl/DataInspect/VisualGrid/SortOrderType.cs
MikhailArkhipov: Yes. It is the default/initial value (no arrow on the column).

src/Package/Impl/DataInspect/VisualGrid/HeaderTextVisual.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Globalization;
+using System.Windows;
+using System.Windows.Media;
+
+namespace Microsoft.VisualStudio.R.Package.DataInspect {
+    /// <summary>
+    /// Text that appears in grid header. Tracks sorting order
+    /// and displays the appropriate sort order glyph.
+    /// </summary>
+    public sealed class HeaderTextVisual : TextVisual {
```
AlexanderSher: Can't find where it is instantiated

src/Host/Client/Impl/rtvs/R/grid.R
```diff
@@ -13,7 +13,28 @@ grid.format <- function(x) {
   sapply(format(x, trim = TRUE, justify = "none"), grid.trim);
 }
 
-grid.data <- function(x, rows, cols) {
+grid.sort.df <- function(x, rows, cols, df.sort.expression) {
+  if(missing(df.sort.expression)) {
+    x.df <- as.data.frame(x)[rows, cols];
+  } else {
+    x.df <- as.data.frame(x)
+    x.df <- x.df[eval(parse(text = df.sort.expression)),][rows, cols];
```
int19h: This feels wrong. The expression is basically required to use `x`, so now that argument name is set in stone. I would rather avoid eval here altogether, and instead make the expression argument a function, to which `x` is passed as an argument. So instead of passing it as `"do.call(order, c(x.df['col_name1'], -x.df['col_name2'], ...))"`, you would pass `function(x) do.call(order, c(x.df['col_name1'], -x.df['col_name2'], ...))`.

int19h: Also, side note: while this code already uses dots in identifiers, I would prefer to avoid it in any new code, and stick to either underscores or camelCase. Most R style guides recommend avoiding dots, to avoid clashing with S3 method names.

src/Package/Impl/DataInspect/VisualGrid/HeaderTextVisual.cs
```diff
@@ -0,0 +1,89 @@
...
+                return new Size(baseSize.Width + 2 * arrowGlyph.Width, baseSize.Height);
+            }
+            return new Size(baseSize.Width + arrowGlyph.Width, baseSize.Height);
+        }
+
+        private void SetArrowDisplay() {
+            string text = Text;
+            switch (SortOrder) {
+                case SortOrderType.None:
+                    SetArrow('\0');
```
AlexanderSher: Maybe non-breaking space (U+00A0)?

src/Package/Impl/DataInspect/VisualGrid/HeaderTextVisual.cs
MikhailArkhipov: CreateGrid()

var visual = new HeaderTextVisual();

src/Host/Client/Impl/rtvs/R/grid.R
MikhailArkhipov: Sure. Code used dots so I kinda kept the style, but I'll make them underscores.

AlexanderSher:
>     public class BooleanToVisibilityConverter : IValueConverter {

Do we use it anywhere? 

---
Refers to: src/Package/Impl/DataInspect/BooleanToVisibilityConverter.cs:10 in 166172c. [](commit_id = 166172c56491dec97baa968ae84f2746791d691e, deletion_comment = False)

MikhailArkhipov:
Doesnt seem like it is used. Delete?

AlexanderSher:
Yep

---
In reply to: [222023684](https://github.com/Microsoft/RTVS/pull/1820#issuecomment-222023684) [](ancestors = 222023684)

AlexanderSher:
:shipit:

PR #1821, opened 2016-05-26
Support showing WPF windows in unit tests

AlexanderSher:
`ContainerHostMethodFixture` can be referenced in test class constructor. For such test, before constructor is called, `ContainerHostMethodFixture` will create WPF window.
`ContainerHostMethodFixture.AddToHost` adds `UIElement` to the window. Up to 9 elements can be added at the same time.
Several tests that run concurrently will share the same window, so parallel execution may be limited by the maximum amount of allowed elements.

To wait until window is updated, use `UIThreadTools.DoEvents` or `UIThreadHelper.Instance.DoEventsAsync`


src/Host/Client/Test/DataInspection/ValuesTest.cs
```diff
@@ -18,6 +18,7 @@
 using Microsoft.R.Host.Client.Test.Script;
 using Microsoft.R.StackTracing;
 using Microsoft.UnitTests.Core.XUnit;
+using Microsoft.UnitTests.Core.XUnit.MethodFixtures;
```
MikhailArkhipov: I am wondering if we should move away with strict naming by folders and simplify our namespaces a bit. Just a thought.

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
```diff
@@ -116,6 +117,12 @@ public class UIThreadHelper {
             return args.Exception;
         }
 
+        public Task DoEventsAsync() {
+            return TaskUtilities.IsOnBackgroundThread()
+                ? _application.Dispatcher.InvokeAsync(() => { }, DispatcherPriority.Background).Task
+                : Task.Run(DoEventsAsync);
```
MikhailArkhipov: Is it reentering?

src/UnitTests/Core/Impl/UI/ContainerHost.cs
```diff
@@ -16,17 +17,34 @@ namespace Microsoft.UnitTests.Core.UI {
     internal class ContainerHost : ContentControl {
         private const int MaxContainerCount = 9;
 
+        private static int _count;
         private static ContainerHost _current;
         private static readonly SemaphoreSlim AvailableSpots = new SemaphoreSlim(MaxContainerCount, MaxContainerCount);
 
+        public static async Task Increment() {
```
MikhailArkhipov: AddRef() ?

MikhailArkhipov:
:+1:

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
AlexanderSher: If method is called from background thread, it schedules a task in `Dispatcher`. If not, it calls itself from background thread.

Issue #1819, opened 2016-05-24
Editor intellisense should always complete on Enter


MikhailArkhipov:
Currently setting controls both REPL and editor and although it does make sense to suppress Enter in REPL since it is execute, it doesn't make much sense in the editor.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/4c5229b7cc2a65de7eacf1dc77f3c27fc9f4f9e8

Issue #1803, opened 2016-05-20
Crash after resetting REPL


huguesv:
This was after using RTVS for a while, including using data grid, variable explorer, debugger.  Reset REPL is the last thing I did before it crashed.
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentOutOfRangeException
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(System.ExceptionArgument, System.ExceptionResource)
   at System.Collections.Generic.List`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].get_Item(Int32)
   at Microsoft.VisualStudio.R.Package.DataInspect.VariableGridHost.SetEvaluation(Microsoft.VisualStudio.R.Package.DataInspect.VariableViewModel)
   at Microsoft.VisualStudio.R.Package.DataInspect.VariableGridHost+<>c__DisplayClass5_0.<EvaluateAsync>b__0()
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(System.Delegate, System.Object, Int32)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(System.Object, System.Delegate, System.Object, Int32, System.Delegate)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Windows.Threading.DispatcherOperation.Invoke()
   at System.Windows.Threading.Dispatcher.ProcessQueue()
   at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr, Int32, IntPtr, IntPtr, Boolean ByRef)
   at MS.Win32.HwndWrapper.WndProc(IntPtr, Int32, IntPtr, IntPtr, Boolean ByRef)
   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(System.Object)
[...]

Issue #1801, opened 2016-05-20
Add R Documentation file template


MikhailArkhipov:
to File | New and to Add New Item. We've been supporting RD for quite some time, and RS has the template.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/4c5229b7cc2a65de7eacf1dc77f3c27fc9f4f9e8

Issue #1610, opened 2016-04-18
Support code folding and sections


AMarthallerMS:
RStudio supports translating certain comments into code folding sections.
https://support.rstudio.com/hc/en-us/articles/200484568-Code-Folding-and-Sections

It would be great if the VS IDE also understood these hints.

MikhailArkhipov:
Similar to C# ```#region```

fabriciovasselai:
That is one of the features I have been advocating the most since the first non-public release. It helps me a lot to organize my code when using VS for C# or even for C++ even it's a little diffrent there. It would be great to have such a feature for RTools too.

I would like to call the attention to a nice behavior of `#regions` in C#: one can even comment the region section, which ends up working as a "region title" for when code is folded. Would be great to have that as well.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/4c5229b7cc2a65de7eacf1dc77f3c27fc9f4f9e8

PR #1823, opened 2016-05-27
#1122 #1610 #1801 #1803 #1819

MikhailArkhipov:
#1122 Intellisense for variables in packages
#1610 Support code folding and sections 
#1801 Add R Documentation file template 
#1803 Crash after resetting REPL 
#1819 Editor intellisense should always complete on Enter


MikhailArkhipov:
Ping... I have more PRs pending :-)

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -92,7 +96,7 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
                             break;
                         }
 
-                        var children = eval.GetChildrenAsync().WaitAndUnwrapExceptions();   // TODO: discuss wait is fine here. If not, how to fix?
+                        var children = eval.GetChildrenAsync().WaitTimeout(_maxWaitTime);   // TODO: discuss wait is fine here. If not, how to fix?
```
int19h: Is this invoked from UI thread?

MikhailArkhipov: Yes, all of them

src/Common/Core/Impl/Extensions/TaskExtensions.cs
```diff
@@ -117,6 +117,16 @@ public static class TaskExtensions {
         }
 
         /// <summary>
+        /// Waits for a task to complete within a given timeout. 
+        /// Returns task result or default(T) if task did not complete
+        /// within the timeout specified.
+        /// </summary>
+        public static T WaitTimeout<T>(this Task<T> task, int msTimeout) {
+            task.Wait(msTimeout);
+            return task.IsCompleted ? task.Result : default(T);
```
int19h: `Result` throws `AggregateException` if there was an exception. It would be preferable to unwrap it here; or better yet, return true/false, and let the caller obtain the result as they see fit.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -111,13 +115,42 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
                                     .ToArray();
                         }
                     }
+                } else {
+                    // May be a package object line mtcars$
+                    variableName = TrimTrailingSelector(variableName);
+                    var sessionProvider = EditorShell.Current.ExportProvider.GetExportedValue<IRSessionProvider>();
+                    var session = sessionProvider.GetOrCreate(GuidList.InteractiveWindowRSessionGuid);
+                    var infoList = session.DescribeChildrenAsync("NULL", variableName, REvaluationResultProperties.HasChildrenProperty,
```
int19h: Use `REnvironments.GlobalEnv` (or whatever is most appropriate from that class) instead of "NULL". I'm actually kinda surprised this even works - looking at the source code of `describe_children` in `eval.R`, if it does, it's purely by accident.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -111,13 +115,42 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
                                     .ToArray();
                         }
                     }
+                } else {
```
int19h: Given this branch, is the one above still needed? It looks like evaluating the head of the expression will always give the right answer, whether the expression involves a package or not.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -175,6 +208,15 @@ class VariableInfo : INamedItemInfo {
                 }
             }
 
+            public VariableInfo(string name, string typeName) {
+                this.Name = name;
+                if (typeName == "closure") {
```
int19h: Or "builtin" (e.g. `c`).

src/Common/Core/Impl/Extensions/TaskExtensions.cs
MikhailArkhipov: OK. If it timed out, there is no result though

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: Old JP's code, fixed.

src/Common/Core/Impl/Extensions/TaskExtensions.cs
int19h: Right. So make this return false if it timed out or true if it didn't, and then the caller can directly access `Result` in a way that unwraps exceptions (e.g. via `GetAwaiter().GetResult()`), or fall back to some default value that is not actually `default(T)`.

Alternatively, make it throw `TimeoutException` if it times out, and the caller has to catch/handle.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: OK, let me check that

Issue #1825, opened 2016-05-27
Variable Explorer visual tests are unstable


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/8c19fd0b3b6ef14da18289139b3f8806eeefb28c

PR #1827, opened 2016-06-01
#1825 + CR Feedback

MikhailArkhipov:
#1825 Variable Explorer visual tests are unstable 

- Skip properties which are different on different machines such as touch/track
- Enumerate WPF properties from type and not from actual instance so dynamic properties specific to the environment won't affect baselines
- Make some properties into DependencyProperty so they actually DO appear in baselines


src/Common/Core/Impl/Extensions/TaskExtensions.cs
```diff
@@ -122,8 +122,16 @@ public static class TaskExtensions {
         /// within the timeout specified.
         /// </summary>
         public static T WaitTimeout<T>(this Task<T> task, int msTimeout) {
-            task.Wait(msTimeout);
-            return task.IsCompleted ? task.Result : default(T);
+            bool result = false;
+            try {
+                task.Wait(msTimeout);
+                result = task.IsCompleted;
+            } catch(AggregateException) { }
```
int19h: `Wait` does not throw anything. `Result` will throw if task completed with an exception.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -84,54 +84,25 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
...
+                var session = sessionProvider.GetOrCreate(GuidList.InteractiveWindowRSessionGuid);
+                IReadOnlyList<IREvaluationResultInfo> infoList = null;
+                try {
+                    infoList = session.DescribeChildrenAsync(REnvironments.GlobalEnv, variableName, REvaluationResultProperties.HasChildrenProperty,
+                            RValueRepresentations.Str(), _maxResults).WaitTimeout(_maxWaitTime);
+                } catch(TimeoutException) { }
+
+                if (infoList != null) {
+                    return infoList
+                                .Where(m => m.Name.StartsWithOrdinal("$") || m.Name.StartsWithOrdinal("@"))
```
int19h: This should probably only show one or the other, depending on whether the user has typed `$` or `@` to get here. Also, take a look at `IRValueInfo.AccessorKind`.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -84,54 +84,25 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
...
-                                    .ToArray();
-                    }
+                // May be a package object line mtcars$
+                variableName = TrimTrailingSelector(variableName);
+                var sessionProvider = EditorShell.Current.ExportProvider.GetExportedValue<IRSessionProvider>();
+                var session = sessionProvider.GetOrCreate(GuidList.InteractiveWindowRSessionGuid);
+                IReadOnlyList<IREvaluationResultInfo> infoList = null;
+                try {
+                    infoList = session.DescribeChildrenAsync(REnvironments.GlobalEnv, variableName, REvaluationResultProperties.HasChildrenProperty,
+                            RValueRepresentations.Str(), _maxResults).WaitTimeout(_maxWaitTime);
```
int19h: While you're at it, can you specify a limit for `Str()` here? The default is unbounded, so you may end up with multi-megabyte strings and such. Also, if we don't actually need the representation, that argument can be `null` to skip it entirely (maybe the caller of this method should specify?).

src/Common/Core/Impl/Extensions/TaskExtensions.cs
MikhailArkhipov: https://msdn.microsoft.com/en-us/library/dd270644(v=vs.110).aspx says it does if task was canceled

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: It may be Ctrl+J rather than typing (i.e. forced completion invoke)

src/Common/Core/Impl/Extensions/TaskExtensions.cs
int19h: Ah, I see. My earlier remarks about this method were rather dealing with the case where the task itself throws - then you only find out in `Result` (and it'll be an `AggregateException`).

The old code was calling `WaitAndUnwrapExceptions`, which unwraps AE. The new one needs to do so as well to avoid silent changes and possibly breakage (if nothing ends up catching it).

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
int19h: But there would still be either `$` or `@` left to the cursor for completion to appear, right? Or do we enable Ctrl+J even after `foo`, and allow you to pick either `$bar` or `@bar`, and put it there complete with the selector?

MikhailArkhipov: If I specify REvaluationResultProperties.AccessorKindProperty where is it returned?

src/Common/Core/Impl/Extensions/TaskExtensions.cs
MikhailArkhipov: So when task completes then ```return task.WaitAndUnwrapExceptions()``` will do that. Now, if ```Wait()``` does throw aggregate, do we want to rethrow? In most cases I think code that calls timeout wait typically only cares if result is valid or a timeout happened.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: ```AccessorKind``` is not exposed in ```IREvaluationResultInfo``` but exposed in ```IRValueInfo``` so gonna cast .

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -91,13 +90,17 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
                 var session = sessionProvider.GetOrCreate(GuidList.InteractiveWindowRSessionGuid);
                 IReadOnlyList<IREvaluationResultInfo> infoList = null;
                 try {
-                    infoList = session.DescribeChildrenAsync(REnvironments.GlobalEnv, variableName, REvaluationResultProperties.HasChildrenProperty,
-                            RValueRepresentations.Str(), _maxResults).WaitTimeout(_maxWaitTime);
+                    infoList = session.DescribeChildrenAsync(REnvironments.GlobalEnv, 
+                               variableName, 
+                               REvaluationResultProperties.HasChildrenProperty | REvaluationResultProperties.AccessorKindProperty,
+                            RValueRepresentations.Str(null), _maxResults).WaitTimeout(_maxWaitTime);
```
int19h: I meant passing `null` instead of `Str` (if we don't need the representation; I'm not sure if that is the case here or not). Otherwise, pass some reasonable value as an argument to `Str` to restrict the maximum length of representation.

int19h:
:+1:

PR #1829, opened 2016-06-01
Microsoft R Client install

MikhailArkhipov:
- Add Menu items for R Client install and switch
- Prompt and install R Client if R is missing
- Remove MRO setup custom action as we offer R Client on start
- Remove R.Actions (not much left there and setup action is gone)
  -- Move R Install code to Core
  -- Move remaining code to Components
- New R Installation tests related to R Client
- Registry mocks change to allow writing values


src/Common/Core/Impl/Install/ISupportedRVersionList.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// Maximal supported R version minor part such as 5 in 4.5
+        /// </summary>
+        int MaxMinorVersion { get; }
+
+        /// <summary>
+        /// Determines if given version falls into the supported range.
+        /// </summary>
+        /// <param name="v"></param>
+        /// <returns></returns>
+        bool IsCompatibleVersion(Version v);
```
int19h: This looks like it should be an extension method on this interface, since it's always going to be defined in the same way, in terms of comparing all the other properties here.

src/Common/Core/Impl/Install/ISupportedRVersionList.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.Common.Core.Install {
+    public interface ISupportedRVersionList {
```
int19h: This is closer to "supported version range"; "list" implies it's a collection, or at least enumerable.

src/Common/Core/Impl/Install/ISupportedRVersionList.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Minimal supported R version major part such as 3 in 3.2
+        /// </summary>
+        int MinMajorVersion { get; }
+        /// <summary>
+        /// Minimal supported R version minor part such as 2 in 3.2
+        /// </summary>
+        int MinMinorVersion { get; }
```
int19h: This is assuming that the third number in the version never matters (i.e. you can't specify at least 3.2.1, for example)  - are we sure that is the case?

src/Common/Core/Impl/Install/RInstallation.cs
```diff
@@ -123,33 +144,96 @@ public static class RInstallation {
             return path;
         }
 
-        public static void GoToRInstallPage() {
-            Process.Start("https://mran.revolutionanalytics.com/download");
+        public static Exception LaunchRClientSetup() {
```
int19h: Why return exceptions, instead of just throwing them?

src/Common/Core/Impl/Install/RInstallation.cs
```diff
@@ -123,33 +144,96 @@ public static class RInstallation {
...
+            var rClientExe = Path.Combine(Path.GetTempPath(), "RClientSetup.exe");
+            try {
+                using (var client = new WebClient()) {
+                    client.DownloadFileAsync(
+                        new Uri("http://go.microsoft.com/fwlink/?LinkId=800048", UriKind.Absolute), rClientExe);
+
+                    var startTime = DateTime.Now;
+                    while (client.IsBusy && (DateTime.Now - startTime).TotalMilliseconds < 5000) {
```
int19h: Since it's called from UI thread when invoking command, this will basically block all UI for as long as it takes to download. I'd rather have this method return a task, and have `DownloadFileCompleted` event mark it as complete.

src/Common/Core/Impl/Install/RInstallation.cs
```diff
@@ -123,33 +144,96 @@ public static class RInstallation {
...
+                    if (!FileSystem.FileExists(rClientExe)) {
+                        return new TimeoutException(Resources.Error_UnableToDownloadRClient_Timeout);
+                    }
+                }
+            } catch (WebException ex) {
+                return ex;
+            }
+
+            ProcessServices.Start(rClientExe);
+            NativeMethods.PostQuitMessage(0);
```
int19h: Won't this close VS when called from `InstallRClientCommand`?

src/Common/Core/Impl/Install/RInstallation.cs
```diff
@@ -123,33 +144,96 @@ public static class RInstallation {
...
+                using (var hkcu = Registry.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Registry64)) {
+                    using (var key = hkcu.OpenSubKey(rtvsKey + "\\" + Toolset.Version, writable: true)) {
+                        object value = key.GetValue("RClientPrompt");
+                        if (value == null) {
+                            key.SetValue("RClientPrompt", 0);
+                            return true;
+                        }
+                    }
+                }
+            } catch (Exception) { }
```
int19h: `when (!IsCriticalException)`

src/Common/Core/Impl/Install/RInstallation.cs
```diff
@@ -322,10 +407,10 @@ public static class RInstallation {
                 foreach (IFileSystemInfo fsi in directories) {
                     string subFolderName = fsi.FullName.Substring(baseRFolder.Length + 1);
                     Version v = GetRVersionFromFolderName(subFolderName);
-                    if (v.Major >= minMajorVersion &&
-                        v.Minor >= minMinorVersion &&
-                        v.Major <= maxMajorVersion &&
-                        v.Minor <= maxMinorVersion) {
+                    if (v.Major >= supportedVersions.MinMajorVersion &&
```
int19h: This check is incorrect. Consider min version as 3.2, max version as 5.0. Then 4.1 will not match, because 1 is not greater than 2, and is not less than 0.

I think we should just use `Version` everywhere, then we can use its own comparison operators.

src/Common/Core/Impl/Install/SupportedRVersionList.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Microsoft.Common.Core.Install {
+    public sealed class SupportedRVersionList : ISupportedRVersionList {
+        // TODO: this probably needs configuration file
+        // or another dynamic source of supported versions.
+        public int MinMajorVersion { get; }
+        public int MinMinorVersion { get; }
+        public int MaxMajorVersion { get; }
+        public int MaxMinorVersion { get; }
+
+        public SupportedRVersionList() : this(3, 2, 3, 9) { }
```
int19h: This should probably be a `public static readonly` instance here or elsewhere.

src/Common/Core/Impl/Install/RInstallation.cs
int19h: Also, 5s is not going to be enough to download a ~9Mb file in many cases.

MikhailArkhipov: For simplicity I wanted to block here so user won't start doing assorted things in VS. However, I can instead make it completely async - probably can reuse long operation class from our log collection thing. Let me try that.

src/Common/Core/Impl/Install/ISupportedRVersionList.cs
MikhailArkhipov: Makes sense

src/Common/Core/Impl/Install/ISupportedRVersionList.cs
MikhailArkhipov: So far it didn't matter. Going forward I don't think we'll be able to block versions, we'll have to figure out how to support then like RS does.

src/Common/Core/Impl/Install/RInstallation.cs
int19h: Preventing them from using the UI might be prudent, but then you should use `RunSynchronouslyOnUIThread` - this will spin up the threaded wait dialog, so that OS doesn't treat VS as unresponsive.

src/Common/Core/Impl/Install/ISupportedRVersionList.cs
MikhailArkhipov: OK

src/Common/Core/Impl/Install/RInstallation.cs
MikhailArkhipov: I think I can return ex.Message instead. Caller only show error messages,

src/Common/Core/Impl/Install/RInstallation.cs
MikhailArkhipov: Yep, that's about what we do in the feedback collection

src/Common/Core/Impl/Install/RInstallation.cs
int19h: If it's solely to display the error, and all callers do that, then perhaps this should be done right here? Then there's nothing to return.

src/Common/Core/Impl/Install/RInstallation.cs
MikhailArkhipov: Ah, the problem was that all that VS progress dialogs and long operations with message pumps are over in the package. I'll see what I can do to plumb them through.

int19h: `RunSynchronouslyOnUIThread` currently lives in `Microsoft.R.Debugger`, which does not reference the package project, but it does reference VS directly. We should probably flow it through `ICoreShell`, since it already has things like `DispatchOnUIThread`.

src/Common/Core/Impl/Install/RInstallation.cs
MikhailArkhipov: I re-did this via MEF export and moved download etc to the package

src/Package/Impl/Logging/DiagnosticLogs.cs
```diff
@@ -74,7 +75,7 @@ internal static class DiagnosticLogs {
             return zipPath;
         }
 
-        private static void CollectRTVSLogs(object o) {
+        private static void CollectRTVSLogs(object o, CancellationToken ct) {
```
int19h: Why do all these need it? I mean, they aren't even async.

src/Package/Impl/RClient/MicrosoftRClientInstaller.cs
```diff
@@ -0,0 +1,54 @@
...
+            try {
+                using (var client = new WebClient()) {
+                    client.DownloadFileAsync(
+                        new Uri("http://go.microsoft.com/fwlink/?LinkId=800048", UriKind.Absolute), rClientExe);
+                    while (client.IsBusy && !ct.IsCancellationRequested) {
+                        Thread.Sleep(200);
+                    }
+                }
+            } catch (WebException ex) {
+                _errorMessage = ex.Message;
```
int19h: Need to reset this field before calling, in case this class is going to be invoked more than once.

src/Package/Impl/Logging/DiagnosticLogs.cs
MikhailArkhipov: Actions may or may not be async but token is still useful. LongOperationNotification (as it is derived from the way VS does long cancellable ops) only checks for the cancellation between actions. So if there is only one action or one of the actions is long, they should be checking for cancellation periodically and exit when appropriate.

int19h:
:+1:

PR #1830, opened 2016-06-02
Fixes #1815 Adds an option to evaluate active bindings to show in Var…

### LINKED ISSUE #1815: Active bindings should be evaluated by default
Active bindings in R are basically the equivalent of the property getter in C#, and are used in that manner for e.g. S4 classes. Right now we avoid actually executing them, and just show `<active binding>` in VE value column and in Locals. But C# evaluates property getters by default, and we should follow suit - in most cases, the code that has to run to produce a value is side-effect-free. Might be worth adding an option, just in case.

karthiknadig:
Adds an option to evaluate active bindings to show in Variable Explorer and debugger tool windows.

### DIFF
--- a/src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
+++ b/src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
@@ -7,8 +7,8 @@ namespace Microsoft.R.DataInspection {
     /// </summary>
     public interface IRActiveBindingInfo : IREvaluationResultInfo {
         /// <summary>
-        /// This is null if evaluate active bindings in RTools Options is set 
-        /// to False.
+        /// This property is set to an instance of <see cref="RValueInfo"/> if <see cref="REvaluationResultProperties.ComputedValueProperty"/> 
+        /// is used while evaluating active bindings. This is null otherwise.
         /// </summary>
         IRValueInfo ComputedValue { get; }
     }
--- a/src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
+++ b/src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
@@ -6,5 +6,9 @@ namespace Microsoft.R.DataInspection {
     /// Describes the result of evaluating an expression that references an active binding. 
     /// </summary>
     public interface IRActiveBindingInfo : IREvaluationResultInfo {
+        /// <summary>
+        /// Value of the active binding after evaluation.
+        /// </summary>
+        IRValueInfo Value { get; }
     }
 }
--- a/src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
+++ b/src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
@@ -7,8 +7,9 @@ namespace Microsoft.R.DataInspection {
     /// </summary>
     public interface IRActiveBindingInfo : IREvaluationResultInfo {
         /// <summary>
-        /// Value of the active binding after evaluation.
+        /// This is null if evaluate active bindings in RTools Options is set 
+        /// to False.
         /// </summary>
-        IRValueInfo Value { get; }
+        IRValueInfo ComputedValue { get; }
     }
 }
--- a/src/Host/Client/Impl/DataInspection/IREvaluationResultInfo.cs
+++ b/src/Host/Client/Impl/DataInspection/IREvaluationResultInfo.cs
@@ -58,11 +58,10 @@ namespace Microsoft.R.DataInspection {
             this IREvaluationResultInfo info,
             REvaluationResultProperties properties,
             string repr,
-            bool evaluateActiveBindings,
             int? maxCount = null,
             CancellationToken cancellationToken = default(CancellationToken)
         ) =>
-            info.Session.DescribeChildrenAsync(info.EnvironmentExpression, info.Expression, properties, repr, evaluateActiveBindings, maxCount, cancellationToken);
+            info.Session.DescribeChildrenAsync(info.EnvironmentExpression, info.Expression, properties, repr, maxCount, cancellationToken);
 
         /// <summary>
         /// If this evaluation result corresponds to an expression that is a valid assignment target (i.e. valid on the
--- a/src/Host/Client/Impl/DataInspection/IREvaluationResultInfo.cs
+++ b/src/Host/Client/Impl/DataInspection/IREvaluationResultInfo.cs
@@ -58,10 +58,11 @@ namespace Microsoft.R.DataInspection {
             this IREvaluationResultInfo info,
             REvaluationResultProperties properties,
             string repr,
+            bool evaluateActiveBindings,
             int? maxCount = null,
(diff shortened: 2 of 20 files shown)


src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
```diff
@@ -6,5 +6,9 @@ namespace Microsoft.R.DataInspection {
     /// Describes the result of evaluating an expression that references an active binding. 
     /// </summary>
     public interface IRActiveBindingInfo : IREvaluationResultInfo {
+        /// <summary>
+        /// Value of the active binding after evaluation.
```
int19h: Can you mention here that this is `null` unless the flag was specified?

src/Host/Client/Impl/DataInspection/IREvaluationResultInfo.cs
```diff
@@ -58,10 +58,11 @@ public static class REvaluationResultInfoExtensions {
             this IREvaluationResultInfo info,
             REvaluationResultProperties properties,
             string repr,
+            bool evaluateActiveBindings,
```
int19h: I'd make it a new flag in `REvaluationResultProperties` instead, since it's effectively about whether the `Value` property of the result is filled or not.

src/Host/Client/Impl/rtvs/R/eval.R
```diff
@@ -201,7 +203,12 @@ describe_children <- function(obj, env, fields, count = NULL, repr = NULL) {
         value <- list(promise = deparse_str(code), expression = item_expr);
       } else if (bindingIsActive(name, obj)) {
         # It's an active binding - we don't want to read it to avoid inadvertently changing program state.
-        value <- list(active_binding = TRUE, expression = item_expr);
+        binding_res <- NULL
```
int19h: Nit: you can do `binding_res <- if (...) ... else NULL`.

src/Package/Impl/Options/R/RToolsOptionsPage.cs
```diff
@@ -116,6 +116,14 @@ public class RToolsOptionsPage : DialogPage {
             set { RToolsSettings.Current.RCodePage = value; }
         }
 
+        [LocCategory("Settings_DebuggingCategory")]
+        [CustomLocDisplayName("Settings_EvaluateActiveBindings")]
+        [LocDescription("Settings_EvaluateActiveBindings_Description")]
+        [DefaultValue(false)]
```
int19h: Let's set it to true by default, like C# does for properties.

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -105,7 +105,8 @@ internal sealed class RToolsSettingsImplementation : IRToolsSettings {
         public DateTime SurveyNewsLastCheck { get; set; }
         public string SurveyNewsFeedUrl { get; set; } = SurveyNewsUrls.Feed;
         public string SurveyNewsIndexUrl { get; set; } = SurveyNewsUrls.Index;
- 
+        public bool EvaluateActiveBindings { get; set; }
```
int19h: (and here as well)

src/Package/Impl/Resources.resx
```diff
@@ -764,7 +764,10 @@ Additional information:
   <data name="HelpSearchWatermark" xml:space="preserve">
     <value>Type search term</value>
   </data>
-  <data name="VariableExplorer_ErrorEnvironment" xml:space="preserve">
-    <value>Error retrieving environment list</value>
+  <data name="Settings_EvaluateActiveBindings" xml:space="preserve">
+    <value>Evalaute active bindings.</value>
```
int19h: Spelling. Also, since this is the name of the property in the property grid, it shouldn't have the trailing dot.

src/Package/Impl/Resources.resx
```diff
@@ -764,7 +764,10 @@ Additional information:
...
+  </data>
+  <data name="Settings_EvaluateActiveBindings_Description" xml:space="preserve">
+    <value>Show result of active binding evaluation in Variable Explorer and debugger tool windows. </value>
```
int19h: This should make it very clear that this involves running code, potentially with side effects. Something like "Automatically evaluate active bindings and display the resulting value. Warning: when this option is on, inspecting active bindings in Variable Explorer and debugger tool windows may change program state."

src/R/Editor/Impl/Data/RSessionDataObject.cs
```diff
@@ -63,12 +64,33 @@ public class RSessionDataObject : IRSessionDataObject {
...
             } else if (DebugEvaluation is IRActiveBindingInfo) {
-                const string ActiveBindingValue = "<active binding>";
+                const string ActiveBindingText = "<active binding>";
                 var activeBinding = (IRActiveBindingInfo)DebugEvaluation;
+                var activeBindingValue = activeBinding.Value;
 
-                Value = ActiveBindingValue;
-                TypeName = ActiveBindingValue;
-                Class = ActiveBindingValue;
+                if(activeBindingValue != null) {
```
int19h: The code inside mostly duplicates `IRValueInfo` branch above - this should probably be refactored into a reusable method. Or maybe reorder the branches so that we do checks for promises and active bindings first, and then check the value, so that the first two branches can tweak the input for the third. E.g.:
```cs
IRValueInfo valueInfo = DebugEvaluation as IRValueInfo;

if (DebugEvaluation is IRPromiseInfo) {
  ...
} else if (DebugEvaluation is IRActiveBindingInfo) {
  ...
  valueInfo = activeBinding.Value;
  ...
}

if (valueInfo != null) { ... }
```

src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
```diff
@@ -6,5 +6,9 @@ namespace Microsoft.R.DataInspection {
...
+        /// </summary>
+        IRValueInfo Value { get; }
```
int19h: Also, there is a mismatch here between the name of the property, and the name of the corresponding JSON entry. Ideally these should be the same, except for the naming convention (I think the only other place where we deviate currently is "type" vs "TypeName"). I'd call it something like `ComputedValue` for clarity.

src/Host/Client/Impl/DataInspection/IRActiveBindingInfo.cs
```diff
@@ -7,8 +7,9 @@ namespace Microsoft.R.DataInspection {
     /// </summary>
     public interface IRActiveBindingInfo : IREvaluationResultInfo {
         /// <summary>
-        /// Value of the active binding after evaluation.
+        /// This is null if evaluate active bindings in RTools Options is set
```
int19h: The doc comment should be referencing a lower-level concept, i.e. `REvaluationResultProperties.ComputedValueProperty` in this case, since this layer is not supposed to be VS-dependent or UI-dependent in general (and will likely be shared with VSCode in the future).

int19h:
:+1: , but it has merge conflicts, so you'll need to rebase before it can be merged.

Issue #1828, opened 2016-06-01
Syntax error squiggle are not visible in RMD


MikhailArkhipov:
Task list is also empty

PR #1832, opened 2016-06-02
#1828 Syntax error squiggle are not visible in RMD

MikhailArkhipov:
The beef is in 

```
 public bool IsTransient {
```

The rest is moving code around and formatting.


int19h:
:+1:

Issue #1088, opened 2016-02-29
Add R file template to SQL project


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/3a0b6b30cfc0b240bcab022c02d7af0107cf89dc

PR #1833, opened 2016-06-03
#1088 Add R file template to SQL project 

MikhailArkhipov:
So they appear in the database project when RTVS is installed. Setup authoring only.

image


int19h:
:+1:

PR #1834, opened 2016-06-04
Microsoft R Client telemetry

MikhailArkhipov:
MRC telemetry
Moved MRC-related code from RInstallation to package so it doesn't pollute Host.Client
Added test for MRC telemetry
Fixed Outliner threading test issue


src/Package/Impl/RClient/MicrosoftRClient.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.Common.Core;
+using Microsoft.Common.Core.OS;
+using Microsoft.Common.Core.Shell;
+using Microsoft.R.Support.Settings;
+using Microsoft.VisualStudio.R.Package.Options.R;
+using Microsoft.VisualStudio.R.Packages.R;
+using Microsoft.Win32;
+
+namespace Microsoft.VisualStudio.R.Package.RClient {
+    internal static class MicrosoftRClient {
```
AlexanderSher: Do we care about UI thread anywhere in this class?

src/R/Editor/Impl/Completion/RCompletionController.cs
```diff
@@ -165,6 +165,15 @@ public sealed class RCompletionController : CompletionController {
                     return false;
                 }
 
+                if(CharExtensions.IsLineBreak(typedChar)) {
+                    // Complete on Enter but onyl if selection does not exactly match
```
AlexanderSher: only

src/Package/Impl/RClient/MicrosoftRClient.cs
MikhailArkhipov: Good question. This is called on UI thread (at package load) and should be. Add assert?

AlexanderSher: Yes, that would be good

---
In reply to: [65924041](https://github.com/Microsoft/RTVS/pull/1834#discussion_r65924041) [](ancestors = 65924041)

AlexanderSher:
:shipit:

PR #1846, opened 2016-06-06
Fixes #1423 Scroll Bars in TreeGrid are themed like VS.

### LINKED ISSUE #1423: VE: scrollbar has different style than other panes
Although VE scroll bar is themed, it has different style (look arrow head, and thumb's width)
It seems like regression.

image

karthiknadig:


int19h:
:+1:

karthiknadig:
Screen shot :) 
image

Issue #1847, opened 2016-06-06
Data grid windows cause errors after REPL reset


huguesv:
```
ip <- installed.packages()
View(ip)
```
Reset REPL.

![image](https://cloud.githubusercontent.com/assets/1696845/15838465/ff0b17a0-2bf2-11e6-9a85-b20a4707fa0c.png)

If you keep playing with the existing data grid window, it will keep generating errors.

Also, if you try to Reset REPL once again, it will get stuck at "Stopping R Session..."

PR #1862, opened 2016-06-07
Handle the possibility of a removed variable in data grid.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/1847

### DIFF
--- a/src/Package/Impl/DataInspect/VariableGridHost.xaml.cs
+++ b/src/Package/Impl/DataInspect/VariableGridHost.xaml.cs
@@ -54,7 +54,7 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
         internal void SetEvaluation(VariableViewModel wrapper) {
             VsAppShell.Current.AssertIsOnMainThread();
 
-            if (wrapper.TypeName == "NULL" && wrapper.Value == "NULL") {
+            if (wrapper.TypeName == null && wrapper.Value == null) {
                 // the variable should have been removed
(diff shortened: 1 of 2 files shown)


src/Package/Impl/DataInspect/VariableGridHost.xaml.cs
```diff
@@ -54,7 +54,7 @@ public partial class VariableGridHost : UserControl {
         internal void SetEvaluation(VariableViewModel wrapper) {
             VsAppShell.Current.AssertIsOnMainThread();
 
-            if (wrapper.TypeName == "NULL" && wrapper.Value == "NULL") {
+            if (wrapper.TypeName == null && wrapper.Value == null) {
```
int19h: Perhaps this should be an OR? Will the other branch actually work if one is null, but the other one isn't?

huguesv: If I set an existing variable to `NULL` or `null` (same result):
```
ip <- NULL
```
then TypeName is `"NULL"` and Value is `null`.  It goes in the else statements and the result is a blank grid.  So that's fine.

int19h: We should probably be checking `TypeName` only then, since it sounds like a null value is generally valid.

huguesv:
Ok, only checking TypeName for `null`.

int19h:
:+1:

Issue #1864, opened 2016-06-08
Project properties value cannot be null error


huguesv:
Installed 2016-06-07.14 build
Right click on a project, properties:
```
Value cannot be null.
Parameter name: stream
```

This doesn't happen on dev machine running in exp hive.

PR #1865, opened 2016-06-08
Project properties fix

huguesv:
The new rule file run.xaml was missing from the installer, so it was going to the embedded resource as a fallback, but that was using the wrong namespace.

On dev machine everything worked since run.xaml is found in the build output rules subdir.

Fixes https://github.com/Microsoft/RTVS/issues/1864


AlexanderSher:
Please update files.swr private repo as well

Issue #1860, opened 2016-06-07
Add background shading in R blocks in RMD


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7f6b400551d23ea0a9c392ac34dfa0492c74708d

Issue #1852, opened 2016-06-06
Crash after pasting R code in R markdown code section


huguesv:
Add new R Markdown file to your project.
Put this in your clipboard.
```
plot(hx, col = ifelse(abs(hx$breaks) < 1.669, 4, 2))
```

Select this line by clicking in the margin.
```
plot(cars)
```

Paste.
VS crashes.

```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ObjectDisposedException
   at Microsoft.VisualStudio.Text.Formatting.Implementation.FormattedLine.ThrowIfInvalid()
   at Microsoft.VisualStudio.Text.Formatting.Implementation.FormattedLine.get_Start()
   at Microsoft.VisualStudio.Text.Editor.Implementation.TextViewLineCollection.FindTextViewLineIndexContainingBufferPosition(Microsoft.VisualStudio.Text.SnapshotPoint)
   at Microsoft.VisualStudio.Text.Editor.Implementation.TextViewLineCollection.GetTextViewLinesIntersectingSpan(Microsoft.VisualStudio.Text.SnapshotSpan)
   at Microsoft.VisualStudio.Text.Editor.Implementation.SolidColorBrushSelectionPainter.Update(Boolean)
   at Microsoft.VisualStudio.Text.Editor.Implementation.SolidColorBrushSelectionPainter.Activate()
   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextSelection.set_IsActive(Boolean)
   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.OnLostKeyboardFocus(System.Object, System.Windows.Input.KeyboardFocusChangedEventArgs)
   at System.Windows.Input.KeyboardFocusChangedEventArgs.InvokeEventHandler(System.Delegate, System.Object)
   at System.Windows.RoutedEventArgs.InvokeHandler(System.Delegate, System.Object)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(System.Object, System.Windows.RoutedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(System.Object, System.Windows.RoutedEventArgs, Boolean)
   at System.Windows.UIElement.RaiseEventImpl(System.Windows.DependencyObject, System.Windows.RoutedEventArgs)
   at System.Windows.UIElement.RaiseTrustedEvent(System.Windows.RoutedEventArgs)
[...]

Issue #1849, opened 2016-06-06
Snippet expansion using TAB doesn't work in R markdown


huguesv:
For example, in a R code section, type `sql` + TAB, TAB and it doesn't expand like in R editor.  Insert code snippet command works ok though.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7f6b400551d23ea0a9c392ac34dfa0492c74708d

Issue #1845, opened 2016-06-06
Pasting R code at the very end of code block in RMD throw exception


MikhailArkhipov:
Paste 
```
print(sample(c(2,5,3), size=3, replace=FALSE))
print(sample(c(2,5,3), size=3, replace=TRUE))
```
right before the trailing triple-tick

![image](https://cloud.githubusercontent.com/assets/12820357/15836731/06a1086a-2beb-11e6-92fe-e7b453cab489.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7f6b400551d23ea0a9c392ac34dfa0492c74708d

PR #1868, opened 2016-06-08
 #1839 #1845 #1849 #1852 #1860 

MikhailArkhipov:
- Better tracking of changes in ``` blocks ```
- Additional tests for changes around back ticks
- Proper insertion of snippets in markdown (use view buffer)
- Project adjacent tracking modes for edges of projection spans


src/Package/Impl/Shell/ThemeColorsProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.ComponentModel.Composition;
+using System.Windows.Media;
+using Microsoft.Common.Wpf.Themes;
+using Microsoft.VisualStudio.PlatformUI;
+
+namespace Microsoft.VisualStudio.R.Package.Shell {
+    [Export(typeof(IThemeColorsProvider))]
+    internal sealed class ThemeColorsProvider : IThemeColorsProvider {
+        public ThemeColorsProvider() {
+            VSColorTheme.ThemeChanged += OnThemeChanged;
```
AlexanderSher: Do you need to unsubscribe?

src/Package/Impl/Shell/ThemeColorsProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public bool IsDarkTheme {
+            get {
+                var defaultBackground = VSColorTheme.GetThemedColor(EnvironmentColors.ToolWindowBackgroundColorKey);
+                return defaultBackground.GetBrightness() < 0.5;
+            }
+        }
+        public Color CodeBackgroundColor {
+            get {
+                return IsDarkTheme ? Color.FromArgb(0xFF, 0x0C, 0x0C, 0x0C) : Color.FromArgb(0xFF, 0xFA, 0xFA, 0xFA);
```
huguesv: Could we make this customizable in tools/options/fonts?  I don't like that the user would have no workaround if they don't like what we picked.

MikhailArkhipov: Unfortunately it is tricky. Tools | Options only expose editor colorable items, not IDE items. And they are not directly themeable. If one uses standard classifications, they change with themes automatically. IDE colors also change, but they are not exposed to Options (only to Theme Editor). So yes, I will try to figure out if colors can be made themeable, but I am not 100% sure at the moment.

int19h: How does HTML do it for their embedded code snippets? They do seem to have colors in Tools | Options for this kind of stuff (e.g. "HTML Razor Code Background").

MikhailArkhipov: Since they are in box, their colors are in VS theme I believe. I am trying to discover where is that.

huguesv: PTVS includes theme specific colors in its package, maybe this helps:
https://github.com/Microsoft/PTVS/tree/master/Python/Product/PythonTools/Theme

MikhailArkhipov: Aha, that is usefull

src/Package/Impl/Shell/ThemeColorsProvider.cs
MikhailArkhipov: Added disposable

src/Package/Impl/Shell/ThemeColorsProvider.cs
MikhailArkhipov: This will be quite a bit of separate work. Opened https://github.com/Microsoft/RTVS/issues/1869

MikhailArkhipov:
Merging for the BB, will address anything new next PR.

PR #1870, opened 2016-06-09
F5 and Ctrl+F5

### LINKED ISSUE #1090: Launch project with debugging (F5)
Implement IVsDebuggableProjectCfg in the project system, and LaunchSuspended in debugger

### LINKED ISSUE #1654: Launch project without debugging (Ctrl+F5)

int19h:
Fix #1090: Launch project with debugging (F5)
Fix #1654: Launch project without debugging (Ctrl+F5)

Implement IDebugLaunchProvider for R projects.

### DIFF
--- a/src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
+++ b/src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
@@ -229,6 +229,7 @@
     <Compile Include="ProjectSystem\PropertyPages\WpfBasedPropertyPage.Designer.cs">
       <DependentUpon>WpfBasedPropertyPage.cs</DependentUpon>
     </Compile>
+    <Compile Include="ProjectSystem\RDebugLaunchProvider.cs" />
     <Compile Include="RClient\MicrosoftRClient.cs" />
     <Compile Include="RClient\MicrosoftRClientInstaller.cs" />
     <Compile Include="Repl\Shiny\StopShinyAppCommand.cs" />
@@ -1190,6 +1191,14 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Generator>MSBuild:Compile</Generator>
       <SubType>Designer</SubType>
+    </Content>
+    <Content Include="Rules\Debugger.xaml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Generator>MSBuild:Compile</Generator>
+      <SubType>Designer</SubType>
+    </Content>
+    <Content Include="Targets\Microsoft.R.targets">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
   <ItemGroup>
--- a/src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
+++ b/src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
@@ -18,7 +18,7 @@ using Microsoft.VisualStudio.ProjectSystem.VS.Debuggers;
 
 namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
     // ExportDebugger must match rule name in ..\Rules\Debugger.xaml.
-    [ExportDebugger("RDebugger")] 
+    [ExportDebugger("RDebugger")]
     [AppliesTo(Constants.RtvsProjectCapability)]
     internal class RDebugLaunchProvider : DebugLaunchProviderBase {
         private readonly ProjectProperties _properties;
@@ -33,8 +33,8 @@ namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
 
         private IRSession Session => _interactiveWorkflow.RSession;
 
-        public override async Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions) {
-            return Session.IsHostRunning && !string.IsNullOrEmpty(await _properties.GetStartupFileAsync());
+        public override Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions) {
+            return Task.FromResult(true);
         }
 
         public override Task<IReadOnlyList<IDebugLaunchSettings>> QueryDebugTargetsAsync(DebugLaunchOptions launchOptions) {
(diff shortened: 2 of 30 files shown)


src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
```diff
@@ -1191,6 +1192,14 @@
       <Generator>MSBuild:Compile</Generator>
       <SubType>Designer</SubType>
     </Content>
+    <Content Include="Rules\Debugger.xaml">
```
MikhailArkhipov: Please add to files.swr in Setup.15 folder for Willow

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+        [ImportingConstructor]
+        public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWorkflowProvider interactiveWorkflowProvider)
+            : base(configuredProject) {
+            _properties = configuredProject.Services.ExportProvider.GetExportedValue<ProjectProperties>();
+            _interactiveWorkflow = interactiveWorkflowProvider.GetOrCreate();
+        }
+
+        private IRSession Session => _interactiveWorkflow.RSession;
+
+        public override async Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions) {
```
MikhailArkhipov: Should we allow F5 and pop up message telling user to specify function to call in the project properties? It may be hard for them to discover how to make F5 working

MikhailArkhipov:
There is some info in OneNote how to build swixproj on desktop if you want

MikhailArkhipov:
:+1:

MikhailArkhipov:
Pending Setup.15 changes

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
int19h: I'll see if I can propagate a better error message to CPS somehow. Right now you get a message box saying something like "Cannot start because project not configured", but then tells you to go and check if you have the right thing as a startup project - which is misleading.

If not, it'll have to be done in `LaunchAsync`, because other things can be called in contexts where UI (esp. blocking) is not expected.

MikhailArkhipov: Yeah, I meant in the LaunchAsync. If there is nothing set, post some error message. Maybe even in REPL.

src/ProjectSystem/Impl/Project/FileSystemMirroringProjectFileGenerator.cs
```diff
@@ -63,6 +64,7 @@ public abstract class FileSystemMirroringProjectFileGenerator : IVsProjectGenera
                             new XProjElement("UserProperties")
                         )
                     ),
+                    new XProjElement("Target", new XAttribute("Name", "Build")),
```
huguesv: Does this mean F5/CTRL-F5 won't work for existing projects?  If yes, what is the symptom?  Start command disabled?
I'm not saying that it needs to work for existing projects.  We're still in preview after all.  We could just document the manual change.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+            // Base implementation will try to launch or attach via the debugger, but there's nothing to launch
+            // in case of Ctrl+F5 - we only want to source the file. So only invoke base if we intend to debug.
+            if (!launchOptions.HasFlag(DebugLaunchOptions.NoDebug)) {
+                await base.LaunchAsync(launchOptions);
+            }
+
+            string startupFile = await _properties.GetStartupFileAsync();
+            if (!string.IsNullOrEmpty(startupFile)) {
+                _interactiveWorkflow.ActiveWindow?.Container.Show(false);
+                _interactiveWorkflow.Operations.SourceFileAsync(startupFile, echo: false)
```
AlexanderSher: Why we don't await `SourceFileAsync`?

src/Setup/Definitions.wxi
```diff
@@ -23,6 +23,7 @@
   <?define ItemTemplatesDir="$(var.BinDir)\ItemTemplates\R" ?>
   <?define ProjectTemplatesDir="$(var.BinDir)\ProjectTemplates\R" ?>
   <?define CpsRulesDir="$(var.BinDir)\Rules" ?>
+  <?define MSBuildTargetsDir="$(var.BinDir)\Targets" ?>
```
AlexanderSher: You also need to make these changes in private repo

src/ProjectSystem/Impl/Project/FileSystemMirroringProjectFileGenerator.cs
int19h: Yes, we'll need to doc the manual workaroud (add `<Target Name="Build"/>`) for now. Later we can do the same thing PTVS did, and implement project upgrade, but that's a can of worms that I really don't want to open right now.

And yes, this makes both Start commands being unavailable.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
int19h: It didn't feel like there's any meaning in awaiting it - the goal of this is to kick the execution off, it shouldn't really care where it goes from there.

huguesv:
👍  Please file follow up items to document manual change to existing rxproj, honoring the Reset REPL option, and possibliy honoring the command line args, or hiding it from UI.

int19h:
I'll add code to reset REPL - just forgot about it, and it's easy enough.

src/Setup/Definitions.wxi
int19h: https://github.com/Microsoft/RTVS-pr/pull/5

huguesv:
You probably need to install debugger.xaml too in cpsrules.wxi

int19h:
Ah, indeed. I missed it because it gets loaded from experimental if it's there.

huguesv:
:+1:

Issue #1887, opened 2016-06-09
EvaluationWrapperTest.FormulaTest failure


huguesv:
I'm using CRAN R 3.3.0

```
Test Name:	EvaluationWrapperTest.FormulaTest
Test FullName:	Microsoft.VisualStudio.R.Package.Test.DataInspect.EvaluationWrapperTest.FormulaTest
Test Source:	C:\Projects\huguesv\RTVS\src\Package\Test\DataInspect\EvaluationWrapperTest.cs : line 87
Test Outcome:	Failed
Test Duration:	0:00:03.623

Result StackTrace:	
at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions.Net45\Execution\XUnit2TestFramework.cs:line 35
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions.Net40\Execution\TestFrameworkProvider.cs:line 42
   at FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\CollectingAssertionStrategy.cs:line 50
   at FluentAssertions.Execution.AssertionScope.Dispose() in C:\projects\fluentassertions-vf06b\Src\Core\Execution\AssertionScope.cs:line 252
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(EquivalencyValidationContext context) in C:\projects\fluentassertions-vf06b\Src\Core\Equivalency\EquivalencyValidator.cs:line 37
   at FluentAssertions.AssertionExtensions.ShouldBeEquivalentTo[T](T subject, Object expectation, Func`2 config, String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\Shared\AssertionExtensions.cs:line 687
   at Microsoft.VisualStudio.R.Package.Test.DataInspect.VariableRHostScript.AssertEvaluationWrapper(IRSessionDataObject rdo, VariableExpectation expectation) in C:\Projects\huguesv\RTVS\src\Package\Test\DataInspect\VariableRHostScript.cs:line 75
[...]

PR #1895, opened 2016-06-10
Make FormulaTest pass on both R 3.2 and 3.3.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/1887


AlexanderSher:
Do we need to handle this anywhere in the product itself?

int19h:
I don't think so. R changed the return value of `str()` for formulas, basically. But we don't really rely on it anywhere (and we shouldn't, precisely because it's not a stable parseable representation) - we just show it to the user.

int19h:
This does however raise the question of what to do about such cases. Right now the test will be run using only one R version, so it will only be testing one of the possible two cases. We'll need to figure something out to make it, and other cases like it, run on different R versions for our nightlies.

AlexanderSher:
:+1:

PR #1896, opened 2016-06-10
Bunch'o'fixes

MikhailArkhipov:


src/Package/Impl/Help/ShowHelpOnCurrentCommand.cs
```diff
@@ -88,7 +88,7 @@ internal sealed class ShowHelpOnCurrentCommand : PackageCommand {
             if (activeReplWindow != null && _activeReplTracker.IsActive) {
                 return activeReplWindow.InteractiveWindow.TextView;
             }
-            return _textViewTracker.GetLastActiveTextView(RContentTypeDefinition.ContentType);
+            return _textViewTracker.LastActiveTextView;
```
AlexanderSher: What if it was SQL script?

MikhailArkhipov: It is context menu command so the context menu itself wouldn't show up.

MikhailArkhipov: I.e. if context menu with this command is up, it is the right content type

AlexanderSher:
:+1:

PR #1900, opened 2016-06-10
Fix ##1069: Placement of R project templates in New Project dialog is weird

int19h:
Adjust priority to place it right after SQL, and next to Python.

image


AlexanderSher:
:+1:

Issue #1901, opened 2016-06-10
Some code pages shown in the code page drop down are not honored


karthiknadig:
Repro steps:
Set Code Page in R Tools Options to ISCII Kannada

```
> Sys.getlocale("LC_CTYPE")
[1] "English_United States.437"
Warning messages:
1: In Sys.setlocale("LC_CTYPE", ".57008") :
  OS reports request to set locale to ".57008" cannot be honored
```

MikhailArkhipov:
Yep, we should .NET list while R actually uses list from C runtime or something that is not entirely documented. So our list is basically optimistic.

int19h:
We assumed that if it has a codepage, it's supported, basically. But apparently MinGW (well, msvcrt.dll) doesn't know about many encodings, not just UTF-8. It might be worth seeing if we can get access to that code, and just make it a hardcoded list, since it's not a component that is going to have new stuff added to it.

MikhailArkhipov:
If it just one or two exotic ones I wouldn't bother. I remember trying common/popular ones and they worked.

int19h:
It looks like it's all the various ISCII varieties, at the very least.

MikhailArkhipov:
OK, I will filter these ones for now

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/2672327dadc2f80501ae313d688a7b2cd7e09468

Issue #1877, opened 2016-06-09
 Add R Script vs. Add > New item > R Script have inconsistent filename casing (.r vs .R)


karthiknadig:
![image](https://cloud.githubusercontent.com/assets/3840081/15945204/1043d0ac-2e46-11e6-9537-b77d16e93de2.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/82d5a2d02357aa6ce25563c5a302c5c8627f71cf

huguesv:
Tested with the build where it's supposed to be fixed, but it's not ;)

I expect it's because Git didn't actually rename the template file name on disk, you have to do that in 2 steps when doing a rename that only changes file name casing.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/2672327dadc2f80501ae313d688a7b2cd7e09468

PR #1903, opened 2016-06-13
#1877 #1901

MikhailArkhipov:
+ small issue in completion on Enter in editor so it only completes when text is not fully typed or when completion is explicitly selected

### DIFF
--- a/src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
+++ b/src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
@@ -11,28 +11,30 @@ using Microsoft.Common.Core;
 
 namespace Microsoft.VisualStudio.R.Package.Options.R.Tools {
     internal sealed class EncodingTypeConverter : TypeConverter {
-        private static readonly int[] _unicodePages = new int[] {
-            Encoding.UTF7.CodePage ,
-            Encoding.UTF8.CodePage,
-            Encoding.Unicode.CodePage,
-            Encoding.UTF32.CodePage,
-            Encoding.BigEndianUnicode.CodePage,
-            12001 // Big Endian UTF32
+        private static readonly int[] _supportedCodePages = new int[] {
+            708,720, 737, 775,
+            850, 852, 855, 857, 860, 861, 862,
+            863, 864, 865, 866, 869, 874,
+            932, 936, 949, 950,
+            1250, 1251, 1252, 1253,1254, 1255, 1256, 1257, 1258,
+            20000, 20127, 20866, 20932, 20936, 20949, 21866,
+            28591, 28592, 28593, 28594, 28595, 28596, 28597,
+            28598, 28599, 28603, 28605,
+            38598
         };
 
-        public override bool GetStandardValuesSupported(ITypeDescriptorContext context) {
-            return true;
-        }
+        public override bool GetStandardValuesSupported(ITypeDescriptorContext context) => true;
 
         public override StandardValuesCollection GetStandardValues(ITypeDescriptorContext context) {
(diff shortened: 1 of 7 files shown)


src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
```diff
@@ -11,25 +11,30 @@
...
 namespace Microsoft.VisualStudio.R.Package.Options.R.Tools {
     internal sealed class EncodingTypeConverter : TypeConverter {
-        private static readonly int[] _unicodePages = new int[] {
-            Encoding.UTF7.CodePage ,
-            Encoding.UTF8.CodePage,
-            Encoding.Unicode.CodePage,
-            Encoding.UTF32.CodePage,
-            Encoding.BigEndianUnicode.CodePage,
-            12001 // Big Endian UTF32
+        private static readonly int[] _supportedCodePages = new int[] {
```
int19h: Out of curiosity, how did you make this list?

src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
```diff
@@ -11,25 +11,30 @@
...
         };
 
-        public override bool GetStandardValuesSupported(ITypeDescriptorContext context) {
-            return true;
-        }
+        public override bool GetStandardValuesSupported(ITypeDescriptorContext context) => true;
 
         public override StandardValuesCollection GetStandardValues(ITypeDescriptorContext context) {
+            var encList = new List<Encoding>();
+            foreach(var cp in _supportedCodePages) {
```
int19h: `_supportedCodePages.Select(cp => Encoding.GetEncoding(cp)).ToList()`

src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
```diff
@@ -11,25 +11,30 @@
...
+                encList.Add(Encoding.GetEncoding(cp));
+            }
+
+            var codePages = encList.OrderBy(e => e.EncodingName)
+                                   .Select(e => e.CodePage);
+
             var codePageList = new List<int>() { 0 };
```
int19h: Also `ToList()`.

src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
int19h: (actually, no need for `ToList` even)

src/Package/Impl/Options/R/Tools/EncodingTypeConverter.cs
int19h: Never mind, I see the test now :)

src/Package/Test/Options/EncodingsTest.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.VisualStudio.R.Package.Test.Utility;
+using Xunit;
+using static System.FormattableString;
+
+namespace Microsoft.VisualStudio.R.Package.Test.Options {
+    [ExcludeFromCodeCoverage]
+    [Collection(CollectionNames.NonParallel)]
+    public class EncodingsTest {
+        [Test]
+        public async Task ValidateEncodings() {
```
int19h: Perhaps make it `[CompositeTest]` with one test per page? So that when it fails, we know right away which one.

MikhailArkhipov: It takes values from the table in options, I didn't want to have separate table in InlineData. If it fails, ```s.Should().NotBeNull().And.Contain(cp``` will provide the code page

Issue #1893, opened 2016-06-10
New project should have startup file automatically set


huguesv:
Maybe `RProjectFileGenerator` can enumerate the .R files in the folder where the msbuild file is being created, and pick one to set as startup file.

An alternative is to change the project template to include an .rxproj (right now it's only an .rproj)

PR #1907, opened 2016-06-13
Add StartupFile property generation to RProjectFileGenerator.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/1893

The advantage of fixing it in the project file generator is that projects created from RStudio that are later opened in VS will have a startup file set automatically.

Changing the template to create the .rxproj with the startup file set as part of the template would allow each template to set whatever startup file they want, instead of relying on the project file generator to pick one.  However, I haven't been able to get this to work, VS complains "The template specified cannot be found. Please check that the full path is correct."  The template looks valid to me, it's likely that we are missing project type registrations of some sort, and VS doesn't know what to do with a .rxproj project template.

Since we only have one project template, with only one .R file in it, the fix in the generator seems appropriate enough, and has the advantage of setting a startup file for RStudio projects.

### DIFF
--- a/src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
+++ b/src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
@@ -337,7 +337,7 @@
     <Compile Include="Packages\Markdown\MdPackage.cs" />
     <Compile Include="Packages\Markdown\MdGuidList.cs" />
     <Compile Include="ProjectSystem\RMsBuildFileSystemFilter.cs" />
-    <Compile Include="ProjectSystem\RProjectFactory.cs" />
+    <Compile Include="ProjectSystem\RProjectFileGenerator.cs" />
     <Compile Include="Interop\CommandTargetToOleShim.cs" />
     <Compile Include="Interop\CommandTargetToOleShimVariantStacks.cs" />
     <Compile Include="Interop\ConnectionPoint.cs" />
@@ -938,6 +938,7 @@
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Xaml" />
     <Reference Include="System.Xml" />
+    <Reference Include="System.Xml.Linq" />
     <Reference Include="WindowsBase" />
     <Reference Include="WindowsFormsIntegration" />
   </ItemGroup>
--- a/src/Package/Impl/ProjectSystem/RProjectFileGenerator.cs
+++ b/src/Package/Impl/ProjectSystem/RProjectFileGenerator.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.IO;
 using System.Runtime.InteropServices;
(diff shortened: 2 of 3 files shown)


int19h:
:+1:

huguesv:
Turns out I need to set the `Condition` on it, otherwise when editing with the project properties UI, it will always set it without the condition, which means no way of setting different values for different configs, as it was designed to be.

huguesv:
If only it was as simple as adding the `Condition` attribute.  Sigh.

If I make it generate a PropertyGroup with the Condition, exactly like the property page UI creates:
```
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <ResetReplOnRun>True</ResetReplOnRun>
    <StartupFile>script.R</StartupFile>
    <CommandLineArgs />
  </PropertyGroup>
```

Then the .sln file that gets created is missing configuration section for some reason.  That means F5/CTRL-F5 is disabled, the platform and config combo box on the main toolbar are blank and disabled.

For comparison, here's a working solution, where the project generator doesn't add the property group at all:
```

Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio 14
VisualStudioVersion = 14.0.25123.0
MinimumVisualStudioVersion = 10.0.40219.1
Project("{DA7A21FA-8162-4350-AD77-A8D1B671F3ED}") = "rproject28", "rproject28\rproject28.rxproj", "{C7F5C54A-119F-49D7-BCBD-09B1BF5D2968}"
EndProject
Global
	GlobalSection(SolutionConfigurationPlatforms) = preSolution
		Debug|Any CPU = Debug|Any CPU
		Release|Any CPU = Release|Any CPU
	EndGlobalSection
	GlobalSection(ProjectConfigurationPlatforms) = postSolution
		{C7F5C54A-119F-49D7-BCBD-09B1BF5D2968}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
		{C7F5C54A-119F-49D7-BCBD-09B1BF5D2968}.Debug|Any CPU.Build.0 = Debug|Any CPU
		{C7F5C54A-119F-49D7-BCBD-09B1BF5D2968}.Release|Any CPU.ActiveCfg = Release|Any CPU
		{C7F5C54A-119F-49D7-BCBD-09B1BF5D2968}.Release|Any CPU.Build.0 = Release|Any CPU
	EndGlobalSection
	GlobalSection(SolutionProperties) = preSolution
		HideSolutionNode = FALSE
	EndGlobalSection
EndGlobal
```


And here's a bad solution, where the project file generator added the condition based PropertyGroup:

```

Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio 14
VisualStudioVersion = 14.0.25123.0
MinimumVisualStudioVersion = 10.0.40219.1
Project("{DA7A21FA-8162-4350-AD77-A8D1B671F3ED}") = "rproject30", "rproject30\rproject30.rxproj", "{3E9E5289-DEB8-455B-8EAD-A404F218C651}"
EndProject
Global
	GlobalSection(SolutionProperties) = preSolution
		HideSolutionNode = FALSE
	EndGlobalSection
EndGlobal

```

Issue #1908, opened 2016-06-13
Grid viewer crashes on 1D vectors


int19h:
Note that this is about vectors that _do_ have dimensions, but the count of them is 1, as opposed to dimensionless vectors.

```r
xs <- 1:10
dim(xs) <- 10
View(xs)
```

Issue #1884, opened 2016-06-09
Changing column names while grid is sorted breaks grid


int19h:
```r
m <- matrix(1:9, 3, 3)
View(m)
```
then sort by any of the columns - e.g. V1. Then:
```r
colnames(m) <- c(1,2,3)
```
You get:
```
Error in `[.data.frame`(x, "V1") : undefined columns selected
```
As with #1883, after this happens, REPL hangs, and cannot even be reset.

MikhailArkhipov:
I think we need to reset sorting on mutation

int19h:
We also need to harden that code in general. Basically, it has to assume that anything it evaluates that is based on user input in any way can throw, and catch/recover accordingly.

PR #1909, opened 2016-06-13
Fix #1883, #1884, #1908 (grid sort and related issues)

### LINKED ISSUE #1883: Grid sort doesn't handle unnamed columns properly
```r
m <- matrix(1:9, 3, 3)
colnames(m) <- c(1, NA, 3)
View(m)
```
Then try to sort by the "NA" column. You will get this error:
```
Error in `[.data.frame`(x, "NA") : undefined columns selected
```

int19h:


src/Package/TestApp/Files/VariableGrid02.tree
```diff
@@ -520,7 +520,7 @@
                                       "Properties": [
                                         {
                                           "Name": "Text",
-                                          "Value": "V1"
+                                          "Value": "[,1]"
```
MikhailArkhipov: RStudio shows V1 that's why it was changed back then. I frankly don't have background why we were different

int19h: We were showing `V1` because that is an auto-generated column name when you convert matrix to data frame. I changed the code to avoid that conversion completely now, so for matrices, there are no auto-generated names anymore, and the code on C# side that generates columns kicks in (same one that produces them for 1D vectors). I prefer this approach, because it's now clear from column names whether they were originally named or not.

MikhailArkhipov:
:+1:

PR #1911, opened 2016-06-14
Display indices instead of NA for headers of rows and columns that do not have a name

int19h:
This was a sub-issue in one of the grid bugs that I have overlooked.


MikhailArkhipov:
:+1:

PR #1912, opened 2016-06-14
Variable Explorer context menu

MikhailArkhipov:


src/Package/TestApp/Utility/VisualTreeExtensions.cs
```diff
@@ -4,47 +4,23 @@
...
-                int childrenCount = VisualTreeHelper.GetChildrenCount(o);
-                for (int i = 0; i < childrenCount; i++) {
-                    var child = VisualTreeHelper.GetChild(o, i);
-                    var inner = await FindChildAsync<T>(child);
-                    if (inner != null) {
-                        return inner;
-                    }
-                }
-                return null;
+            return await UIThreadHelper.Instance.InvokeAsync(() => {
```
AlexanderSher: You don't need async/await here.

src/Package/Impl/Shell/VsContextMenu.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.Windows;
+using Microsoft.R.Components.Controller;
+using Microsoft.VisualStudio.R.Package.Interop;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.R.Package.Shell {
+    public static class VsContextMenu {
+        public static void Show(Guid guid, int id, ICommandTarget commandTarget, Point screenPosition) {
```
AlexanderSher: What's the difference between this one and ICoreShell.ShowContextMenu?

src/Common/Wpf/Impl/Extensions/VisualTreeExtensions.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Windows;
+using System.Windows.Media;
+
+namespace Microsoft.Common.Wpf.Extensions {
+    public static class VisualTreeExtensions {
+        public static T FindChild<T>(DependencyObject o) where T : DependencyObject {
```
AlexanderSher: I think this method requires better name.
It searches for the first child of some type in depth, correct?

src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -219,5 +273,85 @@ public partial class VariableView : UserControl, IDisposable {
                     return KnownMonikers.BulletList;
             }
         }
+        #endregion
+
+        #region ICommandTarget
+        public CommandStatus Status(Guid group, int id) {
```
AlexanderSher: I don't understand why do you need this.
VE is a tool window, which should handle VS commands.

src/Package/Impl/Shell/VsContextMenu.cs
MikhailArkhipov: This one goes via IvsUIShell that has ability to supply command target rather than have menu service send commands via VS chain. I guess we can add optional command target parameter to ICoreShell.ShowContextMenu and have it do it via menu service or via command target.

src/Package/TestApp/Utility/VisualTreeExtensions.cs
MikhailArkhipov: OK. It was async before so I left it.

src/Common/Wpf/Impl/Extensions/VisualTreeExtensions.cs
MikhailArkhipov: FindFirstVisualChildOfType?

AlexanderSher: Ok

---
In reply to: [67010428](https://github.com/Microsoft/RTVS/pull/1912#discussion_r67010428) [](ancestors = 67010428)

src/Package/Impl/DataInspect/VariableView.xaml.cs
MikhailArkhipov: Or expose VisualGrid via some sort of interface. It may be better to do it when we turn grid to a visual component.

AlexanderSher: Do you need a grid or a model from the grid?

---
In reply to: [67012100](https://github.com/Microsoft/RTVS/pull/1912#discussion_r67012100) [](ancestors = 67012100)

MikhailArkhipov: Model is probably fine

MikhailArkhipov:
Also added set directory commands as in RS per users request and re-worded message when R is not found.

MikhailArkhipov:
I think I'd leave alone for now plumbing command target up to the pane, it will be bigger change I think,

PR #1921, opened 2016-06-14
Fix #1229: Loading solution is stuck

### LINKED ISSUE #1229: Loading solution is stuck
repro steps: double click on xyz.sln, which contains a R project

Dump is here: \\pytools\R\Bugs\Project_open_stuck\devenv.DMP

AlexanderSher:
Issue happens when CPS makes UI thread "half-reentrant" - you can reenter it but only using `JoinableTaskFactory`.


int19h:
So basically in VS, the only legal way to post to UI thread is via `JoinableTaskFactory`, otherwise you risk deadlock?

int19h:
:+1:

AlexanderSher:
> So basically in VS, the only legal way to post to UI thread is via JoinableTaskFactory, otherwise you risk deadlock?

If there is CPS method in UI stack, yes

PR #1922, opened 2016-06-14
Fixes #1917 Updates background color when "Markdown Code Background" …

### LINKED ISSUE #1917: Change Markdown Code Background color doesn't apply to already opened files
Open .rmd file
Tools/Options/Fonts and Colors, Text Editor, Markdown Code Background (change to a different color)
Click OK
Notice no change in editor.
Close and reopen .rmd file
Notice new color is used.

I tried with a few other color names, such as 'Markdown Bold Text' (applies to markdown sections), and 'String' (applies to R sections) and both of them applied immediately without having to reload.

karthiknadig:
Updates background color when "Markdown Code Background" is changed


MikhailArkhipov:
I expected editor to fire LayoutChanged on colors change since format definitions can actually change fonts and so on, but I guess it doesn't.

👍

PR #1923, opened 2016-06-14
Fixes #1913 Tooltip will now show plot size.

### LINKED ISSUE #1913: Plot: Display resolution somewhere
From user:
Would be nice if it displayed the resolution somewhere (maybe status line? a tooltip? Right-click? I don’t know… somewhere)

karthiknadig:


karthiknadig:
sample:
image

huguesv:
We should set the tooltip on the <Image> element, so that it doesn't show when the plot is cleared and the watermark is displayed.
It may be possible to do it all in Xaml too, by binding to PlotImage that's on the ViewModel.

PR #1927, opened 2016-06-15
Fix #1924: Sorting a list in the grid does not work

### LINKED ISSUE #1924: Sorting a list in the grid does not work
If a list is opened in a grid, and then sorted, it breaks, because `order()` does not support lists.

In general, lists cannot be meaningfully sorted, because they are heterogeneous, and different data types don't have comparison operators defined between them. So sorting should be disabled for lists.

(But note that dataframe is a special kind of list, which is treated specially, and in particular, is supported by `order`.)

int19h:
Fix a bunch of failing grid unit tests and related issues.

Add better tests for rtvs:::grid_data (not ful coverage yet, more to be added).

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/grid.R
+++ b/src/Host/Client/Impl/rtvs/R/grid.R
@@ -1,20 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See LICENSE in the project root for license information.
 
-grid_trim <- function(str, max_length = 100) {
-  if (nchar(str) > (100 - 3)) {
-    paste(substr(str, 1, 97), '...', sep='');
-  } else {
-    str;
-  }
-}
-
 grid_header_format <- function(x)
 	if (is.na(x)) NULL else format(x)
-
-grid_format <- function(x) {
-  sapply(format(x, trim = TRUE, justify = "none"), grid_trim);
-}
 
 grid_data <- function(x, rows, cols, row_selector) {
   # If it's a 1D vector, turn it into a single-column 2D matrix, then process as such.
@@ -35,9 +23,19 @@ grid_data <- function(x, rows, cols, row_selector) {
   if (!missing(row_selector)) {
       x <- x[row_selector(x),, drop = FALSE]
   }
-  x <- x[rows, cols]
+  x <- x[rows, cols, drop = FALSE]
 
-  data <- sapply(x, grid_format, USE.NAMES = FALSE);
+  # Process and format values column by column, then flatten the resulting list of character vectors.
+  max_length <- 100 - 3
+  data <- c(lapply(1:ncol(x), function(i) {
+    lapply(format(x[,i], trim = TRUE, justify = "none"), function(s) {
+      if (nchar(s) <= max_length) s else paste0(substr(s, 1, max_length), '...', collapse = '')
+    })
+  }), recursive = TRUE)
+
+  # Any names in the original data will flow through, but we don't want them.
+  names(data) <- NULL;
+
   rn <- row.names(x);
   cn <- colnames(x);
 
@@ -61,10 +59,6 @@ grid_data <- function(x, rows, cols, row_selector) {
   vp$col.start <- cols[1];
   vp$col.count <- length(cols);
   vp$col.names <- as.list(x.colnames);
-  data.list <- list();
-  for (i in data) {
-     data.list[length(data.list) + 1] <- as.list(i);
-  }
-  vp$data <- data.list;
+  vp$data <- as.list(data);
   vp;
 }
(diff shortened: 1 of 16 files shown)


int19h:
@MikhailArkhipov  @AlexanderSher can you take a look? I'd like to merge this in before deadline.

The new tests are incomplete - I didn't have the time to cover all the stuff - so I left the old ones for now. This means some duplication between them. I'll continue working on them once this is merged, but the primary goal is to get the necessary product changes in.

src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
```diff
@@ -31,5 +30,10 @@ internal sealed class GridDataSource {
                 }
             }
         }
+
+        public static Task<IGridData<string>> GetGridDataAsync(string expression, GridRange? gridRange, ISortOrder sortOrder = null) =>
+            GetGridDataAsync(
+                VsAppShell.Current.ExportProvider.GetExportedValue<IRSessionProvider>().GetInteractiveWindowRSession(),
```
AlexanderSher: Better access it through IWorkflowProvider

src/Package/Impl/DataInspect/VisualGrid/SortOrder.cs
```diff
@@ -54,6 +47,10 @@ internal sealed class SortOrder : ISortOrder {
             }
         }
 
+        public void Add(ColumnSortOrder order) {
+            _sortOrderList.Add(order);
```
AlexanderSher: Do we need to check if `ColumnSortOrder` is in the list already?

src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
int19h: This is the way that it was accessed originally before I split this up. There's really a lot to refactor here - e.g. this class shouldn't even have this overload at all, the caller should just supply session - but for now I'd just like to fix the issues / failing tests.

src/Package/TestApp/Data/GridDataTest.cs
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        private static IEnumerable<T> ToEnumerable<T>(IRange<T> range) {
+            int start = range.Range.Start, end = start + range.Range.Count;
+            for (int i = start; i < end; ++i) {
+                yield return range[i];
+            }
+        }
+
+        private void ShouldEqual<T>(IGrid<T> actual, T[,] expected) {
```
AlexanderSher: Can you add a task to extract this into Assertions later?

src/Package/Impl/DataInspect/DataSource/GridDataSource.cs
AlexanderSher: ok

---
In reply to: [67211001](https://github.com/Microsoft/RTVS/pull/1927#discussion_r67211001) [](ancestors = 67211001)

src/Host/Client/Impl/rtvs/R/grid.R
```diff
@@ -35,9 +23,19 @@ grid_data <- function(x, rows, cols, row_selector) {
   if (!missing(row_selector)) {
       x <- x[row_selector(x),, drop = FALSE]
   }
-  x <- x[rows, cols]
+  x <- x[rows, cols, drop = FALSE]
+
+  # Process and format values column by column, then flatten the resulting list of character vectors.
+  max_length <- 100 - 3
```
MikhailArkhipov: Probably will need to be configurable

MikhailArkhipov:
👍

AlexanderSher:
:shipit:

src/Package/Impl/DataInspect/VisualGrid/SortOrder.cs
int19h: No, since it's purely a UI thing. Sort itself doesn't care (or break) if you sort on the same column twice - it's not particularly useful, but it's not harmful either. And doing nothing is not really a good thing, because if you sort by A, then by B, then by A, it's not the same as sorting by A then by B (which is what happens when the last add is a no-op). It should behave as sorting by B then by A. 

Also, this is basically a lower-level access hatch for the tests currently, to avoid tying it to all the visual stuff - it's not used in product code. In future refactoring, I intend it to make the only way to work with this class, and move the visual-related logic elsewhere, to the UI layer where it belongs.

src/Package/TestApp/Data/GridDataTest.cs
int19h: I think we need a generic assertion for 2D arrays (including non-zero bounds). Then this code just needs a small helper to convert IGrid to one. I'll file that.

src/Host/Client/Impl/rtvs/R/grid.R
int19h: Definitely. I kept the hardcoded value for now because wiring up the option would touch a lot more stuff.

PR #1929, opened 2016-06-15
This change moves the plot resolution tool tip to the Image UI element.

karthiknadig:
This change moves the plot resolution tool tip to the Image UI element. This is to fix the following problem @huguesv identified:
image
Tool tip is shown when there is no plot.


huguesv:
I would rather do it all from xaml.  This appears to work fine:
```
        <Image Cursor="{Binding Path=LocatorMode, Converter={x:Static rwpf:Converters.TrueIsCrossCursor}}" Source="{Binding Path=PlotImage}" Visibility="{Binding Path=PlotImage, Converter={x:Static rwpf:Converters.NullIsCollapsed}}" MouseLeftButtonUp="Image_MouseLeftButtonUp">
            <Image.ToolTip>
                <TextBlock>
                    <Run Text="{Binding Path=PlotImage.PixelWidth, Mode=OneWay}"/>
                    <Run Text=" x "/>
                    <Run Text="{Binding Path=PlotImage.PixelHeight, Mode=OneWay}"/>
                    <Run Text=" px"/>
                </TextBlock>
            </Image.ToolTip>
        </Image>
```

huguesv:
👍  You could remove the x:Name since it's no longer used.

Issue #1933, opened 2016-06-15
Set Working Directory to Project sometimes fails


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/16095028/b563a19c-32f7-11e6-8794-6e77c7f816d3.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f694afa327ff438ff7c4b4d26879698a003a5cf4

PR #1951, opened 2016-06-15
#1933 Set Working Directory to Project sometimes fails

MikhailArkhipov:


src/Package/Impl/ProjectSystem/Definitions/IProjectSystemServices.cs
```diff
@@ -12,7 +14,7 @@ public interface IProjectSystemServices {
         /// <summary>
         /// Locates project that is currently selected in Solution Explorer
         /// </summary>
-        EnvDTE.Project GetSelectedProject();
+        IVsProject GetSelectedProject();
```
AlexanderSher: Do you need similar change for `GetActiveProject`?

MikhailArkhipov: No at the moment

AlexanderSher:
:+1:

PR #1952, opened 2016-06-16
Fixes #1932 this change passes the selected environment in Variable E…

### LINKED ISSUE #1932: Deleting variables always tries to delete them in current active frame
```r
x <- 1
f <- function() {
  browser()
}
f()
```
Run it, then switch VE to global environment, and try to delete `x`. It doesn't work, because delete sends `rm(x)` to the REPL, but that executes in the current active environment (which is the function), rather than the environment where the variable lives.

karthiknadig:
Fixes #1932 this change passes the selected environment in Variable Explorer to rm()


MikhailArkhipov:
👍

huguesv:
:+1:

PR #1953, opened 2016-06-16
Additional F1 Web search options

MikhailArkhipov:
Per users request

### DIFF
--- a/src/Package/Impl/Help/SearchWebForCurrentCommand.cs
+++ b/src/Package/Impl/Help/SearchWebForCurrentCommand.cs
@@ -33,7 +33,7 @@ namespace Microsoft.VisualStudio.R.Package.Help {
         protected override void Handle(string item) {
             // Bing: search?q=item+site%3Astackoverflow.com
             var encoded = HttpUtility.HtmlEncode(item);
-            var search = "http://" + Invariant($"www.bing.com/search?q={encoded}+R+site%3A{RToolsSettings.Current.HelpSearchSite}");
+            var search = "http://" + Invariant($"www.bing.com/search?q={encoded}+R+site%3A{RToolsSettings.Current.WebHelpSearchString}");
             var browser = VsAppShell.Current.GetGlobalService<IVsWebBrowsingService>(typeof(SVsWebBrowsingService));
             IVsWindowFrame frame;
             browser.Navigate(search, (uint)__VSWBNAVIGATEFLAGS.VSNWB_WebURLOnly, out frame);
--- a/src/Package/Impl/Help/SearchWebForCurrentCommand.cs
+++ b/src/Package/Impl/Help/SearchWebForCurrentCommand.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Diagnostics;
 using System.Text;
-using System.Web;
 using Microsoft.R.Components.InteractiveWorkflow;
 using Microsoft.R.Support.Settings;
 using Microsoft.R.Support.Settings.Definitions;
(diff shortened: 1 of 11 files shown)


src/Package/Impl/Help/SearchWebForCurrentCommand.cs
```diff
@@ -32,11 +36,26 @@ internal sealed class SearchWebForCurrentCommand : HelpOnCurrentCommandBase {
...
             // Bing: search?q=item+site%3Astackoverflow.com
-            var encoded = HttpUtility.HtmlEncode(item);
-            var search = "http://" + Invariant($"www.bing.com/search?q={encoded}+R+site%3A{RToolsSettings.Current.HelpSearchSite}");
-            var browser = VsAppShell.Current.GetGlobalService<IVsWebBrowsingService>(typeof(SVsWebBrowsingService));
-            IVsWindowFrame frame;
-            browser.Navigate(search, (uint)__VSWBNAVIGATEFLAGS.VSNWB_WebURLOnly, out frame);
+            var tokens = RToolsSettings.Current.WebHelpSearchString.Split(new char[] { ' ', '\t' }, StringSplitOptions.RemoveEmptyEntries);
+            bool siteSet = false;
+
+            var sb = new StringBuilder("http://" + Invariant($"www.bing.com/search?q={HttpUtility.HtmlEncode(item)}"));
```
int19h: This should be `UrlEncode`. `HtmlEncode` will do things like `>` to `&gt;`, which will not be correct for an URI.

int19h: Better yet, `Uri.EscapeUriString`.

src/Package/Impl/Help/SearchWebForCurrentCommand.cs
```diff
@@ -32,11 +36,26 @@ internal sealed class SearchWebForCurrentCommand : HelpOnCurrentCommandBase {
...
+            foreach (var t in tokens) {
+                sb.Append('+');
+                if (!siteSet && t.IndexOf('.') > 0) {
+                    sb.Append("site:");
+                    siteSet = true;
+                }
+                sb.Append(t);
```
int19h: `t` will need to be escaped here, too.

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -104,7 +104,8 @@ internal sealed class RToolsSettingsImplementation : IRToolsSettings {
         public string SurveyNewsFeedUrl { get; set; } = SurveyNewsUrls.Feed;
         public string SurveyNewsIndexUrl { get; set; } = SurveyNewsUrls.Index;
         public bool EvaluateActiveBindings { get; set; } = true;
-        public string HelpSearchSite { get; set; } = "stackoverflow.com";
+        public string WebHelpSearchString { get; set; } = "R stackoverflow.com";
```
int19h: Should it be `R site:stackoverflow.com`?

src/Package/Impl/Help/SearchWebForCurrentCommand.cs
```diff
@@ -32,11 +36,26 @@ internal sealed class SearchWebForCurrentCommand : HelpOnCurrentCommandBase {
...
-            IVsWindowFrame frame;
-            browser.Navigate(search, (uint)__VSWBNAVIGATEFLAGS.VSNWB_WebURLOnly, out frame);
+            var tokens = RToolsSettings.Current.WebHelpSearchString.Split(new char[] { ' ', '\t' }, StringSplitOptions.RemoveEmptyEntries);
+            bool siteSet = false;
+
+            var sb = new StringBuilder("http://" + Invariant($"www.bing.com/search?q={HttpUtility.HtmlEncode(item)}"));
+            foreach (var t in tokens) {
+                sb.Append('+');
+                if (!siteSet && t.IndexOf('.') > 0) {
+                    sb.Append("site:");
```
int19h: I'd rather let them specify it explicitly if they want site: or not. Not everything that contains a dot is necessarily a site.

int19h: (As in, just use the tokens as is, without trying to guess what's what. If they want site:foo.bar, they can put it there)

src/Package/Impl/Help/SearchWebForCurrentCommand.cs
MikhailArkhipov: Yeah I thought about this but was not sure... OK then

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
MikhailArkhipov: No, token with dot auto-become site: parameter

src/Package/Impl/Help/SearchWebForCurrentCommand.cs
MikhailArkhipov: OK, makes it simpler

PR #1954, opened 2016-06-16
Fix #1938: REPL steals focus from editor after breaking in debugger

### LINKED ISSUE #1938: REPL steals focus from editor after breaking in debugger
After breaking in debugger, REPL will _repeatedly_ steal focus from the editor, with no ability to override - i.e. if you click on the editor to focus it, the REPL will immediately steal the focus back. It should only steal it once, when the break occurs.

int19h:
Move focus from editor back to REPL no more than once per debugger break mode change.


MikhailArkhipov:
:+1:

PR #1956, opened 2016-06-16
Fix #1916, #1955

### LINKED ISSUE #1916: Data grid sorting - clicking anywhere on column header should work
Right now we only activate sorting if you click on the column name label itself.  It should work in the whitespace around the name as well.

On behalf of user and myself, who both made the same mistake and thought sorting wasn't working.

### LINKED ISSUE #1955: Character columns cannot be sorted in descending order
```r
m <- matrix("foo")
View(m)
```
Then click on the column twice to sort in descending order. The following error appears in the REPL:
```
Error in -x[, 1] : invalid argument to unary operator
```
and it becomes unusable thereafter.

int19h:
Fix #1916: Data grid sorting - clicking anywhere on column header should work

Accept clicks anywhere within the boundaries of the header cell for sorting purposes.

Fix #1955: Character columns cannot be sorted in descending order

Use xtfrm() to convert non-numeric columns to sort-equivalent numeric sequences for sorting purposes. Gracefully handle errors when column cannot be sorted.

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/grid.R
+++ b/src/Host/Client/Impl/rtvs/R/grid.R
@@ -4,6 +4,30 @@
 grid_header_format <- function(x)
 	if (is.na(x)) NULL else format(x)
 
+# Given a 2D data object and indices of columns, returns a vector of row indices indicating the appropriate order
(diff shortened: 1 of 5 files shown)


MikhailArkhipov:
:+1:

PR #1963, opened 2016-06-16
Fixes #1959 This change detaches RSession Mutated event from VariableGridHost

### LINKED ISSUE #1959: Closed grids still attempt to refresh
It seems that closing the grid window simply hides it, but it is still fetching data for the associated background on every session mutation - can be observed in debugger and by watching the logs. For long-running sessions with lots of variables and heavy use of grids, this could result in a significant slowdown.

karthiknadig:
Detaches RSession Mutated event from VariableGridHost when VariableGridWindowPane is closed. Previously the event handler for RSession Mutated in VariableGridHost re-evaluated the expression even after closing the view window.

### DIFF
--- a/src/Package/Impl/DataInspect/VariableGridWindowPane.cs
+++ b/src/Package/Impl/DataInspect/VariableGridWindowPane.cs
@@ -24,5 +24,10 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
             }
             _gridHost.SetEvaluation(evaluation);
         }
+
+        protected override void OnClose() {
+            base.OnClose();
+            _gridHost.CleanUp();
+        }
     }
 }
--- a/src/Package/Impl/DataInspect/VariableGridHost.xaml.cs
+++ b/src/Package/Impl/DataInspect/VariableGridHost.xaml.cs
@@ -31,6 +31,10 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
             _rSession.Mutated += RSession_Mutated;
         }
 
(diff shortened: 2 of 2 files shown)


src/Package/Impl/DataInspect/VariableGridWindowPane.cs
```diff
@@ -24,5 +24,10 @@ internal sealed class VariableGridWindowPane : RToolWindowPane {
             }
             _gridHost.SetEvaluation(evaluation);
         }
+
+        protected override void OnClose() {
+            base.OnClose();
+            _gridHost.CleanUp();
```
AlexanderSher: Please verify that this code is called when user closes the tab and is not called when user toggles between tabs

AlexanderSher: Also verify detaching the tab.

---
In reply to: [67448357](https://github.com/Microsoft/RTVS/pull/1963#discussion_r67448357) [](ancestors = 67448357)

MikhailArkhipov: And when window gets hidden b/c layout changes to debug when debugging starts

karthiknadig: I verified that, detaching details view tab, toggling between tabs, and layout changes when debugging starts does not call OnClose.

AlexanderSher:
:+1:

int19h:
:+1:

Issue #1961, opened 2016-06-16
Searching with F1 or Ctrl+F1 doesn't fire when caret is at the end of a word


MikhailArkhipov:
```read.csv| ``` <- we need to include end

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f1d0a4d931aa23970a69aba66b8f4b6748449c20

PR #1964, opened 2016-06-17
#1961 

MikhailArkhipov:
Searching with F1 or Ctrl+F1 doesn't fire when caret is at the end of a word.

Also switching off couple of tests until I figure out proper mocking behavior of the snippet session.


src/Mocks/VisualStudio/VsEnumStreamMarkersMock.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        public int GetCount(out int pcMarkers) {
+            pcMarkers = _markers.Length;
+            return VSConstants.S_OK;
+        }
+
+        public int Next(out IVsTextStreamMarker ppRetval) {
+            ppRetval = null;
+            if (_index >= 0 && _index < _markers.Length) {
```
AlexanderSher: Is it possible that `_index` is less than 0?

MikhailArkhipov: Not really

MikhailArkhipov: Unless it overflows somehow

AlexanderSher:
:+1:

Issue #1966, opened 2016-06-17
Make window for Shiny or use VS browser for Shiny output


MikhailArkhipov:
With various options.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ebc0cfce5a33964701da07e7c6c5e9047cfdaf35

int19h:
Verified.

PR #1969, opened 2016-06-17
#1966 Make window for Shiny or use VS browser for Shiny output

MikhailArkhipov:


src/Package/Impl/SurveyNews/SurveyNewsService.cs
```diff
@@ -91,9 +93,9 @@ internal class SurveyNewsService : ISurveyNewsService {
             try {
                 if (!string.IsNullOrEmpty(url)) {
                     if (forceCheck) {
-                        _browserLauncher.Navigate(url);
+                        _browserServices.OpenBrowser(WebBrowserRole.News, url);
                     } else {
-                        _browserLauncher.NavigateOnIdle(url);
+                        _browserServices.OpenBrowser(WebBrowserRole.News, url, onIdle: true);
```
AlexanderSher: you can just make it
```
_browserServices.OpenBrowser(WebBrowserRole.News, url, onIdle: !forceCheck);
```

AlexanderSher:
:shipit:

Issue #1967, opened 2016-06-17
Shiny is installed but does not appear in library() completion


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/16159454/b591f76e-3477-11e6-8a79-31ab722d37ad.png)

MikhailArkhipov:
Looks like completion is looking at the top level folder rather than under versioned folder

![image](https://cloud.githubusercontent.com/assets/12820357/16159579/7f552760-3478-11e6-82af-f6f473f08753.png)

MikhailArkhipov:
Specific to R Client.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/eb0e7a8b36a4bbd30734e999eec37cdcd6ce008c

int19h:
Verified.

PR #1972, opened 2016-06-17
#1967 Shiny is installed but does not appear in library() completion

MikhailArkhipov:


src/R/Support/Impl/Help/Packages/UserPackagesCollection.cs
```diff
@@ -33,8 +35,9 @@ public sealed class UserPackagesCollection : PackageCollection {
                     } else {
                         // Multiple library folders. Try and match to R version
                         // specified in the Tools | Options
-                        string version = GetReducedVersion();
-                        string path = Path.Combine(libraryPath, version);
+                        var version = RInstallation.GetRVersion(RToolsSettings.Current.RBasePath);
+                        var versionString = Invariant($"{version.Major}.{version.Minor}");
```
AlexanderSher: Does it work for developer version?

MikhailArkhipov: Yes, folders are 3.4


image

MikhailArkhipov: In RTM we'll go asking R rather than figuring this manually. But at least we not get version # from file rather than from disk path.

Issue #1851, opened 2016-06-06
Solution Explorer should not show temporary files


karthiknadig:
![image](https://cloud.githubusercontent.com/assets/3840081/15839913/2f8d32de-2bfb-11e6-910a-d065c009e038.png)

huguesv:
Similar to https://github.com/Microsoft/RTVS/issues/1307.

MikhailArkhipov:
How do we know it is temporary? Project shows everything that is on disk.

huguesv:
Look at the screenshot.  The icon shows the file as missing.  That shouldn't happen, it should have been removed from solution explorer.

MikhailArkhipov:
So it is not the issue with 'temp' file but rather than file was deleted but is still visible.

karthiknadig:
The file got created when i first did File > New Project > Rproject. I assumed it was a temp file since it  was deleted automatically.

AlexanderSher:
Any additional information? I can't repro it on any of my machines.

Marchen:
I'm also having this issue but I can't repro... It seems that if I use RTVS longer time, number of the temporary files increases.

MikhailArkhipov:
Found it. Happens when we get file names in 8.3 format.

```
 	Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.dll!Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.IO.MsBuildFileSystemWatcherEntries.DeleteDirectory(string relativePath) Line 58	C#
 	Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.dll!Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.IO.MsBuildFileSystemWatcher.DirectoryDeleted.Apply() Line 29	C#
>	Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.dll!Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.IO.MsBuildFileSystemWatcher.Consume() Line 146	C#
 	Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.dll!Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.IO.MsBuildFileSystemWatcher.ConsumeWaitPublish() Line 115	C#
```

![image](https://cloud.githubusercontent.com/assets/12820357/16179002/54009810-360d-11e6-8b7c-8b81c53c1b0f.png)

[...]

PR #1975, opened 2016-06-20
#1851 Solution Explorer should not show temporary files

MikhailArkhipov:
Maybe there is a better fix... Problem is that short path must be stored since after item is deleted, GetShortFilePath and GetLongFilePath no longer work. Alternative is to pass full path and root down to AddEntry...

### DIFF
--- a/src/ProjectSystem/Impl/Utilities/PathExtensions.cs
+++ b/src/ProjectSystem/Impl/Utilities/PathExtensions.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.IO;
+using System.Text;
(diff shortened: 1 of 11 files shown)


src/ProjectSystem/Impl/Utilities/PathExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using System.IO;
+using System.Text;
+using Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.Interop;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
+
+namespace Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.Utilities {
+    public static class PathExtensions {
+        public static string ToLongPath(this string path) {
```
AlexanderSher: I think we can make this part of `IFileSystem` interface (since it's not just an extension method and in fact requires access to file system)

AlexanderSher:
:+1:

Issue #1978, opened 2016-06-20
Error when loading existing project


huguesv:
This is on Dev15 Exp, I haven't tried Dev14.  I have the latest sources.

I got this error message when loading a project:
```
---------------------------
Microsoft Visual Studio
---------------------------
The project system has encountered an error.

 

An exception was thrown while initializing part "Microsoft.VisualStudio.R.Package.Browsers.WebBrowserServices".



A diagnostic log has been written to the following location: "C:\Users\huvalo\AppData\Local\Temp\VsProjectFault_babc54a4-08cd-43ea-9071-806472314ebb.failure.txt".


---------------------------
OK   
---------------------------

```

And the contents of the referenced file is:
```
=====================
6/20/2016 1:34:43 PM
Recoverable
Microsoft.VisualStudio.Composition.CompositionFailedException: An exception was thrown while initializing part "Microsoft.VisualStudio.R.Package.Browsers.WebBrowserServices". ---> System.Runtime.InteropServices.COMException: VsEditorShell constructor must be called on the UI thread.
   at Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread(String callerMemberName)
   at Microsoft.VisualStudio.R.Package.Shell.VsAppShell..ctor()
   at Microsoft.VisualStudio.R.Package.Shell.VsAppShell.<>c.<.cctor>b__62_0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.VisualStudio.R.Package.Shell.VsAppShell.get_Current()
   at Microsoft.VisualStudio.R.Package.Browsers.WebBrowserServices..ctor()
   --- End of inner exception stack trace ---
   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.RuntimePartLifecycleTracker.CreateValue()
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.Create()
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveNext(PartLifecycleState nextState)
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToState(PartLifecycleState requiredState)
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.GetValueReadyToExpose()
[...]

PR #1979, opened 2016-06-20
#1978

MikhailArkhipov:
Avoid using VsAppShell in ctor

### DIFF
--- a/src/Package/Impl/Browsers/WebBrowserServices.cs
+++ b/src/Package/Impl/Browsers/WebBrowserServices.cs
@@ -14,13 +14,14 @@ using Microsoft.VisualStudio.Shell.Interop;
 namespace Microsoft.VisualStudio.R.Package.Browsers {
     [Export(typeof(IWebBrowserServices))]
     internal class WebBrowserServices : IWebBrowserServices {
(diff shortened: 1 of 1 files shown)


src/Package/Impl/Browsers/WebBrowserServices.cs
```diff
@@ -14,13 +14,14 @@
...
         private readonly IProcessServices _ps;
         private readonly IRToolsSettings _settings;
 
+        private readonly IVsWebBrowsingService _wbs;
+        private IVsWebBrowsingService WebBrowserService => _wbs ?? VsAppShell.Current.GetGlobalService<IVsWebBrowsingService>(typeof(SVsWebBrowsingService));
+
         public WebBrowserServices() : 
-            this(VsAppShell.Current.GetGlobalService<IVsWebBrowsingService>(typeof(SVsWebBrowsingService)),
-                 ProcessServices.Current, RToolsSettings.Current) {
+            this(null, ProcessServices.Current, RToolsSettings.Current) {
```
AlexanderSher: Ugly, but I think it works for a quick fix

AlexanderSher:
:+1:

Issue #1976, opened 2016-06-20
vignette's aren't visible in dark color theme


akzaidi:
Vignette fonts are not visible in R Help Pane when using a dark color theme:
<img width="456" alt="vignetteissue" src="https://cloud.githubusercontent.com/assets/1307603/16197839/79a6a1d8-36fb-11e6-8abd-adf61bfbcf90.PNG">

What's the best solution here? Since vigenttes and object help share the same R Help window, I assume they are both going to have same font negative operation applied to them when they're used with a dark color theme, but since vignettes are a bit more complicated this doesn't always work out, as above.

Thanks!

MikhailArkhipov:
Pane is Web browser and is styled via CSS files, there is no font or color reversal. Styling files are in Program Files\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\R Tools for Visual Studio\Themes.

MikhailArkhipov:
Which specific search leads to this page? Thanks

MikhailArkhipov:
So it appears that vignettes include HTML encoded CSS that overrides theming 

```
[...]

PR #1980, opened 2016-06-21
#1976 vignette's aren't visible in dark color theme

MikhailArkhipov:
image


MikhailArkhipov:
Tests incoming, can't access my desktop with the test code at the moment.

src/Package/Impl/Help/HelpVisualComponent.cs
```diff
@@ -38,10 +38,13 @@ internal sealed class HelpVisualComponent : IHelpVisualComponent {
         /// unchanged and only replace browser that is inside it.
         /// </summary>
         private readonly ContentControl _windowContentControl;
+        private readonly IVignetteCodeColorBuilder _codeColorBuilder;
         private IRSession _session;
         private WindowsFormsHost _host;
 
         public HelpVisualComponent() {
+            _codeColorBuilder = VsAppShell.Current.ExportProvider.GetExportedValue<IVignetteCodeColorBuilder>();
```
AlexanderSher: Can you pass `ICoreShell` as a parameter?

MikhailArkhipov: sure, just tried to avoid too much refactoring since this may land in 0.4

src/Package/Impl/Help/HelpVisualComponent.cs
```diff
@@ -156,7 +206,11 @@ internal sealed class HelpVisualComponent : IHelpVisualComponent {
         }
 
         private void OnNavigating(object sender, WebBrowserNavigatingEventArgs e) {
-            if (Browser.Document != null && Browser.Document.Window != null) {
+            // Disconnect browser from the tool window so it does not
+            // flicker when we change page and element styling.
+            _host.Child = null;
```
AlexanderSher: Do we need to check `_host` for null?

src/Package/Impl/Help/HelpVisualComponent.cs
MikhailArkhipov: Actually I can't. Tests create components via reflection 

```
        public ControlTestScript(Type type) {
            ControlWindow.Create(type);
        }
```

We probably should change this and pass some sort of initialization context

src/Package/Impl/Help/VignetteCodeColorBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        public string GetCodeColorsCss() {
+            var sb = new StringBuilder();
+            var workflow = _workflowProvider.GetOrCreate();
+            if (workflow.ActiveWindow?.TextView != null) {
+                var map = _formatMapService.GetClassificationFormatMap(workflow.ActiveWindow.TextView);
+                foreach (var cssCodeProp in _cssPropertyMap) {
+                    var props = map.GetTextProperties(_classificationRegistryService.GetClassificationType(cssCodeProp.ClassificationTypeName));
+
+                    sb.AppendLine(Invariant($"code > span.{cssCodeProp.CssClassName} {{"));
```
AlexanderSher: So these 4 lines can be combined into one:
```{csharp}
$@"code > span.{cssCodeProp.CssClassName} {{
    color: {CssColorFromBrush(props.ForegroundBrush)};
}}
";
```
I think it will be easier to read this way

src/Package/Impl/Help/VignetteCodeColorBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+                    sb.AppendLine(Invariant($"color: {CssColorFromBrush(props.ForegroundBrush)};"));
+                    sb.AppendLine("}");
+                    sb.AppendLine(string.Empty);
+                }
+            }
+            return sb.ToString();
+        }
+
+        private string CssColorFromBrush(System.Windows.Media.Brush brush) {
+            var sb = brush as System.Windows.Media.SolidColorBrush;
```
AlexanderSher: `using System.Windows.Media`?

src/Package/Impl/Help/HelpVisualComponent.cs
AlexanderSher: Ok, I'll look into it

src/Package/Impl/Help/HelpVisualComponent.cs
MikhailArkhipov: Probably not. If we created Browser, we created host and we never set it to null

AlexanderSher: Ok

src/Package/Impl/Help/VignetteCodeColorBuilder.cs
MikhailArkhipov: It conflicts with System.Drawing unfortunately. Help window is Windows.Forms browser

AlexanderSher:
:+1:

PR #1984, opened 2016-06-21
VsAppShell fixes

AlexanderSher:
- Added explicit `EnsureInitialize` method that is called from `RPackage`
- Added exports for `ICoreShell` and `IEditorShell`
- Removed initialization code from `VsAppShell` constructor so that it can be accessed inside constructors from any thread
- Removed `IsUITestEnvironment` that isn't used


src/Languages/Editor/Impl/Shell/IEditorShell.cs
```diff
@@ -45,11 +45,6 @@ public interface IEditorShell: ICoreShell {
         bool IsUnitTestEnvironment { get; }
 
         /// <summary>
-        /// Tells if code runs in UI test environment
-        /// </summary>
-        bool IsUITestEnvironment { get; }
-
```
MikhailArkhipov: It was intended for VS-based tests originally. May come handy some day.

src/Package/Impl/Shell/VsAppShell.cs
```diff
@@ -89,6 +99,16 @@ public sealed class VsAppShell : IApplicationShell, IMainThread, IIdleTimeServic
             }
         }
 
+        private static VsAppShell GetInstance() {
```
MikhailArkhipov: So basically code can import it, but it is useless until package calls Initialize right? Maybe then at least set CompositionService and ExportProvider in GetInstance?

AlexanderSher: Agreed

MikhailArkhipov: Perhaps we also want assert if someone calls any method on non-initialized objects.

AlexanderSher: It will fail anyway (with NRE), but I can add this verification

MikhailArkhipov:
👍

int19h:
:+1:

PR #1988, opened 2016-06-21
More null checks

MikhailArkhipov:


src/Package/Impl/Help/HelpVisualComponent.cs
```diff
@@ -238,15 +238,18 @@ internal sealed class HelpVisualComponent : IHelpVisualComponent {
             // Refresh button clicked. Current document state is 'complete'.
             // We need to delay until it changes to 'loading' and then
             // delay again until it changes again to 'complete'.
-            Browser.Document.Window.Unload -= OnWindowUnload;
+            var window = Browser?.Document?.Window;
```
AlexanderSher: You can extract it into method

MikhailArkhipov: Sure lemme do that

Issue #1989, opened 2016-06-21
Error pasting in r markdown


huguesv:
Right click on project, Add R Markdown
CTRL-A to select all
Put this in your clipboard:
```
`left_join`, `right_join`, `inner_join`
```

CTRL-V to paste

Error:

```
---------------------------
Microsoft Visual Studio
---------------------------
Attempted to introduce overlapping source spans.
---------------------------
OK   
---------------------------
```

PR #1991, opened 2016-06-22
#1989 Error pasting in r markdown

MikhailArkhipov:
Effectively forgotten change. Edge tracking in all spans must match otherwise paste over will grow span in one buffer and shrink in another introducing overlapped spans which is bad.


AlexanderSher:
:+1:

PR #1994, opened 2016-06-23
Improve protocol to simplify remoting

int19h:
Change protocol so that the type of every message (one-way notification, request, response) is encoded in the message name in a uniform way.

Rename associated APIs (send_message etc) for clarity.

### DIFF
--- a/src/Host/Client/Impl/Definitions/IRCallbacks.cs
+++ b/src/Host/Client/Impl/Definitions/IRCallbacks.cs
@@ -51,7 +51,7 @@ namespace Microsoft.R.Host.Client {
         /// </summary>
         /// <param name="url"></param>
         /// <returns></returns>
-        Task Browser(string url);
+        Task WebBrowser(string url);
 
         /// <summary>
         /// Invoked in response of parameter-less library call
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -131,7 +131,10 @@ namespace Microsoft.R.Host.Client {
             }
         }
 
-        private async Task<string> SendAsync(string name, CancellationToken ct, params object[] args) {
+        private async Task<string> NotifyAsync(string name, CancellationToken ct, params object[] args) {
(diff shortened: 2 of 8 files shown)


int19h:
Corresponding changes in R-Host coming shortly.

int19h:
The purpose of this is to enable the broker to figure out how to handle the packet by only looking at its name (so all custom messages sent from e.g. R code do not need to be special-cased). With the change, there are three types of messages:

Notifications: `!Foo`
Requests: `?Foo`
Responses: `:Foo`

So first character alone determines message type.

Also, I got rid of most special character names for the sake of clarity, keeping them only for those messages that constitute the bulk of the typical session: input prompts, evals, console output, and busy status reporting.

src/Host/Client/Impl/Definitions/IRCallbacks.cs
```diff
@@ -51,7 +51,7 @@ public interface IRCallbacks {
         /// </summary>
         /// <param name="url"></param>
         /// <returns></returns>
-        Task Browser(string url);
+        Task WebBrowser(string url);
```
int19h: This is to disambiguate from R debug browser.

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -131,18 +131,22 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
...
             return id;
         }
 
         private async Task<string> RespondAsync(Message request, CancellationToken ct, params object[] args) {
+            Debug.Assert(request.Name.StartsWithOrdinal("?"));
             TaskUtilities.AssertIsOnBackgroundThread();
 
             string id;
-            var message = CreateMessage(out id, ":", request.Id, request.Name, args);
+            var message = CreateMessage(out id, ":" + request.Name.Substring(1), request.Id, args);
```
MikhailArkhipov: Throw if Name is actually shorter than 1 and does not start with '?' ?

MikhailArkhipov:
:+1:

Issue #1996, opened 2016-06-23
Null reference exception when using code snippets for ">" and "=" in editor


jflam:
Repros on 2 machines with 2016-06-22.2 build. Editor window only. Works fine in REPL.

This is a release blocking bug for 0.4.

AlexanderSher:
Fails in `ExpansionClient.GetFieldMarkers()` on `ErrorHandler.ThrowOnFailure(Session.GetSnippetSpan(pts));` because `Session` is `null`

MikhailArkhipov:
That is probably b/c shippet does not actually have any fields. So null check is probably sufficient.

MikhailArkhipov:
Actually no.... This code is old, prev version had _expansionSession which is same thing.

MikhailArkhipov:
OK, old code had same bug but managed to get away since it didn't manually position caret in the first field.

PR #1998, opened 2016-06-23
#1996 Null reference exception when using code snippets for ">" and "…

MikhailArkhipov:
…=" in editor

### DIFF
--- a/src/Package/Impl/Expansions/ExpansionClient.cs
+++ b/src/Package/Impl/Expansions/ExpansionClient.cs
@@ -29,7 +29,6 @@ namespace Microsoft.VisualStudio.R.Package.Expansions {
         private IVsExpansionSession _expansionSession;
         private int _currentFieldIndex = -1;
         private int _markerCount;
-        private bool _earlyEndExpansionHappened = false;
 
         class Marker : TextRange {
             public IVsTextStreamMarker StreamMarker { get; }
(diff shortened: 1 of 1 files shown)


src/Package/Impl/Expansions/ExpansionClient.cs
```diff
@@ -208,15 +198,12 @@ class Marker : TextRange {
...
-
                     hr = expansion.InsertNamedExpansion(pszTitle, pszPath, ts.Value, this, RGuidList.RLanguageServiceGuid, 0, out _expansionSession);
-                    if (_earlyEndExpansionHappened) {
-                        // EndExpansion was called before InsertNamedExpansion returned, so set _expansionSession
-                        // to null to indicate that there is no active expansion session. This can occur when 
-                        // the snippet inserted doesn't have any expansion fields.
-                        _expansionSession = null;
-                        _earlyEndExpansionHappened = false;
+                    // If EndExpansion was called before InsertNamedExpansion returned, so set _expansionSession
+                    // to null to indicate that there is no active expansion session. This can occur when
```
AlexanderSher: Looks like this comment isn't correct - `_expansionSession` isn't set to `null`

AlexanderSher:
:+1:

src/Package/Impl/Expansions/ExpansionClient.cs
MikhailArkhipov: It is in EndExpansion

Issue #1130, opened 2016-02-29
ASCII progress bar for reporting progress in REPL


MikhailArkhipov:
This is an issue in the VS editor around cursor movement. Need to followup with VS.

MikhailArkhipov:
```
testit <- function(x = sort(runif(20)), style = 3, ...){
  pb <- txtProgressBar(..., style = style)
  for(i in c(0, x, 1)) {Sys.sleep(0.1); setTxtProgressBar(pb, i)}
  # Sys.sleep(1)
  close(pb)
}
testit()
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/65b2b8443bb3a313f8325d0722b98a1a0752dbd3

PR #2002, opened 2016-06-24
#1130 ASCII progress bar for reporting progress in REPL

MikhailArkhipov:


src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -203,12 +204,46 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
...
+                if (CurrentWindow != null) {
+                    // If message starts with CR we remember current output buffer
+                    // length so we can continue writing lines into the same spot.
+                    // See txtProgressBar in R.
+                    if (message.StartsWithOrdinal("\r")) {
+                        // Store the message and the initial position. All subsequent 
+                        // messages that start with CR. Will be written into the same place.
+                        _lastMessage = message.Substring(1);
+                        _lastPosition = _lastPosition < 0 ? CurrentWindow.OutputBuffer.CurrentSnapshot.Length : _lastPosition;
+                        message = "|"; // replacement placeholder so we can receive 'buffer changed' event
```
AlexanderSher: Why don't you want to use `\r` as a placeholder?

int19h: It seems that this won't be overwritten if someone just prints a lone `\r`.

MikhailArkhipov: Any placeholder can be used. I am not entirely sure what would happen on ```\r``` as placeholder since generally VS editor doesn't like mixed line endings. Lonely ```\r``` is uncommon (I think) since it is just CR without LF (i.e. print on the same line).

int19h: Lonely `\r` is sometimes used by console apps to simulate an input field, esp. masked. So for example you first output something like:
```
###-##-##
```
And then do a `readline` with `\r` as prompt. That way cursor is back at the beginning of the line, and #'s get overwritten by actual input as user is typing.

MikhailArkhipov: However... I probably need to check for ```\r\n```

MikhailArkhipov: Any example in R? I'd be happy to check if this 'workaround' ;-) works with that

int19h: But anyway, my point was that we shouldn't put some weird character there in this case. I think it's okay to just let `\r` be treated as a newline then, as REPL normally does. So the check for `StartsWithOrdinal` above should simply be extended with `.Length > 1`.

MikhailArkhipov:
Updated change is part of the Tests PR

Issue #2001, opened 2016-06-24
SetBreakpointWhileRunning never completes when running in a group


MikhailArkhipov:
R Host keeps running at 15% CPU. Break All lands in

```
>   Microsoft.R.Host.exe!std::_Wrap_alloc<std::allocator<std::_Container_proxy> >::destroy<std::_Container_proxy>(std::_Container_proxy * _Ptr) Line 900    C++
    Microsoft.R.Host.exe!std::_String_alloc<std::_String_base_types<char,std::allocator<char> > >::_Free_proxy() Line 654   C++
    Microsoft.R.Host.exe!std::_String_alloc<std::_String_base_types<char,std::allocator<char> > >::~_String_alloc<std::_String_base_types<char,std::allocator<char> > >() Line 616  C++
    Microsoft.R.Host.exe!std::basic_string<char,std::char_traits<char>,std::allocator<char> >::~basic_string<char,std::char_traits<char>,std::allocator<char> >() Line 1013 C++
    Microsoft.R.Host.exe!rhost::host::message::~message()   C++
    Microsoft.R.Host.exe!CallBack() Line 550    C++
    R.dll!000000006c71f092()    Unknown
```

```
[...]

PR #2004, opened 2016-06-24
Tests stability

MikhailArkhipov:
Apart from https://github.com/Microsoft/RTVS/issues/2001 All tests pass consistently together on my two machines. Changes:

- Make sure editor tree and documents are disposed of. Otherwise they stay attached to idle and throw when called on wrong thread.
- Idle event can specify which thread to fire at.
- Some tests had hardcoded R 3.2, Now use RInstallation
- RInstallation static didn't work well since registry mocks affected tests that expected real registry when running in parallel
- Export provider should not be disposed since tests dispose it before WPF window is fully closed and all view and editor components are disposed of.


src/Common/Core/Impl/IO/FileSystem.cs
```diff
@@ -8,6 +8,9 @@
 
 namespace Microsoft.Common.Core.IO {
     public sealed class FileSystem : IFileSystem {
+        private static FileSystem _instance = new FileSystem();
+        public static IFileSystem Current => _instance;
```
AlexanderSher: Let's not introduce more singletons.
This class is lightweight and can be instantiated even when it isn't accessible from MEF

src/Host/Client/Test/ExecutionTracing/BreakpointsTest.cs
```diff
@@ -18,6 +18,8 @@
 
 namespace Microsoft.R.ExecutionTracing.Test {
     [ExcludeFromCodeCoverage]
+    [Category.R.ExecutionTracing]
+    [Collection(CollectionNames.NonParallel)]
```
AlexanderSher: It is important for these tests to run in parallel - that simulates multiple instances of RHost running concurrently

src/R/Components/Test/History/RHistoryIntegrationTest.cs
```diff
@@ -158,9 +158,5 @@ public class RHistoryIntegrationTest : IDisposable {
                 text.Should().Be(string.Empty);
             }
         }
-
-        public void Dispose() {
-            (_exportProvider as IDisposable)?.Dispose();
```
AlexanderSher: What is this change for?

src/R/Components/Test/MefCompositionTests.cs
```diff
@@ -56,9 +56,5 @@ public class MefCompositionTests : IDisposable {
         public void InteractiveWindowFactoryService() {
             _exportProvider.GetExportedValue<IInteractiveWindowFactoryService>().Should().NotBeNull();
         }
-
-        public void Dispose() {
-            (_exportProvider as IDisposable)?.Dispose();
```
AlexanderSher: Disposing `ExportProvider` helps detecting static caches of MEF components, which should be considered an anti-pattern (MEF can be forced to rebuild its instances, while static cache will have no idea about it).

src/Common/Core/Impl/IO/FileSystem.cs
MikhailArkhipov: OK. It was for convenience only, but yeah, we can just new it.

src/R/Components/Test/MefCompositionTests.cs
MikhailArkhipov: We need to figure out then when to dispose it. disposing it here causes core editor to throw since although from our point of view the control is closed, it is still getting some code executed (I guess dispatcher frames age getting pushed). I'll see if I can find safe place to disposed of the container.

src/R/Components/Test/History/RHistoryIntegrationTest.cs
MikhailArkhipov: See the other comment wrt export provider disposal

src/Host/Client/Test/ExecutionTracing/BreakpointsTest.cs
MikhailArkhipov: OK, it was attempt to make debugger breakpoints setting tests pass but this they still fail, i can put it back.

src/R/Components/Test/MefCompositionTests.cs
MikhailArkhipov: OK, restored. couple throw in debug, but it doesnt seem to affect the result

src/Common/Core/Impl/Shell/IIdleTimeService.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 
+using System.Threading;
+
 namespace Microsoft.Languages.Editor.Host {
     public interface IIdleTimeService {
-        void DoIdle();
+        void DoIdle(Thread thread = null);
```
AlexanderSher: Can you explain this change?

MikhailArkhipov: When tests run in parallel while EditorShell is a singleton, they may get events fired from a wrong thread. In most cases it doesn't matter though, but I've seen some fail occasionally. However, with changes where editor objects are disposed properly, this may not be that necessary.

src/R/Components/Impl/History/Implementation/RHistoryProvider.cs
```diff
@@ -16,18 +16,29 @@ namespace Microsoft.R.Components.History.Implementation {
...
-        private readonly IFileSystem _fileSystem;
         private readonly IEditorOperationsFactoryService _editorOperationsFactory;
         private readonly IRtfBuilderService _rtfBuilderService;
         private readonly ITextSearchService2 _textSearchService;
         private readonly IRSettings _settings;
         private readonly IContentTypeRegistryService _contentTypeRegistryService;
         private readonly Dictionary<ITextBuffer, IRHistory> _histories;
 
+        private IFileSystem _fileSystem;
+        internal IFileSystem FileSystem {
```
AlexanderSher: Where this change is used?

src/Languages/Editor/Impl/Shell/IEditorShell.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Threading;
```
AlexanderSher: You can undo this change

src/Host/Client/Test/ExecutionTracing/SteppingTest.cs
```diff
@@ -20,6 +19,8 @@
 
 namespace Microsoft.R.ExecutionTracing.Test {
     [ExcludeFromCodeCoverage]
+    [Category.R.ExecutionTracing]
+    [Collection(CollectionNames.NonParallel)]
```
AlexanderSher: Do you need NonParallel here?

src/Host/Client/Impl/Definitions/PlotMessage.cs
```diff
@@ -17,11 +17,11 @@ public struct PlotMessage {
 
         public bool IsClearAll => string.IsNullOrEmpty(FilePath);
 
-        public bool IsPlot => new FileInfo(FilePath).Length > 0;
+        public bool IsPlot => File.Exists(FilePath) && new FileInfo(FilePath).Length > 0;
```
AlexanderSher: Technically, file can exists during first check and be deleted during second change. We need to wrap `new FileInfo(FilePath).Length > 0` into try/catch

src/R/Components/Impl/History/Implementation/RHistoryProvider.cs
MikhailArkhipov: Added so substitute could be provided as in other places, but if you think it won't be needed we can nix it

AlexanderSher: Is it required in any tests right now?

---
In reply to: [68613538](https://github.com/Microsoft/RTVS/pull/2004#discussion_r68613538) [](ancestors = 68613538)

AlexanderSher: AFAIK, in RHistory tests, IFileSystem implementation is provided through MEF, so you can leave it in ctor but mark parameter with AllowDefault

---
In reply to: [68614367](https://github.com/Microsoft/RTVS/pull/2004#discussion_r68614367) [](ancestors = 68614367,68613538)

MikhailArkhipov: IFileSystem is no longer exported

src/Host/Client/Test/ExecutionTracing/SteppingTest.cs
MikhailArkhipov: BreakContinue is failing unless I put it here.

AlexanderSher: Can you please create a bug and assign it to @int19h ? This shouldn't happen.

src/R/Components/Impl/History/Implementation/RHistoryProvider.cs
AlexanderSher: In R.Components.Tests it is explicitly added to the container:
```{csharp}
        protected override void AddValues(CompositionContainer container) {
            base.AddValues(container);
            var coreShell = new TestCoreShell(container);
            var batch = new CompositionBatch()
                .AddValue(FileSystemStubFactory.CreateDefault())
                .AddValue<IRSettings>(RSettingsStubFactory.CreateForExistingRPath())
                .AddValue<ICoreShell>(coreShell)
                .AddValue(coreShell);
            container.Compose(batch);
        }
```

MikhailArkhipov: I guess it should be removed from the container then. Frankly I'd rather have containers use actual exported values rather than manually add stuff since it makes container != runtime container.

AlexanderSher: They aren't identical. Runtime container provides more exports and may provide them in a different time periods.
Injecting `IFileSystem` in constructor simplifies test code. For example, `RInteractiveWorkflow` calls `CreateRHistory` right in constructor, so we have to explicitly call `_exportProvider.GetExportedValue<IRHistoryProvider>().FileSystem = FileSystemStubFactory.CreateDefault()` in test constructors.
I don't see big difference in default value checking inside constructor or in the internal property. In both cases we know that it is null in production and not null in unit tests.

AlexanderSher:
:+1:

Issue #2023, opened 2016-06-28
Blob message support in RHost


karthiknadig:
Host can send messages that may contain blobs (e.g., plot image). Currently we create files in the RHost process and send the file name back and read from it in RTVS.

PR #2035, opened 2016-07-01
#2023  Host can send binary blobs.

karthiknadig:
#2023  Host can send binary blobs. Plots updated to use blobs. Added =r (raw) eval kind. First element in the message json now has the header [id, name, blob_count, <optional>request_id].

### DIFF
--- a/src/Host/Client/Impl/Definitions/IMessageTransport.cs
+++ b/src/Host/Client/Impl/Definitions/IMessageTransport.cs
@@ -9,5 +9,6 @@ namespace Microsoft.R.Host.Client {
     public interface IMessageTransport {
         Task SendAsync(string message, CancellationToken ct = default(CancellationToken));
         Task<string> ReceiveAsync(CancellationToken ct = default(CancellationToken));
+        Task<byte[]> ReceiveRawAsync(CancellationToken ct = default(CancellationToken));
     }
 }
--- a/src/Host/Client/Impl/Definitions/PlotMessage.cs
+++ b/src/Host/Client/Impl/Definitions/PlotMessage.cs
@@ -25,6 +25,6 @@ namespace Microsoft.R.Host.Client {
         /// <summary>
         /// The plot could not be rendered id length is 0.
         /// </summary>
-        public bool IsError => (Data == null) || Data.Length == 0;
+        public bool IsError => Data.Length == 0;
     }
 }
--- a/src/Host/Client/Impl/Definitions/PlotMessage.cs
+++ b/src/Host/Client/Impl/Definitions/PlotMessage.cs
@@ -9,27 +9,22 @@ namespace Microsoft.R.Host.Client {
         public string FilePath { get; }
         public int ActivePlotIndex { get; }
         public int PlotCount { get; }
+        public byte[] Data { get; }
 
-        public PlotMessage(string filePath, int activePlotIndex, int plotCount) {
+        public PlotMessage(string filePath, int activePlotIndex, int plotCount, byte[] data) {
             FilePath = filePath;
             ActivePlotIndex = activePlotIndex;
             PlotCount = plotCount;
+            Data = data;
         }
 
         public bool IsClearAll => string.IsNullOrEmpty(FilePath);
 
-        public bool IsPlot {
-            get {
-                try {
-                    return new FileInfo(FilePath).Length > 0;
-                } catch (IOException) { } catch (AccessViolationException) { }
-                return false;
-            }
-        }
+        public bool IsPlot => Data.Length > 0;
 
         /// <summary>
-        /// A zero-sized file means a blank image, ie. the plot could not be rendered.
+        /// The plot could not be rendered id length is 0.
         /// </summary>
-        public bool IsError => !File.Exists(FilePath) || new FileInfo(FilePath).Length == 0;
+        public bool IsError => (Data == null) || Data.Length == 0;
     }
 }
--- a/src/Host/Client/Impl/Definitions/REvaluationResult.cs
+++ b/src/Host/Client/Impl/Definitions/REvaluationResult.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
 using static System.FormattableString;
+using System.Collections.Generic;
 
 namespace Microsoft.R.Host.Client {
     /// <summary>
@@ -31,6 +32,14 @@ namespace Microsoft.R.Host.Client {
         /// </remarks>
         public JToken Result { get; }
         /// <summary>
+        /// Result of evaluation for 'raw'
+        /// </summary>
+        /// <remarks>
+        /// Computed by serializing the immediate result of evaluation, as if by <c>serialize(x,NULL)</c>
+        /// </remarks>
+        public List<byte[]> Raw { get; }
+
+        /// <summary>
         /// If evaluation failed because of an R runtime error, text of the error message.
         /// Otherwise, <see langword="null"/>.
         /// </summary>
@@ -46,11 +55,19 @@ namespace Microsoft.R.Host.Client {
             Error = error;
             ParseStatus = parseStatus;
             Result = null;
+            Raw = null;
         }
 
         public REvaluationResult(JToken result, string error, RParseStatus parseStatus)
             : this(error, parseStatus) {
             Result = result;
+            Raw = null;
+        }
+
+        public REvaluationResult(JToken result, string error, RParseStatus parseStatus, List<byte[]> raw)
+            : this(error, parseStatus) {
+            Result = result;
+            Raw = raw;
         }
 
         public override string ToString() {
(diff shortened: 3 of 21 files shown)


src/Host/Client/Impl/Definitions/PlotMessage.cs
```diff
@@ -9,27 +9,22 @@ public struct PlotMessage {
...
-            }
-        }
+        public bool IsPlot => Data.Length > 0;
 
         /// <summary>
-        /// A zero-sized file means a blank image, ie. the plot could not be rendered.
+        /// The plot could not be rendered id length is 0.
         /// </summary>
-        public bool IsError => !File.Exists(FilePath) || new FileInfo(FilePath).Length == 0;
+        public bool IsError => (Data == null) || Data.Length == 0;
```
AlexanderSher: In which case `Data == null`?

src/Host/Client/Impl/Definitions/REvaluationResult.cs
```diff
@@ -46,11 +55,19 @@ public struct REvaluationResult {
...
+            Raw = null;
         }
 
         public REvaluationResult(JToken result, string error, RParseStatus parseStatus)
             : this(error, parseStatus) {
             Result = result;
+            Raw = null;
+        }
+
+        public REvaluationResult(JToken result, string error, RParseStatus parseStatus, List<byte[]> raw)
```
AlexanderSher: You can have one single ctor with `List<byte[]> raw = null`

src/Host/Client/Impl/Host/RHost.Message.cs
```diff
@@ -12,32 +16,68 @@ private class Message {
...
+                    if (requestId.Type != JTokenType.String) {
+                        throw ProtocolError($"request_id must be {JTokenType.String}:", this);
+                    }
+                    RequestId = (string)requestId;
                 }
+
+                // header part is done
+                _argsOffset = 1;
+
+                Blobs = new ConcurrentQueue<byte[]>();
```
AlexanderSher: Why don't you want to pass `Blobs` as a parameter? `ReceiveMessageAsync

src/Host/Client/Test/RtvsPackage/BlobsTest.cs
```diff
@@ -0,0 +1,63 @@
...
+        [CompositeTest]
+        [Category.R.RtvsPackage]
+        [InlineData("serialize(1:10, NULL)", "null", new int[] { 62 })]
+        [InlineData("NULL", "null", null)]
+        public async Task Serialize(string expr, string json, int[] sizes) {
+            using (var eval = await _session.BeginEvaluationAsync()) {
+                var res = await eval.EvaluateAsync(expr, REvaluationKind.RawBytes);
+                var actualJson = JsonConvert.SerializeObject(res.Result).ToUnicodeQuotes();
+                actualJson.Should().Be(json);
+                if (sizes != null) {
```
AlexanderSher: Try to avoid conditional logic in tests. I'd rather split this test into two.

src/Host/Client/Test/RtvsPackage/BlobsTest.cs
```diff
@@ -0,0 +1,63 @@
...
+                    res.Raw.Should().NotBeNull();
+                    res.Raw.Count.Should().Be(sizes.Length);
```
AlexanderSher: `res.Raw.Should().Equal(sizes, (a, e) => a.Length == e)`

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -74,7 +75,20 @@ internal static class CsvAppFileIO {
             var dec = CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator;
 
             using (var e = await session.BeginEvaluationAsync()) {
-                await e.EvaluateAsync($"write.table({result.Expression}, qmethod='double', col.names=NA, file={file.ToRPath().ToRStringLiteral()}, sep={sep.ToRStringLiteral()}, dec={dec.ToRStringLiteral()})", REvaluationKind.Normal);
+                var fileinfo = await e.EvaluateAsync($"rtvs:::export_to_csv({result.Expression}, sep={sep.ToRStringLiteral()}, dec={dec.ToRStringLiteral()})", REvaluationKind.Normal);
```
AlexanderSher: Do you need evaluation after you've gotten `fileinfo`?

src/Host/Client/Impl/Definitions/REvaluationResult.cs
```diff
@@ -31,6 +32,14 @@ public struct REvaluationResult {
         /// </remarks>
         public JToken Result { get; }
         /// <summary>
+        /// Result of evaluation for 'raw'
+        /// </summary>
+        /// <remarks>
+        /// Computed by serializing the immediate result of evaluation, as if by <c>serialize(x,NULL)</c>
```
int19h: Is this comment still correct? Looking at the host code, it seems that it just handles `RAWSXP`, and doesn't try to serialize anything else?

src/Host/Client/Impl/Definitions/IMessageTransport.cs
```diff
@@ -9,5 +9,6 @@ namespace Microsoft.R.Host.Client {
     public interface IMessageTransport {
         Task SendAsync(string message, CancellationToken ct = default(CancellationToken));
         Task<string> ReceiveAsync(CancellationToken ct = default(CancellationToken));
+        Task<byte[]> ReceiveRawAsync(CancellationToken ct = default(CancellationToken));
```
int19h: I think we should revisit this. This interface is too low-level now - it's meant to be the message transport, but it's effectively a packet transport now (since you need multiple `Receive` calls to get a single message where blobs are involved).

Let's move `Message` and all related functionality here, so that `RHost` only ever deals with complete message instances. This will also let us do things more efficiently if we ever decide to move to a different protocol that allows mixing binary and non-binary data in a more efficient way.

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
karthiknadig: The second eval is not required. But, in remote scenario if the file size is massive then we might have to consider sending it over the network. For now I will change this to a single eval but this is something we will need to address.

src/Host/Client/Impl/Definitions/IMessageTransport.cs
karthiknadig: Created #2042 to refactor this.

src/Host/Client/Impl/Host/RHost.Message.cs
karthiknadig: We need to parse the header before we know how many blob (slices) we need to receive. I have a task on my plate to refactor some of this code. I will add a note to expose blobs as `IReadOnlyCollection<byte[]>`

src/R/Components/Test/Plots/PlotIntegrationTest.cs
```diff
@@ -223,26 +223,6 @@ public class PlotIntegrationTest : IAsyncLifetime {
 
         [Test(ThreadType.UI)]
         [Category.Plots]
-        public async Task ExportAsPdfInvalidFilename() {
```
karthiknadig: This test was written to verify file paths on the R side. This is no longer necessary. The filename used for export here must satisfy windows path requirements. The save dialog box disallows any invalid path characters. On the R side we use tempfile() to generate a temporary file name. Discussed removal of this test with @huguesv.

src/R/Components/Test/Plots/PlotIntegrationTest.cs
```diff
@@ -296,26 +280,6 @@ public class PlotIntegrationTest : IAsyncLifetime {
 
         [Test(ThreadType.UI)]
         [Category.Plots]
-        public async Task ExportAsImageInvalidFilename() {
```
karthiknadig: See comment for `ExportAsPdfInvalidFilename`

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -135,5 +135,27 @@ export_to_csv <- function(expr, sep, dec) {
...
+	data
+}
+
+# Helper to export current plot to image
+export_to_image <- function(device, width, height, resolution) {
+	filepath <- tempfile('plot_', fileext='.dat')
+	dev.copy(device=device,filename=filepath,width=width,height=height,res=resolution)
+	dev.off()
+	data <- readBin(filepath, 'raw', file.info(filepath)$size)
+	file.remove(filepath)
```
int19h: Use `on.exit` to make sure that this is cleaned up even if something above fails.

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -135,5 +135,27 @@ export_to_csv <- function(expr, sep, dec) {
...
+	data
+}
+
+# Helper to export current plot to pdf
+export_to_pdf <- function(width, height) {
+	filepath <- tempfile('plot_',fileext='.pdf')
+	dev.copy(device=pdf,file=filepath,width=width,height=height)
+	dev.off()
+	data <- readBin(filepath, 'raw', file.info(filepath)$size)
+	file.remove(filepath)
```
int19h: Ditto.

int19h:
:+1:

PR #2052, opened 2016-07-07
Settings property page + Settings.R I/O + tests

MikhailArkhipov:


src/Common/Core/Impl/Extensions/StringExtensions.cs
```diff
@@ -43,6 +43,19 @@ public static class StringExtensions {
             return s.IndexOf(prefix, StringComparison.OrdinalIgnoreCase) >= 0;
         }
 
+        public static string RemoveQuotes(this string s) {
```
AlexanderSher: Let's call it `TrimQuotes`

src/Package/Impl/ProjectSystem/PropertyPages/Settings/SettingsPageControl.cs
```diff
@@ -0,0 +1,188 @@
...
+        private void SetTextToControls() {
+            variableNameLabel.Text = Resources.SettingsPage_VariableNameLabel;
+            variableValueLabel.Text = Resources.SettingsPage_VariableValueLabel;
+            variableTypeLabel.Text = Resources.SettingsPage_VariableTypeLabel;
+
+            variableTypeList.Items.Add(Resources.SettingsPage_VariableType_String);
+            variableTypeList.Items.Add(Resources.SettingsPage_VariableType_Expression);
+            variableTypeList.SelectedIndex = 0;
+
+            explanationText.Text = Resources.SettingsPage_Explanation1
```
AlexanderSher: Can't we use `string.Format` here?

src/Package/Impl/ProjectSystem/PropertyPages/Settings/SettingsPageViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+        public IEnumerable<string> Files => _filesMap.Keys;
+        public SettingsTypeDescriptor TypeDescriptor => new SettingsTypeDescriptor(_settings);
+
+        public string CurrentFile {
+            get {
+                return _currentFile;
+            }
+            set {
+                if (!value.EqualsIgnoreCase(_currentFile)) {
+                    _currentFile = value;
```
AlexanderSher: We should probably change _currentFile in `finally` block, or restore it in `catch` block

src/R/Core/Impl/AST/DataTypes/RString.cs
```diff
@@ -44,5 +44,9 @@ public class RString : RScalar<string> {
         public override int GetHashCode() {
             return this.Value.GetHashCode();
         }
+
+        public override string ToString() {
+            return Value;
```
AlexanderSher: Is it for debug purposes?

MikhailArkhipov: It is partially completed support if we want dropdown-type editing of properties i.e. when Logical gets TRUE/FALSE in the dropdown similar to .NET types. However, I decided not to develop this ability just yet.

src/R/Components/Test/Configuration/ConfigurationSettingsTest.cs
```diff
@@ -0,0 +1,205 @@
...
+                        break;
+                    }
+                    settings.Add(s);
+                }
+            }
+            return settings;
+        }
+
+        private Stream ToStream(string s) {
+            MemoryStream stream = new MemoryStream();
```
AlexanderSher:
```{csharp}
return new MemoryStream(Encoding.UTF8.GetBytes("abc"));
```

src/R/Components/Test/Configuration/ConfigurationSettingsTest.cs
```diff
@@ -0,0 +1,205 @@
...
+            writer.Write(s);
+            writer.Flush();
+
+            stream.Position = 0;
+            return stream;
+        }
+
+        private Stream CreateStream() {
+            MemoryStream stream = new MemoryStream();
+            StreamWriter writer = new StreamWriter(stream);
```
AlexanderSher: `writer` isn't used

src/R/Components/Test/Configuration/ConfigurationSettingsCollectionTest.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.R.Components.Application.Configuration;
+using Microsoft.UnitTests.Core.XUnit;
+
+namespace Microsoft.R.Components.Test.Configuration {
+    [ExcludeFromCodeCoverage]
+    [Category.Configuration]
+    public class ConfigurationSettingsCollectionTest {
+        [Test]
+        public void Test01() {
+            string file = Path.GetTempFileName();
```
AlexanderSher: Let's use `TestFilesFixture` and create file in a test folder - in this case we will be able to verify that file in case of test failing

src/R/Components/Impl/Application/Configuration/Parser/ConfigurationParser.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            return null;
+        }
+
+        private bool IsEndOfStream() {
+            return _bufferedLine == null && _sr.EndOfStream;
+        }
+        private string ReadLine() {
+            var line = _bufferedLine ?? _sr.ReadLine();
+            _lineNumber++;
```
AlexanderSher: Do we need to do `_lineNumber++` if `_sr.ReadLine()` isn't called?

src/R/Components/Impl/Application/Configuration/IO/ConfigurationSettingsWriter.cs
```diff
@@ -0,0 +1,79 @@
...
+                    WriteAttributes(s);
+                    _writer.WriteLine(Invariant($"{s.Name} <- {v}"));
+                    _writer.WriteLine(string.Empty);
+                }
+            }
+            _writer.Flush();
+        }
+
+        private void WriteHeader() {
+            _writer.WriteLine("# Application settings file.");
```
AlexanderSher: Should we put these strings into resources?

src/R/Components/Impl/Application/Configuration/ConfigurationSetting.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+using System.ComponentModel;
+
+namespace Microsoft.R.Components.Application.Configuration {
+    public sealed class ConfigurationSetting : IConfigurationSetting {
```
AlexanderSher: You can derive it from `BindableBase`

src/Package/Test/ProjectSystem/PropertyPages/ProjectSettingsControlTest.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        [Test]
+        public async Task PropertyGridMultiple02() {
+            var shell = Substitute.For<ICoreShell>();
+            var fs = Substitute.For<IFileSystem>();
+            var control = new SettingsPageControl(shell, fs);
+
+            var folder = Path.GetTempPath();
+            var fileName1 = "1123.Settings.R";
```
AlexanderSher: There is no general `TestFilesFixture` in package tests, but you can create your own deriving it from `DeployFilesFixture`.

src/R/Components/Impl/Application/Configuration/IO/ConfigurationSettingsWriter.cs
MikhailArkhipov: Sure. I was not entirely sure we want or need to loc it, but it won;t hurt.

src/R/Components/Impl/Application/Configuration/Parser/ConfigurationParser.cs
MikhailArkhipov: since UnreadLine ```--``` line number, we should be increasing it here in any case i think.

AlexanderSher:
:+1:

PR #2063, opened 2016-07-11
Interactive Window update

karthiknadig:
New pull request. Previous one had a bad merge.
Notes: This pull request updates the interactive window to the most recent version. Which is identical to the one used by python tools and roslyn. It was built out of the clone of roslyn here, https://github.com/zooba/VsInteractiveWindow


AlexanderSher:
Karthik,

What’s the purpose of having 3 identical copies of Interactive Window? And why we merge binaries instead of source code through submodule?

MikhailArkhipov:
We discussed the submodule, we should certainly do it next time we have cycles.

AlexanderSher:
Ok, but what is the purpose of this PR? Why do we need a private copy that is identical to other copies?

MikhailArkhipov:
Hm I thought someone needed it to start making changes...

AlexanderSher:
I've updated it to 1.3, which already has AddToHistory that I needed.

karthiknadig:
@AlexanderSher the purpose was to use the same source files to build the interactive window dll for PTVS and RTVS. I have forwarded you the conversation regarding this.

AlexanderSher:
Ok, I have some issues with accessing my corporate email, I read it as soon as it is restored

karthiknadig:
Decided over e-mail that we will not be taking an update to interactive window at this time.

Issue #2070, opened 2016-07-13
Save As... does not change file name


MikhailArkhipov:
New Script.R
Save As... and change to Script5.R

Observed: Script5.R is created in the project, but current file is still Script.R. It works as 'Make Copy Of'.
Expected: same as in other projects, Save As... changes file name in the active editor tab.

Doesn't seem to be specific to R, SQL and HTML are working same way,

AlexanderSher:
Opened file is in fact Script5.R, but it shows Script.R in header.

AlexanderSher:
Issue happens because CPS treats this action as rename, while file watcher just sees a newly added file

AlexanderSher:
Issue happens because CPS doesn't call `RunningDocumentTable.RenameDocument` for the "Save as" operation

MikhailArkhipov:
@AlexanderSher did we file CPS bug and if yes, what is the number in VSTS

AlexanderSher:
No, there is no bug in VSTS

MikhailArkhipov:
Does it repro in Dev15?

MikhailArkhipov:
Reproes in Dev15

MikhailArkhipov:
VSTS 370378

MikhailArkhipov:
Won't be fixed in RTW. Pushing to Update 1.

MikhailArkhipov:
Punted to Dev16.

PR #2074, opened 2016-07-14
#2060 and other fixes

### LINKED ISSUE #2060: New file does not appear in the project and AF when closing the project
Not sure if this is truly needed, but this how I got there
Install SQL tools
Install AdventureWorks DB

1. Create R project
2. Open SQL Object Explorer (View -> ...)
3. Connect to local instance / AdventureWorks
4. Right click AdventureWorks -> New Query
5. SELECT * FROM Human...
6. Save SQL in the project folder

OBSERVED: file does not appear in the project (Show all does display it)
Close solution

image

```
'path' cannot be an empty string ("") or start with the null character.
Parameter name: path

   at Microsoft.Requires.NotNullOrEmpty(String value, String parameterName)
   at Microsoft.VisualStudio.ProjectSystem.Utilities.PathHelper.MakeRelative(String basePath, String path)
   at Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.Utilities.PathExtensions.ToShortRelativePath(IFileSystem fileSystem, String path, String rootFolder) in F:\RTVS\src\ProjectSystem\Impl\Utilities\PathExtensions.cs:line 14
   at Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.IO.MsBuildFileSystemWatcher.FileCreated.Apply() in F:\RTVS\src\ProjectSystem\Impl\IO\MsBuildFileSystemWatcher.FileCreated.cs:line 31

```

change appears to be OK

image

MikhailArkhipov:
#2060 New file does not appear in the project and AF when closing...
Reverse Ctrl+F1 and F1 bindings
Add SQL Query template
Better handle .sql files written with UTF-8 BOM
Auto-position caret at the REPL prompt on typing


src/ProjectSystem/Impl/IO/MsBuildFileSystemWatcher.cs
```diff
@@ -215,6 +215,9 @@ public sealed partial class MsBuildFileSystemWatcher : IDisposable {
 
             relativePath = PathHelper.MakeRelative(rootDirectory, fullPath);
             try {
+                if (!fileSystem.FileExists(fullPath)) {
```
AlexanderSher: This is really weird. How is it possible that `GetFileAttributes` fails for a newly added file?

MikhailArkhipov: Async stuff. Set bp in debugger, see exception catch getting hit. Repeat stepping of the same code, it succeeds. Not sure where the race is coming from.

MikhailArkhipov: OK, I can remove this if the exception is not causing trouble. CPS exception was caused by Link metadata and appears to go away when I remove Link property from rules. But this is more or less 'best guess'

AlexanderSher: Race may occur right after this method returns `true`, and there is try/catch around this block already, so calling `FileExists` is useless. But we probably need to catch `UnauthorizedAccessException` in addition to `IOException`.

MikhailArkhipov: Not entirely. FileExists check prevents returning false in IsFileAllowed. Returning false may actually explain when sometimes file is added, but does not appear in project explorer. Race occurs when check happens between file is deleted and then re-created. I guess the event queue is not compact i.e. create-delete-create is not collapsed into just created.

MikhailArkhipov: Basically - is there a harm reporting that file is allowed when it actually does not exist i.e. doing optimistic check?

AlexanderSher: I think that `_fileSystem.ToShortRelativePath` will throw an exception on the file that doesn't exist

MikhailArkhipov: OK, let me look into that.

MikhailArkhipov: What throws is 

```
                return filter.IsFileAllowed(relativePath, fileSystem.GetFileAttributes(fullPath));

```

specifically

```
fileSystem.GetFileAttributes
```

since the file does not exist.

AlexanderSher: Sure, but it is catched and method exists with false, so there can't be a popup message with exception.
I think that without debugger race happens in a different place - between this method and  `ToShortRelativePath` calls. This method returns `true`, but  `ToShortRelativePath` throws exception.

MikhailArkhipov: Ah, I think we are talking 3 different things since I found 3 different issues :-)

1. IsFileAllowed throwing - no popup, but I've seen file added but not showing up in Solution Explorer.
2. CPS 'not supported' exception when 'Link' metadata mismatch.
3. The popup in the beginning of the bug - yes, that is about ToShortRelativePath and I am going to look into it.

AlexanderSher: Hm... then the first case requires further investigation. Any `IFileSystemChange` appears in queue only after event has happened. So if `FileCreated` happened but no file was found, there should be another `FileCreated` in the queue that should add a file. Or there should be an error in the log.
Do you have logs for the #2070?

MikhailArkhipov: There are some attached to #2060. Basically I saw bp hit in exception, then repeating same code in debugger would succeed.

AlexanderSher: There are no failures. Log shows that file was added to the project msbuild:
```
[16-7-13_10-33-43]:File mirroring project after changes applied:
                   <?xml version="1.0" encoding="utf-16"?>
                   <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
                     <ItemGroup>
                       <Content Include=".Rhistory" />
                       <Content Include="adcd 1 2 3 4 5 6 7 8 9\markdown.html" />
                       <Content Include="adcd 1 2 3 4 5 6 7 8 9\markdown.rmd" />
                       <Content Include="markdown.html" />
                       <Content Include="markdown.rmd" />
                       <Content Include="query.sql" />
                       <Content Include="rproject10.rproj" />
                       <Content Include="script.R" />
                       <Content Include="script1.R" />
                       <Content Include="Script2.R" />
                       <Content Include="Settings.R" />
                       <Content Include="Settings1.R" />
                       <Content Include="SqlQuery.sql" />
                       <Content Include="SQLQuery1.sql" />
                     </ItemGroup>
                     <ItemGroup>
                       <Folder Include="adcd 1 2 3 4 5 6 7 8 9\" />
                     </ItemGroup>
                     <ItemGroup />
                   </Project>
```

AlexanderSher:
:+1:

Issue #2065, opened 2016-07-11
Blobs: Send blobs from client to RHost


karthiknadig:
Add code to send blobs from the client to RHost. Retrieve and clean-up blobs from R.

PR #2076, opened 2016-07-14
Adds feature to work with blobs in RHost from RTVS and R #2065

karthiknadig:
#2065 
Adds protocol commands to send blob to, get blobs from, and destroy blobs from client. 
Adds functions to create, access and destroy blobs from R.


src/Host/Client/Impl/Definitions/IRBlobHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Client {
+    public interface IRBlobHelper {
```
AlexanderSher: Let's change the name. `IRBlobService` or something like this

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -230,6 +245,56 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
...
+                ? Task.FromCanceled(new CancellationToken(true))
+                : DestroyBlobAsyncBackground(blobIds, ct);
+        }
+
+        private async Task DestroyBlobAsyncBackground(long[] blobIds, CancellationToken ct) {
+            await TaskUtilities.SwitchToBackgroundThread();
+
+            JArray message;
+            var request = BlobRequest.MakeDestroyBlobsRequest(this, out message, blobIds);
+            //  we don't expect a response for this so don't add it to the pending blob requests
```
AlexanderSher: Every request should have matching response - we need to know that command get executed successfully.

src/Host/Client/Impl/Definitions/SendBlobResult.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Client {
+    public class SendBlobResult : BlobResult {
```
int19h: It feels like this, and `GetBlobResult`, do not really offer anything useful to the API client - they won't care about `Kind`, because the methods that are exposed to them return specific subtypes anyway. They'll just want the ID for send, and the data for get. We should make this private to `RHost`, and expose the results directly on private `IRBlobHelper` methods.

src/Host/Client/Impl/Host/RHost.BlobRequest.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.R.Host.Client {
+    partial class RHost {
+        
+        private class BlobRequest {
+            public static readonly string CreateBlobRequestMessageName = "?CreateBlob";
```
int19h: All of these can be `const` instead of `static readonly`.

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -230,6 +245,56 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
             await RespondAsync(request, ct, input);
         }
 
+        public Task<SendBlobResult> SendBlobAsync(byte[] data, CancellationToken ct) {
```
AlexanderSher: These 3 methods are almost identical.
The only difference is the type of request and corresponding result. 
The same thing then propagated to RSession.

From the protocol point of view - what's the difference between blob message and evaluation message?

int19h: They're semantically completely different. An eval message triggers an eval and returns the result; the input is an expression. A send blob message uploads a blob, and returns its ID; the input is a byte array.

AlexanderSher: Semantically - yes, but under the hood, do we use different sockets or message formats for those messages?

src/Host/Client/Impl/Host/RHost.cs
int19h: In our terminology, "Destroy" is technically a notification, not a request. It doesn't require a response because it can always be assumed to either succeed (if host is alive and connection is live) or not matter (if host is dead or connection is going down; in the latter case, host will shut down anyway). And since we always use transports that guarantee delivery, we don't need to implement our own ACK scheme.

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: For example, do we need a separate dictionary of requests for blobs?

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -53,6 +53,18 @@ toJSON <- function(obj) {
   call_embedded("toJSON", obj)
 }
 
+create_blob <- function(obj) {
+    call_embedded("create_blob", obj)
+}
+
+get_blob <- function(blob.id) {
```
int19h: We use underscore instead of dot for identifiers in R code as a matter of style (even though some existing code doesn't follow this). This is because S3 objects use dots in names to identify methods (e.g. `blob.id` would be implementation of S3 generic method `blob` for class `id`), so there's potential for confusion there.

src/Host/Client/Impl/Host/RHost.cs
karthiknadig: Under-the-hood difference is that for evaluation request requires only one websocket message,  whereas for blobs we need two websocket messages; text message [which has the header] followed by binary with the data. We did this to avoid duplicating the header parsing code (which we have for text) for binary.

After I make the change suggested by @int19h (see comment on SendBlobRequest.cs). There is  explicit value returned from send and get, instead of a result object which gets propagated to the caller. The caller does not get to choose how they want the result. That is different from evaluation result, where the result object makes sense. Since the result value depends on the expression and expectation of the caller.

We could create a base request type that manages the request id, and message name. We can then derive `EvaluationRequest` and `BlobRequest` from it, wherefore one dictionary should be enough to manage request and response agnostic of eval or blob.

src/Host/Client/Impl/Definitions/IRBlobHelper.cs
karthiknadig: Will update in the next iteration.

---
In reply to: [70852735](https://github.com/Microsoft/RTVS/pull/2076#discussion_r70852735) [](ancestors = 70852735)

src/Host/Client/Impl/rtvs/R/util.R
karthiknadig: Will update in the next iteration.

---
In reply to: [70860104](https://github.com/Microsoft/RTVS/pull/2076#discussion_r70860104) [](ancestors = 70860104)

AlexanderSher:
Do we have any tests for this?

karthiknadig:
Tests added.

---
In reply to: [233046276](https://github.com/Microsoft/RTVS/pull/2076#issuecomment-233046276) [](ancestors = 233046276)

src/Host/Client/Impl/Host/RHost.GetBlobResult.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.R.Host.Client {
+    partial class RHost {
+        private class GetBlobResult : BlobResult {
+            public IReadOnlyList<Blob> Blobs;
+
+            public GetBlobResult(List<Blob> blobs) : base(BlobRequestKind.Get) {
+                if(blobs == null) {
```
int19h: Nit: you can use `??` to make this more concise:
```cs
Blos = blobs ?? new List<Blob>()
```

src/Host/Client/Impl/Transports/WebSocketMessageTransport.cs
```diff
@@ -64,8 +64,12 @@ public class WebSocketMessageTransport : WebSocketBehavior, IMessageTransport {
         public async Task SendAsync(string message, byte[] data, CancellationToken ct = default(CancellationToken)) {
             await _sendLock.WaitAsync(ct);
             try {
-                Socket.Send(message);
-                Socket.Send(data);
+                if (data.Length > 0) {
+                    Socket.Send(message);
+                    Socket.Send(data);
+                } else {
+                    throw new ArgumentException($"{nameof(data)} cannot be a 0 length array.");
```
int19h: Why not? It would seem that a 0-sized blob would occasionally arise naturally (e.g. when uploading an empty file).

karthiknadig: Ah.. I will revert that one and update the test. I agree, it should be possible to send '0' length data.

PR #2080, opened 2016-07-15
Propagate RHost disconnected state to interactions and evaluations + different fixes

### LINKED ISSUE #1469: Allow BeginInteractionAsync to notify RInteractiveEvaluator that response should be added to REPL

### LINKED ISSUE #1489: Package Manager will send library() and detach() commands to REPL

### LINKED ISSUE #2066: Package.Tests - fix export cardinality mismatch
Currently there are instances from both Package and Components.Test which causes test failures.

AlexanderSher:
This is an infrastructure change that allows handling R disconnected state in code that uses evaluations and interactions. Previously, `EvaluateAsync` and `RespondAsync` either swallowed disconnected state or returned a never ending task. With this change, they will return a task that will get canceled with `RHostDisconnectedException`. Code that doesn't care can simply use `DoNotWait`, which will swallow task cancellation as an expected state. Code that cares can handle this exception instead of general `OperationCanceledException` that can be thrown by client-side cancellation token.

Other changes in the fix:
- Fix #1469 Allow BeginInteractionAsync to notify RInteractiveEvaluator that response should be added to REPL
- Fix #1489 Package Manager will send library() and detach() commands to REPL 
- Fix #2066 Package.Tests - fix export cardinality mismatch
- SessionPool is merged into RSessionProvider


AlexanderSher:
Ping

src/Common/Core/Impl/Tasks/TaskCompletionSourceEx.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        private void SpinUntilCompleted() {
+            if (Task.IsCompleted) {
+                return;
+            }
+
+            var sw = new SpinWait();
+            while (!Task.IsCompleted) {
+                sw.SpinOnce();
```
MikhailArkhipov: What to do if it never completes

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -231,16 +234,19 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
         }
 
         public Task<REvaluationResult> EvaluateAsync(string expression, REvaluationKind kind, CancellationToken ct) {
-            return ct.IsCancellationRequested || _runTask == null || _runTask.IsCompleted
-                ? Task.FromCanceled<REvaluationResult>(new CancellationToken(true))
-                : EvaluateAsyncBackground(expression, kind, ct);
+            return _runTask == null || _runTask.IsCompleted
+                ? TaskUtilities.CreateCanceled<REvaluationResult>(new RHostDisconnectedException())
```
MikhailArkhipov: A bit too long for inline conditional

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -275,7 +288,15 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator {
             } else {
                 result = new REvaluationResult(response[2], error, parseStatus);
             }
-            request.CompletionSource.SetResult(result);
+            request.CompletionSource.TrySetResult(result);
+        }
+
+        private void CancelRemainingEvaluationRequests() {
+            var requests = _evalRequests.Values;
+            _evalRequests.Clear();
```
MikhailArkhipov: I assume this sequence does not have to be atomic (i.e. between Clear() and between setting cancellations there may be calls using values in requests)

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -317,26 +309,27 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
 
-            ClearPendingRequests();
+            ClearPendingRequests(exception);
 
             return Task.CompletedTask;
         }
 
-        private void ClearPendingRequests() {
+        private void ClearPendingRequests(OperationCanceledException exception) {
             RSessionRequestSource requestSource;
```
MikhailArkhipov: Should this be atomic?

src/Host/Client/Impl/Session/RSessionRequestSource.cs
```diff
@@ -5,25 +5,26 @@
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Common.Core.Tasks;
 
 namespace Microsoft.R.Host.Client.Session {
     internal sealed class RSessionRequestSource {
```
MikhailArkhipov: This stuff will benefit from more comments and/or document/diagram explaining the state machine

src/Host/Client/Test/Mocks/RSessionMock.cs
```diff
@@ -31,7 +31,7 @@ public sealed class RSessionMock : IRSession {
 
         public Task<IRSessionEvaluation> BeginEvaluationAsync(CancellationToken cancellationToken = default(CancellationToken)) {
             Evaluation = new RSessionEvaluationMock();
-            BeforeRequest?.Invoke(this, new RRequestEventArgs(Evaluation.Contexts, Prompt, 4096, true));
+            BeforeRequest?.Invoke(this, new RBeforeRequestEventArgs(Evaluation.Contexts, Prompt, 4096, true));
```
MikhailArkhipov: Put named constant somewhere?

src/Common/Core/Impl/Tasks/TaskCompletionSourceEx.cs
AlexanderSher: Then we have problems in Task implementations. `SpinUntilCompleted` is called if `_completed == 1`, which means that task is in process of becoming completed.
I've borrowed this code from `Task` implementation.

---
In reply to: [71067349](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71067349) [](ancestors = 71067349)

src/Host/Client/Test/Mocks/RSessionMock.cs
AlexanderSher: Done

---
In reply to: [71067946](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71067946) [](ancestors = 71067946)

src/Host/Client/Test/Mocks/RSessionMock.cs
```diff
@@ -40,17 +40,17 @@ public sealed class RSessionMock : IRSession {
 
         public Task<IRSessionInteraction> BeginInteractionAsync(bool isVisible = true, CancellationToken cancellationToken = default (CancellationToken)) {
             _inter = new RSessionInteractionMock();
-            BeforeRequest?.Invoke(this, new RRequestEventArgs(_inter.Contexts, Prompt, 4096, true));
+            BeforeRequest?.Invoke(this, new RBeforeRequestEventArgs(_inter.Contexts, Prompt, 4096, true));
             return Task.FromResult(_inter);
         }
 
         public Task CancelAllAsync() {
```
MikhailArkhipov: Can be void?

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: This method is called from 3 places:
- `CancelAllAsync` - should cancel all currently available requests, but it doesn't block any new requests from being added, so we don't need it to be atomic in this case
- `StartHostAsyncBackground` - no new requests can be added when method is called because RHost is not running yet
- `IRCallbacks.Disconnected` - RHost is already stopped, so no new requests can be added

---
In reply to: [71067562](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71067562) [](ancestors = 71067562)

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -41,13 +42,13 @@ public class FilesCompletionProvider : IRCompletionListProvider {
                 throw new ArgumentNullException(nameof(directoryCandidate));
             }
 
-            _shell.CompositionService.SatisfyImportsOnce(this);
+            ImagesProvider = _shell.ExportProvider.GetExportedValueOrDefault<IImagesProvider>();
+            Workflow = _shell.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>().GetOrCreate();
```
MikhailArkhipov: Looks like it doesn't need [Import] above anymore

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -173,34 +177,28 @@ internal class RPackageManager : IRPackageManager {
...
+                    await WrapRException(eval.ExecuteAsync($"options(repos=eval(parse(text={repositories})))"));
+                    var libraries = (await DeparseLibrariesAsync()).ToRStringLiteral();
+                    await WrapRException(eval.ExecuteAsync($".libPaths(eval(parse(text={libraries})))"));
 
-                    var result = await WrapRException(queryFunc(sessionToken.Session));
+                    var result = await WrapRException(queryFunc(eval));
                     return result.Select(p => p.ToObject<RPackage>()).ToList().AsReadOnly();
                 }
-            } catch (MessageTransportException ex) {
+            } catch (RHostDisconnectedException ex) {
```
MikhailArkhipov: You may want to send guidance on new exceptions and their handling. I have new code in remote intellisense support so I guess i need to handle some of these.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflowOperations.cs
```diff
@@ -189,12 +189,10 @@ internal class RInteractiveWorkflowOperations : IRInteractiveWorkflowOperations
 
         public void TryRunShinyApp () {
             _shinyRunningTask = Task.Run(async () => {
-                try {
-                    using (var e = await _workflow.RSession.BeginInteractionAsync()) {
-                        await e.RespondAsync("library(shiny)" + Environment.NewLine + "runApp()" + Environment.NewLine);
-                    }
-                } catch (TaskCanceledException) { } catch (MessageTransportException) { }
-            }).ContinueWith((t) => _shinyRunningTask = null);
+                using (var e = await _workflow.RSession.BeginInteractionAsync()) {
```
MikhailArkhipov: These exceptions no longer come here?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -6,16 +6,13 @@
...
 using Microsoft.R.Components.PackageManager;
 using Microsoft.R.Components.Plots;
 using Microsoft.R.Components.Settings;
 using Microsoft.R.Host.Client;
-using Microsoft.R.Host.Client.Install;
-using Microsoft.VisualStudio.InteractiveWindow;
+using Microsoft.VisualStudio.R.Package.Repl;
 
-namespace Microsoft.VisualStudio.R.Package.Repl {
+namespace Microsoft.R.Components.InteractiveWorkflow.Implementation {
```
MikhailArkhipov: Actually, I'd rather avoid Implementation and Definitions suffixes in namespaces. We need to clean those up.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflowOperations.cs
AlexanderSher: `MessageTransportException` never comes here, `TaskCanceledException` will just mark task canceled, so there is no need to handle it explicitly

---
In reply to: [71068292](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71068292) [](ancestors = 71068292)

src/Host/Client/Test/Mocks/RSessionMock.cs
AlexanderSher: No, this is Mock implementation, real implementation needs to be async

---
In reply to: [71068173](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71068173) [](ancestors = 71068173)

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: There is no harm if Clean will be called concurrently with request processing, but there can be a problem if new request is added after `Clear` is called. I'll look into it.

---
In reply to: [71067479](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71067479) [](ancestors = 71067479)

MikhailArkhipov:
:+1:

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
AlexanderSher: There is no much to add to what is written in this PR description. `MessageTransportException` became private. I can put this description in OneNote somewhere

---
In reply to: [71068272](https://github.com/Microsoft/RTVS/pull/2080#discussion_r71068272) [](ancestors = 71068272)

MikhailArkhipov: I think general guidance on host exception handling in OneNote would be useful. I.e. what is being thrown when and what to handle and what to ignore.

PR #2083, opened 2016-07-15
Initial commit for R-host broker web app.

### LINKED ISSUE #2037: Broker: enumerate R implementations
Endpoint to enumerate all installed R implementations, with version number, info on whether it's MRC/MRS/CRAN, and any other bits of info that might prove useful.

### LINKED ISSUE #2044: Broker configuration
Configure group that is allowed access, timeout to kill host process, etc. Probably a config file. Path to R. Assume single R per remote machine.

int19h:
Fix #2044: Broker configuration

Use .NET Core configuration framework, configured to read options from command line arguments and from config file.

Add specific options for authentication (only users in specific groups are permitted access), and for R interpreters.

Fix #2037: Broker: enumerate R implementations

Enumerate auto-detected R interpreter, and any interpreters explicitly specified in the config.


src/Package/Impl/project.14.0.json
```diff
@@ -12,6 +12,7 @@
     "Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime": "14.2.25123",
     "Microsoft.VisualStudio.TextManager.Interop.12.0": "12.0.30110",
     "Microsoft.VSSDK.BuildTools": "14.2.25201",
+    "System.Collections.Immutable": "1.1.33-beta",
```
int19h: The reason for this change is to ensure that we have the right version for CPS. Previously, we didn't depend on it directly, so CPS brought this version via indirect reference. Now that we also bring in ASP.NET Core via NuGet, it has a direct reference on a different version, which was then preferred, and ended up causing breakage when implementing CPS interface, due to mismatches between versions. Explicitly specifying this in projects that reference CPS makes sure that we always use a single, compatible version.

int19h:
No tests here yet. I want to get this in so that other functionality that goes into the broker can be developed in parallel - in particular, @karthiknadig will need it for HTTP request proxying for help and Shiny apps.

src/Common/Core/Test/project.json
```diff
@@ -2,7 +2,7 @@
   "dependencies": {
     "FluentAssertions": "4.6.1",
     "MicroBuild.Core": "0.2.0",
-    "Newtonsoft.Json": "8.0.3",
+    "Newtonsoft.Json": "9.0.1",
```
huguesv: Does this have implications for installer, and/or assembly redirects?

src/Host/Broker/Impl/Interpreters/InterpretersController.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly IOptions<InterpretersOptions> _interpOptions;
+        private readonly RInstallation rInstallation = new RInstallation();
+
+        public InterpretersController(IOptions<InterpretersOptions> interpOptions) {
+            _interpOptions = interpOptions;
+        }
+
+        [HttpGet]
+        public IEnumerable<InterpreterInfo> Get() {
+            if (_interpOptions.Value.AutoDetect) {
```
huguesv: if AutoDetect, we may potentially return the same interpreter twice?  Is that okay?

src/Host/Broker/Impl/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+        internal static IConfigurationRoot Configuration { get; private set; }
+
+        public static void Main(string[] args) {
+            var configBuilder = new ConfigurationBuilder().AddCommandLine(args);
+            Configuration = configBuilder.Build();
+
+            string configFile = Configuration["configFile"];
+            if (configFile != null) {
+                configBuilder.AddJsonFile(configFile, optional: false);
+                Configuration = configBuilder.Build();
```
huguesv: replace both `configBuilder.Build()` calls with a single one after the `if` statement.

huguesv:
:+1:

src/Common/Core/Test/project.json
int19h: We don't seem to be doing anything special wrt this assembly in the installer. Did you have some specific issue in mind?

src/Host/Broker/Impl/Interpreters/InterpretersController.cs
int19h: The list is supposed to be for explicit configuration, and it's really up to the admin to put entries in there. They can put the same R under two different names even without autodetect.

I'm actually not sure about how this will end up being used exactly. It will become clearer once we wire up the broker from VS, and the interfaces here may well change to something more natural, depending on what that code will look like.

src/Host/Broker/Impl/Program.cs
int19h: The first `Build` is necessary to get a configuration object to read "configFile" from.

huguesv: Oh, right, I missed that!

src/Common/Core/Test/project.json
huguesv: No, just being paranoid, remembering the issues we had with NTVS and Newtonsoft.Json.

src/Package/Impl/project.15.0.json
```diff
@@ -13,6 +13,7 @@
     "Microsoft.VisualStudio.TextManager.Interop.12.0": "12.0.30110",
     "Microsoft.VSSDK.BuildTools": "15.0.25201-Dev15Preview2",
     "System.IO.Compression.ZipFile": "4.0.0",
+    "System.Collections.Immutable": "1.1.33-beta",
```
AlexanderSher: Should we have this version in VS15 as well?

src/Host/Client/Impl/Microsoft.R.Host.Client.csproj
```diff
@@ -198,6 +193,10 @@
       <Project>{8d408909-459f-4853-a36c-745118f99869}</Project>
       <Name>Microsoft.Common.Core</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\R\Core\Impl\Microsoft.R.Core.csproj">
+      <Project>{0c4bce1d-3cb8-4e2a-9252-58784d7f26a5}</Project>
+      <Name>Microsoft.R.Core</Name>
```
AlexanderSher: If I recall correctly, original idea with Host.Client was to reference as few assemblies as possible. Why do want to move installation to R.Core?

MikhailArkhipov: Since Host.Client may become public API, the fewer assemblies the better. Also - R.Core is R parser, it has nothing to do with MRC or R Installation. MRC and R Install actually have more to do with the host...

src/Host/Broker/Impl/Security/RUserAuthorizationHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class RUserAuthorizationHandler : AuthorizationHandler<RUserAuthorizationRequirement> {
+        private readonly IOptions<SecurityOptions> _securityOptions;
+
+        public RUserAuthorizationHandler(IOptions<SecurityOptions> securityOptions) {
+            _securityOptions = securityOptions;
+        }
+
+        protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, RUserAuthorizationRequirement requirement) {
+            if (!context.User.IsInRole(_securityOptions.Value.AllowedGroup)) {
+                return Task.CompletedTask;
```
MikhailArkhipov: Seems to be returning same thing in both cases?

src/R/Core/Impl/Microsoft.R.Core.csproj
```diff
@@ -79,6 +80,12 @@
     <Compile Include="AST\Utility\AstWriter.cs" />
     <Compile Include="AST\Values\MissingValue.cs" />
     <Compile Include="AST\Values\NullValue.cs" />
+    <Compile Include="Install\IMicrosoftRClientInstaller.cs" />
+    <Compile Include="Install\ISupportedRVersionRange.cs" />
```
MikhailArkhipov: Perhaps create new assembly specific to R engine itself? Microsoft.R.Engine? Or move to Support and I will take RData things out of it and into the editor.

src/Host/Client/Impl/Microsoft.R.Host.Client.csproj
int19h: Broker can't depend on Host.Client, because Host.Client might want to depend on some things in broker (like the JSON POCOs).

src/Host/Broker/Impl/Security/RUserAuthorizationHandler.cs
int19h: Yeah, the crucial part here is calling/not calling `Succeed`.

src/R/Core/Impl/Microsoft.R.Core.csproj
int19h: `R.Support` depends on `R.Host.Client`. I need it to be in some place that doesn't, so that broker can reference that, and `Host.Client` can reference broker.

src/Host/Client/Impl/Microsoft.R.Host.Client.csproj
int19h: Basically what we really need is an assembly in the stack of dependencies that has stuff that is related to R, but that does not depend on _execution_ of R code. Host.Client is really mostly about execution, so it would sit above in the stack.

src/Package/Impl/project.15.0.json
int19h: Good catch, it's a higher version there. I think it doesn't matter during build, because it'll just fetch the higher one because it requires it; but I'll adjust it here as well, and check on VS15.

src/Host/Client/Impl/Microsoft.R.Host.Client.csproj
MikhailArkhipov: After I am done with remote intellisense (which should be today) I will take all Host.Client related code from R.Support to R.Editor and R.Support won't be needed and can be reused.
Or we can rename R.Core -> R.Language.Core and reuse R.Core name.

int19h: Okay. I will move this to R.Support then, and add a dependency for now; and remove it after this refactoring happens.

int19h:
Addressed feedback. The shared classes to deal with R interpreters are now in their own assembly.

src/Setup/Assemblies.wxi
```diff
@@ -15,6 +15,7 @@
       <File Source="$(var.BinDir)\Microsoft.R.Editor.dll" />
       <File Source="$(var.BinDir)\Microsoft.R.Host.Client.dll" />
       <File Source="$(var.BinDir)\Microsoft.R.Support.dll" />
+      <File Source="$(var.BinDir)\Microsoft.R.Interpreters.dll" />
```
MikhailArkhipov: Needs to be added to files.swr in the RTVS-pr branch Dev15 setup

MikhailArkhipov: Also to VSIX manifest

MikhailArkhipov:
👍

PR #2084, opened 2016-07-18
Remote intellisense

MikhailArkhipov:
More tests incoming


src/R/Components/Impl/Application/Configuration/ConfigurationSettingCollection.cs
```diff
@@ -25,7 +25,7 @@ public sealed class ConfigurationSettingCollection :
         /// </summary>
         public void Load(string filePath) {
             lock (_lock) {
-                if (!string.IsNullOrEmpty(filePath)) {
+                if (!string.IsNullOrEmpty(filePath) && File.Exists(filePath)) {
```
AlexanderSher: Use `try/catch` here instead

src/R/Support/Impl/Help/IntelliSenseRHost.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Microsoft.R.Support.Help {
+    [Export(typeof(IIntellisenseRHost))]
+    public sealed class IntelliSenseRHost : IIntellisenseRHost {
+        private static readonly Guid SessionId = new Guid("8BEF9C06-39DC-4A64-B7F3-0C68353362C9");
+        private readonly ICoreShell _coreShell;
+        private readonly IRSessionProvider _sessionProvider;
+        private readonly SemaphoreSlim _sessionSemaphore = new SemaphoreSlim(1, 1);
+
+        [ImportingConstructor]
+        public IntelliSenseRHost(ICoreShell coreShell, IRSessionProvider sessionProvider) {
```
AlexanderSher: Do we need a long-running session for IntelliSence, or it will be enough to use session from pool: `IRSessionProvider.BeginEvaluationAsync`?

src/R/Support/Impl/Help/IntelliSenseRHost.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public IRSession Session { get; private set; }
+
+        public void Dispose() {
+            Session?.Dispose();
+            Session = null;
+        }
+
+        public async Task CreateSessionAsync() {
+            await _sessionSemaphore.WaitAsync();
```
AlexanderSher: You can use `BinaryAsyncLock` here

src/R/Support/Impl/Help/IntelliSenseRHost.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Common.Core.Shell;
+using Microsoft.Languages.Editor.Shell;
+using Microsoft.R.Host.Client;
+using Microsoft.R.Support.Settings;
+
+namespace Microsoft.R.Support.Help {
+    [Export(typeof(IIntellisenseRHost))]
+    public sealed class IntelliSenseRHost : IIntellisenseRHost {
```
AlexanderSher: I'd rather call it `IIntellisenseRSession`

src/R/Editor/Test/Utility/FunctionIndexBasedTest.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Microsoft.R.Editor.Test.Utility {
+    [ExcludeFromCodeCoverage]
+    public abstract class FunctionIndexBasedTest : IAsyncLifetime {
+        protected readonly IExportProvider _exportProvider;
+        protected readonly IEditorShell _editorShell;
+        protected readonly IPackageIndex _packageIndex;
+        protected readonly IFunctionIndex _functionIndex;
+
+        protected FunctionIndexBasedTest(REditorMefCatalogFixture catalog) {
```
AlexanderSher: >REditorMefCatalogFixture [](start = 41, length = 24)

`AssemblyMefCatalogFixture` will be enough

src/R/Editor/Test/Completions/RCompletionSourceTest.cs
```diff
@@ -21,20 +20,13 @@
...
-        private readonly IExportProvider _exportProvider;
-
-        public RCompletionSourceTest(REditorMefCatalogFixture catalog) {
-            _exportProvider = catalog.CreateExportProvider();
-        }
-
-        public void Dispose() {
-            _exportProvider.Dispose();
-        }
+    [Collection(CollectionNames.NonParallel)]
```
AlexanderSher: Does it fail when non-parallel?

src/R/Editor/Test/Signatures/ComputeCurrentParameterTest.cs
```diff
@@ -25,40 +22,23 @@ namespace Microsoft.R.Editor.Test.Signatures {
...
-
-        public Task InitializeAsync() {
-            return FunctionIndexUtility.InitializeAsync(_functionIndex);
-        }
-
-        public async Task DisposeAsync() {
-            await FunctionIndexUtility.DisposeAsync(_functionIndex, _exportProvider);
-            _exportProvider.Dispose();
-        }
+    public class ComputeCurrentParameter : FunctionIndexBasedTest {
```
AlexanderSher: `ComputeCurrentParameterTest`

src/R/Editor/Test/Signatures/ComputeCurrentParameterTest.cs
```diff
@@ -25,40 +22,23 @@ namespace Microsoft.R.Editor.Test.Signatures {
...
+        public ComputeCurrentParameter(REditorMefCatalogFixture catalog, EditorTestFilesFixture testFiles) : base(catalog) { }
```
AlexanderSher: Looks like you don't use `testFiles` here anymore

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -2,106 +2,157 @@
...
-            get {
-                if (_basePackages == null)
-                    _basePackages = new BasePackagesCollection(_shell.ExportProvider.GetExportedValue<IFunctionIndex>());
-
-                return _basePackages.Packages;
+        public IEnumerable<IPackageInfo> Packages => _packages.Values;
+
+        public async Task BuildIndexAsync() {
+            _ready = await _buildIndexLock.WaitAsync();
+            if (!_ready) {
```
AlexanderSher: You don't need to use `_ready` field here - the whole purpose of the `BinaryAsyncLock` is to combine checking and lock in one primitive

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -2,106 +2,157 @@
...
+                Debug.WriteLine("R function index (preloaded): {0} ms", (DateTime.Now - startTime).TotalMilliseconds);
+
+                // Populate function index for all remaining packages
+                startTime = DateTime.Now;
+                await BuildRemainingPackagesFunctionListAsync();
+                Debug.WriteLine("R function index (remaining): {0} ms", (DateTime.Now - startTime).TotalMilliseconds);
+
+                await _functionIndex.BuildIndexAsync(this);
+                Debug.WriteLine("R function index total: {0} ms", (DateTime.Now - startTotalTime).TotalMilliseconds);
+                _ready = true;
```
AlexanderSher: You need to call `_buildIndexLock.Release()`

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -2,257 +2,113 @@
...
-                        }
+        public IEnumerable<INamedItemInfo> Functions => _functions;
+
+        public void WriteToDisk() {
+            if (!_saved) {
+                var filePath = GetCacheFilePath();
+                try {
+                    var dir = Path.GetDirectoryName(filePath);
+                    if (!Directory.Exists(dir)) {
+                        Directory.CreateDirectory(dir);
```
AlexanderSher: Do you want to use `IFileSystem` here?

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -2,257 +2,113 @@
...
-                }
+        private async Task<IEnumerable<string>> GetFunctionNamesAsync() {
+            var cached = TryRestoreFromCache();
+            if (cached == null) {
+                try {
+                    var r = await _host.Session.EvaluateAsync<JArray>(Invariant($"as.list(base::getNamespaceExports('{this.Name}'))"), REvaluationKind.Normal);
+                    return r.Select(p => (string)((JValue)p).Value).ToArray();
+                } catch (MessageTransportException) { } catch (TaskCanceledException) { } catch (REvaluationException) { }
+            } else {
+                _saved = true;
```
AlexanderSher: Why `_saved` is set to `true` here?

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -2,257 +2,113 @@
...
-                    };
-                    functions.Add(niAlias);
-                }
+        private async Task<IEnumerable<string>> GetFunctionNamesAsync() {
+            var cached = TryRestoreFromCache();
+            if (cached == null) {
+                try {
+                    var r = await _host.Session.EvaluateAsync<JArray>(Invariant($"as.list(base::getNamespaceExports('{this.Name}'))"), REvaluationKind.Normal);
+                    return r.Select(p => (string)((JValue)p).Value).ToArray();
+                } catch (MessageTransportException) { } catch (TaskCanceledException) { } catch (REvaluationException) { }
```
AlexanderSher: You don't need to catch `MessageTransportException` anymore

src/R/Support/Impl/Help/IntelliSenseRHost.cs
MikhailArkhipov: He, that was what I started with but it was odd to type session.RSession

src/R/Support/Impl/Help/IntelliSenseRHost.cs
MikhailArkhipov: It can be any session as long as it is not interactive and not heavily loaded by other stuff. Fetching functions and descriptions is pretty taxing so the faster the better.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: Yes but for real testability we need interface for StreamReader/Writer which I am going to add

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: So we won't be re-saving data that already came from the disk cache

src/R/Support/Impl/Help/IntelliSenseRHost.cs
int19h: How taxing is it, actually? If the main problem is that it's delaying other evals, then you could make your evals reentrant, like package manager does for its package fetching - then they can be interrupted by other stuff. But it will block the REPL prompt. So if we're talking hundreds of milliseconds long, it's an option. If it's seconds, we need a separate session.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -2,257 +2,113 @@
...
-                        Description = ni.Description,
-                        ItemType = ni.ItemType
-                    };
-                    functions.Add(niAlias);
-                }
+        private async Task<IEnumerable<string>> GetFunctionNamesAsync() {
+            var cached = TryRestoreFromCache();
+            if (cached == null) {
+                try {
+                    var r = await _host.Session.EvaluateAsync<JArray>(Invariant($"as.list(base::getNamespaceExports('{this.Name}'))"), REvaluationKind.Normal);
```
int19h: If you use `REvaluationKind.BaseEnv`, you don't need `base::` anywhere, and there's reduced risk of other breakages.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -2,257 +2,113 @@
...
+                    var list = new List<string>();
+                    using (var sr = new StreamReader(filePath)) {
+                        while (!sr.EndOfStream) {
+                            list.Add(sr.ReadLine().Trim());
                         }
                     }
+                    return list;
                 }
-            } catch (IOException) { }
+            } catch (IOException) { } catch (AccessViolationException) { }
```
int19h: Did you mean `UnauthorizedAccessException`? `AccessViolationException` is for broken unmanaged pointers.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
int19h: Also, this will get all exports from a given namespace, not just functions. Is that intentional? If so, the names of the methods should probably be changed to match. Otherwise, you'll need to use `Filter` and `is.function` to narrow it down.

src/R/Support/Impl/Help/IntelliSenseRHost.cs
AlexanderSher: My question was more about state of the session. `IRSessionProvider.BeginEvaluationAsync` stops the host after evaluation is released, so if there should be some preloaded data or preserved state from previous calls, that's not an option.

---
In reply to: [71205046](https://github.com/Microsoft/RTVS/pull/2084#discussion_r71205046) [](ancestors = 71205046)

MikhailArkhipov: ~1500 ms to start the host
~2000ms to fetch installed package names and descriptions
~8000 ms to fetch function names

The latter is trimmed to 500ms with caching

MikhailArkhipov: ~200-400ms to fetch function signature/description if local.

All of the above will get longer when R is not local

src/R/Support/Impl/Help/Packages/PackageInfo.cs
AlexanderSher: I thought that you've already used `MemoryStream` for testing somewhere

---
In reply to: [71204260](https://github.com/Microsoft/RTVS/pull/2084#discussion_r71204260) [](ancestors = 71204260)

src/R/Support/Impl/Help/Packages/PackageInfo.cs
AlexanderSher: Got it

---
In reply to: [71204403](https://github.com/Microsoft/RTVS/pull/2084#discussion_r71204403) [](ancestors = 71204403)

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: Hmmmm we use that sequence all over... Needs to be fixed in all places then

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: Yes, everything. It is for the completion list so operators and variables also should appear. I can change class name to PackageExports but then I guess I should change IFunctionIndex -> IExportsIndex or something

src/R/Support/Impl/Help/IntelliSenseRHost.cs
MikhailArkhipov: To @AlexanderSher - If session stops it is probably not good since fetching of signatures and quick infos must be quick, it can't wait for another session to spin up.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: I meant to verify that we did actually write file to disk for example. I used MS in drag-drop tests but that's b/c DD already gets stream from clipboard.

src/R/Support/Impl/Help/IntelliSenseRHost.cs
AlexanderSher: Got it. Then let's use long-running session.

---
In reply to: [71210875](https://github.com/Microsoft/RTVS/pull/2084#discussion_r71210875) [](ancestors = 71210875)

src/R/Support/Impl/Help/IntelliSenseRHost.cs
AlexanderSher: `iSession.RSession` seems file. We already have such code in debugger I think.

---
In reply to: [71202975](https://github.com/Microsoft/RTVS/pull/2084#discussion_r71202975) [](ancestors = 71202975)

src/R/Support/Impl/Help/Packages/PackageIndex.cs
AlexanderSher: I meant that you don't need it at all :)
```{csharp}
var ready = await _buildIndexLock.WaitAsync();
if (!ready) {
    try {
//...
    } finally {
        _buildIndexLock.Release();
    }
}
```

---
In reply to: [71199605](https://github.com/Microsoft/RTVS/pull/2084#discussion_r71199605) [](ancestors = 71199605)

MikhailArkhipov: Ah, OK, it caches the result

AlexanderSher:
:shipit:

PR #2086, opened 2016-07-19
Delete all variables is enabled for .GlobalEnv only Fixes #1899.

### LINKED ISSUE #1899: Delete All Variables always applies to global environment
Even if you have a different environment selected in the dropdown, and VE is showing variables from that environment, the command always removes things from .GlobalEnv.

In most cases, this command doesn't actually have much usefulness outside of global environment, so we should probably just disable it if you have anything other than global selected.

karthiknadig:
Delete all variables is enabled for .GlobalEnv only Fixes #1899.

### DIFF
--- a/src/Package/Impl/DataInspect/Commands/DeleteAllVariablesCommand.cs
+++ b/src/Package/Impl/DataInspect/Commands/DeleteAllVariablesCommand.cs
@@ -8,11 +8,17 @@ using Microsoft.R.Host.Client;
 using Microsoft.VisualStudio.R.Package.Commands;
 using Microsoft.VisualStudio.R.Package.Shell;
 using Microsoft.VisualStudio.R.Packages.R;
+using Microsoft.VisualStudio.R.Package.Utilities;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect.Commands {
     internal sealed class DeleteAllVariablesCommand : SessionCommand {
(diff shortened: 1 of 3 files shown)


src/Package/Impl/DataInspect/Commands/DeleteAllVariablesCommand.cs
```diff
@@ -8,13 +8,19 @@
...
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect.Commands {
     internal sealed class DeleteAllVariablesCommand : SessionCommand {
         public DeleteAllVariablesCommand(IRSession session) :
             base(session, RGuidList.RCmdSetGuid, RPackageCommandId.icmdDeleteAllVariables) {
         }
 
+        protected override void SetStatus() {
+            var variableWindowPane = ToolWindowUtilities.FindWindowPane<VariableWindowPane>(0);
+            Enabled = (variableWindowPane?.IsGlobalREnvironment()) ?? true;
```
int19h: This will also make it enabled when there's no environment - is this desirable?

int19h:
:+1:

src/Package/Impl/DataInspect/Commands/DeleteAllVariablesCommand.cs
karthiknadig: I will update this. I think the button should be enabled only when it can be used.

Issue #2051, opened 2016-07-06
Open local viewer on remote HTML/PDF/Word file after RMD publish 


MikhailArkhipov:
May amount to changing 'open viewer' to send file to local and open it there.

PR #2102, opened 2016-07-20
Publish rmarkdown remotely #2051

karthiknadig:
Added helper to publish r markdown remotely and open the output locally using blobs.


src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
             return true;
         }
 
-        public string GetCommandLine(string inputFile, string outputFilePath, PublishFormat publishFormat, Encoding encoding) {
-            string format = GetDocumentTypeString(publishFormat);
+        public async Task Publish(IRSession session, string inputFilePath, string outputFilePath, PublishFormat publishFormat, Encoding encoding) {
+            // Send the markdown to R-Host for processing as a blob
+            byte[] rmdData = File.ReadAllBytes(inputFilePath);
+            var rmdBlobId = await session.SendBlobAsync(rmdData);
```
MikhailArkhipov: I'd make extension to the session for sending files so it would look like session.SendFileAsync() and  hide reading bytes

MikhailArkhipov: Do we need to catch TaskCanceled, Disconnected, etc?

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+            // Perform markdown rendering on the blob
             string outputFile = Path.GetFileName(outputFilePath);
-            string outputFolder = Path.GetDirectoryName(outputFilePath).Replace('\\', '/');
-            // Run rmarkdown::render
-            return Invariant($"rmarkdown::render('{inputFile}', output_format='{format}', output_file='{outputFile}',  output_dir='{outputFolder}', encoding='cp{encoding.CodePage}')");
+            string format = GetDocumentTypeString(publishFormat);
+            string ext = Path.GetExtension(outputFilePath).Substring(1);
+            var publishResult = await session.EvaluateAsync($"rtvs:::rmarkdown_publish(blob_id = {rmdBlobId}, output_filename = '{outputFile}', output_format='{format}', encoding='cp{encoding.CodePage}')", REvaluationKind.Normal);
+            JObject obj = publishResult.Result as JObject;
+            JArray jsonBlobids = obj["blob.ids"] as JArray;
```
MikhailArkhipov: This code appears to be generic enough and not specific to publishing, maybe we should make a class for packing/unpacking blobs so the whole thing would look like

1. session.SendFiles(...)
2. rmarkdown::publish
3. GetFiles()

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -167,4 +167,24 @@ export_to_pdf <- function(width, height) {
...
+    rmdpath <- tempfile('rmd_', fileext = '.rmd');
+    on.exit(unlink(rmdpath));
+    writeBin(get_blob(blob_id), rmdpath);
+
+    # Create temp directory to collect the generated files
+    output_filepath <- paste0(tempdir(), '/', output_filename);
+
+    rmarkdown::render(rmdpath, output_format = output_format, output_file = output_filepath,  output_dir = tempdir(), encoding = encoding);
+
+	output_id <- create_blob(readBin(output_filepath, 'raw', file.info(output_filepath)$size));
```
int19h: Seems to be some whitespace mixup here - different indentation levels.

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -167,4 +167,24 @@ export_to_pdf <- function(width, height) {
...
+
+	output_files <- list();
```
int19h: You can do this all in one go like so:
```r
... <- list(blob_ids = ..., file_names = ...)
```
Also, underscores rather than dots for identifiers.

src/Package/Impl/Publishing/Definitions/IMarkdownFlavorPublishHandler.cs
```diff
@@ -2,13 +2,15 @@
...
 using Microsoft.Markdown.Editor.Flavor;
+using Microsoft.R.Host.Client;
 
 namespace Microsoft.VisualStudio.R.Package.Publishing.Definitions {
     public interface IMarkdownFlavorPublishHandler {
         MarkdownFlavor Flavor { get; }
         string RequiredPackageName { get; }
         bool FormatSupported(PublishFormat format);
-        string GetCommandLine(string inputFile, string outputFile, PublishFormat publishFormat, Encoding encoding);
+        Task Publish(IRSession session, string inputFilePath, string outputFilePath, PublishFormat publishFormat, Encoding encoding);
```
int19h: `PublishAsync` (since it returns a task)

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
int19h: Let's wait until we have more similar uses before we refactor it into an extension. I suspect most usages will actually just read the blobs directly rather than writing them to disk. @AlexanderSher, did anything similar come up in sync stuff?

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+
+            // Perform markdown rendering on the blob
             string outputFile = Path.GetFileName(outputFilePath);
-            string outputFolder = Path.GetDirectoryName(outputFilePath).Replace('\\', '/');
-            // Run rmarkdown::render
-            return Invariant($"rmarkdown::render('{inputFile}', output_format='{format}', output_file='{outputFile}',  output_dir='{outputFolder}', encoding='cp{encoding.CodePage}')");
+            string format = GetDocumentTypeString(publishFormat);
+            string ext = Path.GetExtension(outputFilePath).Substring(1);
+            var publishResult = await session.EvaluateAsync($"rtvs:::rmarkdown_publish(blob_id = {rmdBlobId}, output_filename = '{outputFile}', output_format='{format}', encoding='cp{encoding.CodePage}')", REvaluationKind.Normal);
+            JObject obj = publishResult.Result as JObject;
```
int19h: Use `(T)` for casting where the cast is not expected to fail (so that it throws early rather than producing `null`).

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+            var rmdBlobId = await session.SendBlobAsync(rmdData);
+
+            // Perform markdown rendering on the blob
             string outputFile = Path.GetFileName(outputFilePath);
-            string outputFolder = Path.GetDirectoryName(outputFilePath).Replace('\\', '/');
-            // Run rmarkdown::render
-            return Invariant($"rmarkdown::render('{inputFile}', output_format='{format}', output_file='{outputFile}',  output_dir='{outputFolder}', encoding='cp{encoding.CodePage}')");
+            string format = GetDocumentTypeString(publishFormat);
+            string ext = Path.GetExtension(outputFilePath).Substring(1);
+            var publishResult = await session.EvaluateAsync($"rtvs:::rmarkdown_publish(blob_id = {rmdBlobId}, output_filename = '{outputFile}', output_format='{format}', encoding='cp{encoding.CodePage}')", REvaluationKind.Normal);
```
int19h: Use `ToRStringLiteral` for all strings, to ensure proper escaping.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+            var data = await session.GetBlobAsync(blobids);
+
+            // Save it locally 
+            string outputFolder = Path.GetDirectoryName(outputFilePath);
+            foreach (Blob blob in data) {
+                string fullPath = Path.Combine(outputFolder, fileNameMap[blob.BlobId]);
+                File.WriteAllBytes(fullPath, blob.Data);
+            }
+
+            // Clean up all the blobs that were created during publishing
```
int19h: This stuff should probably be in a `finally` block, since I/O etc above can throw.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+            long[] blobids2 = new long[blobids.Length + 1];
+            blobids2[0] = rmdBlobId;
+            Array.ConstrainedCopy(blobids, 0, blobids2, 1, blobids.Length);
+            await session.DestroyBlobAsync(blobids);
```
int19h: `blobids2`?

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+
+            // Save it locally 
+            string outputFolder = Path.GetDirectoryName(outputFilePath);
+            foreach (Blob blob in data) {
+                string fullPath = Path.Combine(outputFolder, fileNameMap[blob.BlobId]);
+                File.WriteAllBytes(fullPath, blob.Data);
+            }
+
+            // Clean up all the blobs that were created during publishing
+            long[] blobids2 = new long[blobids.Length + 1];
```
int19h: `Array.Resize` will be a tad shorter and clearer.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
             string outputFile = Path.GetFileName(outputFilePath);
-            string outputFolder = Path.GetDirectoryName(outputFilePath).Replace('\\', '/');
-            // Run rmarkdown::render
-            return Invariant($"rmarkdown::render('{inputFile}', output_format='{format}', output_file='{outputFile}',  output_dir='{outputFolder}', encoding='cp{encoding.CodePage}')");
+            string format = GetDocumentTypeString(publishFormat);
+            string ext = Path.GetExtension(outputFilePath).Substring(1);
+            var publishResult = await session.EvaluateAsync($"rtvs:::rmarkdown_publish(blob_id = {rmdBlobId}, output_filename = '{outputFile}', output_format='{format}', encoding='cp{encoding.CodePage}')", REvaluationKind.Normal);
+            JObject obj = publishResult.Result as JObject;
+            JArray jsonBlobids = obj["blob.ids"] as JArray;
+            JArray jsonFileNames = obj["file.names"] as JArray;
```
MikhailArkhipov: Or at least separate functions. Plus, this needs to be testable so separate methods would be much more useful.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+            }
+
+            // Get the published content from R-Host
+            var data = await session.GetBlobAsync(blobids);
+
+            // Save it locally 
+            string outputFolder = Path.GetDirectoryName(outputFilePath);
+            foreach (Blob blob in data) {
+                string fullPath = Path.Combine(outputFolder, fileNameMap[blob.BlobId]);
+                File.WriteAllBytes(fullPath, blob.Data);
```
MikhailArkhipov: Add method to IFileSystem and pass IFileSystem down instead so it can be substituted in tests

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
MikhailArkhipov: I'd try and do some higher level API. Something along the line of

```
using(var fts = new FileTransferSession(...)) {
   await fts.SendFileAsync(...)
   ...
   await fts.GetFileNames(...)
   await fts.WriteFiles(...)
}
```

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
AlexanderSher: There will be functionality to send specific file to the remote machine, so there will be other scenarios with file reading/writing

---
In reply to: [71459410](https://github.com/Microsoft/RTVS/pull/2102#discussion_r71459410) [](ancestors = 71459410)

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +29,45 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
             return true;
         }
 
-        public string GetCommandLine(string inputFile, string outputFilePath, PublishFormat publishFormat, Encoding encoding) {
-            string format = GetDocumentTypeString(publishFormat);
+        public async Task Publish(IRSession session, string inputFilePath, string outputFilePath, PublishFormat publishFormat, Encoding encoding) {
+            // Send the markdown to R-Host for processing as a blob
+            byte[] rmdData = File.ReadAllBytes(inputFilePath);
```
AlexanderSher: `try/catch` + `IFileSystem`

karthiknadig:
I think we can add a file transfer service/helper when we have a better picture of how the sync works. For now I added extension methods on IRSession to do this.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -29,45 +30,48 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
-            string format = GetDocumentTypeString(publishFormat);
-            string ext = Path.GetExtension(outputFilePath).Substring(1);
-            var publishResult = await session.EvaluateAsync($"rtvs:::rmarkdown_publish(blob_id = {rmdBlobId}, output_filename = '{outputFile}', output_format='{format}', encoding='cp{encoding.CodePage}')", REvaluationKind.Normal);
-            JObject obj = publishResult.Result as JObject;
-            JArray jsonBlobids = obj["blob.ids"] as JArray;
-            JArray jsonFileNames = obj["file.names"] as JArray;
+            try {
+                // Send the markdown to R-Host for processing as a blob
+                rmdBlobId = await session.SendFileAsBlobAsync(fs, inputFilePath);
+                if (rmdBlobId <= 0) {
```
int19h: Does this indicate an error? I would expect the managed API to use exceptions for these.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -29,45 +30,48 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
-                string fullPath = Path.Combine(outputFolder, fileNameMap[blob.BlobId]);
-                File.WriteAllBytes(fullPath, blob.Data);
+                // Get the published content from R-Host
+                outputRmdBlobId = jsonBlobids[0].Value<long>();
+                string fullOutputFilePath = Path.Combine(Path.GetDirectoryName(outputFilePath), jsonFileNames[0].Value<string>());
+                await session.GetFileFromBlobAsync(fs, outputRmdBlobId, fullOutputFilePath);
+
+            } catch (IOException ex) {
+                coreShell.ShowErrorMessage(ex.Message);
+            } catch (RException) {
```
int19h: We should probably say something to the user when this happens. This would indicate that the R expression doing the export has failed with an error, which can also happen in all the usual corner cases (I/O failure etc).

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -29,45 +30,48 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+            } catch (OperationCanceledException) {
+            } finally {
+                // Clean-up blobs that are no longer needed
+                await session.DestroyBlobAsync(new long[] { rmdBlobId, outputRmdBlobId });
```
int19h: Need to catch OCE on this as well, in case host goes down while we're cleaning up. You can use `SilenceException` for that to avoid another try/catch.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +30,48 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+                if (rmdBlobId <= 0) {
+                    return;
+                }
+
+                // Perform markdown rendering on the blob
+                JObject jsonObj = await RMarkdownRenderAsync(session, rmdBlobId, outputFilePath, publishFormat, encoding);
+
+                // Markdown results should be in a single blob (file) output. The response for markdown rendering should contain 
+                // the blob id and the output file name.
+                JArray jsonBlobids = (JArray)jsonObj["blob_ids"];
```
MikhailArkhipov: This is still too much details. Why does publishing have to know all the gory details about JSON, specific property names and so on. 

There is nothing more permanent than temporary changes. Think about deployment of files on F5 or implementation of rtvs::fetch that we need to provide so users can get their files. I'd rather do proper file transfer service.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +30,48 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+
+            try {
+                // Send the markdown to R-Host for processing as a blob
+                rmdBlobId = await session.SendFileAsBlobAsync(fs, inputFilePath);
+                if (rmdBlobId <= 0) {
+                    return;
+                }
+
+                // Perform markdown rendering on the blob
+                JObject jsonObj = await RMarkdownRenderAsync(session, rmdBlobId, outputFilePath, publishFormat, encoding);
```
MikhailArkhipov: This should return opaque class that would hide JSON details. Code can hold on it and then pass it over to another method that would create file on disk and return the path

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
int19h: Working with JSON on this side (and creating it on R side) allows us to avoid putting all that functionality into the host and the broker, thereby keeping them simple, and allowing to change this code without breaking compatibility. That was the whole point of introducing blobs as a first-class entity - it's low-level enough that anything (not just file transfer) can be done with it + eval.

MikhailArkhipov: That's fine. But there should be another layer hiding blobs from the high level code. This code should not care how exactly transfer is implemented.

MikhailArkhipov: API calls should look like
```
var token = session.SendFile(...) // don't mention blobs, it is implementation detail.
var result = RMarkdownRenderAsync(token, ...)
var filePath = result.GetFilePath()
```

said ```token``` can be a class that hides necessary information.
Ex compare StreamReader/Writer and Win32 ```CreateFile```

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
karthiknadig: Looking again this is not needed. I will remove that check.

MikhailArkhipov:
👍

src/Host/Client/Impl/Definitions/IRBlobData.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Client {
+    public interface IRBlobData : IRBlob {
+        RBlobKind Kind { get; }
```
int19h: Is this actually necessary? It feels like it simply duplicates the interface hierarchy - i.e. one can check whether it's a file or not via `... is IRBlobFile`.

karthiknadig: I agree. I will remove it.

src/Host/Client/Impl/Definitions/IRBlob.cs
```diff
@@ -0,0 +1,8 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Client {
+    public interface IRBlob {
```
int19h: If this implements `Dispose` and destroys the blob inside that, it can be conveniently used with `using`.

src/Host/Client/Impl/Definitions/FileTransferSession.cs
```diff
@@ -0,0 +1,59 @@
...
+            var result = await _session.GetBlobAsync(new long[] { blob.Id });
+            _fs.FileWriteAllBytes(filePath, result[0].Data);
+
+            if (doCleanUp) {
+                _cleanup.Add(blob);
+            }
+        }
+
+        public async void Dispose() {
+            await _session.DestroyBlobAsync(_cleanup.Where(b => { return b != null; }).Select(b => b.Id)).SilenceException<OperationCanceledException>();
```
int19h: You can use `DoNotWait` here to avoid `async` and `await`. It will automatically silence OCE for you, too.

int19h: No need for `{}` and `return`, since it's a single expression.

int19h: Actually, can `_cleanup` ever contain nulls? It looks like you always add a freshly new'd object to it, so it should be null-free.

src/Host/Client/Impl/Definitions/FileTransferSession.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Common.Core;
+using Microsoft.Common.Core.IO;
+using System.Threading.Tasks;
+using System.Linq;
+
+namespace Microsoft.R.Host.Client {
+    public class FileTransferSession : IDisposable {
+        private readonly IRSession _session;
+        private readonly IFileSystem _fs;
+        private List<IRBlob> _cleanup;
```
int19h: `readonly ... = new List ...`

src/Host/Client/Impl/Definitions/IRBlob.cs
int19h: I think `IRBlobInfo` would be a more appropriate name (and `IRBlobData` could then become `IRBlob`). It would indicate that this is an object that _describes_ a blob, but doesn't actually contain one; whereas the actual blob can be both described and accessed. It would also make things more consistent where it meshes with files - e.g. right now you have `IRBlobFileInfo` derived from `IRBlob`, but with this change it'll be `IRBlobFileInfo : IRBlobInfo`.

src/Host/Client/Impl/Definitions/IRBlobFile.cs
```diff
@@ -0,0 +1,8 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Client {
+    public interface IRBlobFile : IRBlobData {
```
int19h: Should this require `IRBlobFileInfo` as well?

src/Host/Client/Impl/Definitions/BlobData.cs
```diff
@@ -0,0 +1,36 @@
...
+        private readonly byte[] _data;
+        private readonly RBlobKind _kind;
+
+        public BlobData(long id, byte[] data, RBlobKind kind = RBlobKind.Unspecified) {
+            _id = id;
+            _data = data;
+            _kind = kind;
+        }
+
+        public long Id {
```
int19h: Nit: you can use `=> _id` to shorten one-line accessors like these.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +30,26 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
 
-        public string GetCommandLine(string inputFile, string outputFilePath, PublishFormat publishFormat, Encoding encoding) {
-            string format = GetDocumentTypeString(publishFormat);
-            string outputFile = Path.GetFileName(outputFilePath);
-            string outputFolder = Path.GetDirectoryName(outputFilePath).Replace('\\', '/');
-            // Run rmarkdown::render
-            return Invariant($"rmarkdown::render('{inputFile}', output_format='{format}', output_file='{outputFile}',  output_dir='{outputFolder}', encoding='cp{encoding.CodePage}')");
+        public async Task PublishAsync(IRSession session, ICoreShell coreShell,  IFileSystem fs, string inputFilePath, string outputFilePath, PublishFormat publishFormat, Encoding encoding) {
+            try {
+                using(var fts = new FileTransferSession(session, fs)) {
```
int19h: Nit: formatting/whitespace.

src/Package/Impl/Publishing/Handlers/RmdPublishHandler.cs
```diff
@@ -25,12 +30,26 @@ internal sealed class RmdPublishHandler : IMarkdownFlavorPublishHandler {
...
+                    var rmd = await fts.SendFileAsync(inputFilePath);
+                    var result = await RMarkdownRenderAsync(session, rmd, outputFilePath, publishFormat, encoding);
+                    await fts.FetchFileAsync(result, outputFilePath);
```
int19h: We currently compute `outputFilePath` as `inputFilePath` with a different extension (see the call site for `PublishAsync`). But this only works when both input and output files are on the same local filesystem (indeed, we get the input path from the text buffer, which will always be local). As soon as local and remote are different, this will break. 

Furthermore, we currently use `outputFilePath` for two completely different things. First, we pass it to `RMarkdownRenderAsync`, which will use it as a destination on the remote file system to write the output to. Then we fetch the produced file as a blob, and write it to `outputFilePath` onto the local file system. That doesn't make sense to me.

It feels like the remote path is actually irrelevant, and should just be another temp filename, like the input `rmd`, since we immediately load that file into a blob anyway.

int19h: In fact, do we actually derive any benefit from having `RMarkdownRenderAsync` return just the blob ID (with file info), rather than the blob, given that it unconditionally creates a blob anyway? It seems that there's no reason for the caller to call it without calling `FetchFileAsync` later to get the data. And if they forget to call it, they will leave a dangling blob in the host, so potentially it's dangerous, too.

@MikhailArkhipov, what do you think? Is there value in maintaining the split here? Or would it be better to change `RMarkdownRenderAsync` to just take input and output paths (on local FS), and create and use a `FileTransferSession` inside as an implementation detail?

int19h: To clarify, I think that there are two valid approaches here.

First one is when `RMarkdownRenderAsync` is strictly a remote-side affair: it takes a remote path for input file, another remote path for output file, runs markdown conversion, and that's that. If the caller needs to do it for local files, the onus is on them to use `FileTransferSession` to upload the input file before the conversion, and to download the output after. In this approach, `RMarkdownRenderAsync` doesn't do anything with blobs at all. This is more low-level, but provides more flexibility (e.g. input might be generated by some R code directly on remote FS, rather than uploaded).

The second approach is when `RMarkdownRenderAsync` works on the local side: i.e. it takes a local path for input file, and a local path for output file, and does all the magic that needs to happen to place the input on the remote FS, run conversion, and get the output back. So it uses `FileTransferSession` and blobs internally, but there's nothing in the API itself that indicates that. The returned task simply indicates completion, and is not a blob ID or anything like that - when it's complete, you just use the output file that you've specified. This is more high-level, and is easier to use, at the expense of flexibility.

MikhailArkhipov: Probably second. Project system is local and text buffer is local so publishing call should indeed hide sending files back and forth.

int19h:
:+1:

PR #2107, opened 2016-07-20
Namespaces cleanup

MikhailArkhipov:
Mostly removed .Definitions namespace from R.Core and R.Editor. No code logic change. Reformat few files and added test categories to some tests that lacked them


AlexanderSher:
:shipit:

PR #2115, opened 2016-07-25
Fixes #2114 - a crash when CSV export is not possible.

### LINKED ISSUE #2114: Export to CSV should not be available on types that cannot be coerced to data.frames
`m` should not have the export to Excel button on it.
image

When I click the button, write.table fails to coerce `m` to `data.frame`. 
```
Error in as.data.frame.default(x[[i]], optional = TRUE, stringsAsFactors = stringsAsFactors) : 
  cannot coerce class ""lm"" to a data.frame
```  

This later causes a crash in RTVS when we attempt save the exported content (which is NULL) to the file.

karthiknadig:
Added a bool filed to RValueInfo to indicate if a variable can be exported to CSV.

### DIFF
--- a/src/Host/Client/Impl/DataInspection/REvaluationResultProperties.cs
+++ b/src/Host/Client/Impl/DataInspection/REvaluationResultProperties.cs
@@ -41,23 +41,45 @@ namespace Microsoft.R.DataInspection {
         FlagsProperty = 1 << 10,
         /// <seealso cref="IRActiveBindingInfo.ComputedValue"/>
         ComputedValueProperty = 1 << 11,
+        /// <seealso cref="IRValueInfo.CanExportToCSV"/>
+        CanExportToCsvProperty = 1 << 12,
         /// <seealso cref="IRValueInfo.HasChildren"/>
         HasChildrenProperty = ExpressionProperty | LengthProperty | AttributeCountProperty | SlotCountProperty | NameCountProperty | FlagsProperty,
     }
 
+    internal class REvaluationResultFieldNames {
+        public const string ReprFieldName = "repr";
+        public const string ExpressionFieldName = "expression";
+        public const string AccessorKindFieldName = "kind";
+        public const string TypeFieldName = "type";
+        public const string ClassesFieldName = "classes";
+        public const string LengthFieldName = "length";
+        public const string SlotCountFieldName = "slot_count";
+        public const string AttributeCountFieldName = "attr_count";
+        public const string NameCountFieldName = "name_count";
+        public const string DimFieldName = "dim";
+        public const string FlagsFieldName = "flags";
+        public const string ComputedValueFieldName = "computed_value";
+        public const string CanExportToCsvFieldName = "to_csv";
+        public const string ErrorFieldName = "error";
+        public const string PromiseFieldName = "promise";
+        public const string ActiveBindingFieldName = "active_binding";
+    }
+    
     internal static class REvaluationResultPropertiesExtensions {
         private static readonly Dictionary<REvaluationResultProperties, string> _mapping = new Dictionary<REvaluationResultProperties, string> {
-            [ExpressionProperty] = "expression",
-            [AccessorKindProperty] = "kind",
-            [TypeNameProperty] = "type",
-            [ClassesProperty] = "classes",
-            [LengthProperty] = "length",
-            [SlotCountProperty] = "slot_count",
-            [AttributeCountProperty] = "attr_count",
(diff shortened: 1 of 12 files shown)


src/Host/Client/Impl/DataInspection/REvaluationResultProperties.cs
```diff
@@ -41,23 +41,45 @@ public enum REvaluationResultProperties : ulong {
         FlagsProperty = 1 << 10,
         /// <seealso cref="IRActiveBindingInfo.ComputedValue"/>
         ComputedValueProperty = 1 << 11,
+        /// <seealso cref="IRValueInfo.CanCoerceToDataFrame"/>
+        CanCoerceToDataFrameProperty = 1 << 12,
         /// <seealso cref="IRValueInfo.HasChildren"/>
         HasChildrenProperty = ExpressionProperty | LengthProperty | AttributeCountProperty | SlotCountProperty | NameCountProperty | FlagsProperty,
     }
 
+    internal class REvaluationResultFieldNames {
```
int19h: Let's make it a nested class inside `REvaluationResult` (can use `partial` to keep it in a separate file), rename it to `FieldNames`, and drop the `FieldName` prefix as redundant. So external clients would do `REvaluationResult.FieldNames.Repr` etc, and inside the class and its children, you can just do `FieldNames.Repr`.

src/Host/Client/Impl/rtvs/R/eval.R
```diff
@@ -67,6 +67,17 @@ describe_object <- function(obj, res, fields, repr = NULL) {
...
+  
+  if (field('to_df')) {
+    can_coerce_to_df <- function(obj){
+      df_test <- function(objclass){
+        res <- getS3method('as.data.frame', objclass, optional = TRUE);
+        if (is.null(res)) { FALSE } else { TRUE }
+      }
+      any(as.logical(lapply(class(obj), df_test)), na.rm = TRUE);
+    }
+    res$to_df <- FALSE_if_error(can_coerce_to_df(obj));
```
int19h: For a more compact representation, use `NULL` here if there is an error, and then on the other side, when parsing JSON, assume that missing field represents the default value of false. I believe other boolean flags here use this approach.

src/Host/Client/Impl/rtvs/R/eval.R
```diff
@@ -67,6 +67,17 @@ describe_object <- function(obj, res, fields, repr = NULL) {
       res$dim <- as.list(dim);
     }
   }
+  
+  if (field('to_df')) {
+    can_coerce_to_df <- function(obj){
+      df_test <- function(objclass){
+        res <- getS3method('as.data.frame', objclass, optional = TRUE);
+        if (is.null(res)) { FALSE } else { TRUE }
```
int19h: `!is.null`

src/Host/Client/Impl/rtvs/R/eval.R
```diff
@@ -67,6 +67,17 @@ describe_object <- function(obj, res, fields, repr = NULL) {
...
+      }
+      any(as.logical(lapply(class(obj), df_test)), na.rm = TRUE);
```
int19h: Do we actually get any `NA` to remove?

src/Package/Impl/DataInspect/VariableViewModel.cs
```diff
@@ -59,7 +59,7 @@ public sealed class VariableViewModel : RSessionDataObject, IIndexedItem {
             }
 
             var tableCaps = (ViewerCapabilities.Table | ViewerCapabilities.List);
-            CanShowOpenCsv = CanShowDetail && (_detailsViewer.Capabilities & tableCaps) != 0 && result.Length > 0;
+            CanShowOpenCsv = CanShowDetail && result.CanCoerceToDataFrame && (_detailsViewer.Capabilities & tableCaps) != 0 && result.Length > 0;
```
int19h: While you're here, can you also change the use of `&` to `Enum.HasFlag` for the sake of better clarity, and `tableCaps` to const?

src/Host/Client/Impl/rtvs/R/eval.R
int19h: Also, why is `as.logical` needed here? Aren't we already producing a vector of booleans?

int19h: Ah, I see, it always returns a list. `sapply` or `vapply` instead of `lapply` should do the trick.

int19h:
:+1:

Issue #2082, opened 2016-07-15
Optimize protocol for better broker performance


int19h:
In a multi-user system, broker becomes the weak link, as it is handling all packets for all active sessions of all users. It is important that the protocol between the host and the broker, and the broker and the clients, is designed such that it minimizes the amount of time spent handling each packet.

Currently, the protocol is suboptimal in that regard, because message headers are combined with message payload in a single JSON document, such that getting to message name, id etc requires parsing the whole thing. For expensive messages, like eval requests for VE and data grid, this means spending a lot of time just parsing.

We should split the message metadata from the message data in such a way that the latter is not needed to be interpreted by the broker in any way to process and tunnel messages. 

Furthermore, in light of #2081, since we'll need some binary data to delineate messages on the pipes anyway, we might as well switch to fixed-length binary encoding for message metadata, which will further simplify header parsing.

Also need investigate using BSON for JSON payload, to reduce channel saturation.

PR #2116, opened 2016-07-26
Optimize protocol for better broker performance [RTVS]

int19h:
https://github.com/Microsoft/RTVS/issues/2082

Switch protocol to use a single binary websocket packet for every message, to allow message header to be inspected without the overhead JSON parsing.

Refactor request handling.


int19h:
Corresponding R-Host change:
https://github.com/Microsoft/R-Host/pull/73

int19h:
Ping.

src/Host/Client/Impl/Definitions/IRBlobService.cs
```diff
@@ -7,8 +7,8 @@
 
 namespace Microsoft.R.Host.Client {
     public interface IRBlobService {
-        Task<long> SendBlobAsync(byte[] data, CancellationToken cancellationToken = default(CancellationToken));
-        Task<IReadOnlyList<IRBlob>> GetBlobAsync(IEnumerable<long> blobIds, CancellationToken cancellationToken = default(CancellationToken));
-        Task DestroyBlobAsync(IEnumerable<long> blobIds, CancellationToken cancellationToken = default(CancellationToken));
+        Task<ulong> CreateBlobAsync(byte[] data, CancellationToken cancellationToken = default(CancellationToken));
+        Task<byte[]> GetBlobAsync(ulong blobId, CancellationToken cancellationToken = default(CancellationToken));
+        Task DestroyBlobAsync(ulong[] blobIds, CancellationToken cancellationToken = default(CancellationToken));
```
karthiknadig: Why change this from IEnumerable to []?

src/Host/Client/Impl/Definitions/Message.cs
```diff
@@ -0,0 +1,185 @@
...
+                offset += sizeof(ulong);
+
+                RequestId = BitConverter.ToUInt64(data, offset);
+                offset += sizeof(ulong);
+
+                int term = Array.IndexOf<byte>(data, 0, offset);
+                if (term < 0) {
+                    throw new IndexOutOfRangeException();
+                }
+                Name = Encoding.UTF8.GetString(data, offset, term - offset);
```
karthiknadig: Should we consider using ulong for message type?

src/Host/Client/Impl/Definitions/REvaluationResult.cs
```diff
@@ -56,22 +59,28 @@ public struct REvaluationResult {
...
-            Raw = raw;
+            RawResult = rawResult;
         }
 
         public override string ToString() {
-            var sb = new StringBuilder((Result ?? "").ToString());
+            var sb = new StringBuilder();
+
+            if (RawResult != null) {
+                sb.Append("<raw>");
```
karthiknadig: Number of bytes along with raw might be helpful. <raw>[1234]

src/Host/Client/Impl/Extensions/RawEvalResultExtensions.cs
```diff
@@ -7,16 +7,10 @@
...
-        }
-
         public static void SaveRawDataToFile(this REvaluationResult result, string filepath) {
-            if (result.HasRawData()) {
+            if (result.RawResult != null) {
                 using (BinaryWriter writer = new BinaryWriter(File.OpenWrite(filepath))) {
-                    foreach (var data in result.Raw) {
-                        writer.Write(data);
-                    }
+                    writer.Write(result.RawResult);
```
karthiknadig: File.WriteAllBytes, I had used the binarywriter since there was a possibility that we send single file as multiple blobs.

src/Host/Client/Impl/Definitions/IRBlobService.cs
int19h: No particular reason, I just happened to fork before the change from array to enumerable happened in master. I'll change it to enumerable.

I'll also rename the method to `DestroyBlobsAsync`, since it's operating on multiple blobs.

src/Host/Client/Impl/Definitions/Message.cs
int19h: I started down that path, but there were more extensive changes on the host side for this (and other things) that I didn't want to get into, given the tight schedule.

For now, I think this will do even if we do need super-fast name checking in the broker, because the name is positioned at a fixed location inside the message, and most interesting messages have short names. So one can cast the pointer to name to a pointer to some appropriate integer type and compare (e.g. if you need to spot evals, you'd cast to int16 and compare against `?=`).

int19h:
Addressed feedback (going to rebase right before merging, so as to not mess up comments).

int19h:
Rebase was small enough, so I went ahead with it. All ready to merge as soon as I have sign-off.

karthiknadig:
:+1:

src/Host/Client/Impl/Extensions/RawEvalResultExtensions.cs
```diff
@@ -7,18 +7,9 @@
...
-            if (result.HasRawData()) {
-                using (BinaryWriter writer = new BinaryWriter(File.OpenWrite(filepath))) {
-                    foreach (var data in result.Raw) {
-                        writer.Write(data);
-                    }
-                    writer.Flush();
-                }
+        public static void SaveRawDataToFile(this REvaluationResult result, string filePath) {
+            if (result.RawResult != null) {
+                File.WriteAllBytes(filePath, result.RawResult);
```
AlexanderSher: Any try/catch required?

int19h:
Thanks!

src/Host/Client/Impl/Extensions/RawEvalResultExtensions.cs
int19h: Not here for sure, since it's just a thin wrapper (actually, I don't even know if this has much value now - before it dealt with multiple blobs, but now it's just `WriteAllBytes` under a different name). But call sites should be validated.

PR #2117, opened 2016-07-27
Remote package lock state check

karthiknadig:
Moved the code that checks for the package lock state to R-host. Added a rtvs util function to retrieve lock state. Generalized the search for dlls in a library.

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/util.R
+++ b/src/Host/Client/Impl/rtvs/R/util.R
@@ -198,3 +198,24 @@ rmarkdown_publish <- function(blob_id, output_format, encoding) {
     rmarkdown::render(rmdpath, output_format = output_format, output_file = output_filepath,  output_dir = tempdir(), encoding = encoding);
     readBin(output_filepath, 'raw', file.info(output_filepath)$size);
 }
+
+package_lock_state <- function(package_name, package_lib) {
+    file_path_i386 <- paste0(package_lib, '/', package_name, '/libs/i386/', package_name, '.dll');
+    file_path_x64 <- paste0(package_lib, '/', package_name, '/libs/x64/', package_name, '.dll');
+    
+    lock_state <- if (file.exists(file_path_x64)) {
+        call_embedded('get_package_lock_state', file_path_x64);
+    } else {
+        'unlocked'
+    }
+
+    if (identical(lock_state, 'unlocked')) {
+        lock_state <- if (file.exists(file_path_i386)) {
+        call_embedded('get_package_lock_state', file_path_i386);
+        } else {
+            'unlocked'
+        }
+    }
+
+    lock_state
+}
--- a/src/Host/Client/Impl/rtvs/R/util.R
+++ b/src/Host/Client/Impl/rtvs/R/util.R
@@ -200,16 +200,23 @@ rmarkdown_publish <- function(blob_id, output_format, encoding) {
 }
 
 package_lock_state <- function(package_name, lib_path) {
-    dlls <- dir(path = paste0(lib_path,'/', package_name), pattern = '.dll', recursive = TRUE, ignore.case = TRUE, full.names = TRUE)
-    
-    lock_state <- 'unlocked'
-    for(dll in dlls) {
-        if(identical(lock_state, 'unlocked')) {
-            lock_state <- call_embedded('get_package_lock_state', dll);
-        } else {
-            break;
-        }
+    files <- dir(path = paste0(lib_path,'/', package_name), pattern = '\\.', recursive = TRUE, ignore.case = TRUE, full.names = TRUE)
+    lock_state <- call_embedded('get_file_lock_state', files);
+    factor(lock_state, levels = c(0, 1, 2), labels = c('unlocked', 'locked_by_r_session', 'locked_by_other'));
+}
+
+package_uninstall <- function(package_name, lib_path) {
+    lock_state <- package_lock_state(package_name, lib_path);
+    if (lock_state == 'unlocked') {
+        remove.packages(package_name, lib = lib_path);
     }
-    
+    lock_state
+}
+
+package_update <- function(package_name, lib_path) {
+    lock_state <- package_uninstall(package_name, lib_path)
+    if (lock_state == 'unlocked') {
+        install.packages(package_name, lib = lib_path);
+    }
     lock_state
 }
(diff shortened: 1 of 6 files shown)


src/R/Components/Impl/PackageManager/IRPackageManager.cs
```diff
@@ -105,6 +105,6 @@ public interface IRPackageManager : IDisposable {
         /// <param name="name">Package name.</param>
         /// <param name="libraryPath">Library path (in any format).</param>
         /// <returns>Lock state.</returns>
-        PackageLockState GetPackageLockState(string name, string libraryPath);
+        Task<PackageLockState> GetPackageLockState(string name, string libraryPath);
```
int19h: `...Async`

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -144,21 +144,15 @@ internal class RPackageManager : IRPackageManager {
...
-        public PackageLockState GetPackageLockState(string name, string libraryPath) {
-            string dllPath = GetPackageDllPath(name, libraryPath);
-            if (!string.IsNullOrEmpty(dllPath)) {
-                var processes = RestartManager.GetProcessesUsingFiles(new string[] { dllPath }).ToArray();
-                if (processes != null) {
-                    if (processes.Length == 1 && processes[0].Id == _interactiveWorkflow.RSession.ProcessId) {
-                        return PackageLockState.LockedByRSession;
-                    }
+        public async Task<PackageLockState> GetPackageLockState(string name, string libraryPath) {
+            var pkgStateresult =  await _interactiveWorkflow.RSession.EvaluateAsync($"rtvs:::package_lock_state({name.ToRStringLiteral()}, {libraryPath.ToRStringLiteral()})", REvaluationKind.Normal);
```
int19h: Use `EvaluateAsync<string>()` extension method to get a typed value back directly.

Better yet, you can use [this](http://www.newtonsoft.com/json/help/html/t_newtonsoft_json_converters_stringenumconverter.htm), and then do `EvaluateAsync<PackageLockState>()`. If you change `rtvs:::package_lock_state` to return a factor instead of a string, then you won't even need the attribute, since enum-as-int is default serialization behavior.

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -194,3 +194,18 @@ rmarkdown_publish <- function(blob_id, output_format, encoding) {
     rmarkdown::render(rmdpath, output_format = output_format, output_file = output_filepath,  output_dir = tempdir(), encoding = encoding);
     readBin(output_filepath, 'raw', file.info(output_filepath)$size);
 }
+
+package_lock_state <- function(package_name, lib_path) {
```
AlexanderSher: If you're making this remote, it will be better to add another 2 methods here - `package_install` and `package_uninstall` and call this function from them directly. Both functions can return if they have succeeded or failed because of the lock.

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -68,17 +68,31 @@ internal class RPackageManager : IRPackageManager {
...
                 using (var request = await _interactiveWorkflow.RSession.BeginInteractionAsync()) {
-                    if (string.IsNullOrEmpty(libraryPath)) {
-                        await request.UninstallPackageAsync(name);
-                    } else {
-                        await request.UninstallPackageAsync(name, libraryPath);
+                    using (var eval = await _interactiveWorkflow.RSession.BeginEvaluationAsync()) {
+                        return await eval.EvaluateAsync<PackageLockState>($"rtvs:::package_uninstall({name.ToRStringLiteral()}, {libraryPath.ToRStringLiteral()})", REvaluationKind.Normal);
+                    }
+                }
+            } catch (RHostDisconnectedException ex) {
```
AlexanderSher: You don't need to catch this exception. I have a changeset that removes such code in other methods

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -68,17 +68,31 @@ internal class RPackageManager : IRPackageManager {
...
+                throw new RPackageManagerException(Resources.PackageManager_TransportError, ex);
+            }
+        }
+
+        public async Task<PackageLockState> UpdatePackageAsync(string name, string libraryPath) {
+            if (!_interactiveWorkflow.RSession.IsHostRunning) {
```
AlexanderSher: You don't need this check

karthiknadig:
Ping

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -198,3 +198,25 @@ rmarkdown_publish <- function(blob_id, output_format, encoding) {
...
+    files <- dir(path = paste0(lib_path,'/', package_name), pattern = '\\.', recursive = TRUE, ignore.case = TRUE, full.names = TRUE)
+    lock_state <- call_embedded('get_file_lock_state', files);
+    factor(lock_state, levels = c(0, 1, 2), labels = c('unlocked', 'locked_by_r_session', 'locked_by_other'));
```
int19h: Can you make a helper "definition" for this outside of the function, so that it's easy to locate and reuse? Something like this:
```r
as.lock_state <- function(x)
  factor(x, levels = c(0, 1, 2), labels = c('unlocked', 'locked_by_r_session', 'locked_by_other'));
```
and then it can be used like the standard conversion functions:
```r
as.lock_state(state)
```
And this can be our standard pattern for enums in R.

src/R/Components/Test/PackageManager/PackageManagerIntegrationTest.cs
```diff
@@ -269,8 +269,10 @@ public class PackageManagerIntegrationTest : IAsyncLifetime {
             var abn = pkgs.Should().ContainSingle(pkg => pkg.Package == "abn").Which;
             var cairo = pkgs.Should().ContainSingle(pkg => pkg.Package == "Cairo").Which;
 
-            _workflow.Packages.GetPackageLockState(abn.Package, abn.LibPath).Should().Be(PackageLockState.LockedByRSession);
-            _workflow.Packages.GetPackageLockState(cairo.Package, cairo.LibPath).Should().Be(PackageLockState.LockedByRSession);
+            var abn_state = await _workflow.Packages.GetPackageLockStateAsync(abn.Package, abn.LibPath);
```
int19h: camelCase

int19h:
:+1:

int19h:
:+1:

PR #2122, opened 2016-07-28
SQL stored procedure development & deployment

MikhailArkhipov:
Demo video is incoming. Meanwhile

1. Template with nested .Template.sql and .Query.sql files for the sproc boilerplate and the query editor
2. Publishing dialog, supports R placement inline (in ```@script```) and in a separate table with fetching code at the execution time
3. Tests
4. Also: cleaned up WPF resources names (i.e. ResourceKey has 'Key' suffix) and fixed theming of buttons


src/Package/Impl/ProjectSystem/ProjectSettings.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.Common.Core;
+
+namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
+    internal static class ProjectSettings {
+        public const string SettingsFilePattern = "*settings*.r";
```
int19h: It seems that the pattern doesn't quite match the method below (it would match e.g. xsettingsy.r, but the method will return false).

src/Package/Impl/Sql/Commands/PublishSProcCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
+using Microsoft.Common.Core;
+using Microsoft.Common.Core.Shell;
+using Microsoft.Languages.Core.Settings;
+using Microsoft.VisualStudio.R.Package.Commands;
+using Microsoft.VisualStudio.R.Package.ProjectSystem;
+using Microsoft.VisualStudio.R.Package.Sql.Publish;
+using Microsoft.VisualStudio.Shell.Interop;
```
int19h: This looks like it could use some "remove and sort" love.

src/Package/Impl/Sql/Commands/PublishSProcCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+    [ExportCommandGroup("AD87578C-B324-44DC-A12A-B01A6ED5C6E3")]
+    [AppliesTo(ProjectConstants.RtvsProjectCapability)]
+    internal sealed class PublishSProcCommand : ICommandGroupHandler {
+        private readonly ICoreShell _coreShell;
+        private readonly IProjectSystemServices _pss;
+        private readonly IWritableSettingsStorage _settingsStorage;
+
+        [ImportingConstructor]
+        public PublishSProcCommand(ICoreShell coreShell, IProjectSystemServices pss, [Import(AllowDefault = true)]IWritableSettingsStorage settingsStorage) {
+            if (settingsStorage == null) {
```
int19h: Does every user of `IWritableSettingsStorage` have to do this check and fallback, and if so, how many do we have? Is this worth turning into a helper? Maybe an extension on `ICoreShell`?

src/Package/Impl/Sql/Commands/PublishSProcCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+        private void Handle() {
+            var project = _pss.GetSelectedProject<IVsHierarchy>()?.GetDTEProject();
+            if (project != null) {
+                if (SqlPublishDialogViewModel.GetDatabaseProjectsInSolution(_pss).Count > 0) {
+                    var rFiles = _pss.GetProjectFiles(project).Where(x => Path.GetExtension(x).EqualsIgnoreCase(".R"));
+                    var sqlFiles = new HashSet<string>(_pss.GetProjectFiles(project).Where(x => Path.GetExtension(x).EqualsIgnoreCase(".sql")));
+                    var sprocFiles = rFiles.Where(x =>
+                                sqlFiles.Contains(x.ToQueryFilePath(), StringComparer.OrdinalIgnoreCase) &&
+                                sqlFiles.Contains(x.ToSProcFilePath(), StringComparer.OrdinalIgnoreCase));
+                    if (sprocFiles.FirstOrDefault() != null) {
```
int19h: `if (sprocFiles.Any())`

src/Package/Impl/Sql/DragDrop/DataObject.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Languages.Editor.DragDrop;
+
+namespace Microsoft.VisualStudio.R.Package.Sql.DragDrop {
+    public static class DataObject {
+        public static string GetPlainText(this IDataObject dataObject) {
+            var flags = dataObject.GetFlags();
+            if ((flags & DataObjectFlags.ProjectItems) != 0) {
+                var item = dataObject.GetProjectItems().FirstOrDefault();
+                if (item != null && Path.GetExtension(item.FileName).EqualsIgnoreCase(".r")) {
+                    return GetFileContent(item.FileName).Replace("'", "''");
```
int19h: Will it gracefully handle everything else other than quotes? I'm particularly curious about `\0` and other chars in 0-31 range.

src/Package/Impl/Sql/DragDrop/DataObject.cs
```diff
@@ -0,0 +1,31 @@
...
+                }
+            }
+            return string.Empty;
+        }
+
+        private static string GetFileContent(string file) {
+            try {
+                return File.ReadAllText(file).Trim();
```
int19h: This will read the file as UTF-8 (I think it'll also check for UTF-16 BOMs). Aren't .R files normally stored on disk in CP_ACP?

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+            CreateStoredProcedures(settings, targetProject, targetFolder, targetProjectItem);
+        }
+
+        /// <summary>
+        /// Create SQL file that defines table template that will hold R code
+        /// </summary>
+        private void CreateRCodeTable(SqlSProcPublishSettings settings, EnvDTE.Project targetProject, string targetFolder, EnvDTE.ProjectItem targetProjectItem) {
+            var creatTableScriptFile = Path.Combine(targetFolder, CreateRCodeTableScriptName);
+            using (var sw = new StreamWriter(creatTableScriptFile)) {
+                sw.WriteLine(Invariant($"CREATE TABLE {settings.TableName}"));
```
int19h: Might want brackets around the table name as well.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+            using (var sw = new StreamWriter(postDeploymentScript)) {
+                sw.WriteLine(Invariant($"INSERT INTO {settings.TableName}"));
+
+                for (int i = 0; i < settings.Files.Count; i++) {
+                    var filePath = settings.Files[i];
+
+                    var sprocName = settings.SProcNames[filePath];
+                    if (!string.IsNullOrEmpty(sprocName)) {
+                        var content = GetRFileContent(filePath);
+                        sw.Write(Invariant($"VALUES ('{sprocName}', '{content}')"));
```
int19h: `sProcName` needs quote escaping as well. Probably worth making it a helper somewhere, actually.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        /// <summary>
+        /// Generates SQL post deployment script that pushes R code into a table
+        /// as well as 
+        /// </summary>
+        private void CreatePostDeploymentScript(SqlSProcPublishSettings settings, EnvDTE.Project targetProject, string targetFolder, EnvDTE.ProjectItem targetProjectItem) {
+            var postDeploymentScript = Path.Combine(targetFolder, PostDeploymentScriptName);
+
+            using (var sw = new StreamWriter(postDeploymentScript)) {
+                sw.WriteLine(Invariant($"INSERT INTO {settings.TableName}"));
```
int19h: Just in case we ever want to extend the schema of that table, specifying field names here would keep the scripts compatible.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+            var sprocTemplateFile = rFilePath.ToSProcFilePath();
+            var sprocTemplate = GetSqlFileContent(sprocTemplateFile);
+
+            var format =
+@"BEGIN
+DECLARE @RCodeQuery NVARCHAR(max);
+DECLARE @RCode NVARCHAR(max);
+DECLARE @ParmDefinition NVARCHAR(max);
+
+SET @RCodeQuery = 'SELECT @RCodeOUT = RCode FROM {0} WHERE SProcName = ''{1}''';
```
int19h: Brackets and escaping.

int19h: And you might want `N` prefix on the string literal, in case any Unicode chars sneak in.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+                return _fs.ReadAllText(filePath).Replace("'", "''");
+            }
+            return string.Empty;
+        }
+
+        private string GetSqlFileContent(string filePath) {
+            if (_fs.FileExists(filePath)) {
+                var sb = new StringBuilder();
+                foreach (var line in _fs.FileReadAllLines(filePath)) {
+                    if (!line.TrimStart().StartsWithOrdinal("--")) {
```
int19h: Why do we want to remove comments? It might be undesirable for templating, because SP source will end up in the database, and they will probably prefer to keep comments for the ease of debugging on-server.

Syntactically, it's not as simple as that, either - `--` comments can be at the end of the line, and there are also `/**/` comments. Also I think string literals can span multiple lines?

src/Package/Impl/Sql/Publish/SqlPublishDialogViewModel.cs
```diff
@@ -0,0 +1,101 @@
...
+            Settings.CodePlacement = (RCodePlacement)_settingsStorage.GetInteger(CodePlacementSettingName, (int)RCodePlacement.Inline);
+        }
+
+        private void PopulateProjectList(IProjectSystemServices pss) {
+            TargetProjects = GetDatabaseProjectsInSolution(pss);
+            if (TargetProjects.Count > 0) {
+                var index = -1;
+                if (!string.IsNullOrEmpty(Settings.TargetProject)) {
+                    var indices = TargetProjects.IndexWhere(x => x.EqualsIgnoreCase(Settings.TargetProject));
+                    index = indices.Count() > 0 ? indices.First() : -1;
```
int19h: `indices.DefaultIfEmpty(-1).First()`

src/Package/Impl/Sql/Publish/SqlSProcPublishSettings.cs
```diff
@@ -0,0 +1,85 @@
...
+                var sprocName = GetSProcNameFromTemplate(file);
+                _sprocNameMap[file] = sprocName;
+            }
+        }
+
+        private string GetSProcNameFromTemplate(string rFilePath) {
+            var sprocTemplateFile = rFilePath.ToSProcFilePath();
+            if (!string.IsNullOrEmpty(sprocTemplateFile) && _fs.FileExists(sprocTemplateFile)) {
+                var content = _fs.ReadAllText(sprocTemplateFile);
+                var str = "CREATE PROCEDURE";
```
int19h: This needs to be more flexible (e.g. arbitrary whitespace between "CREATE" and "PROCEDURE"), and more strict with respect to inputs like "XCREATE PRODECUREY".

And it feels like this would be better as a regex instead of manual loops. Something like:
```
\bCREATE\s+PROCEDURE\s+(?'name'[^\s]+)
```

int19h: Also need to handle bracketed names (potentially with whitespace and other fun stuff inside).

src/Package/Test/Files/SqlCode1.table.baseline.sql
```diff
@@ -0,0 +1,18 @@
+﻿CREATE PROCEDURE _PROCEDURENAME_
+AS
+BEGIN
+DECLARE @RCodeQuery NVARCHAR(max);
+DECLARE @RCode NVARCHAR(max);
+DECLARE @ParmDefinition NVARCHAR(max);
+
+SET @RCodeQuery = 'SELECT @RCodeOUT = RCode FROM RCodeTable WHERE SProcName = ''_PROCEDURENAME_''';
```
int19h: 'N'

src/Package/Impl/Sql/DragDrop/DataObject.cs
MikhailArkhipov: VS stores everything as UTF-8 without BOM by default. So it depends where the file was created. In this case it most probably is created by RTVS. However, yes, it might come from elsewhere theoretically.

src/Package/Impl/Sql/Commands/PublishSProcCommand.cs
MikhailArkhipov: I thought about that... Technically this is language services storage but I guess it is OK to use in general. No reason to create yet another one. I'll make it a member on core shell.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
MikhailArkhipov: Mostly to strip template comments so it appears more compact, but it really doesn't matter, we can keep them

src/Package/Impl/Sql/DragDrop/DataObject.cs
MikhailArkhipov: I am not entirely sure they would be allowed in R code that is inside ```N'...'``` in SQL. Although N is technically Unicode, I am not sure if odd characters would survive when code is passed down to RScript.exe. So basically I don't expect anything funny here. How common would that be?

int19h: Oh, it wouldn't be legal R code. But it could still be in an .R file, since it's user input, so it can be anything at all. Highly unlikely, but we might want to check what happens so as to not end with weird SQL output that doesn't execute because it's not syntactically valid, just as a matter of general input sanitization.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -67,7 +67,7 @@ internal sealed class SProcGenerator {
         private void CreateRCodeTable(SqlSProcPublishSettings settings, EnvDTE.Project targetProject, string targetFolder, EnvDTE.ProjectItem targetProjectItem) {
             var creatTableScriptFile = Path.Combine(targetFolder, CreateRCodeTableScriptName);
             using (var sw = new StreamWriter(creatTableScriptFile)) {
-                sw.WriteLine(Invariant($"CREATE TABLE {settings.TableName}"));
+                sw.WriteLine(Invariant($"CREATE TABLE [{settings.TableName}]"));
```
veikkoeeva: A note: square brackets surrounding the table name is a SQL Server convention to allow using reserved words as table names.

veikkoeeva:
I happened to stumble this PR by semi-accident, but I wonder if cross-vendor capability would be desired here? It looks like some of the names are surrounded with `[ ]` and the SQL syntax is SQL Server specific too. I gather from [here](https://github.com/Microsoft/RTVS/pull/2122/files#diff-deca3aacbd1f33a97b2db245711b82d7R13) cross-vendorness would be desirable.

In that case instead of one `CreateRCodeTable.sql` you might want to introduce a convention like `CreateRCodeTable_SqlServer.sql`, `CreateRCodeTable_PostgreSql.sql` etc. that will have the database centric templates and make it so supplying a script following the set conventions is all that is needed to bring a new backend in.

There's precedent to this way in [Microsoft Orleans repo](https://github.com/dotnet/orleans/tree/master/src/OrleansSQLUtils). If you open the scripts, you see the notes about the interface contract the application layers honors. Otherwise one is free to tweak the scripts (i.e. taking into account physical and other deployment specific things). There's also differences between the scripts between MySQL and SQL Server and even in the SQL Server script just recently there were differenct counstructs used to support versions going as back as SQL Server 2000.

The system needs some supporting code, it's basically [here](https://github.com/dotnet/orleans/tree/master/src/OrleansSQLUtils/Storage), a small cross-platform ORM ([RelationalStorage being the core of it](https://github.com/dotnet/orleans/blob/master/src/OrleansSQLUtils/Storage/RelationalStorage.cs)), if one wants to see it like that. If nothing else, it might give some ideas what could be needed to support cross-vendor ADO.NET supported databases using a similar technique with templates.

src/R/Wpf/Impl/Microsoft.R.Wpf.csproj
```diff
@@ -36,6 +36,7 @@
     <Compile Include="..\..\..\GlobalAssemblyInfo.cs">
       <Link>Properties\GlobalAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="BindableBase.cs" />
```
AlexanderSher: Looks like this will be the third one, cause we already have one in Package and one in `Microsoft.Common.Wpf`

src/R/Wpf/Impl/Brushes.cs
```diff
@@ -16,31 +18,32 @@ public static class Brushes {
...
         public static object ToolWindowBorderKey { get; set; } = SystemColors.WindowBrushKey;
         public static object ToolWindowButtonDownBorderKey { get; set; } = SystemColors.WindowBrushKey;
         public static object ToolWindowButtonDownKey { get; set; } = SystemColors.WindowBrushKey;
         public static object ToolWindowButtonHoverActiveBorderKey { get; set; } = SystemColors.WindowBrushKey;
         public static object ToolWindowButtonHoverActiveKey { get; set; } = SystemColors.WindowBrushKey;
-        public static object UIText { get; set; } = SystemColors.ControlTextBrushKey;
+        public static object UITextKey { get; set; } = SystemColors.ControlTextBrushKey;
         public static object WindowTextKey { get; set; } = SystemColors.WindowTextBrushKey;
-        public static object WindowKey { get; set; } = SystemColors.WindowColorKey;
+        public static object WindowKey { get; set; } = SystemColors.WindowBrushKey;
```
AlexanderSher: Is it used anywhere?

src/ProjectSystem/Impl/Project/FileSystemMirroringProjectSourceItemProviderExtensionBase.cs
```diff
@@ -118,7 +118,9 @@ public abstract class FileSystemMirroringProjectSourceItemProviderExtensionBase
         private async Task<ProjectItem> GetMsBuildItemByProjectItem(IProjectItem projectItem) {
             using (var access = await _projectLockService.ReadLockAsync()) {
                 var project = await access.GetProjectAsync(_configuredProject);
-                return project.GetItemsByEvaluatedInclude(projectItem.EvaluatedInclude).FirstOrDefault(pi => StringComparer.OrdinalIgnoreCase.Equals(pi.ItemType, projectItem.ItemType));
+                var item =  project.GetItemsByEvaluatedInclude(projectItem.EvaluatedInclude)
+                    .FirstOrDefault(pi => StringComparer.OrdinalIgnoreCase.Equals(pi.ItemType, projectItem.ItemType));
```
AlexanderSher: Is it just formatting?
`pi => pi.ItemType.EqualsIgnoreCase(projectItem.ItemType)

src/Package/Impl/Sql/Publish/SProcGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+                        _fs.WriteAllText(sprocFile, template);
+                        targetProjectItem.ProjectItems.AddFromFile(sprocFile);
+                    }
+                }
+            }
+        }
+
+        private string GetRFileContent(string filePath) {
+            if (_fs.FileExists(filePath)) {
+                return _fs.ReadAllText(filePath).Replace("'", "''");
```
AlexanderSher: try/catch around file access

src/Package/Impl/Sql/DragDrop/DropHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+        public DragDropPointerEffects HandleDraggingOver(DragDropInfo dragDropInfo) => DragDropPointerEffects.Copy | DragDropPointerEffects.Track;
+        public bool IsDropEnabled(DragDropInfo dragDropInfo) => true;
+        #endregion
+
+        private void HandleDrop(DragDropInfo dragDropInfo) {
+            var dataObject = dragDropInfo.Data;
+
+            var text = dataObject.GetPlainText();
+            var line = _wpfTextView.TextViewLines.GetTextViewLineContainingYCoordinate(dragDropInfo.Location.Y);
+            line = line ?? _wpfTextView.TextViewLines.LastVisibleLine;
```
AlexanderSher: Do we want to insert text at the end of the file when there is no visible line?

int19h:
@veikkoeeva The context of this change is adding some R support to the Visual Studio tooling around Microsoft SQL Server 2016, and specifically its new [R integration feature](https://msdn.microsoft.com/en-us/library/mt604845.aspx?f=255&MSPPError=-2147217396). Similar in principle to what Visual Studio already has for SQL/CLR, but for R, essentially. Hence the use of MS-specific syntax - these snippets are for templates to be used in VS Database projects (that target SQL Server specifically), to utilize a feature that is specific to SQL Server.

src/Package/Impl/Shell/VsAppShell.cs
```diff
@@ -394,6 +399,19 @@ public sealed class VsAppShell : IApplicationShell, IMainThread, IIdleTimeServic
                 }
             }
         }
+
+        public IWritableSettingsStorage SettingsStorage {
```
AlexanderSher: >IWritableSettingsStorage [](start = 15, length = 24)

How this will be related to `REditorSettings` and `IRSettings`?

int19h:
This does remind me, though - while most SQL Server documentation and generated code uses `[]` for name escaping, it actually also conditionally supports the standard `""` quotes for that purpose, too (and [that setting](https://technet.microsoft.com/en-us/library/ms176027%28v=sql.105%29.aspx?f=255&MSPPError=-2147217396) is on by default). So where we're handling `CREATE PROCEDURE` in the SQL source, it needs to be able to parse that, too.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
int19h: It's actually caught at the top of the call chain:

https://github.com/Microsoft/RTVS/pull/2122/files/fccd17657673eb092e3c88b76d387581e897692d#diff-1e118299ea36daf74a89a210c373f8feR58

AlexanderSher: Then `_fs.FileExists(filePath)` is not needed

---
In reply to: [72572665](https://github.com/Microsoft/RTVS/pull/2122#discussion_r72572665) [](ancestors = 72572665)

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
```diff
@@ -53,6 +53,7 @@ internal sealed class RProjectLoadHooks {
         private readonly IEnumerable<Lazy<IVsProject>> _cpsIVsProjects;
         private readonly IRInteractiveWorkflowProvider _workflowProvider;
         private readonly IInteractiveWindowComponentContainerFactory _componentContainerFactory;
+        private readonly IProjectItemDependencyProvider _dependencyProvider;
```
AlexanderSher: This field isn't used anywhere except constructor.

src/Package/Impl/ProjectSystem/ProjectSettings.cs
MikhailArkhipov: Yes the pattern is wider but it is narrowed to ```SettingsABC.r``` or ```Debug.Settings.r```

src/Package/Impl/Sql/Publish/SProcGenerator.cs
MikhailArkhipov: Yes that is the point. Table name can technically be anything

MikhailArkhipov: As for Ms SQL specific syntax, yes the target is sproc written in R and executed in MS SQL as in https://msdn.microsoft.com/en-us/library/mt591996.aspx

src/R/Wpf/Impl/Microsoft.R.Wpf.csproj
MikhailArkhipov: I think I moved it but let me double check

src/R/Wpf/Impl/Brushes.cs
MikhailArkhipov: Just added for consistency. I think there are few others not used though - do we want to remove all unused?

src/Package/Impl/Sql/Publish/SqlSProcPublishSettings.cs
MikhailArkhipov: I thought about this, generally it is a template with limited customization. I don't know how far we want to go short of parsing the thing properly with all possible escapes etc. I guess I can at least handle whitespace and square brackets

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
MikhailArkhipov: I pass it down to FileSystemMirroringProject ctor...

src/Package/Impl/Sql/Publish/SProcGenerator.cs
MikhailArkhipov: Remnant from a variant when I used to store list of files between sessions. Removed.

src/Package/Impl/Shell/VsAppShell.cs
MikhailArkhipov: ```REditorSettings``` uses it to store stuff, but it is named/typed, contains commonly used stuff exposed to user somewhere. This one is low level, anyone can use it to store things. Originally I imported it directly, but @int19h suggested to have a shell member instead.

src/Package/Impl/Sql/Publish/SProcGenerator.cs
MikhailArkhipov: Yes, called handles it with the error message in ```OKButton_Click```

src/ProjectSystem/Impl/Project/FileSystemMirroringProjectSourceItemProviderExtensionBase.cs
MikhailArkhipov: Yes, just a break in a long line

src/R/Wpf/Impl/Microsoft.R.Wpf.csproj
MikhailArkhipov: I moved it from the package and there is only one in ```Microsoft.Common.Wpf```

src/Package/Impl/Sql/DragDrop/DataObject.cs
MikhailArkhipov: Replace them with spaces I guess?

src/Package/Impl/Sql/DragDrop/DropHandler.cs
MikhailArkhipov: I believe there is always at least one visible line - where the caret is.

src/R/Wpf/Impl/Brushes.cs
AlexanderSher: I think its fine to leave it as is

---
In reply to: [72653040](https://github.com/Microsoft/RTVS/pull/2122#discussion_r72653040) [](ancestors = 72653040)

MikhailArkhipov:
OK, added more tests, helper for quoting, quoting option to the dialog, regex for 'CREATE PROCEDURE'.

src/Package/Impl/Sql/Extensions/SqlStringExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Linq;
+using static System.FormattableString;
+
+namespace Microsoft.VisualStudio.R.Package.Sql {
+    internal static class SqlStringExtensions {
+        public static string ToSqlName(this string name, SqlQuoteType quoteType) {
```
int19h: This should also escape the closing quote character if it occurs in the name (doubling does the trick for both quote and closing bracket).

src/Package/Impl/Sql/Extensions/SqlStringExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+                return name;
+            }
+            if (quoteType == SqlQuoteType.Quote) {
+                return Invariant($"\"{name}\"");
+            }
+            return Invariant($"[{name}]");
+        }
+
+        private static bool HasSpaces(this string s) {
+            return s.FirstOrDefault(c => char.IsWhiteSpace(c)) != default(char);
```
int19h: `s.Any(c => char.IsWhiteSpace(c))`

src/Package/Impl/Sql/Extensions/SqlStringExtensions.cs
MikhailArkhipov: Generally there is no intent to handle all kinds of weird names :-)

int19h:
:+1:

AlexanderSher:
:+1:

Issue #2093, opened 2016-07-19
Handle project loading when R is disconnected


AlexanderSher:
- Show yellow message bar in Solution Explorer (or find any other way to show notification)
- Disable automatic RData loading for remote scenarios for local files

MikhailArkhipov:
Project opens fine. Editor opens. REPL behavior is s expected, getting appropriate error message. Need to handle intellisense better since it is attempting to start the session and typing is slow. Although not entirely legal case, but some may need/want to edit R without actually having the connection up. Obviously, intellisense will work only partially.

MikhailArkhipov:
#2251 more or less takes care of found issues. Otherwise there is no feature work.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/d132ef920675674d6ff2a8c7649fb3fdde98402f

Issue #2043, opened 2016-07-06
Provide indication for the connection to R


AlexanderSher:

MikhailArkhipov:
Still in progress?

AlexanderSher:
No, done in #2133

Issue #2030, opened 2016-06-30
Handle RHost disconnected state in UI


AlexanderSher:
Interactive Window, Package Manager, Variable Explorer, Project loading
Probably will require refactoring the pull of sessions

MikhailArkhipov:
Output of a message to REPL should be sufficient for now

MikhailArkhipov:
Project loading was covered in #2093. Otherwise UI looks reasonble:

![image](https://cloud.githubusercontent.com/assets/12820357/18359619/8c24af2e-75af-11e6-93bc-f823e34fde7d.png)

PR #2133, opened 2016-08-01
Handling RHost disconnected state

AlexanderSher:
- #2030: Handle RHost disconnected state in UI
- #2043: Provide indication for the connection to R
- #2093: Handle project loading when R is disconnected
- Separated `RHost` from local process creation
- Added `IAsyncCommandRange` interface for the range of commands that can be dynamically provided to the menus
- Fixed unit-test concurrency issues for TestVsAppShell


src/Host/Client/Impl/ExecutionTracing/RExecutionTracer.cs
```diff
@@ -171,18 +171,16 @@ internal sealed class RExecutionTracer : IRExecutionTracer {
...
-                    } catch (RException) {
-                        _stepTcs.TrySetResult(false);
-                        return false;
-                    }
-                    return true;
+                try {
+                    // EvaluateAsync will push a new toplevel context on the context stack before
+                    // evaluating the expression, so tell browser_set_debug to skip 1 toplevel context
+                    // before locating the target context for step-out.
+                    await Session.ExecuteAsync("rtvs:::browser_set_debug(1, 1)", REvaluationKind.Normal, cancellationToken);
```
int19h: I would prefer this to be inside `BeginEvaluation` (or `BeginInteraction` if we're get rid of the former), to ensure that it always happens at the REPL prompt. Otherwise it might interact with things in an unpredictable way.

src/Host/Client/Impl/Host/IRHostBrokerConnector.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+
+namespace Microsoft.R.Host.Client.Host {
+    public interface IRHostBrokerConnector : IRHostConnector {
+        /// <summary>
+        /// String token that allows to identify broker that is used to connect to RHost
+        /// For local connections, it is a path to the RHost
+        /// For remote connections, it is a URL
+        /// </summary>
+        string BrokerId { get; }
```
int19h: Maybe actually make it an URL always, and it'll just be `file://` for local? As I recall, you can tease a proper Windows path back out of an URI if it was created with one to begin with.

src/Host/Client/Impl/Host/LocalRHostConnector.cs
```diff
@@ -0,0 +1,174 @@
...
+                    } catch (InvalidOperationException) { }
+                }
+                _log.RHostProcessExited();
+            });
+
+            var host = new RHost(name, callbacks, transportTcs.Task.Result, process.Id, cts);
+            process.Exited += delegate { host.Dispose(); };
+            return host;
+
+            //using (this)
```
int19h: ?

AlexanderSher: Sorry, garbage :)

---
In reply to: [73022424](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73022424) [](ancestors = 73022424)

src/Host/Client/Impl/Host/IRHostBrokerConnector.cs
AlexanderSher: Do you want to use change property type to `Url` as well?

---
In reply to: [73022290](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73022290) [](ancestors = 73022290)

src/Host/Client/Impl/ExecutionTracing/RExecutionTracer.cs
AlexanderSher: Ok.

---
In reply to: [73022035](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73022035) [](ancestors = 73022035)

src/Host/Client/Impl/Host/IRHostBrokerConnector.cs
int19h: Sure. And if we feel like it, we can have a broker connector factory later, that would automatically determine the right connector based on Uri.Schema.

---
In reply to: [73025921](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73025921) [](ancestors = 73025921,73022290)

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
```diff
@@ -5,19 +5,28 @@
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.R.Host.Client.Host;
 
 namespace Microsoft.R.Host.Client {
     public interface IRSessionProvider : IDisposable {
-        IRSession GetOrCreate(Guid guid);
+        IRSession GetOrCreate(Guid guid, IRHostBrokerConnector brokerConnector);
```
MikhailArkhipov: ```IRHostBrokerConnector``` is passed pretty much to every method. This means that either methods belong to ```IRHostBrokerConnector``` or it is a settable property on ```IRSessionProvider``` or may be imported internally by the implementor of the ```IRSessionProvider```. Can different calls can receive different instances of the ```IRHostBrokerConnector``` ?

src/Host/Client/Impl/Host/IRHostConnector.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Client.Host {
+    public interface IRHostConnector {
+        Task<RHost> ConnectToRHost(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000, CancellationToken cancellationToken = default(CancellationToken));
```
MikhailArkhipov: Just ```Connect```

src/Host/Client/Impl/Host/IRHostBrokerConnector.cs
```diff
@@ -0,0 +1,19 @@
...
+
+        event EventHandler BrokerIdChanged;
```
MikhailArkhipov: This is really ```broker changed``` and not `broker is the same but its ID changed` so ```BrokerChanged``` is more appropriate

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -50,6 +51,12 @@ public class RSessionProvider : IRSessionProvider {
             }
         }
 
+        public Task RestartSessions(IRHostBrokerConnector brokerConnector) {
+            var sessions = _sessions.Values.Where(s => s.BrokerConnector.Equals(brokerConnector)).ToList();
```
MikhailArkhipov: Should it be comparing IDs?

src/Host/Client/Impl/Host/RHostBrokerConnector.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.R.Interpreters;
+
+namespace Microsoft.R.Host.Client.Host {
+    public sealed class RHostBrokerConnector : IRHostBrokerConnector {
```
MikhailArkhipov: Is it a singleton service? Maybe just export it via MEF then?

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
AlexanderSher: Yes, if there is more than one REPL. `IRHostBrokerConnector` is assigned to the `RInteractiveWorkflow`, while `IRSessionProvider` is a global export

---
In reply to: [73030385](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73030385) [](ancestors = 73030385)

src/Host/Client/Impl/Host/RHostBrokerConnector.cs
AlexanderSher: No, it is one for every RInteractiveWorkflow

---
In reply to: [73032105](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73032105) [](ancestors = 73032105)

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: That's a good question. `BrokerID` belongs to the broker (broker's URL), but theoretically we may want to have several broker connectors that work with the local broker. So I think we should compare the instance itself rather than its broker ID

---
In reply to: [73031834](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73031834) [](ancestors = 73031834)

src/Host/Client/Impl/Host/IRHostBrokerConnector.cs
AlexanderSher: Agreed

---
In reply to: [73031160](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73031160) [](ancestors = 73031160)

src/Host/Client/Impl/Host/RHostBrokerConnector.cs
MikhailArkhipov: So then it should be [internal?] part of the workflow rather than passed to every session operation?

AlexanderSher: It isn't passed to the session operations, it is passed to the session provider operations. Session gets its connector in constructor and uses internally later. 

---
In reply to: [73033191](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73033191) [](ancestors = 73033191)

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
MikhailArkhipov: Oh boy. OK. I guess we then may consider at some point reversing composition. From most callers perspective session provider belongs to a particular REPL (usually the active one) or set of sessions that user is currently operating upon. That makes broker primary and session provider its property rather than the other way around. Then caller simply talks to the broker about getting sessions.

AlexanderSher: You don't usually talk to the broker or session provider. There are just 8 usages of the `IRSessionProvider` interface in the whole product code - the rest are in test code. We will have to decide how secondary sessions should work when user has more than one REPL (i.e., what should happen with intellisense when user switches from one REPL to another)

---
In reply to: [73035733](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73035733) [](ancestors = 73035733)

AlexanderSher: We probably may want to import IRSessionProvider to the RInteractiveWorkflow, and then define properties for REPL session, Intellisense session, packages, etc.

---
In reply to: [73038600](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73038600) [](ancestors = 73038600,73035733)

int19h:
:+1:

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
```diff
@@ -60,79 +60,38 @@ internal class RPackageManager : IRPackageManager {
             }
```
karthiknadig: Should this be changed to follow the pattern used for uninstall and update?

MikhailArkhipov:
👍

src/R/Components/Impl/PackageManager/Implementation/RPackageManager.cs
AlexanderSher: Yes, it should, but I haven't found `rtvs:::package_install`. Though I think it should be a separate change

---
In reply to: [73064953](https://github.com/Microsoft/RTVS/pull/2133#discussion_r73064953) [](ancestors = 73064953)

PR #2135, opened 2016-08-01
Ability to publish stored procs to DACPAC and directly to database

MikhailArkhipov:
Per SQL team feedback

sql1

sql2


src/Package/Impl/project.15.0.json
```diff
@@ -2,6 +2,7 @@
   "dependencies": {
     "DataConnectionDialog": "1.2.0",
     "MicroBuild.Core": "0.2.0",
+    "Microsoft.SqlServer.DacFx.x86": "130.3370.2",
```
AlexanderSher: Is it the same version for VS14 and VS15?

src/Package/Impl/Logging/OutputWindowLog.cs
```diff
@@ -23,31 +23,33 @@ internal sealed class OutputWindowLogWriter : IActionLogWriter {
...
             EnsurePaneVisible();
-            _pane.OutputStringThreadSafe(message);
-
+            _pane?.OutputStringThreadSafe(message);
             return Task.CompletedTask;
         }
 
         public void Flush() { }
+
+        internal object WriteAsync(string error) {
```
AlexanderSher: What is this for?

src/Package/Impl/Sql/Commands/PublishSProcOptionsCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+        [ImportingConstructor]
+        public PublishSProcOptionsCommand(IApplicationShell appShell, IProjectSystemServices pss, IProjectConfigurationSettingsProvider pcsp) {
+            _appShell = appShell;
+            _pss = pss;
+            _pcsp = pcsp;
+        }
+
+        public CommandStatusResult GetCommandStatus(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, string commandText, CommandStatus progressiveStatus) {
+            if (commandId == RPackageCommandId.icmdPublishSProcOptions) {
+                return new CommandStatusResult(true, commandText, CommandStatus.Enabled | CommandStatus.Supported);
```
AlexanderSher: Does it mean that command should be available on every project node?

src/Package/Impl/Sql/Commands/PublishSProcOptionsCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Microsoft.VisualStudio.R.Package.Sql {
+    [ExportCommandGroup("AD87578C-B324-44DC-A12A-B01A6ED5C6E3")]
+    [AppliesTo(ProjectConstants.RtvsProjectCapability)]
+    internal sealed class PublishSProcOptionsCommand : ICommandGroupHandler {
+        private readonly IApplicationShell _appShell;
+        private readonly IProjectSystemServices _pss;
+        private readonly IProjectConfigurationSettingsProvider _pcsp;
+
+        [ImportingConstructor]
+        public PublishSProcOptionsCommand(IApplicationShell appShell, IProjectSystemServices pss, IProjectConfigurationSettingsProvider pcsp) {
```
AlexanderSher: Do you need `IApplicationShell`', or `ICoreShell` is enough?

src/Package/Impl/Sql/Publish/SProcProjectFilesGenerator.cs
```diff
@@ -0,0 +1,96 @@
...
+            item.Properties.Item("BuildAction").Value = "PostDeploy";
+        }
+
+        private IReadOnlyDictionary<string, string> CreateStoredProcedureFiles(SqlSProcPublishSettings settings, EnvDTE.Project targetProject, string targetFolder, EnvDTE.ProjectItem targetProjectItem) {
+            var g = new SProcScriptGenerator(_fs);
+
+            var sprocFiles = targetProject.GetSProcFiles(_pss);
+            var sprocMap = g.CreateStoredProcedureScripts(settings, sprocFiles);
+
+            foreach (var sprocName in sprocMap.Keys) {
```
AlexanderSher: You can enumerate through `sprocMap` itself

src/Package/Impl/Sql/Publish/SProcProjectFilesGenerator.cs
```diff
@@ -0,0 +1,96 @@
...
+        public SProcProjectFilesGenerator(IProjectSystemServices pss, IFileSystem fs) {
+            _pss = pss;
+            _fs = fs;
+        }
+
+        /// <summary>
+        /// Generates SQL scripts for the deployment of R code into SQL database.
+        /// Writes scripts to files and pushes files into the target database project.
+        /// </summary>
+        public void Generate(SqlSProcPublishSettings settings, EnvDTE.Project targetProject) {
```
AlexanderSher: This method can fail with `IOException` - do we handle it anywhere?

src/Package/Impl/Sql/Publish/SProcPublisher.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        private void CreateDacPac(SqlSProcPublishSettings settings, IEnumerable<string> sprocFiles, string dacpacPath) {
+            var project = _pss.GetSelectedProject<IVsHierarchy>()?.GetDTEProject();
+            var g = new SProcScriptGenerator(_fs);
+            var scripts = g.CreateStoredProcedureScripts(settings, sprocFiles);
+            var builder = _dacServices.GetBuilder(_appShell);
+            builder.Build(dacpacPath, project.Name, scripts.Values);
+        }
+
+        private EnvDTE.Project GetProjectByName(string projectName) {
```
AlexanderSher: Don't we have something like this in `IProjectSystemServices`?

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        private readonly IFileSystem _fs;
+
+        public SProcScriptGenerator(IFileSystem fs) {
+            _fs = fs;
+        }
+
+        public string CreateRCodeTableScript(SqlSProcPublishSettings settings) {
+            var sb = new StringBuilder();
+            sb.AppendLine(Invariant($"CREATE TABLE {settings.TableName.ToSqlName(settings.QuoteType)}"));
```
AlexanderSher: >AppendLine [](start = 15, length = 10)

Don't you want to just use formatted string:?
```
        public string CreateRCodeTableScript(SqlSProcPublishSettings settings) => 
            Invariant($@"CREATE TABLE {settings.TableName.ToSqlName(settings.QuoteType)}
(
{SProcColumnName} NVARCHAR(64),
{RCodeColumnName} NVARCHAR(max)
)");
        }
```

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+            sb.AppendLine(Invariant($"{RCodeColumnName} NVARCHAR(max)"));
+            sb.AppendLine(")");
+            return sb.ToString();
+        }
+
+        /// <summary>
+        /// Generates SQL post deployment script that pushes R code into a table
+        /// as well as 
+        /// </summary>
+        public string CreatePostDeploymentScript(SqlSProcPublishSettings settings, IReadOnlyDictionary<string, string> sprocMap) {
```
AlexanderSher: Do you use it anywhere as a dictionary? If it just a collection of string pairs, I'd rather add a simple type for this pair

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+                sb.AppendLine(string.Empty);
+                i++;
+            }
+            return sb.ToString();
+        }
+
+        /// <summary>
+        /// Creates code for stored procedures
+        /// </summary>
+        public IReadOnlyDictionary<string, string> CreateStoredProcedureScripts(SqlSProcPublishSettings settings, IEnumerable<string> sprocFiles) {
```
AlexanderSher: This method can fail with `IOException` - do we handle it anywhere?

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+DECLARE @ParmDefinition NVARCHAR(max);
+
+SET @RCodeQuery = N'SELECT @RCodeOUT = RCode FROM {0} WHERE SProcName = ''{1}''';
+SET @ParmDefinition = N'@RCodeOUT NVARCHAR(max) OUTPUT';
+
+EXEC sp_executesql @RCodeQuery, @ParmDefinition, @RCodeOUT=@RCode OUTPUT;
+SELECT @RCode;
+";
+            var declarations = string.Format(CultureInfo.InvariantCulture, format, codeTableName.ToSqlName(quoteType), sprocName);
+            sprocTemplate = sprocTemplate.Replace("BEGIN", declarations);
```
AlexanderSher: Why do you want to use `Replace` instead of providing another format placeholder?

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+            sprocTemplate = sprocTemplate.Replace("N'_RCODE_'", "@RCode");
+
+            var sqlQuery = GetSqlFileContent(rFilePath.ToQueryFilePath()).Trim();
+            return sprocTemplate.Replace(InputQueryTemplate, sqlQuery);
+        }
+
+        private string GetRFileContent(string filePath) {
+            return _fs.ReadAllText(filePath).Replace("'", "''");
```
AlexanderSher: >Replace("'", "''"); [](start = 45, length = 19)

Shouldn't we skip sequential quote marks (''`)?

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
```diff
@@ -0,0 +1,315 @@
...
+                    Settings.QuoteType = (SqlQuoteType)value;
+                    SetProperty(ref _selectedQuoteTypeIndex, value);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Publishing targets
+        /// </summary>
+        public IReadOnlyList<string> Targets {
```
AlexanderSher: >Targets [](start = 37, length = 7)

It is usually a bad idea to replace collection that is used as a data source for `ItemsControl`. Not only it is a slow operation, it also not handled very well, so you may end up with unbound UI.

Please use `BatchObservableCollection` (use `RPackageManagerViewModel` for an example)

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
```diff
@@ -0,0 +1,315 @@
...
+        public Task InitializationTask { get; private set; } = Task.CompletedTask;
+
+        /// <summary>
+        /// Target types: DACPAC, database, project
+        /// </summary>
+        public IReadOnlyList<string> TargetTypeNames { get; private set; }
+        public int SelectedTargetTypeIndex {
+            get { return _selectedTargetTypeIndex; }
+            set {
+                if (_selectedTargetTypeIndex != value) {
```
AlexanderSher: `SelectTargetTypeAsync` will be called from View using `DoNotWait`, but in tests you can call easily wait until it is finished

---
In reply to: [73166302](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73166302) [](ancestors = 73166302)

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
```diff
@@ -0,0 +1,315 @@
...
+                } catch (NotImplementedException) { } catch (COMException) { }
+            }
+
+            if (projects.Count == 0) {
+                projects.Add(Resources.SqlPublishDialog_NoDatabaseProjects);
+            }
+            return projects;
+        }
+
+        private Task GetDatabaseConnectionsAsync(Action<IReadOnlyList<DbConnectionData>> action) {
```
AlexanderSher: Don't you want to use `async/await` here:
```{csharp}
        private async Task GetDatabaseConnectionsAsync(Action<IReadOnlyList<DbConnectionData>> action) {
            var project = _pss.GetSelectedProject<IVsHierarchy>();
            if (project != null) {
                await TaskUtilities.SwitchToBackgroundThreadAsync();
                var dbConnections = await project.GetDbConnections(_pcsp);
                var connections = dbConnections.Select(c => new DbConnectionData { Name = c.Name, ConnectionString = c.Value }).ToList();
                if (connections.Count == 0) {
                    connections.Add(new DbConnectionData { Name = Resources.SqlPublishDialog_NoDatabaseConnections });
                }

                await _coreShell.SwitchToMainThreadAsync();
                action(connections);
            }
        }
```

AlexanderSher: You actually don't even need an `action` callback and `SwitchToMainThreadAsync` - just make this method return `connections` - caller is in UI thread already, so you don't need to switch context.

---
In reply to: [73168059](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73168059) [](ancestors = 73168059)

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
```diff
@@ -0,0 +1,315 @@
...
+                    break;
+
+                case PublishTargetType.Database:
+                    InitializationTask = PopulateConnectionsListAsync();
+                    TargetHasName = false;
+                    break;
+            }
+        }
+
+        private Task PopulateConnectionsListAsync() {
```
AlexanderSher: I'd add `_.coreShell.AssertIsOnMainThread` at the beginning of this method

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
```diff
@@ -0,0 +1,315 @@
...
+                    var indices = _targetProjects.IndexWhere(x => x.EqualsIgnoreCase(Settings.TargetProject));
+                    index = indices.DefaultIfEmpty(-1).First();
+                }
+                Targets = _targetProjects;
+                ForceSelectedTargetIndexUpdate();
+                SelectedTargetIndex = index >= 0 ? index : 0;
+            }
+        }
+
+        private void ForceSelectedTargetIndexUpdate() {
```
AlexanderSher: What is this for?

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
```diff
@@ -0,0 +1,81 @@
...
+                var uiShell = _appShell.GetGlobalService<IVsUIShell>(typeof(SVsUIShell));
+                var guid = RGuidList.RCmdSetGuid;
+                var o = new object();
+                uiShell.PostExecCommand(ref guid, RPackageCommandId.icmdPublishSProc, 0, ref o);
+            }
+            Close();
+        }
+
+        private void OKButton_Click(object sender, RoutedEventArgs e)  => Close(true);
+        private void CancelButton_Click(object sender, RoutedEventArgs e) => Close(false);
```
AlexanderSher: `Close(false)` is simply `Close()`

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
```diff
@@ -0,0 +1,81 @@
...
+        private void TableName_TextChanged(object sender, TextChangedEventArgs e) => _model.UpdateState();
+
+        private void CodePlacementList_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            var s = CodePlacementList.SelectedItem as string;
+            var placement = s.EqualsOrdinal(Package.Resources.SqlPublishDialog_RCodeInline) ? RCodePlacement.Inline : RCodePlacement.Table;
```
AlexanderSher: Can you move this part into `SqlPublishOptionsDialogViewModel`?

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
```diff
@@ -0,0 +1,81 @@
...
+            _model.Settings.CodePlacement = placement;
+            _model.UpdateState();
+        }
+
+        private void QuoteTypetList_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            var s = QuoteTypeList.SelectedItem as string;
```
AlexanderSher: Can you move this code to view-model as well?

src/Package/Test/Sql/PublishOptionsDialogModelTest.cs
```diff
@@ -0,0 +1,229 @@
...
+            model.TargetTypeNames.Should().HaveCount(3);
+            model.SelectedTargetTypeIndex.Should().Be(0);
+
+            model.QuoteTypeNames.Should().HaveCount(3);
+            model.SelectedQuoteTypeIndex.Should().Be(0);
+
+            model.CodePlacementNames.Should().HaveCount(2);
+            model.SelectedCodePlacementIndex.Should().Be(0);
+
+            model.Targets.Should().HaveCount(0);
```
AlexanderSher: `BeEmpty()`

src/Setup/Assemblies.wxi
```diff
@@ -18,6 +18,10 @@
       <File Source="$(var.BinDir)\Microsoft.R.Interpreters.dll" />
       <File Source="$(var.BinDir)\Microsoft.Data.ConnectionUI.dll" />
       <File Source="$(var.BinDir)\Microsoft.Data.ConnectionUI.Dialog.dll" />
+      <File Source="$(var.BinDir)\Microsoft.SqlServer.Dac.dll" />
```
AlexanderSher: Don't forget to add those assemblies to `Create-FilesSwr.ps1` in low-impact installer.

src/Package/Impl/project.15.0.json
MikhailArkhipov: There is only one version on nuget so I assume yes

src/Package/Impl/Logging/OutputWindowLog.cs
MikhailArkhipov: No idea. It has been around for a while. Removed it.

src/Package/Impl/Sql/Commands/PublishSProcOptionsCommand.cs
MikhailArkhipov: Yes. It is sort of 'right click anywhere in the project'.

src/Package/Impl/Sql/Commands/PublishSProcOptionsCommand.cs
MikhailArkhipov: The command doesn't need it, but the dialog does. I calls

```
                var uiShell = _appShell.GetGlobalService<IVsUIShell>(typeof(SVsUIShell));
```

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: ```PublishSProcCommand``` handles all exceptions in ```Handle()``` and shows the error message.

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
MikhailArkhipov: Good catch. Actually, none of these handlers are needed anymore, it is all handled when selectedIndex changes which is TwoWay bound to model

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: Ha, I didn't know ```$@``` could be combined...

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: sprocTemplate is actually loaded from the file which user could have edited - they may changed proc name and add more to WITH RESULTS. Technically, they can delete BEGIN and other stuff, but I am not sure we want to go into full checking of the template validity after editing. SQL will squiggle bad stuff.

AlexanderSher: Got it. Can there be more than one "BEGIN"?

---
In reply to: [73189641](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73189641) [](ancestors = 73189641)

src/Package/Impl/Sql/Publish/SProcPublisher.cs
MikhailArkhipov: Not currently, but I will move this one there

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
MikhailArkhipov: When collection is replaced, selection needs to be rest. But I can do it in the Targets setter

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: I hope not :-) I am not T-SQL expert really but from what I see if there are multiple statements, there has to be ```GO``` between them. Since template intent to have code for a single procedure there shouldn't be more unless user types more of them. Not sure what would be the useful scenario though.

AlexanderSher: Can we ask someone from SQL team to take a look? Cause I'm not an SQL expert too :)

---
In reply to: [73195831](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73195831) [](ancestors = 73195831)

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: I think no. This is to escape R code into SQL string. So in ```x <- ''``` (empty string) we do want to turn it into ```x <- ''''```

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: Yes, I have a thread with them where I show them demo, UI and the SQL code.

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
AlexanderSher: What about `"'"` combination?

---
In reply to: [73196025](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73196025) [](ancestors = 73196025)

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
```diff
@@ -0,0 +1,56 @@
...
+            _appShell = appShell;
+            _pss = pss;
+            _model = new SqlPublishOptionsDialogViewModel(appShell, pss, appShell.SettingsStorage, pcsp);
+
+            Title = Package.Resources.SqlPublishDialog_Title;
+            DataContext = _model;
+            _model.UpdateState();
+        }
+
+        private void Close(bool saveSettings) {
```
AlexanderSher: This parameter isn't needed anymore - we can just call method `SaveSettingsAndClose`

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
```diff
@@ -0,0 +1,306 @@
...
+                } catch (NotImplementedException) { } catch (COMException) { }
+            }
+
+            if (projects.Count == 0) {
+                projects.Add(Resources.SqlPublishDialog_NoDatabaseProjects);
+            }
+            return projects;
+        }
+
+        private IReadOnlyList<DbConnectionData> GetDatabaseConnections() {
```
AlexanderSher: I don't get it - how did it happen that this not async anymore?

src/Package/Impl/Sql/Publish/SProcScriptGenerator.cs
MikhailArkhipov: That too. SQL uses single quotes for strings so any single quotes inside SQL string must be doubled. Double quotes are in SQL are used like ```[ ]``` i.e. to surround names with spaces or names that are keywords ie``` "SELECT"```

AlexanderSher: Got it

---
In reply to: [73197779](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73197779) [](ancestors = 73197779)

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
MikhailArkhipov: Still working on it

AlexanderSher: Ah, ok

---
In reply to: [73198718](https://github.com/Microsoft/RTVS/pull/2135#discussion_r73198718) [](ancestors = 73198718)

src/Package/Impl/Sql/Publish/SqlPublishOptionsDialogViewModel.cs
MikhailArkhipov: The problem is this chain is that methods call things on IProjectSystemServices that all require main thread. Ex ```GetDbConnections``` calls 

```
            var configuredProject = hierarchy?.GetConfiguredProject();
```
which goes through ```IVsHierarchy``` that is not thread friendly. So up until 

```
await provider.OpenProjectSettingsAccessAsync
```
context has to be main thread

AlexanderSher:
:+1:

Issue #2121, opened 2016-07-27
CsvAppFileIO should be using IFileSystem and IProcessServices


int19h:
It also uses `VsAppShell.Current` all over the place, that should be injected instead as well.

PR #2136, opened 2016-08-01
#2121 CsvAppFileIO changed to use file system and process services

karthiknadig:
CsvAppFileIO changed to use IApplicationShell, IFileSystem, and IProcessServices

### DIFF
--- a/src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
+++ b/src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
@@ -10,6 +10,8 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using Microsoft.Common.Core;
+using Microsoft.Common.Core.IO;
+using Microsoft.Common.Core.OS;
 using Microsoft.R.Components.ContentTypes;
 using Microsoft.R.Components.Extensions;
 using Microsoft.R.Components.InteractiveWorkflow;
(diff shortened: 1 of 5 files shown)


MikhailArkhipov:
Usually this is only needed if code becomes testable but since it is calling GetGlobalService and GetExportedValue, it is still not :-)

But OK.

MikhailArkhipov:
👍

src/Package/Impl/DataInspect/Office/CsvAppFileIO.cs
```diff
@@ -28,68 +30,67 @@ internal static class CsvAppFileIO {
...
             var variableName = result.Name ?? _variableNameReplacement;
-            var csvFileName = MakeCsvFileName(variableName);
+            var csvFileName = MakeCsvFileName(appShell, pss, variableName);
 
-            var pss = VsAppShell.Current.ExportProvider.GetExportedValue<IProjectSystemServices>();
             var file = pss.GetUniqueFileName(folder, csvFileName, "csv", appendUnderscore: true);
 
             string currentStatusText;
-            var statusBar = VsAppShell.Current.GetGlobalService<IVsStatusbar>(typeof(SVsStatusbar));
+            var statusBar = appShell.GetGlobalService<IVsStatusbar>(typeof(SVsStatusbar));
```
AlexanderSher: Is it VS status bar, or it is a status bar of some dialog window?

MikhailArkhipov: VS status bar

AlexanderSher: Hm... probably it is not part of this change, but since we have `IStatusBar`, probably we should use it instead. That will allow to move this code to the `Components` assembly.

karthiknadig:
Created a work item to make the changes suggested by @MikhailArkhipov and @AlexanderSher.

Issue #2147, opened 2016-08-03
Incorrect parameter suggested in arguments with nested function calls


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/17379347/e483ad0a-5976-11e6-9759-97dafb4e7002.png)

Should be signature of ```paste```

MikhailArkhipov:
Specific to the last empty argument as in ```func1(func2(...), |)```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5aead99fab615886b9099b1953e7f76513e69f17

Issue #2146, opened 2016-08-03
Drag & drop of sql file creates code that doesn't work for multiple line sql query


huguesv:
Create a SQL query and add to your project, use multiple lines:
```
SELECT *
FROM ProspectiveBuyer
```

Drag & drop that file from explorer to .R file.
Run that generated code and execute a query:
```
library(RODBC)
source('~/visual studio 2015/projects/rproject1/rproject1/settings.r')

channel <- RODBC::odbcConnect(dbConnection)


query = iconv(readLines('~/visual studio 2015/projects/rproject1/rproject1/sqlquery.sql', encoding = 'UTF-8', warn = FALSE), from = 'UTF-8', to = 'ASCII', sub = '')
InputDataSet <- sqlQuery(channel, query)

```

You'll get an error, because the `query` is a list (one item per line in the original .sql file), and it's only looking at the first line. 
```
> InputDataSet
[1] "42000 263 [Microsoft][ODBC SQL Server Driver][SQL Server]Must specify table to select from."
[2] "[RODBC] ERROR: Could not SQLExecDirect 'SELECT *'"                                          
> 
```

We need to make sure that `query` is a string with the full query.

```
> query
[1] "SELECT *"              "FROM ProspectiveBuyer"
> 
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5aead99fab615886b9099b1953e7f76513e69f17

Issue #2058, opened 2016-07-11
Goto definition should take selection into account


MikhailArkhipov:
```
I love the new F12 and Alt-F12 features (in version 0.4), but I think there’s a small improvement to make (so hence the frown rather than a smile)
I think it’s natural when wanting to see the definition of a function used in a piece of code to double-click the function name and then hit F12. But double-clicking has the effect of highlighting the function’s name and putting the caret just after the last character of the function name and in that case F12 does not work. Surely F12 should work in this case?
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5aead99fab615886b9099b1953e7f76513e69f17

Issue #2019, opened 2016-06-28
Putting $ evaluates the code and shows an error.


Marchen:
On 0.4, typing $ seems to evaluate the object just before the $ and shows an error if it doesn't exist.
But this is not true for existing object like `iris$`.

![default](https://cloud.githubusercontent.com/assets/3863577/16416984/66a7eb48-3d7f-11e6-8334-a326c0655934.PNG)

MikhailArkhipov:
In this particular case intellisense attempts to enumerate members of something$iris and fails. I guess the error can be hidden though.

Marchen:
The error was shown just after typing something$, not after typing iris$.

MikhailArkhipov:
Evaluation is by design since we need to know members and that is achieved by actual code evaluation. However, we should hide the error. ```rtvs:::describe_children``` appears to evaluating with visible output.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5aead99fab615886b9099b1953e7f76513e69f17

PR #2148, opened 2016-08-03
Assorted fixes

### LINKED ISSUE #2138: Replace TwoWay binding by event handlers in SQL publishing
Because of potential issues with WPF not sending back changes

MikhailArkhipov:
#2138 Replace TwoWay binding by event handlers in SQL publishing
#1990 Wrong reported default for Use multiline selection
#2058 Goto definition should take selection into account
#2019 Putting $ evaluates the code and shows an error.
#2146 Drag & drop of sql file creates code that doesn't work for mult… 
#2147 Incorrect parameter suggested in arguments with nested function…

### DIFF
--- a/src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
+++ b/src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
@@ -50,5 +50,21 @@ namespace Microsoft.VisualStudio.R.Package.Sql.Publish {
         private void OKButton_Click(object sender, RoutedEventArgs e) => SaveSettingsAndClose();
         private void CancelButton_Click(object sender, RoutedEventArgs e) => Close();
         private void TableName_TextChanged(object sender, TextChangedEventArgs e) => _model.UpdateState();
+
+        private void TargetTypeList_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            _model.SelectedTargetTypeIndex = TargetTypeList.SelectedIndex;
+        }
+
+        private void TargetList_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            _model.SelectedTargetIndex = TargetList.SelectedIndex;
+        }
+
+        private void CodePlacementList_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            _model.SelectedCodePlacementIndex = CodePlacementList.SelectedIndex;
+        }
+
+        private void QuoteTypeList_SelectionChanged(object sender, SelectionChangedEventArgs e) {
+            _model.SelectedQuoteTypeIndex = QuoteTypeList.SelectedIndex;
+        }
     }
 }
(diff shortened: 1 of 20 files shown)


src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
```diff
@@ -18,27 +19,53 @@ namespace Microsoft.VisualStudio.R.Package.Sql.Publish {
...
+            return dialog;
+        }
+
+        private SqlPublshOptionsDialog(IApplicationShell appShell, IProjectSystemServices pss, IProjectConfigurationSettingsProvider pcsp) :
             this(appShell, pss, new FileSystem(), pcsp) {
             InitializeComponent();
+
+            TargetTypeList.SelectedIndex = _model.SelectedTargetTypeIndex;
+            TargetList.SelectedIndex = _model.SelectedTargetIndex;
+            CodePlacementList.SelectedIndex = _model.SelectedQuoteTypeIndex;
```
AlexanderSher: You still have one-way binding, so this shouldn't be needed

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
```diff
@@ -18,27 +19,53 @@ namespace Microsoft.VisualStudio.R.Package.Sql.Publish {
...
     /// </summary>
     public partial class SqlPublshOptionsDialog : DialogWindow {
-        private readonly SqlPublishOptionsDialogViewModel _model;
         private readonly IApplicationShell _appShell;
         private readonly IProjectSystemServices _pss;
+        private readonly IProjectConfigurationSettingsProvider _pcsp;
+        private SqlPublishOptionsDialogViewModel _model;
 
-        public SqlPublshOptionsDialog(IApplicationShell appShell, IProjectSystemServices pss, IProjectConfigurationSettingsProvider pcsp) :
+        public static async Task<SqlPublshOptionsDialog> CreateAsync(
```
AlexanderSher:
```{csharp}
public static async Task<SqlPublshOptionsDialog> CreateAsync(
            IApplicationShell appShell, IProjectSystemServices pss, 
            IProjectConfigurationSettingsProvider pcsp) => CreateAsync(appShell, pss, new FileSystem(), pcsp)
```
And you need only one ctor in this case

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
MikhailArkhipov: I removed all bindings to the index since selection change notification is all we need

MikhailArkhipov: Oh, this one is for the initial selection only

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
MikhailArkhipov: It was used in tests, but I can get away without it

src/Package/Impl/Sql/Publish/SqlPublshOptionsDialog.xaml.cs
AlexanderSher: Got it

---
In reply to: [73415152](https://github.com/Microsoft/RTVS/pull/2148#discussion_r73415152) [](ancestors = 73415152)

AlexanderSher:
:+1:

Issue #2143, opened 2016-08-02
Intellisense not showing up for newly installed packages


karthiknadig:
Repro steps:
1. Install package, say 'RODBC'
2. when doing library() it does not show up.
3. Later it does not show any functions from thew newly installed package.

MikhailArkhipov:
This basically same as #1604. We need some knowledge that new package got installed or loaded from REPL. We do look at ```library(...)``` statements in the editor and can try and discover package but in REPL we don't.

MikhailArkhipov:
RS overrides ```install.packages``` and ```remove.packages```. We need to do the same, probably map over to ```rtvs:::package_uninstall``` and ```rtvs:::package_update``` and have them sending events

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/16bbfb33f3c03917ba44003ed507f46f3920cfac

Issue #1604, opened 2016-04-16
Package manager doesn't auto-update if package installed manually


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/14583466/87ecacc6-03d7-11e6-97eb-aec624b69a3d.png)

flipping tabs recovers

MikhailArkhipov:
See also #2143

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/9d74642f2e4ae7f0208b5b71f1d92bc26352e416

PR #2149, opened 2016-08-04
#1604 Package manager doesn't auto-update if package installed manually

MikhailArkhipov:
Same notifications as needed for #2143. Cleaned up and consolidated some duplicate code. Tests pass. Manual test scenarios are welcome.

### DIFF
--- a/src/Host/Client/Impl/Definitions/IRCallbacks.cs
+++ b/src/Host/Client/Impl/Definitions/IRCallbacks.cs
@@ -73,5 +73,8 @@ namespace Microsoft.R.Host.Client {
         /// </summary>
         /// <returns></returns>
         void ViewObject(string expression, string title);
+
+        void PackagesInstalled();
+        void PackagesRemoved();
     }
 }
--- a/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
@@ -62,7 +62,7 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
 
             _session.Mutated += RSessionMutated;
             _session.PackagesInstalled += OnPackagesInstalled;
-            _session.PackagesRemoved += OnPackagesInstalled;
+            _session.PackagesRemoved += OnPackagesRemoved;
         }
 
         public ReadOnlyObservableCollection<object> Items { get; }
--- a/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
@@ -96,7 +96,9 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
         }
 
         public async Task ReloadItemsAsync() {
-            switch (_selectedTab) {
+                await _coreShell.SwitchToMainThreadAsync();
+
+                switch (_selectedTab) {
                 case SelectedTab.AvailablePackages:
                     await ReloadAvailablePackagesAsync();
                     break;
@@ -412,6 +414,8 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
         }
 
         public async Task SwitchToInstalledPackagesAsync() {
+            await _coreShell.SwitchToMainThreadAsync();
+
             if (await SetTabAsync(SelectedTab.InstalledPackages)) {
                 await ReloadInstalledAndLoadedPackagesAsync();
                 await ReplaceItemsAsync();
--- a/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
+++ b/src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
@@ -96,9 +96,10 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
         }
 
         public async Task ReloadItemsAsync() {
-                await _coreShell.SwitchToMainThreadAsync();
+            await _coreShell.SwitchToMainThreadAsync();
 
-                switch (_selectedTab) {
+            var startingTab = _selectedTab;
+            switch (_selectedTab) {
                 case SelectedTab.AvailablePackages:
                     await ReloadAvailablePackagesAsync();
                     break;
@@ -107,13 +108,13 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
                     break;
                 case SelectedTab.LoadedPackages:
                     await ReloadLoadedPackagesAsync();
-                     break;
+                    break;
                 case SelectedTab.None:
                     return;
                 default:
                     throw new ArgumentOutOfRangeException();
             }
-            await ReplaceItemsAsync();
+            await ReplaceItemsAsync(startingTab);
         }
 
         public void SelectPackage(IRPackageViewModel package) {
@@ -147,6 +148,7 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
             }
 
             BeforeLoadUnload(package);
+            var startingTab = _selectedTab;
 
             try {
                 var libPath = await GetLibPath();
@@ -158,7 +160,7 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
             }
 
             await ReloadInstalledAndLoadedPackagesAsync();
-            await AfterLoadUnloadAsync(package);
+            await AfterLoadUnloadAsync(package, startingTab);
         }
 
         public async Task UpdateAsync(IRPackageViewModel package) {
@@ -172,25 +174,28 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
                 return;
             }
 
+            var startingTab = _selectedTab;
             BeforeLoadUnload(package);
 
             await UpdateImplAsync(package);
-            await AfterLoadUnloadAsync(package);
+            await AfterLoadUnloadAsync(package, startingTab);
         }
 
-        private async Task ReplaceItemsAsync() {
-            switch (_selectedTab) {
-                case SelectedTab.AvailablePackages:
-                    await ReplaceItemsAsync(_availablePackages);
-                    break;
-                case SelectedTab.InstalledPackages:
-                    await ReplaceItemsAsync(_installedPackages);
-                    break;
-                case SelectedTab.LoadedPackages:
-                    await ReplaceItemsAsync(_loadedPackages);
-                    break;
+        private async Task ReplaceItemsAsync(SelectedTab startingTab) {
+            if (startingTab == _selectedTab) {
+                switch (_selectedTab) {
+                    case SelectedTab.AvailablePackages:
+                        await ReplaceItemsAsync(_availablePackages);
+                        break;
+                    case SelectedTab.InstalledPackages:
+                        await ReplaceItemsAsync(_installedPackages);
+                        break;
+                    case SelectedTab.LoadedPackages:
+                        await ReplaceItemsAsync(_loadedPackages);
+                        break;
+                }
+                IsLoading = false;
             }
-            IsLoading = false;
         }
 
         private async Task UpdateImplAsync(IRPackageViewModel package) {
@@ -238,6 +243,7 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
             }
 
             BeforeLoadUnload(package);
+            var startingTab = _selectedTab;
 
             if (package.IsLoaded) {
                 try {
@@ -266,7 +272,7 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
                 await ReloadInstalledAndLoadedPackagesAsync();
             }
 
-            await AfterLoadUnloadAsync(package);
+            await AfterLoadUnloadAsync(package, startingTab);
         }
 
         public async Task LoadAsync(IRPackageViewModel package) {
@@ -274,7 +280,9 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
             if (package.IsLoaded) {
                 return;
             }
+
             BeforeLoadUnload(package);
+            var startingTab = _selectedTab;
 
             try {
                 await _packageManager.LoadPackageAsync(package.Name, package.LibraryPath.ToRPath());
@@ -285,7 +293,7 @@ namespace Microsoft.R.Components.PackageManager.Implementation.ViewModel {
             }
 
             await ReloadLoadedPackagesAsync();
-            await AfterLoadUnloadAsync(package);
+            await AfterLoadUnloadAsync(package, startingTab);
         }
 
         public async Task UnloadAsync(IRPackageViewModel package) {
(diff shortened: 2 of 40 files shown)


src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -59,6 +61,8 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             Items = new ReadOnlyObservableCollection<object>(_items);
 
             _session.Mutated += RSessionMutated;
+            _session.PackagesInstalled += OnPackagesInstalled;
+            _session.PackagesRemoved += OnPackagesInstalled;
```
karthiknadig: OnPackagesRemoved?

src/Host/Client/Impl/Definitions/IRCallbacks.cs
```diff
@@ -73,5 +73,8 @@ public interface IRCallbacks {
         /// </summary>
         /// <returns></returns>
         void ViewObject(string expression, string title);
+
+        void PackagesInstalled();
```
AlexanderSher: Do we really need 2 callbacks? There are no parameters, so I'd rather use one `PackagesChanged`

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -438,30 +412,37 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
         }
 
         public async Task SwitchToInstalledPackagesAsync() {
-            await _coreShell.SwitchToMainThreadAsync();
-            if (_selectedTab == SelectedTab.InstalledPackages) {
-                return;
+            if (await SetTabAsync(SelectedTab.InstalledPackages)) {
```
AlexanderSher: Since it is a public method, I'd like to have `await _coreShell.SwitchToMainThreadAsync();` in the beginning of it anyway

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -321,25 +303,19 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
+        private void BeforeLoadUnload(IRPackageViewModel package) {
             if (_selectedTab == SelectedTab.InstalledPackages || _selectedTab == SelectedTab.LoadedPackages) {
                 IsLoading = true;
             }
+            package.IsChanging = true;
         }
 
-        private void AfterLoadUnload(IRPackageViewModel package) {
+        private async Task AfterLoadUnloadAsync(IRPackageViewModel package) {
+            await ReplaceItemsAsync();
```
AlexanderSher: We need to check current tab, otherwise, if user has switched from `LoadedPackages` to `AvailablePackages`, we will remove progress ring while available packages are loading.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -91,22 +95,23 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             }
         }
 
-        public void ReloadItems() {
-            _coreShell.AssertIsOnMainThread();
+        public async Task ReloadItemsAsync() {
```
AlexanderSher: >ReloadItemsAsync [](start = 26, length = 16)

Please add `_coreShell.SwitchToMainThreadAsync` to the beginning of this method

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -173,23 +170,25 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
+        private async Task ReplaceItemsAsync() {
+            switch (_selectedTab) {
+                case SelectedTab.AvailablePackages:
+                    await ReplaceItemsAsync(_availablePackages);
+                    break;
+                case SelectedTab.InstalledPackages:
+                    await ReplaceItemsAsync(_installedPackages);
+                    break;
+                case SelectedTab.LoadedPackages:
+                    await ReplaceItemsAsync(_loadedPackages);
```
AlexanderSher: >ReplaceItemsAsync [](start = 26, length = 17)

I don't think `ReplaceItemsAsync` should change `IsLoading` to false - this method has no information about the context and can be called by ReloadItemsAsync from callback of the main session while packages on the opened tab are still loaded from the background session

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: Good catch. Fixed

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: The code in most places looked like

```
                ReplaceItems(_installedPackages);    
                IsLoading = false;    
```

If we don't set it, don't the panel continue showing spinning wheel?

MikhailArkhipov: I can add a flag as a parameter with default to reset.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: ```ReplaceItemsAsync``` checks what tab is the current

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: OK

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: OK

src/Host/Client/Impl/Definitions/IRCallbacks.cs
MikhailArkhipov: I though about that, but basically few reasons:

1. We may choose to pass list of packages in the notification so we can process Remove faster as it only removes items from collection.
2. Same in the intellisense package index

I made them the same for now just to save some time. We can optimize package manager and the package index in completion some time later.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: Or maybe check state of the semaphores? like

```
IsLoading = !_availableLock.IsCompleted
```

and use proper lock depending on the current tab?

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -388,40 +364,38 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
 
-        private async Task LoadAvailablePackagesAsync() {
+        private async Task LoadAvailablePackagesAsync(bool updateStateOnly = false) {
             await TaskUtilities.SwitchToBackgroundThread();
-            var availablePackages = await _packageManager.GetAvailablePackagesAsync();
+
             var vmAvailablePackages = new List<IRPackageViewModel>();
+            IReadOnlyList<RPackage> availablePackages = null;
+            if (updateStateOnly) {
+                availablePackages = _availablePackagesCache;
```
AlexanderSher: What's the purpose of this collection? We have `_availablePackages` already.

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -388,40 +364,38 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
                 }
+            } catch (RPackageManagerException ex) {
+                _coreShell.DispatchOnUIThread(() => AddErrorMessage(ex.Message));
+            } finally {
+                _availableLock.Release();
             }
         }
 
-        private async Task LoadAvailablePackagesAsync() {
+        private async Task LoadAvailablePackagesAsync(bool updateStateOnly = false) {
```
AlexanderSher: Not sure I understand this. If available packages are loaded,  state of every package will be updated in `ReloadInstalledAndLoadedPackagesAsync`, so if `updateStateOnly == true`, there is no reason to call this method at all

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: I though originally that they are different but actually only in ordering. There is a bit of a mismatch with IList/IReadOnlyList, in ```LoadAvailablePackagesAsync``` let me untangle that.

MikhailArkhipov: _availablePackages is ```IList<IRPackageViewModel>``` while `GetAvailablePackagesAsync` returns ```IReadOnlyList<RPackage>```

MikhailArkhipov: So basically cache allowed to avoid calls to `GetAvailablePackagesAsync`

src/Host/Client/Impl/Definitions/IRCallbacks.cs
AlexanderSher: Hm... ok, let's leave it as it is

---
In reply to: [73555637](https://github.com/Microsoft/RTVS/pull/2149#discussion_r73555637) [](ancestors = 73555637)

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -575,28 +560,17 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
-                return;
-            }
-
-            IsLoading = true;
-
-            await ReloadInstalledAndLoadedPackagesAsync();
-            if (_selectedTab == SelectedTab.LoadedPackages) {
-                IsLoading = false;
-                ReplaceItems(_loadedPackages);
+            if (await SetTabAsync(SelectedTab.LoadedPackages)) {
```
AlexanderSher: >SetTabAsync [](start = 22, length = 11)

Please add `await _coreShell.SwitchToMainThreadAsync();` at the beginning of this method

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -388,40 +366,38 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
-            }
-
-            _selectedTab = SelectedTab.AvailablePackages;
-            IsLoading = !_availableLock.IsCompleted;
-
-            await EnsureAvailablePackagesLoadedAsync();
-            if (_selectedTab == SelectedTab.AvailablePackages) {
-                IsLoading = false;
-                ReplaceItems(_availablePackages);
+            if (await SetTabAsync(SelectedTab.AvailablePackages)) {
```
AlexanderSher: >SetTabAsync [](start = 22, length = 11)

Please add `await _coreShell.SwitchToMainThreadAsync();` at the beginning of this method

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
AlexanderSher: It checks which tab is current, while we need to check if current tab is the one that was updated. If while updating user has switched the tab, we shouldn't call `ReplaceItemsAsync`, otherwise we will populate the list with outdated items.

---
In reply to: [73554727](https://github.com/Microsoft/RTVS/pull/2149#discussion_r73554727) [](ancestors = 73554727)

AlexanderSher: Here is the test (add it to `RPackageManagerViewModelTest`):

```{csharp}
        [Test(ThreadType.UI)]
        public async Task SwitchFromInstalledToAvailableWhileLoadingInstalled() {
            // We need real repo for this test
            await TestRepositories.SetCranRepoAsync(_workflow.RSession);

            _packageManagerViewModel.SwitchToLoadedPackagesAsync().DoNotWait();
            var t1 = _packageManagerViewModel.SwitchToInstalledPackagesAsync();
            var t2 = _packageManagerViewModel.SwitchToAvailablePackagesAsync();

            await t1;
            _packageManagerViewModel.IsLoading.Should().BeTrue();

            await t2;
            _packageManagerViewModel.IsLoading.Should().BeFalse();
        }
```

```{csharp}
        public static Task SetCranRepoAsync(IRExpressionEvaluator eval) {
            var code = $"options(repos=list(CRAN=\"[http://cran.r-project.org\](http://cran.r-project.org\)"))";
            return eval.ExecuteAsync(code);
        }
```

---
In reply to: [73560560](https://github.com/Microsoft/RTVS/pull/2149#discussion_r73560560) [](ancestors = 73560560,73554727)

karthiknadig:
should the rtvs:::package_uninstall and rtvs:::package_update be changed to use this version of install and remove?

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
MikhailArkhipov: Got it

MikhailArkhipov:
They already end up calling overloaded ```install.packages``` anyway

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -438,30 +424,37 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
...
+        private async Task ReloadAvailablePackagesAsync() {
             IsLoading = true;
-
             await ReloadInstalledAndLoadedPackagesAsync();
-            if (_selectedTab == SelectedTab.InstalledPackages) {
-                IsLoading = false;
-                ReplaceItems(_installedPackages);
-            }
+            // Updating state only for performance reasons
+            await EnsureAvailablePackagesLoadedAsync(updateStateOnly: true);
```
AlexanderSher: >EnsureAvailablePackagesLoadedAsync [](start = 18, length = 34)

I still think that you don't need to call `EnsureAvailablePackagesLoadedAsync` here - `ReloadInstalledAndLoadedPackagesAsync` should take care of updating available packages that are installed and loaded or will trigger `EnsureAvailablePackagesLoadedAsync(false)` if available packages aren't available yet. Do you have a specific scenario where this call is required?

MikhailArkhipov: Yes, you are right. Seems to work OK. I removed it and the associated stuff like cache

AlexanderSher:
:+1:

PR #2157, opened 2016-08-08
Export to Excel restrictions relaxed.  Fixes #2155

### LINKED ISSUE #2155: Allow xts (and other object types) to be transferred to Excel
Hello,

In the Variable Explorer window, currently only data frames can be opened in Excel. There are other R types, which are widely used, and are quite similar to a data frame. Time series objects (package xts) come to mind as an example. Could these be supported as well?

Thanks in advance

karthiknadig:
If a variable can be coerced to data.frame, and length is greater than 1, export to excel button is shown for that variable.
image

### DIFF
--- a/src/Package/Impl/DataInspect/VariableViewModel.cs
+++ b/src/Package/Impl/DataInspect/VariableViewModel.cs
@@ -61,7 +61,7 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
                 ShowDetailCommandTooltip = Resources.ShowDetailCommandTooltip;
             }
 
-            CanShowOpenCsv = result.CanCoerceToDataFrame && (result.Length > 1 || (result.Length == 1 && result.TypeName == "S4"));
+            CanShowOpenCsv = result.CanCoerceToDataFrame && (result.Length > 1 || result.TypeName == "S4");
             if (CanShowOpenCsv) {
                 OpenInCsvAppCommand = new DelegateCommand(OpenInCsvApp, o => CanShowOpenCsv);
                 OpenInCsvAppCommandTooltip = Resources.OpenCsvAppCommandTooltip;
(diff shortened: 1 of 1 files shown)


src/Package/Impl/DataInspect/VariableViewModel.cs
```diff
@@ -61,8 +61,7 @@ public sealed class VariableViewModel : RSessionDataObject, IIndexedItem {
                 ShowDetailCommandTooltip = Resources.ShowDetailCommandTooltip;
             }
 
-            const ViewerCapabilities tableCaps = (ViewerCapabilities.Table | ViewerCapabilities.List);
-            CanShowOpenCsv = CanShowDetail && result.CanCoerceToDataFrame && _detailsViewer.Capabilities.HasFlag(tableCaps) && result.Length > 0;
+            CanShowOpenCsv = result.CanCoerceToDataFrame && result.Length > 1;
```
int19h: Can you also add a check for typename "S4" here? Those will usually have a length of 1, but they can still redefine `as.data.frame` - and if they do, we should show export for them.

karthiknadig: will do.

---
In reply to: [73923629](https://github.com/Microsoft/RTVS/pull/2157#discussion_r73923629) [](ancestors = 73923629)

int19h:
:+1:

src/Package/Impl/DataInspect/VariableViewModel.cs
```diff
@@ -61,7 +61,7 @@ public sealed class VariableViewModel : RSessionDataObject, IIndexedItem {
                 ShowDetailCommandTooltip = Resources.ShowDetailCommandTooltip;
             }
 
-            CanShowOpenCsv = result.CanCoerceToDataFrame && result.Length > 1;
+            CanShowOpenCsv = result.CanCoerceToDataFrame && (result.Length > 1 || (result.Length == 1 && result.TypeName == "S4"));
```
int19h: No need to do the `==1` check for S4 - they are length 1 by default, but can override to mean something else, and there's no reason to suppress export for those that do override length (but provide `as.data.frame`).

int19h:
:+1:

PR #2159, opened 2016-08-08
Adds feature to fetch a file from remote RHost.

karthiknadig:
Adds feature to fetch a file from remote RHost. This file will be saved to user's download folder.


src/Common/Core/Impl/IO/FileSystem.cs
```diff
@@ -85,9 +86,42 @@ public sealed class FileSystem : IFileSystem {
...
+                return Marshal.PtrToStringUni(knownFolderPath);
+            } else {
+                return string.Empty;
+            }
+        }
+
         private static class NativeMethods {
             public const int MAX_PATH = 260;
 
+            public class KnownFolderflags {
```
int19h: Nit: "Flags"

Also, looks like it could be `[Flags] enum`?

src/Host/Client/Impl/Definitions/IRCallbacks.cs
```diff
@@ -76,5 +76,11 @@ public interface IRCallbacks {
 
         void PackagesInstalled();
         void PackagesRemoved();
+
+        /// <summary>
+        /// Called when user invokes rtvs:::fetch_file() in R.
+        /// </summary>
+        /// <param name="id"></param>
+        Task<string> SaveFile(string filename, byte[] data);
```
int19h: ...Async

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -579,5 +579,10 @@ internal sealed class RSession : IRSession, IRCallbacks {
         void IRCallbacks.PackagesRemoved() {
             PackagesRemoved?.Invoke(this, EventArgs.Empty);
         }
+
+        Task<string> IRCallbacks.SaveFile(string filename, byte[] data) {
+            var callback = _callback;
+            return callback?.SaveFile(filename, data);
```
AlexanderSher: What should happen if `callback` is null?

src/Common/Core/Impl/IO/FileSystem.cs
```diff
@@ -85,9 +86,42 @@ public sealed class FileSystem : IFileSystem {
...
 
+        public string GetDownloadsPath(string fileName) {
+            if (string.IsNullOrWhiteSpace(fileName)) {
+                return GetKnownFolder(KnownFolderGuids.Downloads);
+            } else {
+                return Path.Combine(GetKnownFolder(KnownFolderGuids.Downloads), fileName);
+            }
+        }
+
+        private string GetKnownFolder(string knownFolder) {
```
AlexanderSher: Is it `GetKnownFolder` or `GetKnownFolderPath`?

int19h:
:+1:

PR #2171, opened 2016-08-13
Adds remote F5 feature. Updates properties page for Run.

karthiknadig:
Adds remote F5 feature. Updates properties page for Run. Adds context menu command to set R Script as StartUp file.

### DIFF
--- a/src/Common/Core/Impl/IO/FileSystem.cs
+++ b/src/Common/Core/Impl/IO/FileSystem.cs
@@ -7,6 +7,8 @@ using System.IO;
 using System.IO.Compression;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Threading;
+
 using Microsoft.Extensions.FileSystemGlobbing;
 
 namespace Microsoft.Common.Core.IO {
@@ -73,10 +75,10 @@ namespace Microsoft.Common.Core.IO {
         public string CompressDirectory(string path) {
             Matcher matcher = new Matcher(StringComparison.InvariantCultureIgnoreCase);
             matcher.AddInclude("*.*");
-            return CompressDirectory(path, matcher, (p) => { });
+            return CompressDirectory(path, matcher, new Progress<string>((p) => { }), CancellationToken.None);
         }
 
-        public string CompressDirectory(string path, Matcher matcher, Action<string> callback) {
+        public string CompressDirectory(string path, Matcher matcher, IProgress<string> progress, CancellationToken ct) {
             string zipFilePath = Path.GetTempFileName();
             using (FileStream zipStream = new FileStream(zipFilePath, FileMode.Create)) 
             using (ZipArchive archive = new ZipArchive(zipStream, ZipArchiveMode.Create)) {
@@ -91,7 +93,10 @@ namespace Microsoft.Common.Core.IO {
 
                     var files = matcher.GetResultsInFullPath(dir);
                     foreach (var file in files) {
-                        callback?.Invoke(file);
+                        if (ct.IsCancellationRequested) {
+                            return string.Empty;
+                        }
+                        progress?.Report(file);
                         string entryName = file.MakeRelativePath(dir).Replace('\\', '/');
                         archive.CreateEntryFromFile(file, entryName);
                     }
--- a/src/Common/Core/Impl/IO/FileSystem.cs
+++ b/src/Common/Core/Impl/IO/FileSystem.cs
@@ -4,55 +4,34 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Compression;
 using System.Runtime.InteropServices;
 using System.Text;
 
 namespace Microsoft.Common.Core.IO {
     public sealed class FileSystem : IFileSystem {
-        public IFileSystemWatcher CreateFileSystemWatcher(string path, string filter) {
-            return new FileSystemWatcherProxy(path, filter);
-        }
+        public IFileSystemWatcher CreateFileSystemWatcher(string path, string filter) => new FileSystemWatcherProxy(path, filter);
+        
+        public IDirectoryInfo GetDirectoryInfo(string directoryPath) => new DirectoryInfoProxy(directoryPath);
+        
+        public bool FileExists(string path) => File.Exists(path);
 
-        public IDirectoryInfo GetDirectoryInfo(string directoryPath) {
-            return new DirectoryInfoProxy(directoryPath);
-        }
+        public string ReadAllText(string path) => File.ReadAllText(path);
+        
+        public void WriteAllText(string path, string content) => File.WriteAllText(path, content);
+        
+        public IEnumerable<string> FileReadAllLines(string path) => File.ReadLines(path);
+        
+        public void FileWriteAllLines(string path, IEnumerable<string> contents) => File.WriteAllLines(path, contents);
 
-        public bool FileExists(string path) {
-            return File.Exists(path);
-        }
+        public byte[] FileReadAllBytes(string path) => File.ReadAllBytes(path);
+        
+        public void FileWriteAllBytes(string path, byte[] bytes) => File.WriteAllBytes(path, bytes);
 
-        public string ReadAllText(string path) {
-            return File.ReadAllText(path);
-        }
+        public bool DirectoryExists(string path) => Directory.Exists(path);
 
-        public void WriteAllText(string path, string content) {
-            File.WriteAllText(path, content);
-        }
-
-        public IEnumerable<string> FileReadAllLines(string path) {
-            return File.ReadLines(path);
-        }
-
-        public void FileWriteAllLines(string path, IEnumerable<string> contents) {
-            File.WriteAllLines(path, contents);
-        }
-
-        public byte[] FileReadAllBytes(string path) {
-            return File.ReadAllBytes(path);
-        }
-
-        public void FileWriteAllBytes(string path, byte[] bytes) {
-            File.WriteAllBytes(path, bytes);
-        }
-
-        public bool DirectoryExists(string path) {
-            return Directory.Exists(path);
-        }
-
-        public FileAttributes GetFileAttributes(string path) {
-            return File.GetAttributes(path);
-        }
-
+        public FileAttributes GetFileAttributes(string path) => File.GetAttributes(path);
+        
         public string ToLongPath(string path) {
             var sb = new StringBuilder(NativeMethods.MAX_PATH);
             NativeMethods.GetLongPathName(path, sb, sb.Capacity);
@@ -70,20 +49,66 @@ namespace Microsoft.Common.Core.IO {
             return new FileVersionInfo(fvi.FileMajorPart, fvi.FileMinorPart);
         }
 
-        public void DeleteFile(string path) {
-            File.Delete(path);
+        public void DeleteFile(string path) => File.Delete(path);
+
+        public void DeleteDirectory(string path, bool recursive) => Directory.Delete(path, recursive);
+
+        public string[] GetFileSystemEntries(string path, string searchPattern, SearchOption options) => Directory.GetFileSystemEntries(path, searchPattern, options);
+        
+        public void CreateDirectory(string path) => Directory.CreateDirectory(path);
+
+        public string CompressFile(string path) {
+            string compressedFilePath = Path.GetTempFileName();
+            using (FileStream sourceFileStream = File.OpenRead(path)) {
+                using (FileStream compressedFileStream = File.Create(compressedFilePath)) {
+                    using (GZipStream stream = new GZipStream(compressedFileStream, CompressionLevel.Optimal)) {
+                        // 81920 is the default compression buffer size
+                        sourceFileStream.CopyTo(compressedFileStream, 81920);
+                    }
+                }
+            }
+            return compressedFilePath;
         }
 
-        public void DeleteDirectory(string path, bool recursive) {
-            Directory.Delete(path, recursive);
-        }
+        public string CompressDirectory(string path) =>
+            CompressDirectory(path, (p) => true);
 
-        public string[] GetFileSystemEntries(string path, string searchPattern, SearchOption options) {
-            return Directory.GetFileSystemEntries(path, searchPattern, options);
-        }
+        public string CompressDirectory(string path, Func<string, bool> filter) {
+            string zipFilePath = Path.GetTempFileName();
+            using (FileStream zipStream = new FileStream(zipFilePath, FileMode.Create)) {
+                using (ZipArchive archive = new ZipArchive(zipStream, ZipArchiveMode.Create)) {
+                    Stack<string> dirs = new Stack<string>();
+                    dirs.Push(path);
+                    while (dirs.Count > 0) {
+                        string dirPath = dirs.Pop();
+                        var subdirs = Directory.GetDirectories(dirPath);
+                        foreach (string subdir in subdirs) {
+                            bool addDir = true;
+                            if (filter != null) {
+                                addDir = filter(subdir);
+                            }
+                            if (addDir) {
+                                dirs.Push(subdir);
+                            }
+                        }
 
-        public void CreateDirectory(string path) {
-            Directory.CreateDirectory(path);
+                        var files = Directory.GetFiles(dirPath);
+                        foreach (string file in files) {
+                            bool addFile = true;
+                            if (filter != null) {
+                                addFile = filter(file);
+                            }
+
+                            if (addFile) {
+                                string entryName = file.Remove(0, path.Length + 1).Replace('\\', '/');
+                                archive.CreateEntryFromFile(file, entryName);
+                            }
+                        }
+                    }
+                }
+            }
+
+            return zipFilePath;
         }
 
         public string GetDownloadsPath(string fileName) {
--- a/src/Common/Core/Impl/IO/FileSystem.cs
+++ b/src/Common/Core/Impl/IO/FileSystem.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.IO.Compression;
 using System.Runtime.InteropServices;
 using System.Text;
+using Microsoft.Extensions.FileSystemGlobbing;
 
 namespace Microsoft.Common.Core.IO {
     public sealed class FileSystem : IFileSystem {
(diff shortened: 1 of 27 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -61,6 +61,7 @@ internal sealed class RSession : IRSession, IRCallbacks {
         public int MaxLength { get; private set; } = 0x1000;
         public bool IsHostRunning => _isHostRunning;
         public Task HostStarted => _initializationTcs?.Task ?? Task.FromCanceled(new CancellationToken(true));
+        public bool IsRemoteSession { get; }
```
AlexanderSher: Just `IsRemote`

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -90,6 +91,8 @@ internal sealed class RSession : IRSession, IRCallbacks {
             });
 
             _afterHostStartedTask = TaskUtilities.CreateCanceled(new RHostDisconnectedException());
+
+            IsRemoteSession = brokerConnector.IsRemoteConnection();
```
AlexanderSher: This should be updated each time RHost is connected

src/Package/Impl/ProjectSystem/Commands/SetAsStartUpRScriptCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+        public Task<CommandStatusResult> GetCommandStatusAsync(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, string commandText, CommandStatus progressiveStatus) {
+            if (commandId == RPackageCommandId.icmdSetAsStartUpRScript && nodes.IsSingleNodePath()) {
+                return Task.FromResult(new CommandStatusResult(true, commandText, CommandStatus.Enabled | CommandStatus.Supported));
+            }
+            return Task.FromResult(CommandStatusResult.Unhandled);
+        }
+
+
+        public async Task<bool> TryHandleCommandAsync(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut) {
+            VsAppShell.Current.AssertIsOnMainThread();
```
AlexanderSher: Import `ICoreShell` from importing constructor

src/Package/Impl/ProjectSystem/ProjectProperties.cs
```diff
@@ -96,6 +104,111 @@ public ProjectProperties(ConfiguredProject configuredProject)
...
+            var runProps = await this.GetConfigurationRunPropertiesAsync();
+            await runProps.TransferProjectOnRun.SetValueAsync(val);
+        }
+
+        /// <summary>
+        /// Gets all the R script files in the current project.
+        /// </summary>
+        public IEnumerable<string> GetRFilePaths() {
+            IList<string> files = new List<string>();
+            string projectDir = Path.GetDirectoryName(this.ConfiguredProject.UnconfiguredProject.FullPath);
```
AlexanderSher: Don't use `this`

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
-            if (!FileSystem.FileExists(file)) {
+        private async Task SourceFileAsync(bool transferFiles, string file, string errorMessage) {
+            bool fExists = false;
+            if (transferFiles && Session.IsRemoteSession) {
+                fExists = await Session.EvaluateAsync<bool>(Invariant($"file.exists({file.ToRPath().ToRStringLiteral()})"), REvaluationKind.Normal);
+            } else {
+                fExists = FileSystem.FileExists(file);
+            }
+
+            if (!fExists) {
```
AlexanderSher: Since you output information about errors, you need to catch `RHostDisconnectedException` every time you call `EvaluateAsync`

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -45,6 +46,7 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
         internal IFileSystem FileSystem { get; set; } = new FileSystem();
 
         private IRSession Session => _interactiveWorkflow.RSession;
+        private IRCallbacks Callbacks => (IRCallbacks)_interactiveWorkflow.RSession;
```
AlexanderSher: `IRCallbacks` is designed to be used only by `RHost`. Why do you need it here?

src/Package/Impl/source.extension.vsixmanifest
```diff
@@ -22,6 +22,8 @@
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.VisualStudio.ImageCatalog.dll" />
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.Data.ConnectionUI.dll" />
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.Data.ConnectionUI.Dialog.dll" />
+    <Asset Type="Microsoft.VisualStudio.Assembly" Path="zlib.dll" />
```
AlexanderSher: Do we need to add these assemblies to setup?

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
+            }
+            private static string[] separator = { ";"};
+
+            public bool Match(string path) {
+                var result = false;
+                foreach (string pattern in filterPatterns) {
+                    if (pattern == "*.*") {
+                        result = true;
+                    } else if (pattern.StartsWith("*.") || pattern.StartsWith("*.")) {
+                        result = Path.GetExtension(path) == Path.GetExtension(pattern);
```
AlexanderSher: This will get false positive for the "*.abc.r" and *.def.r"

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
+                filterPatterns = new List<string>(patterns);
+            }
+            private static string[] separator = { ";"};
+
+            public bool Match(string path) {
+                var result = false;
+                foreach (string pattern in filterPatterns) {
+                    if (pattern == "*.*") {
+                        result = true;
+                    } else if (pattern.StartsWith("*.") || pattern.StartsWith("*.")) {
```
AlexanderSher: Conditions look identical

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
+        }
+
+        private async Task<string> GetStartupFileAsync(bool transferFiles, EnvDTE.Project project) {
+            if (transferFiles && Session.IsRemoteSession) { // remote
+                var projectName = Path.GetFileNameWithoutExtension(project.FullName);
+                var remotePath = (await _properties.GetRemoteProjectPathAsync()).ToRPath();
+                var startUpFile = (await _properties.GetStartupFileAsync()).ToRPath();
+                return remotePath + projectName + "/" + startUpFile;
+            } else { // local
+                var projDir = Path.GetDirectoryName(project.FullName);
```
AlexanderSher: Do we need to do anything if `transferFiles` is `true`, but `Session.IsRemoteSession` is false

src/Common/Core/Impl/IO/FileSystem.cs
```diff
@@ -69,20 +49,66 @@ public sealed class FileSystem : IFileSystem {
...
+
+        public void DeleteDirectory(string path, bool recursive) => Directory.Delete(path, recursive);
+
+        public string[] GetFileSystemEntries(string path, string searchPattern, SearchOption options) => Directory.GetFileSystemEntries(path, searchPattern, options);
+        
+        public void CreateDirectory(string path) => Directory.CreateDirectory(path);
+
+        public string CompressFile(string path) {
+            string compressedFilePath = Path.GetTempFileName();
+            using (FileStream sourceFileStream = File.OpenRead(path)) {
```
int19h: Nit: you can skip curly braces for nested `using` blocks to avoid excessive indentation (and VS will respect that when doing auto-format). So:
```cs
using (...)
using (...)
using (...) {
 ...
}
```

src/Host/Client/Impl/Host/RHostBrokerConnectorExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using Microsoft.Common.Core;
+
+namespace Microsoft.R.Host.Client.Host {
+    public static class RHostBrokerConnectorExtensions {
+        public static bool IsRemoteConnection(this IRHostBrokerConnector brokerConnector) {
+            return brokerConnector.BrokerUri.Scheme.StartsWithIgnoreCase("http");
```
int19h: It should be anything other than "file" - most of the time it will actually be "https".

src/Package/Impl/ProjectSystem/ProjectProperties.cs
```diff
@@ -2,23 +2,31 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.ComponentModel.Composition;
+using System.IO;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.ProjectSystem;
 #if VS14
 using Microsoft.VisualStudio.ProjectSystem.Utilities;
+using System.Collections.Generic;
```
int19h: This doesn't look like it's VS14-specific?

src/Package/Impl/ProjectSystem/ProjectProperties.cs
```diff
@@ -96,6 +104,111 @@ public ProjectProperties(ConfiguredProject configuredProject)
...
+                    var subdir = info as IDirectoryInfo;
+                    if (subdir != null) {
+                        dirs.Push(subdir);
+                    } else if (info.FullName.EndsWithIgnoreCase(".R")) {
+                        files.Add(info.FullName.Remove(0, projDir.FullName.Length + 1));
+                    }
+                }
+            }
+
+            return (IEnumerable<string>)files;
```
int19h: Nit: unnecessary cast.

src/Package/Impl/ProjectSystem/ProjectProperties.cs
```diff
@@ -96,6 +104,111 @@ public ProjectProperties(ConfiguredProject configuredProject)
...
+        /// </summary>
+        public IEnumerable<string> GetRFilePaths() {
+            IList<string> files = new List<string>();
+            string projectDir = Path.GetDirectoryName(this.ConfiguredProject.UnconfiguredProject.FullPath);
+            var projDir = _fileSystem.GetDirectoryInfo(projectDir);
+
+            Stack<IDirectoryInfo> dirs = new Stack<IDirectoryInfo>();
+            dirs.Push(projDir);
+
+            while (dirs.Count > 0) {
```
int19h: It might be worth making this (enumerating files recursively) a helper extension method on `IDirectoryInfo`.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
                 }
             }
         }
 
-        private async Task SourceFileAsync(string file, string errorMessage) {
-            if (!FileSystem.FileExists(file)) {
+        private async Task SourceFileAsync(bool transferFiles, string file, string errorMessage) {
+            bool fExists = false;
+            if (transferFiles && Session.IsRemoteSession) {
+                fExists = await Session.EvaluateAsync<bool>(Invariant($"file.exists({file.ToRPath().ToRStringLiteral()})"), REvaluationKind.Normal);
```
int19h: No need for `Invariant` - `EvaluateAsync` provides overloads that take `FormattableString`, and will do this for you.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
+                        }
                     }
                 }
             }
         }
 
-        private async Task SourceFileAsync(string file, string errorMessage) {
-            if (!FileSystem.FileExists(file)) {
+        private async Task SourceFileAsync(bool transferFiles, string file, string errorMessage) {
+            bool fExists = false;
```
int19h: Nit: no Hungarian notation.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
int19h: It looks like this is to output to the REPL. I believe this should go through `IRInteractiveWorkflow.ActiveWindow.InteractiveWindow.Write...`?

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -85,31 +87,126 @@ public RDebugLaunchProvider(ConfiguredProject configuredProject, IRInteractiveWo
...
+
+        private class TransferFileFilter {
+            List<string> filterPatterns;
+            public TransferFileFilter(string filterString) {
+                string[] patterns = filterString.Split(separator, StringSplitOptions.RemoveEmptyEntries);
+                filterPatterns = new List<string>(patterns);
+            }
+            private static string[] separator = { ";"};
+
+            public bool Match(string path) {
```
int19h: Have a look at this:

https://www.nuget.org/packages/Microsoft.Extensions.FileSystemGlobbing

https://docs.asp.net/projects/api/en/latest/autoapi/Microsoft/Extensions/FileSystemGlobbing/Matcher/index.html#matcher-class

src/Package/Impl/source.extension.vsixmanifest
karthiknadig: I have not yet added these assemblies to setup.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
karthiknadig: It should be fine to have `transferFiles` be `true` while `Session.IsRemoteSession` is `false`. This is the case where the user is locally editing and testing their code before switching to remote host. We just don't transfer when user is in local mode.

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
karthiknadig: I filter i had was temporary. I have used the FIleSystemGlobbing extension in this iteration.

---
In reply to: [74678172](https://github.com/Microsoft/RTVS/pull/2171#discussion_r74678172) [](ancestors = 74678172)

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
karthiknadig: The filtering is now done via globbing extension.

---
In reply to: [74677192](https://github.com/Microsoft/RTVS/pull/2171#discussion_r74677192) [](ancestors = 74677192)

src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
```diff
@@ -983,6 +984,7 @@
   </ItemGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
+    <Reference Include="Microsoft.Extensions.FileSystemGlobbing, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, processorArchitecture=MSIL" />
```
int19h: >Microsoft.Extensions.FileSystemGlobbing [](start = 24, length = 39)

This needs to be added to source.manifest.vsix and Setup.

src/Common/Core/Impl/IO/IFileSystem.cs
```diff
@@ -28,5 +30,9 @@ public interface IFileSystem {
         void DeleteDirectory(string path, bool recursive);
         string[] GetFileSystemEntries(string path, string searchPattern, SearchOption options);
         void CreateDirectory(string path);
+
+        string CompressFile(string path);
+        string CompressDirectory(string path);
+        string CompressDirectory(string path, Matcher matcher, Action<string> callback);
```
int19h: >callback [](start = 78, length = 8)

Since this is effectively a progress callback now, consider using IProgress<T> here for the sake of clarity. At call site, Progress<T> provides an easy way to wrap a lambda, so it should be almost as concise.

int19h: Also, perhaps a cancellation token (and check it before processing each entry in the implementation)? At some point we might want to do a fancier UI here, probably with a progress bar and a cancel button, and then it'll come in handy.

---
In reply to: [74853811](https://github.com/Microsoft/RTVS/pull/2171#discussion_r74853811) [](ancestors = 74853811)

src/Package/Impl/ProjectSystem/ProjectProperties.cs
```diff
@@ -92,10 +99,106 @@ public ProjectProperties(ConfiguredProject configuredProject)
...
+
+        /// <summary>
+        /// Gets all the R script files in the current project.
+        /// </summary>
+        public IEnumerable<string> GetRFilePaths() {
+            IList<string> rFiles = new List<string>();
+            string projectDir = Path.GetDirectoryName(ConfiguredProject.UnconfiguredProject.FullPath);
+            var allFiles = _fileSystem.GetDirectoryInfo(projectDir).GetAllFiles();
+            
+            foreach(var file in allFiles) {
```
int19h: >foreach [](start = 12, length = 7)

Now that the recursion logic is elsewhere, this loop can become `allFiles.Where(file => file.FullName.EndsWithIgnoreCase(".R"))` (and the caller can call `ToList` or `ToArray` if they need a snapshot, or not if they don't).

karthiknadig:
Ping.

src/SetupRHost/Product.wxs
```diff
@@ -49,6 +49,8 @@
     <DirectoryRef Id="INSTALLFOLDER">
       <Component Id="ProductAssemblies" Guid="{AFE50A8E-1F80-409E-A14C-6D940625C78F}">
         <File Source="$(var.BinDir)\Microsoft.R.Host.exe" />
+        <File Source="$(var.BinDir)\zip.dll" />
```
int19h: I don't recall if this was in one of the past iterations or not, but can you also check that there are `<VSIXSourceItem>` entries for these DLLs inside Microsoft.VisualStudio.R.Package.csproj? If not, they should be added.

int19h: Never mind, wrong list for host dependencies.

int19h:
:+1:

AlexanderSher:
:+1:

Issue #2081, opened 2016-07-15
Use pipes to communicate between broker and host


int19h:
Using websockets the way we do now presents a potential security issue in a multi-user environment, since it is possible for one user to connect to a socket that is intended for another by racing against the original connection.

Instead, we should use a transport mechanism that does not offer an attack window between process start and establishing the connection. It seems that using stdin/stdout for this purpose would fit the bill, and could also simplify the host code significantly, as there's no websocket dependency anymore.

smortaz:
how is jupyter doing this with their protocol (which i assume they've done
& redone & addressed some of the same types of issues)?

int19h:
On a multi-user system, they have the same problem, potentially. It may be mitigated by using authentication - the problem is that there's no websocket library for C++ that supports it.

There are also various schemes utilizing one-off generated auth tokens. The problem is that there's no good way to transmit them from the broker to the host. Command line and environment variables can both be inspected by other processes, so they do not constitute a secure transmission channel.

Issue #2040, opened 2016-07-06
Switch to ASP.NET Core WebSocketClient in RTVS


int19h:
websocket-sharp doesn't support integrated Windows authentication (NTLM), which we need to connect to the broker. Need to switch to this:

https://github.com/aspnet/WebSockets/blob/dev/src/Microsoft.AspNetCore.WebSockets.Client/WebSocketClient.cs

int19h:
https://github.com/Microsoft/RTVS/pull/2173

Issue #2039, opened 2016-07-05
Broker: host protocol pipe


int19h:
Endpoint for the websocket pipe that connects to R host behind a specific session, and provides two-way transmission of host protocol packets.

(There are two endpoints here - one exposed to the clients, and the other one that the hosts use to connect to the broker.)

int19h:
https://github.com/Microsoft/RTVS/pull/2173

Issue #2038, opened 2016-07-05
Broker: session management


int19h:
Endpoint to enumerate existing sessions, create a new session (for given R interpreter, command line etc), and delete a session.

int19h:
https://github.com/Microsoft/RTVS/pull/2173

PR #2173, opened 2016-08-15
Add remoting support via broker

int19h:
- Basic interpreter and session management.
- Basic authentication and authorization.
- Host/broker communication via stdin/out.
- All RTVS code is now working through the broker for local connections.
- Various associated unit test fixes.


int19h:
This does not include tests for the broker functionality - it's indirectly tested for local scenarios by our existing test suite, but there will also be separate tests for broker itself.

src/Host/Broker/Impl/Lifetime/PingController.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Microsoft.R.Host.Broker.Lifetime {
```
int19h: This isn't actually used by VS side at the moment, and lifetime is managed by PID watching instead. I kept it around in case we need it later.

src/Host/Broker/Impl/NativeMethods.cs
```diff
@@ -0,0 +1,210 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.R.Host.Broker {
```
int19h: Most of definitions here are unused (or rather used by `ProcessHelpers`, which is unused - see comment there). We will likely need to switch from `Process` to direct Win32 calls for security purposes, though, and then it'll come in handy.

src/Host/Broker/Impl/Pipes/ProcessHelpers.cs
```diff
@@ -0,0 +1,119 @@
...
+
+using System;
+using System.ComponentModel;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Security.Principal;
+using Microsoft.Win32.SafeHandles;
+
+namespace Microsoft.R.Host.Broker.Pipes {
+    internal static class ProcessHelpers {
```
int19h: This is not used currently, and `Session` uses `Process` class instead. However, it's likely that this code will need to be revived to avoid storing credentials in memory. Keeping it around as is for now as a prototype for future work.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -0,0 +1,196 @@
...
+            _process.BeginErrorReadLine();
+
+            _pipe = new MessagePipe(messageLogger);
+            var hostEnd = _pipe.ConnectHost();
+
+            ClientToHostWorker(_process.StandardInput.BaseStream, hostEnd).DoNotWait();
+            HostToClientWorker(_process.StandardOutput.BaseStream, hostEnd).DoNotWait();
+        }
+
+        //public void StartHost() {
```
int19h: Alternative code that used `ProcessHelpers`. Didn't get it to fully work yet, but this will be the basis for future rework here.

src/Host/Broker/Impl/Startup/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            CreateWebHost().Run();
+        }
+
+        public static IWebHost CreateWebHost() {
+            var webHostBuilder = new WebHostBuilder()
+                .UseLoggerFactory(_loggerFactory)
+                .UseConfiguration(Configuration)
+                .UseKestrel(options => {
+                    //options.UseConnectionLogging();
```
int19h: This actually logs entire HTTP traffic (including contents), so it's very verbose, and should not be on by default; but can be helpful occasionally when debugging, so keeping it commented out for that purpose.

src/Host/Client/Impl/Host/IRHostConnector.cs
```diff
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.R.Host.Client.Host {
-    public interface IRHostConnector {
+    public interface IRHostConnector : IDisposable {
```
int19h: This entire area (`IRHostConnector`, `IRHostBrokerConnector` and their respective implementations) should be considered a crude prototype for the real thing.

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -0,0 +1,120 @@
...
+            IsDisposed = true;
+        }
+
+        protected abstract Task ConnectToBrokerAsync();
+
+        public async Task PingAsync() {
+            (await _broker.PostAsync("/ping", new StringContent(""))).EnsureSuccessStatusCode();
+        }
+
+        private async Task PingWorker() {
```
int19h: Not used currently, but remains an alternative way to maintain broker lifetime if desired.

src/Host/Client/Impl/Transports/WebSocketClient.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
int19h: This entire file is from:

https://github.com/aspnet/WebSockets/tree/8b25152a67c64d03e86ed2b2347650c78b441529/src/Microsoft.AspNetCore.WebSockets.Client

They don't offer a NuGet package, but since it's a self-contained class, there are no modifications necessary to just embed it directly into the project.

src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
```diff
@@ -1375,6 +1379,118 @@
       <VSIXSourceItem Include="$(OutputPath)\websocket-sharp.dll" />
       <VSIXSourceItem Include="$(OutputPath)\Microsoft.Data.ConnectionUI.Dialog.dll" />
       <VSIXSourceItem Include="$(OutputPath)\Microsoft.Data.ConnectionUI.dll" />
+
+<VSIXSourceItem Include="$(OutputPath)\Microsoft.Extensions.Caching.Abstractions.dll" />
```
int19h: Also needs to be added to the installer - working on it.

src/Common/Core/Impl/Telemetry/TelemetryServiceBase.cs
```diff
@@ -32,13 +32,13 @@ public abstract class TelemetryServiceBase : ITelemetryService, IDisposable {
         /// </summary>
         public bool IsEnabled {
             get {
-                return this.TelemetryRecorder.IsEnabled;
+                return this.TelemetryRecorder?.IsEnabled == true;
```
AlexanderSher: What is the case when `TelemetryRecorder` is null?

int19h: I'm not entirely sure about the cause, but I was hitting this in the tests sporadically. Doesn't seem to be related to the changes.

src/Host/Broker/Impl/project.json
```diff
@@ -1,20 +1,28 @@
...
     "Microsoft.Extensions.Configuration.CommandLine": "1.0.0",
     "Microsoft.Extensions.Configuration.Json": "1.0.0",
     "Microsoft.Extensions.Logging.Console": "1.0.0",
     "Microsoft.Extensions.Logging.Debug": "1.0.0",
+    "Microsoft.Extensions.Logging.TraceSource": "1.0.0",
     "Microsoft.Extensions.Options.ConfigurationExtensions": "1.0.0",
-    "Newtonsoft.Json": "9.0.1"
+    "Microsoft.Tpl.Dataflow": "4.5.24",
+    "Newtonsoft.Json": "9.0.1",
+    "Odachi.AspNetCore.Authentication.Basic": "1.0.0"
```
AlexanderSher: What is this package for?

int19h: HTTP Basic auth - Kestrel doesn't come with its own.

src/Host/Broker/Impl/Lifetime/LifetimeManager.cs
```diff
@@ -0,0 +1,61 @@
...
+                process.Exited += delegate {
+                    _logger.LogInformation($"Parent process {pid} exited, shutting down.");
+                    Program.Exit();
+                };
+            }
+
+            Ping();
+        }
+
+        public void Ping() {
```
AlexanderSher: So ping should be called before `_options.PingTimeout.Value` passed, otherwise process will exit, correct?

int19h: Yep - but only if the timeout value was actually specified (currently the local connector doesn't use it, and neither connector actually sends pings).

Not sure if we will actually need it for lifetime purposes at all. For local scenario, having broker monitor VS process proved to be just as reliable, and doesn't pollute the logs with constant pings. Ping itself is still useful as a quick way to verify if server is up or not, though.

AlexanderSher: Don't you want to use `Threading.Timer` instead? IMO it will be simpler

```{csharp}
if (_options.PingTimeout != null) {
    _timer = new System.Threading.Timer(o => {
                _timer = null;
                Program.Exit();
            }, null, _options.PingTimeout.Value, System.Threading.Timeout.Infinite);
}
```

```{csharp}
public void Ping() {
    if (_options.PingTimeout != null) {
        _timer?.Change(_options.PingTimeout.Value, 0);
    }
}
```

---
In reply to: [74829720](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74829720) [](ancestors = 74829720)

src/Host/Broker/Impl/Logging/FileLogger.cs
```diff
@@ -0,0 +1,55 @@
...
+        public FileLogger(string category, StreamWriter writer) {
+            _category = category;
+            _writer = writer;
+        }
+
+        public void Dispose() {
+            _writer = null;
+        }
+
+        private sealed class Scope : IDisposable {
```
AlexanderSher: Do you need this class?

int19h: Yes, for BeginScope. We can do something more useful here later if we want to - I haven't used scopes in our own logging so far, but that will likely change.

---
In reply to: [74831777](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74831777) [](ancestors = 74831777)

src/Common/Core/Impl/Telemetry/TelemetryServiceBase.cs
int19h: I'm not entirely sure about the cause, but I was hitting this in the tests sporadically. Doesn't seem to be related to the changes.

---
In reply to: [74828037](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74828037) [](ancestors = 74828037)

src/Host/Broker/Impl/project.json
int19h: HTTP Basic auth - Kestrel doesn't come with its own. 

---
In reply to: [74828213](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74828213) [](ancestors = 74828213)

src/Host/Broker/Impl/Lifetime/LifetimeManager.cs
int19h: I don't really have a strong preference either way. But since this code may well be gone completely, it's probably not worth polishing.

---
In reply to: [74831517](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74831517) [](ancestors = 74831517,74829720)

src/Common/Core/Impl/Telemetry/TelemetryServiceBase.cs
AlexanderSher: There are `TestTelemetryRecorder` and `FileTelemetryRecorder` that should be created for the tests. Do you remember what tests were failing?

---
In reply to: [74832126](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74832126) [](ancestors = 74832126,74828037)

int19h: Not anymore, sorry. But undoing the change and doing a few full tests runs should flush them out.

---
In reply to: [74832919](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74832919) [](ancestors = 74832919,74832126,74828037)

src/Host/Broker/Impl/Logging/FileLogger.cs
AlexanderSher: I'd rather return `Disposable.Empty` from `BeginScope` then - it will be easier to understand that it isn't doing anything right now.

---
In reply to: [74832075](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74832075) [](ancestors = 74832075,74831777)

int19h: Ah, I didn't know we had such a thing. Yep, I'll switch to it.

---
In reply to: [74833266](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74833266) [](ancestors = 74833266,74832075,74831777)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
```diff
@@ -0,0 +1,173 @@
...
+        private readonly BufferBlock<byte[]> _hostMessages = new BufferBlock<byte[]>();
+        private readonly BufferBlock<byte[]> _clientMessages = new BufferBlock<byte[]>();
+
+        private ConcurrentDictionary<ulong, byte[]> _sentPendingRequests = new ConcurrentDictionary<ulong, byte[]>();
+        private Queue<byte[]> _unsentPendingRequests = new Queue<byte[]>();
+
+        private byte[] _handshake;
+        private IMessagePipeEnd _hostEnd, _clientEnd;
+
+        private abstract class PipeEnd : IMessagePipeEnd {
```
AlexanderSher: Any reason to make this class private?

int19h: Mainly so that only pipe can create them, since they don't really make sense separate from the pipe. Keeping it all private allows to enforce the invariant that each pipe has exactly two ends, and each pipe end has exactly one owner pipe.

---
In reply to: [74833429](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74833429) [](ancestors = 74833429)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+using Microsoft.Extensions.Logging;
+using Microsoft.R.Host.Client;
+
+namespace Microsoft.R.Host.Broker.Pipes {
+    public class MessagePipe {
+        private readonly ILogger _logger;
+        private readonly BufferBlock<byte[]> _hostMessages = new BufferBlock<byte[]>();
```
int19h: Side note: this is likely to be the bottleneck, since all messages between host and VS flow through here. If there's some implementation of an async concurrent pipe or pipe-like structure that has better perf than BufferBlock, or if we can hand-roll such a thing, it's probably worth investigating.

src/Host/Broker/Impl/Microsoft.R.Host.Broker.csproj
```diff
@@ -42,16 +75,38 @@
     <Compile Include="..\..\..\GlobalAssemblyInfo.cs">
       <Link>Properties\GlobalAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="..\..\Client\Impl\Definitions\Message.cs">
+      <Link>Protocol\Message.cs</Link>
```
int19h: Referencing it like this is a hack until we can untangle dependencies. We'll probably need a new shared assembly that contains just the protocol definitions - specifically, Message from Host.Client, and all the *Info structs from Host.Broker. That can then be referenced by both Host.Client and Host.Broker.

src/Host/Broker/Impl/Pipes/MessagePipe.cs
AlexanderSher: So `HostEnd` is created only once, and its lifetime is similar to MessagePipe itself, while `ClientEnd` is created every time there is a new request. Correct?






---
In reply to: [74833662](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74833662) [](ancestors = 74833662,74833429)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
AlexanderSher: Probably we should add it as a comment :)

---
In reply to: [74833910](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74833910) [](ancestors = 74833910)

src/Common/Core/Impl/Telemetry/TelemetryServiceBase.cs
AlexanderSher: Ok, let's create a separate task for this. I want to find out, probably another singleton issue.

---
In reply to: [74833030](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74833030) [](ancestors = 74833030,74832919,74832126,74828037)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
int19h: Exactly. I'll add a comment.

---
In reply to: [74837833](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74837833) [](ancestors = 74837833,74833662,74833429)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
int19h: Sure.

---
In reply to: [74837927](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74837927) [](ancestors = 74837927,74833910)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
AlexanderSher: In this case it looks like `HostEnd` and `MessagePipe` should be combined into one - `HostEnd` is just a wrapper around `MessagePipe` methods. Also, I'd create `IMessagePipeClientEnd : IMessagePipeEnd, IDisposable` to remove 
```{csharp}
            public override void Dispose() {
                throw new InvalidOperationException("Host end of the pipe should not be disposed.");
            }
```

---
In reply to: [74838114](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74838114) [](ancestors = 74838114,74837833,74833662,74833429)

src/Common/Core/Impl/Telemetry/TelemetryServiceBase.cs
int19h: [https://github.com/Microsoft/RTVS/issues/2174](https://github.com/Microsoft/RTVS/issues/2174)

---
In reply to: [74838090](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74838090) [](ancestors = 74838090,74833030,74832919,74832126,74828037)

src/Host/Broker/Impl/Pipes/WebSocketPipeAction.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            using (var pipe = _session.ConnectClient()) {
+                Task wsToPipe = WebSocketToPipeWorker(socket, pipe, context.RequestAborted);
+                Task pipeToWs = PipeToWebSocketWorker(socket, pipe, context.RequestAborted);
+                await Task.WhenAll(wsToPipe, pipeToWs);
+            }
+        }
+
+        private static async Task WebSocketToPipeWorker(WebSocket socket, IMessagePipeEnd pipe, CancellationToken ct) {
+            var cancellationToken = new CancellationToken();
```
AlexanderSher: it looks like `ct` should be used in `ReceiveAsync` instead of this one

src/Host/Broker/Impl/Pipes/WebSocketPipeAction.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            const int blockSize = 0x10000;
+            var buffer = new MemoryStream(blockSize);
+
+            while (true) {
```
AlexanderSher: Do we need to check for `ct` here as well?

src/Host/Broker/Impl/Pipes/MessagePipe.cs
int19h: Makes sense. How about IOwnedMessagePipeEnd, since the only thing that's different between these two is transfer of ownership?

---
In reply to: [74839003](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74839003) [](ancestors = 74839003,74838114,74837833,74833662,74833429)

int19h: ... on the other hand, if pipe implements IMessagePipeEnd directly, access to it may be gained directly, and multiple hosts might be attached to the same pipe by mistake. Right now they all have to go through ConnectHost, which enforces a single end (obviously someone can call the method once and then share the returned value, but I think it's less likely to happen accidentally).

---
In reply to: [74839548](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74839548) [](ancestors = 74839548,74839003,74838114,74837833,74833662,74833429)

src/Host/Broker/Impl/Pipes/WebSocketPipeAction.cs
```diff
@@ -0,0 +1,71 @@
...
+                    pipe.Write(buffer.ToArray());
+                    buffer.SetLength(0);
+                }
+            }
+        }
+
+        private static async Task PipeToWebSocketWorker(WebSocket socket, IMessagePipeEnd pipe, CancellationToken ct) {
+            var cancellationToken = new CancellationToken();
+
+            while (true) {
```
AlexanderSher: Do we want to check for `ct`?

src/Host/Broker/Impl/Pipes/WebSocketPipeAction.cs
int19h: Yep, this is left over from when there was no token passed in.

---
In reply to: [74839294](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74839294) [](ancestors = 74839294)

src/Host/Broker/Impl/Pipes/WebSocketPipeAction.cs
int19h: Not strictly necessary since ReceiveAsync will do so, but it won't hurt.

---
In reply to: [74839492](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74839492) [](ancestors = 74839492)

int19h:
Associated R-Host change is https://github.com/Microsoft/R-Host/pull/79

int19h:
The following issues will be resolved by this PR:

https://github.com/Microsoft/RTVS/issues/2081
https://github.com/Microsoft/RTVS/issues/2039
https://github.com/Microsoft/RTVS/issues/2038
https://github.com/Microsoft/RTVS/issues/2040

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -474,6 +473,16 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
         private async Task RunWorker(CancellationToken ct) {
             TaskUtilities.AssertIsOnBackgroundThread();
 
+            // Spin until the worker task is registered.
+            while (_runTask == null) {
+                await Task.Yield();
```
AlexanderSher: Looks like a hack.
Probably we replace these 3 tasks with 3 `TaskCompletionSourceEx` instances and cancel them in `finally` block

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.R.Host.Broker.Logging;
+using Microsoft.R.Host.Broker.Pipes;
+
+namespace Microsoft.R.Host.Broker.Sessions {
+    public class SessionManager {
+        private readonly InterpreterManager _interpManager;
+        private readonly LoggingOptions _loggingOptions;
+        private readonly ILogger _hostOutputLogger, _messageLogger;
+
+        private readonly Dictionary<string, List<Session>> _sessions = new Dictionary<string, List<Session>>();
```
AlexanderSher: How sessions are removed from this dictionary?

src/Host/Client/Impl/Host/LocalRHostConnector.cs
```diff
@@ -31,134 +32,149 @@ internal sealed class LocalRHostConnector : IRHostConnector {
...
+                throw new ObjectDisposedException(typeof(LocalRHostConnector).FullName);
             }
 
-            psi.Arguments += Invariant($" --rhost-connect ws://127.0.0.1:{server.Port}");
+            await TaskUtilities.SwitchToBackgroundThread();
 
-            if (!ShowConsole) {
-                psi.CreateNoWindow = true;
+            try {
+                await _connectSemaphore.WaitAsync();
```
AlexanderSher: >_connectSemaphore [](start = 22, length = 17)

You can use `BinaryAsyncLock` here

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -0,0 +1,117 @@
...
+            _broker.DefaultRequestHeaders.Accept.Clear();
+            _broker.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        }
+
+        protected abstract HttpClientHandler GetHttpClientHandler();
+
+        protected abstract void ConfigureWebSocketRequest(HttpWebRequest request);
+
+        public virtual void Dispose() {
+            IsDisposed = true;
```
AlexanderSher: There are 2 classes that can be used here - `DisposeToken` and `DisposableBag`. Both are thread-safe.

src/Host/Client/Impl/Host/UICredentials.cs
```diff
@@ -0,0 +1,26 @@
...
+            var userName = new StringBuilder(NativeMethods.CREDUI_MAX_USERNAME_LENGTH + 1);
+            var password = new StringBuilder(NativeMethods.CREDUI_MAX_PASSWORD_LENGTH + 1);
+
+            // TODO: figure out how to actually make login info persist. CREDUI_FLAGS_EXPECT_CONFIRMATION and CredUIConfirmCredentials?
+            bool save = false;
+            int err = NativeMethods.CredUIPromptForCredentials(
+                IntPtr.Zero, uri.Host.ToString(), IntPtr.Zero, 0, userName, userName.Capacity, password, password.Capacity, ref save,
+                NativeMethods.CREDUI_FLAGS_EXCLUDE_CERTIFICATES/* | NativeMethods.CREDUI_FLAGS_DO_NOT_PERSIST*/);
+            if (err != 0) {
+                throw new UnauthorizedAccessException("No credentials entered.");
```
AlexanderSher: Do we handle this exception?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -300,7 +314,14 @@ internal sealed class RSession : IRSession, IRCallbacks {
             using (var evaluation = await evaluationSource.Task) {
                 // Load RTVS R package before doing anything in R since the calls
                 // below calls may depend on functions exposed from the RTVS package
-                await LoadRtvsPackage(evaluation);
+                string libPath;
+                if (BrokerConnector.BrokerUri.Scheme == "file") {
+                    libPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().GetAssemblyPath());
```
AlexanderSher: Will we change this as part of #2175?

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            rCommandLineArguments = rCommandLineArguments ?? string.Empty;
+
+            var request = new SessionCreateRequest {
+                InterpreterId = _interpreterId,
+                CommandLineArguments = rCommandLineArguments,
+            };
+            var requestContent = new StringContent(JsonConvert.SerializeObject(request), Encoding.UTF8, "application/json");
+
+            (await _broker.PutAsync($"/sessions/{name}", requestContent, cancellationToken)).EnsureSuccessStatusCode();
```
AlexanderSher: >(await _broker.PutAsync($"/sessions/{name}", requestContent, cancellationToken)).EnsureSuccessStatusCode(); [](start = 12, length = 107)

Maybe we should extract this call into private method? We probably should wrap `HttpRequestException`.

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Security;
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Client.Host {
+    internal sealed class RemoteRHostConnector : RHostConnector {
+        public RemoteRHostConnector(Uri brokerUri)
```
AlexanderSher: Can't find where this ctor is called

src/Host/Client/Impl/Transports/WebSocketMessageTransport.cs
```diff
@@ -2,52 +2,62 @@
...
-            _socket.OnMessage += (sender, args) => OnMessage(args);
-            _socket.OnClose += (sender, args) => OnClose(args);
         }
 
         public async Task<Message> ReceiveAsync(CancellationToken ct = default(CancellationToken)) {
-            return await await _incomingMessages.ReceiveAsync(ct);
+            const int blockSize = 0x10000;
+            var buffer = new MemoryStream(blockSize);
+
+            while (true) {
```
AlexanderSher: let's check for `ct` here as well

src/Host/Client/Impl/Transports/WebSocketMessageTransport.cs
```diff
@@ -2,52 +2,62 @@
...
+                } catch (SocketException ex) {
+                    throw new MessageTransportException(ex);
+                } catch (WebSocketException ex) {
+                    throw new MessageTransportException(ex);
+                }
+
+                buffer.SetLength(index + wsrr.Count);
+
+                if (wsrr.CloseStatus != null) {
+                    throw new OperationCanceledException("Connection closed by host.");
```
AlexanderSher: `RHostDisconnectedException`?

src/Host/Client/Test/DataInspection/ValuesTest.cs
```diff
@@ -31,13 +31,14 @@ public class ValuesTest : IAsyncLifetime {
         private const REvaluationResultProperties AllFields = unchecked((REvaluationResultProperties)~0);
 
         private readonly MethodInfo _testMethod;
+        private readonly IRHostBrokerConnector _brokerConnector = new RHostBrokerConnector(name: nameof(ValuesTest));
```
AlexanderSher: We already have `ctor` in this test, so lets initialize this field there as well

src/Host/Client/Test/ExecutionTracing/BreakpointsTest.cs
```diff
@@ -39,6 +40,7 @@ public class BreakpointsTest : IAsyncLifetime {
         public async Task DisposeAsync() {
             await _session.StopHostAsync();
             _sessionProvider.Dispose();
+            _brokerConnector?.Dispose();
```
AlexanderSher: `_brokerConnector` can't be null

src/Host/Client/Test/ExecutionTracing/DebugReplTest.cs
```diff
@@ -39,6 +40,7 @@ public class DebugReplTest : IAsyncLifetime {
         public async Task DisposeAsync() {
             await _session.StopHostAsync();
             _sessionProvider.Dispose();
+            _brokerConnector?.Dispose();
```
AlexanderSher: `_brokerConnector` can't be null

src/Host/Protocol/Impl/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+﻿using System.Runtime.CompilerServices;
+
+#if SIGN
+[assembly: InternalsVisibleTo("Microsoft.R.Host.Protocol.Test, PublicKey=002400000480000094000000060200000024000052534131000400000100010007D1FA57C4AED9F0A32E84AA0FAEFD0DE9E8FD6AEC8F87FB03766C834C99921EB23BE79AD9D5DCC1DD9AD236132102900B723CF980957FC4E177108FC607774F29E8320E92EA05ECE4E821C0A5EFE8F1645C4C0C93C1AB99285D622CAA652C1DFAD63D745D6F2DE5F17E5EAF0FC4963D261C8A12436518206DC093344D5AD293")]
+#else
+[assembly: InternalsVisibleTo("Microsoft.R.Host.Protocol.Test")]
```
AlexanderSher: Do we have such project?

src/Package/Test/FakeFactories/TestRInteractiveWorkflowProviderFactory.cs
```diff
@@ -44,7 +45,7 @@ public static class TestRInteractiveWorkflowProviderFactory {
             shell = shell ?? VsAppShell.Current;
             settings = settings ?? RToolsSettings.Current;
 
-           return new TestRInteractiveWorkflowProvider(sessionProvider, connectionsProvider, historyProvider, packagesProvider, plotsProvider, activeTextViewTracker, debuggerModeTracker, brokerConnector, shell, settings);
+            return new TestRInteractiveWorkflowProvider(sessionProvider, connectionsProvider, historyProvider, packagesProvider, plotsProvider, activeTextViewTracker, debuggerModeTracker, brokerConnector, shell, settings) { TestName = testName };
```
AlexanderSher: Where do you need this?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -50,14 +52,17 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
             , IRSettings settings
             , Action onDispose) {
 
+            _name = name;
```
AlexanderSher: This field isn't used anywhere

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
```diff
@@ -49,14 +49,21 @@ internal class VsRInteractiveWorkflowProvider : IRInteractiveWorkflowProvider {
             _shell = shell;
         }
 
+        public void Dispose() {
+            if (_instanceLazy?.IsValueCreated == true) {
```
AlexanderSher: `Dispose` can be called from any thread, so this is not enough. We know that if `_instanceLazy` isn't null, `_instanceLazy.Value` will be requested, so there is no reason to verify `IsValueCreated`. On the other hand, `GetOrCreate` should fail if `VsRInteractiveWorkflowProvider` is disposed.

src/Host/Client/Impl/Host/RHost.cs
int19h: I'm starting to think that the entire existing arrangement is a hack, and we should just use a semaphore here to ensure that there's only one worker task... but this at least fixes the issue that was there. FWIW, this isn't directly related to other changes, it's a race that we had before that I've ran into while doing multiple verification test runs.

Can you file a bug detailing what you think is the best approach to rewrite this part? And we can do a separate PR for that then.

---
In reply to: [74950109](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74950109) [](ancestors = 74950109)

src/Host/Broker/Impl/Sessions/SessionManager.cs
int19h: They aren't, since there's no session delete functionality here yet - that one is TBD.

---
In reply to: [74966786](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74966786) [](ancestors = 74966786)

src/Host/Client/Impl/Host/RHostConnector.cs
int19h: I was actually thinking of creating a class that wraps HttpClient, and exposes various endpoints as strongly typed methods, basically mirroring similar methods in the controller on the other side. 

What is the most appropriate exception type for wrapping - RHostDisconnectedException?

---
In reply to: [74971357](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74971357) [](ancestors = 74971357)

src/Host/Client/Impl/Session/RSession.cs
int19h: Yes, although how exactly depends on the option that we choose to take.

---
In reply to: [74969930](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74969930) [](ancestors = 74969930)

src/Host/Client/Impl/Host/UICredentials.cs
int19h: It gets wrapped into a WebException, which is in turn wrapped into HttpRequestException that gets thrown out of HttpClient. Which we don't handle yet (per the other comment).

I'm thinking that this one is actually wrong, though. Error here normally indicates that user closed the dialog box without clicking OK, so it should really be OCE, and we should propagate it all the way out of the connector as such.

---
In reply to: [74969361](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74969361) [](ancestors = 74969361)

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
int19h: In the constructor of RHostBrokerConnector. 

---
In reply to: [74971528](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74971528) [](ancestors = 74971528)

src/Host/Client/Impl/Transports/WebSocketMessageTransport.cs
int19h: That's the one that RHost throws to its clients, but it doesn't expect to see it from the transport. Should probably be MessageTransportException.

---
In reply to: [74971825](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74971825) [](ancestors = 74971825)

src/Host/Client/Test/DataInspection/ValuesTest.cs
int19h: You mean InitializeAsync, or the actual constructor? If the latter, why is that preferable?

---
In reply to: [74972244](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74972244) [](ancestors = 74972244)

src/Host/Protocol/Impl/Properties/AssemblyInfo.cs
int19h: Not yet :) for that matter, we don't have any internals here yet, either.

---
In reply to: [74973032](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74973032) [](ancestors = 74973032)

src/Package/Test/FakeFactories/TestRInteractiveWorkflowProviderFactory.cs
int19h: This is for cases where Create is called with null broker - there are some tests that do it, and rely on the workflow provider to create the broker itself. Hence, name needs to be passed here to make the broker log distinct.

---
In reply to: [74973702](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74973702) [](ancestors = 74973702)

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
int19h: I'll use DisposableBag instead, and add the workflow to it from CreateRInteractiveFlow. 

---
In reply to: [74974926](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74974926) [](ancestors = 74974926)

src/Host/Client/Test/DataInspection/ValuesTest.cs
AlexanderSher: In constructor. 2 reasons: better stack trace in case of exception and explicit order of instantiation

---
In reply to: [75011847](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75011847) [](ancestors = 75011847,74972244)

src/Host/Client/Impl/Host/RHostConnector.cs
AlexanderSher: I think yes. Unless there should be some special exception type in this case.

---
In reply to: [75007386](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75007386) [](ancestors = 75007386,74971357)

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Ok

---
In reply to: [75007685](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75007685) [](ancestors = 75007685,74969930)

src/Host/Client/Test/DataInspection/ValuesTest.cs
int19h: Order is actually explicit for instance fields as well (it's always textual order, since they cannot depend on each other). I didn't know about the stack trace, though. Will do.

---
In reply to: [75027791](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75027791) [](ancestors = 75027791,75011847,74972244)

AlexanderSher: > Order is actually explicit for instance fields as well

Yes, compiler defines it explicitly, but in case of derived classes I have to think about it. If we have `A : B` and `B : C`, it will first initialize fields for `A`, then for `B`, then for `C`, then it will call constructor for `C`, then for `B`, then for `A`. I just want the order to be simpler :)

---
In reply to: [75028449](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75028449) [](ancestors = 75028449,75027791,75011847,74972244)

src/Package/Test/FakeFactories/TestRInteractiveWorkflowProviderFactory.cs
AlexanderSher: Can we call it `BrokerConnectorName`?

---
In reply to: [75012197](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75012197) [](ancestors = 75012197,74973702)

src/Host/Client/Impl/Host/LocalRHostConnector.cs
```diff
@@ -31,134 +33,150 @@ internal sealed class LocalRHostConnector : IRHostConnector {
...
-            psi.EnvironmentVariables["PATH"] = rBinPath + ";" + Environment.GetEnvironmentVariable("PATH");
+        protected override async Task ConnectToBrokerAsync() {
+            DisposableBag.ThrowIfDisposed();
+            await TaskUtilities.SwitchToBackgroundThread();
 
-            if (name != null) {
-                psi.Arguments += " --rhost-name " + name;
+            try {
+                if (await _connectLock.WaitAsync()) {
+                    if (!_isConnected) {
```
AlexanderSher: You don't need `_isConnected` anymore

src/Host/Client/Impl/Host/LocalRHostConnector.cs
```diff
@@ -31,134 +33,150 @@ internal sealed class LocalRHostConnector : IRHostConnector {
...
-            string rBinPath = Path.Combine(_rHome, RBinPathX64);
+            _name = name;
+            _rhostDirectory = rhostDirectory ?? Path.GetDirectoryName(typeof(RHost).Assembly.GetAssemblyPath());
+            _rHome = rHome;
+        }
 
-            if (!File.Exists(rhostExe)) {
-                throw new RHostBinaryMissingException();
+        public override void Dispose() {
+            if (!DisposableBag.TryMarkDisposed()) {
```
AlexanderSher: If you have a `DisposableBag`, you don't have to override `Dispose` method. You can create create method `KillBrokerProcess` and then call `DisposableBag.Add(KillBrokerProcess)` in constructor or in `ConnectToBrokerWorker` after `_brokerProcess` field is assigned.

src/Package/Test/FakeFactories/TestRInteractiveWorkflowProviderFactory.cs
int19h: Sure

---
In reply to: [75031291](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75031291) [](ancestors = 75031291,75012197,74973702)

src/Host/Client/Impl/Host/LocalRHostConnector.cs
int19h: It's used to detect when broker process goes away - see the delegate assigned to _process.Exited in ConnectToBrokerWorker

---
In reply to: [75031624](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75031624) [](ancestors = 75031624)

int19h: Although I guess I could use _brokerProcess for this instead.

---
In reply to: [75033083](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75033083) [](ancestors = 75033083,75031624)

AlexanderSher: Sorry, I meant that you can use `_connectLock.IsCompleted` instead. And in current place it isn't needed, because `WaitAsync` will return `true` if `Release` was called and `false` otherwise. So here 
```{csharp}
                if (!(await _connectLock.WaitAsync())) {
```
is enough

---
In reply to: [75033083](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75033083) [](ancestors = 75033083,75031624)

int19h: Side note: this is basically best effort recovery in case broker crashes, or user kills it from Task Manager etc. There are going to be unavoidable races here, because if the process crashes, Exited event is delivered asynchronously, so in the worst case, we won't know until we send the HTTP request. I contemplated adding a loop around that whole thing so that if request fails, it checks if process is still alive and tries to restart it, but I'm not sure if it's a good idea - if the broker does crash repeatedly, we don't want to keep restarting it over and over again. 

---
In reply to: [75033980](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75033980) [](ancestors = 75033980,75033083,75031624)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
int19h: So what's the conclusion here? Do I leave it as is, and just add IOwnedMessagePipeEnd for client end?

---
In reply to: [74839770](https://github.com/Microsoft/RTVS/pull/2173#discussion_r74839770) [](ancestors = 74839770,74839548,74839003,74838114,74837833,74833662,74833429)

src/Host/Client/Impl/Host/LocalRHostConnector.cs
```diff
@@ -31,135 +33,137 @@ internal sealed class LocalRHostConnector : IRHostConnector {
...
+                    } catch (Exception) {
+                    } finally {
+                        process?.Dispose();
+                    }
+                }
             }
         }
+
+        private void DisposeBrokerProcess() {
+            _brokerProcess?.Kill();
```
AlexanderSher: Do you need try/catch here?

int19h: Yep.

---
In reply to: [75045542](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75045542) [](ancestors = 75045542)

src/Host/Broker/Impl/Pipes/MessagePipe.cs
AlexanderSher: Let's add `IOwnedMessagePipeEnd`, and also I'd like MessagePipe to control if end is created or not. That would be more straightforward and you won't need `PipeEnd` base class.

---
In reply to: [75035019](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75035019) [](ancestors = 75035019,74839770,74839548,74839003,74838114,74837833,74833662,74833429)

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: Ok

---
In reply to: [75006462](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75006462) [](ancestors = 75006462,74950109)

src/Host/Client/Impl/Host/LocalRHostConnector.cs
AlexanderSher: Maybe we can have some limited amount of attempts?

---
In reply to: [75034380](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75034380) [](ancestors = 75034380,75033980,75033083,75031624)

int19h: Let's see how frequently this is an issue first. One attempt approach solves the "I killed broker from Task Manager" problem, but will make repeating broker crashes more visible, which is probably what we want at least while we're testing it. 

---
In reply to: [75047226](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75047226) [](ancestors = 75047226,75034380,75033980,75033083,75031624)

AlexanderSher: Ok

---
In reply to: [75047979](https://github.com/Microsoft/RTVS/pull/2173#discussion_r75047979) [](ancestors = 75047979,75047226,75034380,75033980,75033083,75031624)

AlexanderSher:
:shipit:

PR #2178, opened 2016-08-17
Broker session management improvements

int19h:
Don't register session until it has been started.

When handling pipe request, check for non-existent sessions or those for which the process is not running, and report error accordingly.

Add host process PID when logging host messages and stderr in broker.

Enable stderr and packet logging by default for local connections.

### DIFF
--- a/src/Host/Broker/Impl/Sessions/SessionManager.cs
+++ b/src/Host/Broker/Impl/Sessions/SessionManager.cs
@@ -46,8 +46,8 @@ namespace Microsoft.R.Host.Broker.Sessions {
         public Session CreateSession(IIdentity user, string id, Interpreter interpreter, SecureString password, string commandLineArguments) {
             Session session;
 
+            List<Session> userSessions;
             lock (_sessions) {
-                List<Session> userSessions;
                 _sessions.TryGetValue(user.Name, out userSessions);
                 if (userSessions == null) {
                     _sessions[user.Name] = userSessions = new List<Session>();
@@ -60,13 +60,17 @@ namespace Microsoft.R.Host.Broker.Sessions {
                 }
 
                 session = new Session(this, user, id, interpreter, commandLineArguments);
-                userSessions.Add(session);
             }
 
             session.StartHost(
                 password,
                 _loggingOptions.LogHostOutput ? _hostOutputLogger : null,
                 _loggingOptions.LogPackets ? _messageLogger : null);
+
+            lock (_sessions) {
+                userSessions.Add(session);
+            }
+
             return session;
         }
     }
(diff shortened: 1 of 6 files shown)


src/Host/Client/Impl/Host/LocalRHostConnector.cs
```diff
@@ -93,6 +93,8 @@ public LocalRHostConnector(string name, string rHome, string rhostDirectory = nu
                         FileName = rhostBrokerExe,
                         UseShellExecute = false,
                         Arguments =
+                            $" --logging:logHostOutput true" +
```
AlexanderSher: Maybe we can make it configurable?

int19h: Logging host output always is a good idea for local broker, because it's the only way to figure out what went wrong with the host if it crashed/terminated before it had a chance to write its own log.

Logging packets can be made configurable, but given that it's not configurable for our Host and Host.Client logs right now, it's probably not worth it unless those become configurable also.

For the time being, I think we'll want logging anyway to help diagnose issues. We can revisit later to decide how (if at all) this needs to be dialed down, and whether any of this needs to be user-configured.

int19h:
Note that with this change: https://github.com/Microsoft/R-Host/pull/81, the host no longer prints all packets to stderr by default. So with these two changes combined, one flag is for logging packets only, and other is for logging other host output (such as error messages) only.

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -60,13 +60,17 @@ public class SessionManager {
...
-                userSessions.Add(session);
             }
 
             session.StartHost(
                 password,
                 _loggingOptions.LogHostOutput ? _hostOutputLogger : null,
                 _loggingOptions.LogPackets ? _messageLogger : null);
+
+            lock (_sessions) {
+                userSessions.Add(session);
```
AlexanderSher: `session.StartHost` is a long operation. Is it possible that there will be concurrent access to the `userSessions`?

src/Host/Client/Impl/Host/LocalRHostConnector.cs
AlexanderSher: Ok

---
In reply to: [75219094](https://github.com/Microsoft/RTVS/pull/2178#discussion_r75219094) [](ancestors = 75219094)

src/Host/Broker/Impl/Sessions/SessionManager.cs
int19h: Good point. I haven't considered this. I think I'll revert it, and we'll just say that if you expect the session to be running, you need to wait until CreateSession completes. For broker clients, this means that you wait until PUT request returns a response before doing GET for a pipe. If you do a GET before PUT completes, the error handling code I've added there should still give reasonable error reporting in that case.

(This change was due to an issue Karthik observed, but it had a different root cause; I just figured I'd patch this one up as well.)

---
In reply to: [75219393](https://github.com/Microsoft/RTVS/pull/2178#discussion_r75219393) [](ancestors = 75219393)

AlexanderSher:
:shipit:

PR #2179, opened 2016-08-18
Fix broker authentication handling:

int19h:
- Validate credentials, and store them in credential manager if valid.
- Reuse stored credentials if possible.
- Allow user to retry credential input if not authorized, or cancel the operation.

Remove web service interfaces from Host.Protocol, and controllers in Host.Broker don't implement them anymore. They are for internal use within Host.Client now.


int19h:
With respect to web service interface removal, this was a pattern that I wanted to try, but it doesn't work out because controllers may want to return lower-level action results, or accept extra `[FromServices]` parameters, all of which is internal and not part of the REST API surface. So in the most general case, it's not possible to have the controller implement the same interface as the `HttpClient` wrapper on the other side.

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -28,38 +30,68 @@ internal abstract class RHostConnector : IRHostConnector {
...
+        /// <summary>
+        /// Called whenever HTTP authentication succeeds.
+        /// </summary>
+        protected abstract void OnAuthenticationSucceeded();
+
+        /// <summary>
+        /// Called whenever HTTP authentication fails (i.e. server returns HTTP 403).
+        /// </summary>
+        /// <returns>
+        /// <see langword="true"/> if <see cref="Credentials"/> were updated, and HTTP request should be retried.
```
int19h: The comment is misleading now, since `OnAuthenticationFailed` doesn't actually update credentials. I'll replace this with a simpler arrangement and adjust the comment.

int19h:
Ping - we need this for the bug bash tomorrow.

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -83,35 +107,62 @@ internal abstract class RHostConnector : IRHostConnector {
...
-                // Unwrap all the extraneous layers, and rethrow the cancellation.
-                var oce = ex?.InnerException?.InnerException as OperationCanceledException;
-                if (oce != null) {
-                    throw oce;
+            var sessions = new SessionsWebService(HttpClient);
+
+            while (true) {
+                bool isValidCredentials = true;
+                try {
+                    UpdateCredentials();
```
AlexanderSher: So if `UpdateCredentials` fails, we will still call `OnCredentialsValidated` with `isValidCredentials == true`?

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
```diff
@@ -3,32 +3,68 @@
...
+            var passwordBuilder = new StringBuilder(CREDUI_MAX_PASSWORD_LENGTH + 1);
+
+            bool save = false;
+
+            int flags = CREDUI_FLAGS_EXCLUDE_CERTIFICATES | CREDUI_FLAGS_PERSIST | CREDUI_FLAGS_EXPECT_CONFIRMATION | CREDUI_FLAGS_GENERIC_CREDENTIALS;
+            if (_ignoreSavedCredentials) {
+                flags |= CREDUI_FLAGS_ALWAYS_SHOW_UI;
+            }
+
+            int err = CredUIPromptForCredentials(IntPtr.Zero, _authority, IntPtr.Zero, 0, userNameBuilder, userNameBuilder.Capacity, passwordBuilder, passwordBuilder.Capacity, ref save, flags);
```
AlexanderSher: Not part of this review, but do we want to show UI on such a low level? I'd rather think that it is a responsibility of Connections Manager to show this UI.

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
```diff
@@ -3,32 +3,68 @@
...
 
-        protected override void ConfigureWebSocketRequest(HttpWebRequest request) {
-            request.AuthenticationLevel = AuthenticationLevel.MutualAuthRequested;
-            request.Credentials = CredentialCache.DefaultNetworkCredentials;
+        protected override void UpdateCredentials() {
+            string userName, password;
+            GetCredentials(out userName, out password);
+
+            _credentials.UserName = userName;
+            _credentials.Password = password;
```
AlexanderSher: Is it correct that we can change `UserName` and `Password` more than once?

src/Host/Client/Impl/Host/RHostConnector.cs
int19h: Good catch. It's harmless with the current implementation, but it's not what I had in mind.

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
int19h: I considered this, but either way it will have to be tied to HTTP connection handling in a similar manner, since every HTTP request can potentially fail to authorize, and then we need to prompt for credentials. We might want to define some kind of interface for credential handling, similar to `UpdateCredentials` and `OnCredentialsValidated`, and have the remote connector use that.

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
int19h: It's intentional - if they were incorrect, this will change them to the new user input after asking the user.

Changing them directly on the instance of `NetworkCredential` is kinda weird, but it seems to be the only way - `HttpClientHandler.Credentials` is read-only once the first request is made, so the only way to change credentials afterwards is by mutating the instance directly here.

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
AlexanderSher: I'd like to have such interface even instead of `UpdateCredentials` and `OnCredentialsValidated`, but that isn't part of this review either.

---
In reply to: [75424500](https://github.com/Microsoft/RTVS/pull/2179#discussion_r75424500) [](ancestors = 75424500)

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
AlexanderSher: In this case, we can remove `credentials` parameter from `CreateHttpClient` and set them for `LocalRHostConnector` in `UpdateCredentials` as well

---
In reply to: [75424572](https://github.com/Microsoft/RTVS/pull/2179#discussion_r75424572) [](ancestors = 75424572)

int19h: There may be scenarios where we want them to be something other than `NetworkCredentials` (for example, if we do decide to revisit NTLM auth), so I figured I'd let the derived classes handle the creation of the credential instance. Passing them in via argument ensures that this property is set on `HttpClientHandler` before requests begin.

src/Host/Client/Impl/Host/RemoteRHostConnector.cs
int19h: Keep in mind that some connectors (such as local) treat authentication as implementation detail - it's set up for the sake of security, but it's something that is never meant to be configured or exposed outside of the connector. Any future refactoring must take that into consideration.

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -83,35 +107,69 @@ internal abstract class RHostConnector : IRHostConnector {
...
+                    isValidCredentials = true;
+
+                    await sessions.PutAsync(name, new SessionCreateRequest {
+                        InterpreterId = _interpreterId,
+                        CommandLineArguments = rCommandLineArguments,
+                    });
+                    break;
+                } catch (UnauthorizedAccessException) {
+                    isValidCredentials = false;
+                    continue;
```
AlexanderSher: Do you need `continue` here?

AlexanderSher:
:shipit:

PR #2192, opened 2016-08-23
Implementation of ConnectionManager UI + several bug fixes

AlexanderSher:


src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -100,13 +106,12 @@ internal abstract class RHostConnector : IRHostConnector {
             }
         }
 
-        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000, CancellationToken cancellationToken = new CancellationToken()) {
+        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, int timeout = 3000, CancellationToken cancellationToken = new CancellationToken()) {
```
int19h: The change wrt `rCommandLineArguments` implies that every host process started from a given broker shares the same command line. While this is the way we're using it currently, it's not actually an inherent limitation on the other end - it's perfectly valid to spin up multiple hosts with different command lines - and it might come useful in the future.  So I would prefer the API that works with broker directly to not impose this limitation on its clients.

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -89,6 +71,16 @@ internal sealed class RToolsSettingsImplementation : IRToolsSettings {
             }
         }
 
+        public ConnectionInfo[] Connections {
+            get { return _connections; }
+            set {
+                using (var p = RPackage.Current.GetDialogPage(typeof(RToolsOptionsPage)) as RToolsOptionsPage) {
```
int19h: Nit: `(T)`

src/R/Components/Impl/ConnectionManager/IConnection.cs
```diff
@@ -8,8 +8,9 @@ namespace Microsoft.R.Components.ConnectionManager {
     public interface IConnection {
         Uri Id { get; }
         string Name { get; }
+        string Path { get; }
```
int19h: What's the distinction between `Id`, `Name` and `Path`? For remote connections in particular.

int19h: It looks like `Path` is always an `Uri` internally, too - any reason to not expose it on the interface?

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        private void ButtonConnect_Click(object sender, RoutedEventArgs e) {
+            Model?.ConnectAsync(GetConnection(e)).DoNotWait();
+        }
+
+        private void ButtonAdd_Click(object sender, RoutedEventArgs e) {
+            Model?.AddNew();
+        }
+
+        private static IConnectionViewModel GetConnection(RoutedEventArgs e) => ((IConnectionViewModel)((FrameworkElement)e.Source).DataContext);
```
int19h: Can this cast fail if the context is `UnsetValue` or some other funky WPF stub? Would `as` be preferable?

src/R/Components/Impl/ConnectionManager/IConnection.cs
AlexanderSher: The difference is that `Path` is what user has entered in UI. `string` -> `Uri` -> `string` conversion will change original input.

---
In reply to: [75948756](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75948756) [](ancestors = 75948756)

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+            _connectionManager = connectionManager;
+            _coreShell = coreShell;
+            _items = new BatchObservableCollection<object>();
+            Items = new ReadOnlyObservableCollection<object>(_items);
+
+            UpdateConnections();
+        }
+
+        public Task<int> Search(string searchString, CancellationToken cancellationToken) {
+            throw new System.NotImplementedException();
```
int19h: Nit: unnecessary qualification.

src/Host/Client/Impl/Host/RHostConnector.cs
AlexanderSher: Then there is a question how do we want to specify this command line args. Currently they are part of connection settings. If we want one HostConnector to be able to connect with different CommandLine parameters, how user will provide them?

---
In reply to: [75946453](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75946453) [](ancestors = 75946453)

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
AlexanderSher: DataContext should always be specified for list items. But since its WPF, I think it's better to be safe :)

---
In reply to: [75949111](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75949111) [](ancestors = 75949111)

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
AlexanderSher: Search is disabled for now. I'm not really sure if we need it at all - how many connections can one user have?

---
In reply to: [75949265](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75949265) [](ancestors = 75949265)

src/Host/Client/Impl/Host/RHostConnector.cs
int19h: The UI can stay for now. I just think the lower level API should be more flexible in that regard, given that this flexibility is inherent in the design of other components.

Thinking of possible future scenarios - some of the helper sessions might not want to apply the user-specified command line at all (in fact, I'm not sure that the ones that we currently have actually need them), while others might specify some arguments of their own.

---
In reply to: [75949601](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75949601) [](ancestors = 75949601,75946453)

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
int19h: Probably not more than a dozen, I agree. @smortaz, @jflam? 

---
In reply to: [75950043](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75950043) [](ancestors = 75950043,75949265)

int19h: Also, if this method stays, it should be SearchAsync.

---
In reply to: [75950555](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75950555) [](ancestors = 75950555,75950043,75949265)

src/Host/Client/Impl/Host/RHostConnector.cs
AlexanderSher: Good point. I'll revert this

---
In reply to: [75950345](https://github.com/Microsoft/RTVS/pull/2192#discussion_r75950345) [](ancestors = 75950345,75949601,75946453)

AlexanderSher:
Addressed CR comments

int19h:
:+1:

PR #2193, opened 2016-08-24
Fix assembly strong name issues

int19h:
Remove references to Odachi.AspNetCore.Authentication.Basic, and move the corresponding code directly into the broker assembly, to avoid strong name / no strong name mix-up issues.


int19h:
Also see https://github.com/Microsoft/RTVS-pr/pull/11

AlexanderSher:
:+1:

PR #2195, opened 2016-08-24
2 bug fixes

AlexanderSher:
- `libPath` for local and remote were accidentally swapped
- `ActiveConnection` should be compared with the new connection rather than with `BrokerConnector.BrokerUri` (they are actually the same)


int19h:
:+1:

PR #2199, opened 2016-08-24
Multiple plot windows and plot history window

huguesv:
Matching changes for https://github.com/Microsoft/R-Host/pull/83

You can now have more than one plot window.

When an 'ide' graphics device is created, it is associated with a plot window. When it's closed (`dev.off` or reset session), the plot window remains but is ready to be reused (shows a watermark).

There's a concept of 'active' plot window/device, which matches the standard `dev.cur`/`dev.set`. There is a 'checked' command that shows which one is active, and allows to activate a plot device. There's a command to create a new plot window, which internally does the same as typing `dev.new` in repl.

You can cut/copy/paste or drag & drop plots across plot windows.

There is a plot history window. It displays a thumbnail for each plot present across all plot windows/devices. The list is grouped by device. Double-click or use the Activate command to show the selected plot's window, and render that plot in the window.  You can cut/copy or drag from the plot history window to the plot windows.

Actions on the selected plot from the plot history window are limited to activating or deleting a plot. No multiple selection for now in plot history.

The auto-hide option in plot history window is a convenience for people that may want to have that window be large and floating on top of VS. When it's enabled, double-click on a plot will activate that plot window, and automatically close the plot history window. That option is not being saved in RTools settings yet.

Thumbnails in the plot history can be made larger/smaller. The thumbnail size is not being saved in RTools settings yet.


src/Host/Client/Impl/Definitions/PlotDeviceProperties.cs
```diff
@@ -0,0 +1,20 @@
...
+        public int Height { get; }
+        public int Resolution { get; }
+
+        public PlotDeviceProperties(int width, int height, int resolution) {
+            Width = width;
+            Height = height;
+            Resolution = resolution;
+        }
+
+        public static PlotDeviceProperties CreateDefault() {
```
int19h: Since this is a struct, with true copy semantics, this could be a readonly static field.

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -429,19 +429,32 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
                             case "!Plot":
                                 await _callbacks.Plot(
                                     new PlotMessage(
-                                        message.GetString(0, "xaml_file_path"),
-                                        message.GetInt32(1, "active_plot_index"),
-                                        message.GetInt32(2, "plot_count"),
+                                        message.GetString(0, "device_id").SafeParseGuid(),
```
int19h: Do we actually want to ignore GUID parsing errors here? It looks like if it happens, it's a protocol contract violation (i.e. a "should never happen" case).

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -46,31 +49,71 @@ public static class RSessionEvaluationCommands {
...
+        public static Task SelectPlotAsync(this IRExpressionEvaluator evaluation, Guid deviceId, Guid plotId) {
+            Debug.Assert(deviceId != Guid.Empty);
+            Debug.Assert(plotId != Guid.Empty);
+            var script = Invariant($"rtvs:::graphics.ide.selectplot({deviceId.ToString().ToRStringLiteral()}, {plotId.ToString().ToRStringLiteral()})\n");
+            return evaluation.ExecuteAsync(script);
+        }
+
+        public static async Task ActivatePlotDeviceAsync(this IRExpressionEvaluator evaluation, Guid deviceId) {
+            Debug.Assert(deviceId != Guid.Empty);
+            await evaluation.ExecuteAsync(Invariant($"rtvs:::graphics.ide.setactivedeviceid({deviceId.ToString().ToRStringLiteral()})"));
```
int19h: `ExecuteAsync` and `EvaluateAsync` have overloads that take `FormattableString` and apply `Invariant` to it, so you don't have to do it at call site.

src/R/Components/Impl/Plots/Implementation/PlotClipboardData.cs
```diff
@@ -0,0 +1,63 @@
...
+        public PlotClipboardData(Guid deviceId, Guid plotId, int? processId, bool cut) {
+            DeviceId = deviceId;
+            PlotId = plotId;
+            ProcessId = processId;
+            Cut = cut;
+        }
+
+        // Using .NET serialization wasn't working, error unable to locate the assembly that contained the type
+        // so this is using 'manual' serialization and storing a string type in the clipboard instead.
+        public static PlotClipboardData Parse(string text) {
```
int19h: Perhaps use Json.NET here to serialize `PlotClipboardData` directly to/from string, and then use that as the clipboard representation?

src/R/Components/Impl/Plots/Implementation/PlotClipboardData.cs
```diff
@@ -0,0 +1,63 @@
...
+                }
+
+                return new PlotClipboardData(deviceId, plotId, processId, cut);
+            }
+
+            return null;
+        }
+
+        public override string ToString() {
+            var process = ProcessId.HasValue ? ProcessId.Value.ToString() : "";
```
int19h: This is redundant - it is legal to call `ToString` on `Nullable<T>`, and the result is an empty string if `HasValue` is false.

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
+                var tasks = visualComponents.Select(v => v?.ViewModel.RefreshDeviceNameAsync());
+                await Task.WhenAll(tasks);
+
+                _interactiveWorkflow.ActiveWindow?.Container.UpdateCommandStatus(false);
+
+                if (deviceChanged) {
+                    ActiveDeviceChanged?.Invoke(this, EventArgs.Empty);
+                }
+
+            } catch (RException) {
```
int19h: Does this need to also swallow OCE?

src/R/Components/Impl/Plots/Implementation/View/RPlotDeviceControl.xaml.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        private void UserControl_DragOver(object sender, DragEventArgs e) {
+            UserControl_DragEnter(sender, e);
+        }
+
+        private void Image_MouseMove(object sender, MouseEventArgs e) {
+            if (_dragSurface.IsMouseMoveStartingDrag(e)) {
+                var data = new DataObject(
+                    PlotClipboardData.Format,
+                    new PlotClipboardData(Model.DeviceId, Model.ActivePlotId, Model.SessionProcessId, false).ToString()
```
int19h: No null check on `Model` here, like there is in other places where it's used. Is it certain to not be null at this point?

src/Package/Impl/ToolWindows/PlotDeviceWindowPane.cs
```diff
@@ -0,0 +1,88 @@
...
+            var viewModel = new RPlotDeviceViewModel(_plotManager, _session, _instanceId);
+            //controller.AddCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdNewPlotWindow, PlotDeviceCommandFactory.NewPlotDevice(_plotManager.InteractiveWorkflow));
+            controller.AddCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdActivatePlotWindow, PlotDeviceCommandFactory.ActivatePlotDevice(_plotManager.InteractiveWorkflow, viewModel));
+            controller.AddCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdClearPlots, PlotDeviceCommandFactory.RemoveAllPlots(_plotManager.InteractiveWorkflow, viewModel));
+            controller.AddCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdRemovePlot, PlotDeviceCommandFactory.RemoveCurrentPlot(_plotManager.InteractiveWorkflow, viewModel));
+            controller.AddCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdNextPlot, PlotDeviceCommandFactory.NextPlot(_plotManager.InteractiveWorkflow, viewModel));
+            controller.AddCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdPrevPlot, PlotDeviceCommandFactory.PreviousPlot(_plotManager.InteractiveWorkflow, viewModel));
```
AlexanderSher: - Let's make `PlotDeviceCommandFactory` an instance. You'll be able to pass `InteractiveWorkflow` and `viewModel` to its ctor and avoid providing them for every command
- We have `CommandAsyncToOleMenuCommandShimFactory` that with standard guids. We probably can add an extension method that will do the same

src/R/Components/Impl/Controller/AsyncCommandController.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Common.Core;
+
+namespace Microsoft.R.Components.Controller {
+    public class AsyncCommandController : ICommandTarget {
+        public Dictionary<Guid, Dictionary<int, IAsyncCommand>> CommandMap { get; } = new Dictionary<Guid, Dictionary<int, IAsyncCommand>>();
+
+        public AsyncCommandController() {
```
AlexanderSher: Nit: empty ctor not required

src/R/Components/Impl/Controller/AsyncCommandController.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Common.Core;
+
+namespace Microsoft.R.Components.Controller {
+    public class AsyncCommandController : ICommandTarget {
+        public Dictionary<Guid, Dictionary<int, IAsyncCommand>> CommandMap { get; } = new Dictionary<Guid, Dictionary<int, IAsyncCommand>>();
```
AlexanderSher: I'd rather create a `Key` type and make dictionary flat:
```{csharp}
    private struct Key : IEquatable<Key> {
        public Guid Group { get; }
        public int Id { get; }

        public Key(Guid group, int id) {
            Group = group;
            Id = id;
        }

        public override bool Equals(object obj) => obj is Key && Equals((Key) obj);

        public bool Equals(Key other) => Group.Equals(other.Group) && Id == other.Id;

        public override int GetHashCode() {
            unchecked {
                return (Group.GetHashCode()*397) ^ Id;
            }
        }
    }
```

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -61,25 +61,41 @@ internal sealed class RSessionCallback : IRSessionCallback {
         /// Displays R plot in the host app-provided window
         /// </summary>
         public async Task Plot(PlotMessage plot, CancellationToken ct) {
```
AlexanderSher: `PlotAsync`

src/R/Components/Impl/Plots/PlotDeviceCommandFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public static IAsyncCommand ExportAsImage(IRInteractiveWorkflow interactiveWorkflow, IRPlotDeviceViewModel viewModel) {
+            return new PlotDeviceExportAsImageCommand(interactiveWorkflow, viewModel);
+        }
+
+        public static IAsyncCommand ExportAsPdf(IRInteractiveWorkflow interactiveWorkflow, IRPlotDeviceViewModel viewModel) {
+            return new PlotDeviceExportAsPdfCommand(interactiveWorkflow, viewModel);
+        }
+
+        public static IAsyncCommand Copy(IRInteractiveWorkflow interactiveWorkflow, IRPlotDeviceViewModel viewModel, bool cut) {
```
AlexanderSher: Let's either call it `CutCopy` or have 2 methods instead of `bool` parameter

src/R/Components/Impl/Plots/Implementation/RPlotDeviceVisualComponent.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            var control = new RPlotDeviceControl {
+                DataContext = ViewModel,
+            };
+            control.ContextMenuRequested += Control_ContextMenuRequested;
+
+            Control = control;
+
+            ViewModel.DeviceNameChanged += ViewModel_DeviceNameChanged;
+            ViewModel.LocatorModeChanged += ViewModel_LocatorModeChanged;
```
AlexanderSher: You can use `DisposableBag` here

src/R/Components/Impl/Plots/IRPlotHistoryVisualComponent.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.R.Components.Plots.ViewModel;
+using Microsoft.R.Components.View;
+
+namespace Microsoft.R.Components.Plots {
+    public interface IRPlotHistoryVisualComponent : IVisualComponent {
+        IRPlotHistoryViewModel ViewModel { get; }
```
AlexanderSher: `IVisualComponent` is a facade interface between UI and non-UI parts of component. ViewModel belongs to UI, so we shouldn't expose it to the VisualComponent. Of course, there is a `IWpfTextView`, which is exposed and is also a UI component, but that should also be refactored later.

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -20,162 +24,253 @@ internal class RPlotManager : IRPlotManager {
...
-                } else if (plot.IsError) {
-                    VisualComponent.Error();
-                }
+        public async Task DeviceDestroyedAsync(Guid deviceId) {
+            IRPlotDeviceVisualComponent component;
+            if (_assignedVisualComponents.TryGetValue(deviceId, out component)) {
+                // Remove the plots from history before the device view model gets unassigned
+                HistoryVisualComponent?.ViewModel.RemoveAll(deviceId);
+
+                await component.ViewModel.UnassignAsync();
```
AlexanderSher: `IRPlotDeviceVisualComponent` has `UnassignAsync` method

src/Host/Client/Impl/Definitions/IRSessionCallback.cs
```diff
@@ -35,7 +36,18 @@ public interface IRSessionCallback {
...
         /// </summary>
         /// <returns>Location where the user clicked.</returns>
-        Task<LocatorResult> Locator(CancellationToken ct);
+        Task<LocatorResult> Locator(Guid deviceId, CancellationToken ct);
+
+        /// <summary>
+        /// Device is being created, so create/assign a plot window for it.
+        /// </summary>
+        /// <returns>Properties for the plot device window, such as width, height and resolution.</returns>
+        Task<PlotDeviceProperties> PlotDeviceCreate(Guid deviceId, CancellationToken ct);
```
MikhailArkhipov: I'd rather add ```IPlotDevice: IDisposable``` and do ```Task<IPlotDevice> CreateDevice()``` and destroy is via `IPlotDevice.DestroyAsync()`. This will also eliminate need to expose guids.

src/Host/Client/Impl/Definitions/PlotDeviceProperties.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Client {
+    public struct PlotDeviceProperties {
```
MikhailArkhipov: Can be ```IPlotDevice``` and properties on it.

src/Host/Client/Impl/Definitions/PlotMessage.cs
```diff
@@ -6,13 +6,19 @@
 
 namespace Microsoft.R.Host.Client {
```
MikhailArkhipov: Perhaps in `.Plotting` or 'Plots' namespace

src/Host/Client/Impl/Definitions/IRSessionCallback.cs
```diff
@@ -35,7 +36,18 @@ public interface IRSessionCallback {
         /// Set locator mode in the plot window
```
MikhailArkhipov: Maybe we should start splitting `IRSessionCallback` and have separate `IRSessionPlotCallback`? @int19h what do you think?

src/Host/Client/Impl/Program.cs
```diff
@@ -160,9 +160,18 @@ class Program : IRCallbacks {
             }
         }
 
-        public async Task<LocatorResult> Locator(CancellationToken ct) {
-            await Console.Error.WriteLineAsync("Locator called.");
+        public async Task<LocatorResult> Locator(Guid deviceId, CancellationToken ct) {
+            await Console.Error.WriteLineAsync($"Locator called for {deviceId}.");
```
MikhailArkhipov: `Invariant($"...")`

src/Host/Client/Impl/Program.cs
```diff
@@ -160,9 +160,18 @@ class Program : IRCallbacks {
...
             return LocatorResult.CreateNotClicked();
         }
+
+        public async Task<PlotDeviceProperties> PlotDeviceCreate(Guid deviceId, CancellationToken ct) {
+            await Console.Error.WriteLineAsync($"PlotDeviceCreate called for {deviceId}.");
+            return PlotDeviceProperties.CreateDefault();
+        }
+
+        public async Task PlotDeviceDestroy(Guid deviceId, CancellationToken ct) {
+            await Console.Error.WriteLineAsync($"PlotDeviceDestroy called for {deviceId}.");
```
MikhailArkhipov: ```Invariant()```

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -20,162 +24,253 @@ internal class RPlotManager : IRPlotManager {
...
+            if (visualComponent != null) {
+                visualComponent.Container.Show(focus: false, immediate: false);
+                await ProcessPlotMessage(visualComponent.ViewModel, plot);
+            }
 
-            PlotChanged?.Invoke(this, null);
+            PlotMessageReceived?.Invoke(this, EventArgs.Empty);
 
-            return Task.CompletedTask;
+            _interactiveWorkflow.ActiveWindow?.Container.UpdateCommandStatus(false);
```
AlexanderSher: Is it intentional that you want to update command status of the REPL window?

src/Host/Client/Test/IdeGraphicsDeviceTest.cs
```diff
@@ -650,13 +663,20 @@ public class IdeGraphicsDeviceTest : IDisposable {
         }
 
         private static string[] Interactive(string code) {
-            return code.Split(new[] { '\n' }, StringSplitOptions.RemoveEmptyEntries);
+            return code.Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries);
```
MikhailArkhipov: `CharExtensions.LineBreakChars`

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
+            await Task.CompletedTask;
+        }
+
+        private async void RSession_Disconnected(object sender, EventArgs e) {
+            await RSessionDisconnectedAsync();
+        }
+
+        private async Task RSessionDisconnectedAsync() {
+            foreach (var visualComponent in _assignedVisualComponents.Values) {
+                await visualComponent.UnassignAsync();
```
AlexanderSher: Do you need UI thread for that?

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
+                _unassignedVisualComponents.Add(visualComponent);
```
AlexanderSher: >_unassignedVisualComponents [](start = 16, length = 27)

This list can be accessed from multiple threads. We need to either use locks or concurrent collections

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
+        private void RSession_Connected(object sender, EventArgs e) {
+            RSessionConnectedAsync().DoNotWait();
+        }
+
+        private async Task RSessionConnectedAsync() {
+            // TOOD: anything needs to be done here?
+            await Task.CompletedTask;
+        }
+
+        private async void RSession_Disconnected(object sender, EventArgs e) {
```
AlexanderSher: We shouldn't use `async void`. Please use `RSessionDisconnectedAsync().DoNotWait();` instead

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
         }
 
-        public void EndLocatorMode(LocatorResult result) {
-            var tcs = _locatorTcs;
-            _locatorTcs = null;
-            tcs?.SetResult(result);
-            _interactiveWorkflow.Shell.DispatchOnUIThread(() => SetLocatorModeUI(false));
-            LocatorModeChanged?.Invoke(this, null);
+        private async void RSession_Mutated(object sender, EventArgs e) {
+            await RSessionMutatedAsync();
```
AlexanderSher: `RSessionMutatedAsync().DoNotWait()`

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
-                VisualComponent.SetLocatorMode(locatorMode);
+        private async Task RSessionMutatedAsync() {
+            try {
+                var deviceId = await InteractiveWorkflow.RSession.GetActivePlotDeviceAsync();
+                var deviceChanged = ActiveDeviceId != deviceId;
+                ActiveDeviceId = deviceId;
+
+                // Update all the devices in parallel
+                var visualComponents = _visualComponents.Values.ToArray();
+                var tasks = visualComponents.Select(v => v?.ViewModel.RefreshDeviceNameAsync());
```
AlexanderSher: This is the only place where this method is called. And under the hood it calls `GetPlotDeviceNumAsync`. Can we make one request to the host for all deviceId's and the pass it here as a parameter?

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
...
+            } else {
+                // We didn't find the requested visual component.
+                // Our internal state of unassigned visual components must be out of sync.
+                Debug.Assert(false, "Could not find instance of plot window.");
+            }
+        }
 
-        public Task ExportToMetafileAsync(string outputFilePath) =>
-            ExportAsync(outputFilePath, _interactiveWorkflow.RSession.ExportToMetafileAsync(outputFilePath, PixelsToInches(_lastPixelWidth), PixelsToInches(_lastPixelHeight), _lastResolution));
+        private async Task ProcessPlotMessage(IRPlotDeviceViewModel viewModel, PlotMessage plot) {
```
AlexanderSher: >ProcessPlotMessage [](start = 27, length = 18)

Does this method require UI thread?

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -188,35 +283,143 @@ internal class RPlotManager : IRPlotManager {
             }
         }
 
-        public Task ExportToBitmapAsync(string deviceName, string outputFilePath) =>
-            ExportAsync(outputFilePath, _interactiveWorkflow.RSession.ExportToBitmapAsync(deviceName, outputFilePath, _lastPixelWidth, _lastPixelHeight, _lastResolution));
+        private async Task<IRPlotDeviceVisualComponent> GetVisualComponentForDevice(Guid deviceId) {
```
AlexanderSher: Can this method be called concurrently?

src/Host/Client/Impl/Definitions/IRSessionCallback.cs
int19h: If we want to refactor things here, the right approach would be to keep only the core callbacks that the host itself requires (ReadConsole etc), and add an extensible callback registration mechanism for everything else. But it's a relatively big change, so maybe not at this point.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
int19h: Note that this is a callback corresponding to a host message. Our naming convention for those so far has been to use the same name as the message (mostly), without `Async` - this is stemming from the fact that the first callbacks were all corresponding to functions that R calls, like `ReadConsole`, and the desire to establish a clear mapping between those functions, and corresponding C# callbacks.

MikhailArkhipov:
BTW, what is the related 0.5/RTM task?

src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceActivateCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+                return CommandStatus.SupportedAndEnabled;
+            }
+        }
+
+        public async Task<CommandResult> InvokeAsync() {
+            try {
+                await ViewModel.ActivateDeviceAsync();
+            } catch (RPlotManagerException ex) {
+                InteractiveWorkflow.Shell.ShowErrorMessage(ex.Message);
+            } catch (OperationCanceledException) {
```
AlexanderSher: How this exception may happen?

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
AlexanderSher: Not really. If method is called with `DoNotWait`, it will be swallowed automatically

---
In reply to: [76161851](https://github.com/Microsoft/RTVS/pull/2199#discussion_r76161851) [](ancestors = 76161851)

src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceNewCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+                return CommandStatus.SupportedAndEnabled;
+            }
+        }
+
+        public async Task<CommandResult> InvokeAsync() {
+            try {
+                await InteractiveWorkflow.Plots.NewDeviceAsync(-1);
+            } catch (RPlotManagerException ex) {
+                InteractiveWorkflow.Shell.ShowErrorMessage(ex.Message);
+            } catch (OperationCanceledException) {
```
AlexanderSher: You catch OCE exceptions in commands. `RHostDisconnectedException` is derived from OCE, so do you want to convert it to `RPlotManagerException` in `PlotsManager`, or do you want it to be swallowed on this level?

src/R/Components/Impl/Plots/Implementation/Commands/PlotDevicePasteCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+
+                return CommandStatus.Supported;
+            }
+        }
+
+        public async Task<CommandResult> InvokeAsync() {
+            if (Clipboard.ContainsData(PlotClipboardData.Format)) {
+                var source = PlotClipboardData.Parse((string)Clipboard.GetData(PlotClipboardData.Format));
+                if (source != null) {
+                    if (source.ProcessId == InteractiveWorkflow.RSession.ProcessId) {
```
AlexanderSher: How does it work for remote connections?

src/Host/Client/Impl/Definitions/PlotDeviceProperties.cs
huguesv: Done.

src/Host/Client/Impl/Host/RHost.cs
huguesv: true, I think I added this across the board after I had issues with `graphics.ide.getactivedeviceid` where I sometimes returned NULL.  I could just leave it for the code that checks the result of `graphics.ide.getactivedeviceid` or also change that implementation to return empty guid instead of NULL.

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
huguesv: Alright, I'm removing all Invariant in those calls in this file.

src/R/Components/Test/Plots/PlotIntegrationTest.cs
```diff
@@ -603,25 +1057,56 @@ public class PlotIntegrationTest : IAsyncLifetime {
...
-            var eas = new EventTaskSource<IRPlotManager>((o, h) => o.PlotChanged += h, (o, h) => o.PlotChanged -= h);
-            var plotChangedTask = eas.Create(_workflow.Plots);
+            var eas = new EventTaskSource<IRPlotManager>((o, h) => o.PlotMessageReceived += h, (o, h) => o.PlotMessageReceived -= h);
+            var task = eas.Create(_workflow.Plots);
+            await action();
+            await task;
+        }
+
+        private async Task WaitForMutated(Func<Task> action) {
+            var eas = new EventTaskSource<IRSession, EventArgs>((o, h) => o.Mutated += h, (o, h) => o.Mutated -= h);
```
AlexanderSher: >EventTaskSource [](start = 26, length = 15)

You can use `EventTaskSources.IRSession.Mutated` - no need to create new `EventTaskSource` instance every time

src/R/Components/Test/Plots/PlotIntegrationTest.cs
```diff
@@ -603,25 +1057,56 @@ public class PlotIntegrationTest : IAsyncLifetime {
...
             await action();
-            await plotChangedTask;
+            await task;
         }
 
-        private async Task WaitForLocatorModeChangedAsync(Action action) {
-            var eas = new EventTaskSource<IRPlotManager>((o, h) => o.LocatorModeChanged += h, (o, h) => o.LocatorModeChanged -= h);
-            var locatorModeChangedTask = eas.Create(_workflow.Plots);
+        private async Task WaitForLocatorModeChangedAsync(IRPlotDeviceViewModel vm, Action action) {
+            var eas = new EventTaskSource<IRPlotDeviceViewModel>((o, h) => o.LocatorModeChanged += h, (o, h) => o.LocatorModeChanged -= h);
```
AlexanderSher: Same thing here - let's add a static nested class to `Microsoft.R.Components.Test.EventTaskSources`

AlexanderSher:
```{csharp}
            public static readonly EventTaskSource<Microsoft.R.Components.Plots.ViewModel.IRPlotDeviceViewModel, EventArgs> LocatorModeChanged =
                new EventTaskSource<Microsoft.R.Components.Plots.ViewModel.IRPlotDeviceViewModel, EventArgs>(
                    (o, e) => o.LocatorModeChanged += e,
                    (o, e) => o.LocatorModeChanged -= e);

```

---
In reply to: [76331752](https://github.com/Microsoft/RTVS/pull/2199#discussion_r76331752) [](ancestors = 76331752)

src/R/Components/Test/Plots/PlotIntegrationTest.cs
AlexanderSher: With that, you don't have to add these helper methods, which should simplify test code as well

---
In reply to: [76331453](https://github.com/Microsoft/RTVS/pull/2199#discussion_r76331453) [](ancestors = 76331453)

src/R/Components/Test/Plots/PlotIntegrationTest.cs
```diff
@@ -595,6 +1034,21 @@ public class PlotIntegrationTest : IAsyncLifetime {
...
+
+            foreach (string script in scripts) {
+                await WaitForDeviceCreatedAsync(async delegate {
+                    var result = await eval.ExecuteCodeAsync(script.EnsureLineBreak());
+                    result.IsSuccessful.Should().BeTrue();
+                });
+            }
+        }
+
+        private Task<IRPlotDeviceViewModel> GetActiveViewModelAsync() {
```
AlexanderSher: I think we can make this an extension method in  product code.

src/R/Components/Impl/Plots/Implementation/ViewModel/RPlotDeviceViewModel.cs
```diff
@@ -0,0 +1,313 @@
...
+            _plotManager = plotManager;
+            _session = session;
+            _deviceNum = -1;
+            InstanceId = instanceId;
+            _showWatermark = true;
+            ActivePlotIndex = -1;
+            PlotCount = 0;
+        }
+
+        public bool HasPlot {
```
AlexanderSher: Do you have an active binding that is based on this property?

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: I had asked that very question to @int19h and we figured it was unnecessary, as long as we waited for all of them at once.

int19h: To clarify, if you don't await on the call (and ultimately on the underlying `SendNotificationAsync` call in the host), all these packets are going to be pushed down the websocket connection as soon as they come in, without waiting for ACK from the other side between sends. So there isn't really any benefit derived from making special-purpose batch messages, since you can always batch any two or more messages with `Task.WhenAll` like this (and better yet, you can batch different kinds of requests that way).

AlexanderSher: It isn't completely unnecessary, cause current implementation creates cross-layer loop: business logic -> visual component -> view model -> business logic. Also, I don't know if it matters from the performance perspective to have one request instead of many.

---
In reply to: [76335141](https://github.com/Microsoft/RTVS/pull/2199#discussion_r76335141) [](ancestors = 76335141)

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: oops, did it right for connected but not disconnected event.  fixed, and i'm removing the connected event since i didn't find a reason have it yet..

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
int19h: Low-level perf perspective is the one I had in mind. There is some overhead in terms of the extra message header and WebSocket packet frame if you send more than one, of course, but it's on the order of tens of bytes - so negligible for large messages. And for small messages, it would only matter if you constantly send hundreds of them, which I don't think is the case here. 

The bigger problem is latency, when you serialize messages, i.e. send one, and then await before sending another. Send will return only when it got an ACK from the other side, so the channel ends up empty for the duration if nothing else is sending, wasting throughput - and this will be more noticeable on slow remote links. If you don't await on one send before sending another, you're making optimal use of the channel by saturating it. So for any higher-level APIs that we have that ultimately end up sending packets, avoiding unnecessary sequencing where there's no data dependency (i.e. avoiding `await` on individual calls as much as possible, and instead bundling them with `WhenAll` at points where you have to get the result), should be the optimal pattern for perf - and batching won't add much to it.

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: fixed

src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceNewCommand.cs
huguesv: My thought was that any cancellation triggered by the user (click reset or close VS) can be silenced in the commands.  

I'm unclear about `RHostDisconnectedException` though.  Right now it's getting wrapped in a `RPlotManagerException` which means it will popup an error message box.

AlexanderSher: `Reset` will raise `RHostDisconnectedException`

---
In reply to: [76337779](https://github.com/Microsoft/RTVS/pull/2199#discussion_r76337779) [](ancestors = 76337779)

src/R/Components/Impl/Plots/Implementation/Commands/PlotDevicePasteCommand.cs
huguesv: I don't know yet.  If we do get a process id, then it should work.  If process id is null, then it will be equal and so will proceed and fail with an error like "source device not found" instead of the more specific error I have here.

AlexanderSher: Looks like now `ProcessId` is always null

---
In reply to: [76339363](https://github.com/Microsoft/RTVS/pull/2199#discussion_r76339363) [](ancestors = 76339363)

src/R/Components/Test/Plots/PlotIntegrationTest.cs
huguesv: done

src/R/Components/Test/Plots/PlotIntegrationTest.cs
huguesv: done

src/Host/Client/Impl/Program.cs
huguesv: done

src/Host/Client/Impl/Program.cs
huguesv: done

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: Thanks, I moved it to visual component container.

src/Host/Client/Test/IdeGraphicsDeviceTest.cs
huguesv: done

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: Yes. Added AssertIsOnMainThread.

src/Host/Client/Impl/Host/RHost.cs
huguesv: I'm changing them all back to `Guid.Parse`, removing the extension method, and handling `null` in the only place where it can be returned, which is where we call `rtvs:::graphics.ide.getactivedeviceid()`.

src/R/Components/Impl/Controller/AsyncCommandController.cs
huguesv: done

src/R/Components/Impl/Plots/Implementation/RPlotDeviceVisualComponent.cs
huguesv: done here and the other classes.

src/R/Components/Impl/Plots/PlotDeviceCommandFactory.cs
huguesv: done

huguesv:
I just pushed a major update which introduces IRPlotDevice and IRPlot objects, which are owned by IRPlotManager.  ViewModel classes are no longer exposed outside of visual components. They update themselves by registering to events on IRPlotManager and IRPlotDevice.

src/R/Components/Impl/Plots/Implementation/RPlotDevice.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        public void Clear() {
+            var plots = _plots.ToArray();
+            _plots.Clear();
+
+            ActiveIndex = -1;
+            PlotCount = 0;
+
+            PlotAddedOrUpdated?.Invoke(this, new RPlotEventArgs(null));
```
AlexanderSher: Why do you call `PlotAddedOrUpdated` here?

src/R/Components/Impl/Plots/Implementation/RPlotDeviceVisualComponent.cs
```diff
@@ -0,0 +1,248 @@
...
+            await _viewModel.CopyPlotFromAsync(sourceDeviceId, sourcePlotId, isMove);
+        }
+
+        public void CopyToClipboard(bool cut) {
+            var data = PlotClipboardData.Serialize(new PlotClipboardData(_viewModel.Device.DeviceId, _viewModel.Device.ActivePlot.PlotId, cut));
+            Clipboard.Clear();
+            Clipboard.SetData(PlotClipboardData.Format, data);
+        }
+
+        public void ClickPlot(int x, int y) {
```
AlexanderSher: Is this a test-only method?

src/R/Components/Impl/Plots/Implementation/ViewModel/RPlotDeviceViewModel.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        public async Task CopyPlotFromAsync(Guid sourceDeviceId, Guid sourcePlotId, bool isMove) {
+            _shell.AssertIsOnMainThread();
+
+            if (_device == null) {
+                await _plotManager.NewDeviceAsync(InstanceId);
+            }
+
+            Debug.Assert(_device != null);
+            await _plotManager.CopyOrMovePlotFromAsync(sourceDeviceId, sourcePlotId, _device, isMove);
```
AlexanderSher: This method has no UI interaction - does it belong to ViewModel?

AlexanderSher: Or maybe I'm wrong and it has?

---
In reply to: [77024481](https://github.com/Microsoft/RTVS/pull/2199#discussion_r77024481) [](ancestors = 77024481)

src/R/Components/Impl/Plots/Implementation/ViewModel/RPlotDeviceViewModel.cs
```diff
@@ -0,0 +1,279 @@
...
+            tcs?.SetResult(result);
+
+            LocatorMode = false;
+            LocatorModeChanged?.Invoke(this, EventArgs.Empty);
+
+            _device.LocatorMode = LocatorMode;
+        }
+
+        private void DeviceNumChanged(object sender, RPlotDeviceEventArgs e) {
+            DeviceNameChanged?.Invoke(this, EventArgs.Empty);
```
AlexanderSher: Why do you need this event chain?

src/R/Components/Test/Plots/RPlotIntegrationTest.cs
```diff
@@ -0,0 +1,1121 @@
...
+
+                device2.ActivePlot.Image.Should().HaveSamePixels(plot1to10);
+            }
+        }
+
+        [Test(ThreadType.UI)]
+        [Category.Plots]
+        public async Task DeviceCut() {
+            _plotDeviceVisualComponentContainerFactory.DeviceProperties = new PlotDeviceProperties(600, 500, 96);
+            using (await _workflow.GetOrCreateVisualComponent(_componentContainerFactory)) {
```
AlexanderSher: Looks like this can be moved to `InitializeAsync`

src/R/Components/Impl/Plots/Implementation/RPlotDevice.cs
huguesv: I agree this is confusing.  This is the event that `RPlotDeviceViewModel` uses to update its current plot.  I've changed it to only fire a new event `Cleared`, which can be processed more efficiently than multiple `PlotRemoved` events (and I've removed the firing of `PlotAddedOrUpdated` and `PlotRemoved`).

src/R/Components/Impl/Plots/Implementation/RPlotDeviceVisualComponent.cs
huguesv: Yes, to test can simulate a click.

src/R/Components/Impl/Plots/Implementation/ViewModel/RPlotDeviceViewModel.cs
huguesv: It doesn't. As per our conversation, I'll make the commands that don't need to modify the view model go directly to plot manager instead.

src/R/Components/Impl/Plots/Implementation/ViewModel/RPlotDeviceViewModel.cs
huguesv: Ultimately, it's so that the visual component can call on its container to update the tool window caption.

Issue #2224, opened 2016-08-26
Working directory is wrong for remote connections


int19h:
When we connect, we set current working directory to "C:\Users\...\Documents". The problem is that we determine that path on VS side of things, so it is not necessarily valid for the remote machine. And if that path doesn't exist, setting the working directory fails, and then everything that tries to get the working directory also fails (and crashes VS).

int19h:
Also see #2223 - probably same root cause.

int19h:
So, the problem is that right now we actually let the user specify the default working directory in R Tools | Options, and then just blindly apply that setting to all connections. Obviously this isn't working for remote.

First question: do remote connections even need a user-settable working directory? Or do we just assume that it's always `~`, so that users remain boxed in their own folders? If so, we can just stop applying this option to remote connections.

We could go one step further, and have this become a per-connection setting. And then either only allow it for local connections, or for both local and remote - but in the latter case, we need to figure out what it actually means for remote, and how it should be constrained.

I'm going to do #1 for now because it's very easy and fixes the immediate problem. But we need to discuss were we go from there. @jflam?

MikhailArkhipov:
I think #1 is fine. We will see if users need to set the directory differently from usage/feedback.

int19h:
Sounds good, then https://github.com/Microsoft/RTVS/pull/2225 was the fix.

Issue #2223, opened 2016-08-25
Opening a project on a remote connection attempts to call setwd() against the local directory which may not exist on the server


jflam:
In this case it's trying to set the working directory to c:\users\jflam\src\demo directory which doesn't exist on the server.

![image](https://cloud.githubusercontent.com/assets/8877/17989817/1250cc2a-6ae5-11e6-9fd5-7be9c3bcfc5f.png)

int19h:
We should be using the remote directory that is going to be mapped for that project... the catch is that until you do F5, that directory probably doesn't even exist.

I can unblock this for now by not doing setwd on project load if connection is remote, but we need to figure out what is the right thing to do here for a proper fix.

int19h:
This was unblocked by #2225, but we still need to decide if there is some working directory that we do need to set here, and if so, what it is; otherwise, this is fixed. @jflam?

int19h:
Any updates on the desired behavior?

MikhailArkhipov:
Probably just skipping setwd for now

int19h:
Sounds good. I'll close this one then, and we can open a new one if some other desired behavior emerges in testing.

PR #2225, opened 2016-08-26
Unblock #2223, #2224

int19h:
Do not change working directory when starting host or loading project if connection is remote (and local paths are not applicable).

### DIFF
--- a/src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
+++ b/src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
@@ -126,17 +126,22 @@ namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
                 return;
             }
 
-            var rdataPath = Path.Combine(_projectDirectory, DefaultRDataName);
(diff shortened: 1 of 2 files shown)


src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
```diff
@@ -126,14 +126,18 @@ internal sealed class RProjectLoadHooks {
...
+                    if (loadDefaultWorkspace) {
+                        await evaluation.LoadWorkspaceAsync(rdataPath);
+                    }
 
-                await evaluation.SetWorkingDirectoryAsync(_projectDirectory);
+                    await evaluation.SetWorkingDirectoryAsync(_projectDirectory);
+                }
             }
 
             _toolsSettings.WorkingDirectory = _projectDirectory;
```
AlexanderSher: This should go under `if (!_session.IsRemote) {` as well

int19h: Yep. Updated.

AlexanderSher:
:+1:

PR #2242, opened 2016-09-01
Remote help

karthiknadig:
These changes allow ? and ?? help to work for remote scenarios.


src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+namespace Microsoft.R.Host.Broker.RemoteUri {
+    public class RemoteUriHelper {
+        public static async Task HandlerAsync(HttpContext context) {
+            string url = null;
+            string method = null;
+            Dictionary<string, string> headers = null;
+            Stream content = null;
+
+            BinaryReader reader = new BinaryReader(context.Request.Body);
+            url = reader.ReadString();
```
MikhailArkhipov: nit: no need to set to null

```
var url = ...
var method = ...
```

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+            method = reader.ReadString();
+            headers = JsonConvert.DeserializeObject<Dictionary<string, string>>(reader.ReadString());
+
+            HttpClient client = new HttpClient();
+            Uri uri = new Uri(url);
+
+            HttpMethod httpMethod = new HttpMethod(method);
+            HttpRequestMessage req = new HttpRequestMessage(httpMethod, uri);
+
+            foreach (var key in headers.Keys) {
```
MikhailArkhipov:
```
foreach(var kvp in headers) {
 req.Headers.Add(kvp.Key, kvp.Value);
}
```

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+                var val = headers[key];
+                req.Headers.Add(key, val);
+            }
+
+            if (httpMethod == HttpMethod.Post) {
```
MikhailArkhipov: May be worth turning into some sort of a helper method or extension

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+            var httpResponse = await client.SendAsync(req);
+            var respHeaders = new Dictionary<string, string>();
+
+            foreach (var pair in httpResponse.Headers) {
+                StringBuilder value = new StringBuilder("");
+                foreach (var val in pair.Value) {
+                    value.Append(val);
+                    value.Append(",");
+                }
+                respHeaders.Add(pair.Key, value.ToString());
```
MikhailArkhipov: OK to have trailing comma?

src/Host/Broker/Impl/Startup/Startup.cs
```diff
@@ -62,7 +64,13 @@ SecurityManager securityManager
             });
 
             app.Use((context, next) => {
-                if (!context.User.Identity.IsAuthenticated) {
+                if (context.Request.Path.Value.Contains("/remoteuri")) {
```
MikhailArkhipov: `Contains()` or `EndsWith`?

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Threading.Tasks;
+using System.Net.Http;
+using System.IO;
+using Newtonsoft.Json;
+using Microsoft.AspNetCore.Http;
+using System.Net.Http.Headers;
+
+namespace Microsoft.R.Host.Broker.RemoteUri {
+    public class RemoteUriHelper {
+        public static async Task HandlerAsync(HttpContext context) {
```
MikhailArkhipov: May benefit from more descriptive name - what exactly does it `handle`

src/Host/Broker/Impl/Startup/Startup.cs
```diff
@@ -11,6 +11,8 @@
 using Microsoft.R.Host.Broker.Security;
 using Microsoft.R.Host.Broker.Sessions;
 using Odachi.AspNetCore.Authentication.Basic;
```
MikhailArkhipov: Sort

src/Host/Client/Impl/BrokerServices/RemoteUriResponse.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.R.Host.Client.BrokerServices {
+    public class RemoteUriResponse {
+        public Dictionary<string, string> Headers { get; set; }
```
MikhailArkhipov: if it is public' I'd prefer `IReadOnlyDictionary`

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Common.Core;
+using Newtonsoft.Json;
+
+namespace Microsoft.R.Host.Client.BrokerServices {
+    public class WebServer {
+        private readonly string _newHost;
+        public string Host => _newHost;
```
MikhailArkhipov: Just `public string Host {get;}`, no need for supporting variable

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        private readonly int _newPort;
+        public int Port => _newPort;
```
MikhailArkhipov: As above

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+            _service = new RemoteUriWebService(httpClient);
+            Random r = new Random();
+
+            // if remote port is between 10000 and 32000, select a port in the same range.
+            // R Help uses ports in that range.
+            int localPortMin = (_remotePort >= 10000 && _remotePort <= 32000)? 10000: 49152;
+            int localPortMax = (_remotePort >= 10000 && _remotePort <= 32000) ? 32000 : 65535;
+
+            while(true) {
+                _listener = new HttpListener();
```
MikhailArkhipov: HttpListener is disposable so sounds like you need `using` or `_listener?.Dispose()`

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+                if (ct.IsCancellationRequested) {
+                    _listener.Stop();
+                    break;
+                }
+
+                HttpListenerContext context = await _listener.GetContextAsync();
+
+                RemoteUriResponse resp = null;
+                using (var requestStream = await CreateRequestAsync(context.Request, RemoteHost, RemotePort)) {
+                    var responseStream = await _service.PostAsync(requestStream);
```
MikhailArkhipov: does responseStream need to be disposed?

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+                foreach (var pair in resp.Headers) {
+                    string value = pair.Value;
+                    value = value.Replace(remoteHostPort, localHostPort);
+                    webHeaders.Add($"{pair.Key}:{value}");
+                }
+
+                context.Response.Headers = webHeaders;
+                await resp.Content.CopyToAsync(context.Response.OutputStream);
+                await context.Response.OutputStream.FlushAsync();
+                context.Response.OutputStream.Close();
```
MikhailArkhipov: Does `Close` need to be on `finally { }`?

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -185,5 +185,13 @@ internal abstract class RHostConnector : IRHostConnector {
             var host = new RHost(name, callbacks, transport, null, cts);
             return host;
         }
+
+        public string HandleUrl(string url, CancellationToken ct) {
```
AlexanderSher: This method doesn't belong to the RHostConnector. It is either should be implemented in derived class, or should have no dependency on the `IsRemote`
`

src/Host/Broker/Impl/Startup/Startup.cs
AlexanderSher: Shouldn't we define controller instead?

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Net.Http;
+using System.IO;
+using Newtonsoft.Json;
+using Microsoft.AspNetCore.Http;
+using System.Net.Http.Headers;
+
+namespace Microsoft.R.Host.Broker.RemoteUri {
+    public class RemoteUriHelper {
```
AlexanderSher: Why do you want to do all these things manually instead of using MVC?

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Common.Core;
+using Newtonsoft.Json;
+
+namespace Microsoft.R.Host.Client.BrokerServices {
+    public class WebServer {
```
AlexanderSher: Why is it so complicated?
I'd expect to get a new remote Url that matches local Url, and then broker will handle it. And ideally, it should not be a broker, but the RHost itself.
@int19h should take a look on this PR.

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -571,8 +573,9 @@ internal sealed class RSession : IRSession, IRCallbacks {
         /// <param name="url"></param>
         /// <returns></returns>
         Task IRCallbacks.WebBrowser(string url) {
```
AlexanderSher: >WebBrowser [](start = 25, length = 10)

AFAIK, this method is called for any URL, not just help. I don't think we should bother broker if it isn't a help URL.

AlexanderSher: Thinking a bit more about it. If we define a broker-rhost URL conversion protocol. something like:
```
[http://localhost:1212/myhelp.html](http://localhost:1212/myhelp.html) <-> [http://brokerurl:2020/localurl/1212/myhelp.html](http://brokerurl:2020/localurl/1212/myhelp.html)
```
RHost can make this conversion when open_url is called, and then broker can handle it as a normal web request.
In this case, we won't have to make any changes in VS part at all

---
In reply to: [77244150](https://github.com/Microsoft/RTVS/pull/2242#discussion_r77244150) [](ancestors = 77244150)

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
int19h: "Handler" here is the ASP.NET term for middleware handler.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
int19h: MVC stuff will just get in the way here. This just needs to forward HTTP requests, maintaining them as much intact as possible, so the fewer steps of the pipeline they have to go through before being handled, the better. MVC action result handling is also less than helpful for this purpose.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
int19h: Is there any particular reason to restrict this to Post? It should definitely also handle at least Put, but I think it could be method-blind and just forward the body as is - it should be empty for other methods anyway (and if not, it's not really this code's problem).

src/Host/Broker/Impl/Startup/Startup.cs
int19h: The only thing it needs is routing, which is available separately of full-fledged controllers. But it should use `MapRoute`.

https://docs.asp.net/en/latest/fundamentals/routing.html#using-routing-middleware

int19h: Also, can you make it separate from authentication in the pipeline (i.e. two separate `Use` calls, or separate `Use` and `MapRoute`)?

src/Host/Client/Impl/BrokerServices/RemoteUriResponse.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.R.Host.Client.BrokerServices {
+    public class RemoteUriResponse {
```
int19h: This looks like trimmed-down `HttpResponseMessage` - why not use that?

Also, this assembly is intended for DTOs that are shared between the broker and the client to use at the API boundary. Looks like the broker doesn't actually use this anywhere, so it shouldn't be here.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+                content.Position = 0;
+
+                req.Content = new StreamContent(content);
+            }
+
+            // send request to the local server
+            var httpResponse = await client.SendAsync(req);
+            var respHeaders = new Dictionary<string, string>();
+
+            foreach (var pair in httpResponse.Headers) {
```
int19h: I'm confused about the custom serialization scheme here. Why not just use the response itself? i.e. copy the response headers that are received from the call above to `Context.Response.Headers`, and the body to `Context.Response.Body`?

Since this is invoked from a middleware handler, it has the opportunity to decide whether anything else in the pipeline is going to touch this request or not (by calling or not calling `next()` from `Use`). So you can ensure that nothing else touches the data after it's properly formatted in the response.

src/Host/Client/Impl/BrokerServices/RemoteUriResponse.cs
int19h: It's a DTO, so there isn't really any concern about invariants etc here.

src/Host/Client/Impl/BrokerServices/WebServer.cs
int19h: We have discussed this approach at length. The arrangement with the local server is the only one that works reliably. Doing it on broker side actually ends up being more complicated, because then you have to rewrite URLs in the content (so that `/foo` becomes `/remoteuri/8080/foo` etc - don't forget that you'd need to encode the port also, since there may be multiple servers run by multiple users on the broker). Doing this reliably for HTML and JS and whatever else might be on the wire is pretty much impossible in general.

Host is definitely the wrong place to do it, because it would require every new host instance to open a port of its own for direct remote connections. We don't want to tell server admins that there may potentially be several dozen random ports open on their box when multiple users are running their stuff.

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -0,0 +1,155 @@
...
+        private IRemoteUriWebService RemoteUriService => _service;
+
+        private WebServer(string remoteHostIp, int remotePort, HttpClient httpClient) {
+            _newHost = IPAddress.Loopback.ToString();
+            _remoteHost = remoteHostIp;
+            _remotePort = remotePort;
+            _service = new RemoteUriWebService(httpClient);
+            Random r = new Random();
+
+            // if remote port is between 10000 and 32000, select a port in the same range.
```
int19h: Do we actually care about ports being in the same range? It's not going to be the same port anyway.

src/Host/Client/Impl/Session/RSession.cs
int19h: Calling this methods for all URLs is intentional, because it's also used for Shiny etc. Think of it as a proxy - because the URL is issued by the machine on which the broker and the hosts are running, to interpret it correctly, it may need to be fetched on that machine - not just for `localhost` URLs like Help and Shiny, but also potentially for URLs of other intranet stuff that's only accessible from broker but not from client. Always going through the broker makes sure that it always just works.

src/Host/Client/Impl/project.json
```diff
@@ -1,6 +1,7 @@
 ﻿{
   "dependencies": {
     "MicroBuild.Core": "0.2.0",
+    "Microsoft.AspNet.WebApi.Core": "5.2.3",
```
int19h: This is a non-ASP.NET Core assembly (it's from an older ASP.NET version). What is it needed for?

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Ok, but if broker will be just a proxy, why can't we just return proxy url from RHost itself?

---
In reply to: [77257971](https://github.com/Microsoft/RTVS/pull/2242#discussion_r77257971) [](ancestors = 77257971)

src/Host/Client/Impl/BrokerServices/WebServer.cs
karthiknadig: Yes. This is how we decide to open the help url in R Help window.

src/Host/Client/Impl/project.json
karthiknadig: removed it.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -1,78 +1,154 @@
...
+using Microsoft.Common.Core;
 
 namespace Microsoft.R.Host.Broker.RemoteUri {
     public class RemoteUriHelper {
         public static async Task HandlerAsync(HttpContext context) {
-            BinaryReader reader = new BinaryReader(context.Request.Body);
-            var url = reader.ReadString();
-            var method = reader.ReadString();
-            Dictionary<string, string> headers = JsonConvert.DeserializeObject<Dictionary<string, string>>(reader.ReadString());
+            var url = context.Request.Headers["x-rtvs-url"];
```
int19h: Since it's used on both sides, let's put it as a named constant somewhere in the `Microsoft.R.Host.Protocol` assembly.

Also, I believe the naming convention for these would make it something like "X-RTVS-URL". Maybe worth clarifying, too - something like "X-RTVS-Requested-URL".

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -1,78 +1,154 @@
...
+
+            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url);
+            request.Method = context.Request.Method;
+            request.UseDefaultCredentials = true;
```
int19h: I believe this will use Windows authentication, if one is requested. I'm not sure how this will mesh with headers that came from the client (because they will already contain authentication headers if client decided to authenticate). If it overrides them, it'll break things. If it doesn't, it's probably a no-op and can be removed.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -1,78 +1,154 @@
...
 namespace Microsoft.R.Host.Broker.RemoteUri {
     public class RemoteUriHelper {
         public static async Task HandlerAsync(HttpContext context) {
-            BinaryReader reader = new BinaryReader(context.Request.Body);
-            var url = reader.ReadString();
-            var method = reader.ReadString();
-            Dictionary<string, string> headers = JsonConvert.DeserializeObject<Dictionary<string, string>>(reader.ReadString());
+            var url = context.Request.Headers["x-rtvs-url"];
+
+            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url);
```
int19h: This needs to be disposed.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -1,78 +1,154 @@
...
+
+            if (context.Request.ContentLength > 0) {
+                Stream reqStream = await request.GetRequestStreamAsync();
+                await context.Request.Body.CopyToAsync(reqStream);
+                await reqStream.FlushAsync();
+            }
 
-            HttpClient client = new HttpClient();
-            Uri uri = new Uri(url);
+            HttpWebResponse response = (HttpWebResponse)await request.GetResponseAsync();
```
int19h: This also needs to be disposed.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -1,78 +1,154 @@
...
-                var val = headers[key];
-                req.Headers.Add(key, val);
+            Stream respStream = response.GetResponseStream();
+            await respStream.CopyToAsync(context.Response.Body);
+            await context.Response.Body.FlushAsync();
+        }
+
+        private static long GetLong(string value) {
+            long ret;
+            if (long.TryParse(value, out ret)) {
```
int19h: `TryParse` will write the default value into the out-parameter if it fails to parse, so you can just `return ret` unconditionally.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -1,78 +1,154 @@
...
+            }
+            return new DateTime();
+        }
 
-                req.Content = new StreamContent(content);
+        private static void SetRequestHeaders(HttpWebRequest request, IHeaderDictionary requestHeaders) {
+            // copy headers to avoid messing with original request headers
+            Dictionary<string, string> headers = new Dictionary<string, string>();
+            foreach (var pair in requestHeaders) {
+                headers.Add(pair.Key, pair.Value);
```
int19h: It looks like `pair.Value` here is an instance of `StringValues`, rather than `string`, and so can contain multiple values, but has an implicit conversion operator to `string`. I'm not entirely clear on what this does if there are actually multiple values, but I suspect it might either throw, or drop all values except for the first, neither of which is good.

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -5,14 +5,180 @@
...
+            long ret;
+            if (long.TryParse(value, out ret)) {
+                return ret;
+            }
+            return 0;
+        }
+
+        private static DateTime GetDateTime(string value) {
+            DateTime ret;
+            if (DateTime.TryParse(value, out ret)) {
```
int19h: Same thing here, no need for `if`.

Also, these conversion methods should probably be named something like `ToDateTimeOrDefault` (for consistency with existing `ToDateTime` and `TryToDateTime`, and LINQ methods like `FirstOrDefault`). And probably become public extension methods somewhere in `Microsoft.Common.Core`.

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -64,14 +51,18 @@ private class RemoteUriResponse {
...
             }
+            _listener.Close();
         }
 
-        private async Task DoWorkAsync(CancellationToken ct) {
-            if (_listener.IsListening) {
-                // There is already a thread listening.
-                return;
+        public static void Stop(int port) {
+            if (Servers.ContainsKey(port)) {
```
int19h: This is better done with `TryGetValue` to avoid double lookup.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        private static void SetRequestHeaders(HttpWebRequest request, IHeaderDictionary requestHeaders) {
+            // copy headers to avoid messing with original request headers
+            Dictionary<string, string> headers = new Dictionary<string, string>();
+            foreach (var pair in requestHeaders) {
+                headers.Add(pair.Key, pair.Value);
+            }
+
+            if (headers.ContainsKey("Accept")) {
```
int19h: I'm kinda surprised they didn't just wire the properties directly to the corresponding headers. Oh well... might be worth revisiting in the future to see if there's a way to go even lower in level and avoid those helpful request objects that seem to hinder more than help in this case. For now, though, let's stick with this.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
karthiknadig: Looks like X- is discouraged https://tools.ietf.org/html/rfc6648. I will use RTVS-Requested-URL for now.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
karthiknadig: HttpWebRequest does not look like it implements Dispose or Close

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
karthiknadig: Will do.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
int19h: Interesting. Something needs to dispose the request stream somewhere. I guess `GetResponse` does that.

Issue #2214, opened 2016-08-25
Switching to the broken remote link doesn't allow to switch back


AlexanderSher:
REPL becomes frozen

AlexanderSher:
`ConnectionManager.ConnectAsync` throws `RHostDisconnectedException` that should be handled

MikhailArkhipov:
#2239 may fix this

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/fff25f88abe22f5a6df42ca02e4b7c298fcf4a2f

PR #2244, opened 2016-09-02
Merge RHostBrokerProvider into RSessionProvider

AlexanderSher:
This PR is part of the work on #2214, but I want it to be submitted separately.


src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -86,5 +89,33 @@ private class IsolatedRSessionEvaluation : IRSessionEvaluation {
...
+        public Task<bool> TrySwitchBroker(string name, string path = null) {
+            path = path ?? new RInstallation().GetRInstallPath();
+
+            Uri uri;
+            if (!Uri.TryCreate(path, UriKind.Absolute, out uri)) {
+                return Task.FromResult(false);
+            }
+
+            if (uri.IsFile) {
+                BrokerConnector.SwitchToLocalBroker(name, uri.LocalPath);
```
MikhailArkhipov: I think connector can decide which method to use technically based on the URI

AlexanderSher: Connector will be removed completely, in the next PR this method will have contain the whole implementation

---
In reply to: [77388081](https://github.com/Microsoft/RTVS/pull/2244#discussion_r77388081) [](ancestors = 77388081)

MikhailArkhipov:
:+1:

PR #2246, opened 2016-09-03
Rework connections UI

MikhailArkhipov:
And related changes


src/R/Wpf/Impl/Converters.cs
```diff
@@ -28,6 +28,7 @@ public static class Converters {
         public static IValueConverter TrueIsNotHidden { get; } = LambdaConverter.Create<bool>(x => x ? Visibility.Visible : Visibility.Hidden);
         public static IValueConverter FalseIsHidden { get; } = LambdaConverter.Create<bool>(x => !x ? Visibility.Hidden : Visibility.Visible);
         public static IValueConverter FalseIsNotHidden { get; } = LambdaConverter.Create<bool>(x => !x ? Visibility.Visible : Visibility.Hidden);
+        public static IValueConverter FalseIsGrayedOut { get; } = LambdaConverter.Create<bool>(x => x ? 1.0 : 0.25);
```
AlexanderSher: Is it for transparency

src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -60,5 +60,11 @@ public static class IOExtensions {
             return files;
         }
 
+        public static string TrimTrailingSlash(this string path) {
+            if (path.EndsWithOrdinal("\\")) {
```
AlexanderSher: We should probably handle both types of slashes:
```{csharp}
if (!string.IsNullOrEmpty(path) && (path.EndsWithOrdinal("\\") || path.EndsWithOrdinal("/"))) {
    return path.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
}
return path;
```

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -102,10 +103,9 @@ public Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rComm
         }
 
         public Task<bool> TrySwitchBroker(string name, string path = null) {
-            path = path ?? new RInstallation().GetRInstallPath();
-
+            path = path ?? new RInstallation().GetCompatibleEngines().FirstOrDefault()?.InstallPath;
             Uri uri;
-            if (!Uri.TryCreate(path, UriKind.Absolute, out uri)) {
+            if (string.IsNullOrEmpty(path) || !Uri.TryCreate(path, UriKind.Absolute, out uri)) {
```
AlexanderSher: You don't need this verification - `Uri.TryCreate` handles both nulls and empty strings

src/R/Components/Impl/ConnectionManager/Implementation/LocalConnections.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.R.Components.ConnectionManager.Implementation {
+    /// <summary>
+    /// Provides information on locally installed interpreters
+    /// </summary>
+    internal sealed class LocalConnections : IEnumerable<IConnection> {
```
AlexanderSher: Do you use this type anywhere?

src/R/Components/Impl/ConnectionManager/IConnection.cs
```diff
@@ -2,15 +2,14 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
-using System.Threading.Tasks;
 
 namespace Microsoft.R.Components.ConnectionManager {
-    public interface IConnection {
+    public interface IConnection: IConnectionInfo {
```
AlexanderSher: Where do you use that `IConnection` is derived from  `IConnectionInfo`?

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -125,14 +123,13 @@ internal sealed class RToolsSettingsImplementation : IRToolsSettings {
...
-        public RToolsSettingsImplementation(IRInteractiveWorkflowProvider workflowProvider) {
+        public RToolsSettingsImplementation() {
             _workingDirectory = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
-            _workflowProvider = workflowProvider;
         }
 
         private async Task SetMirrorToSession() {
-            var sessions = _workflowProvider.GetOrCreate().RSessions.GetSessions();
+            var provider = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>();
+            var sessions = provider.GetOrCreate().RSessions.GetSessions();
```
AlexanderSher: let's make a helper method for it

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -103,18 +107,13 @@ internal class ConnectionManager : IConnectionManager {
...
-
-        public async Task ConnectAsync(IConnection connection) {
+        public async Task ConnectAsync(IConnectionInfo connection) {
             var sessions = _sessionProvider.GetSessions().ToList();
             if (sessions.Any()) {
                 await Task.WhenAll(sessions.Select(s => s.StopHostAsync()));
             }
 
-            if (ActiveConnection != null && ActiveConnection.Id != connection.Id) {
+            if (ActiveConnection != null && (!ActiveConnection.Path.EqualsIgnoreCase(connection.Path)  || _sessionProvider.BrokerUri.IsLoopback)) {
```
AlexanderSher: Why do we want to special case loopback?

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -153,25 +152,71 @@ internal class ConnectionManager : IConnectionManager {
...
+                // No connections, may be first use or connections were somehow removed.
+                // Add local connections so there is at least something available.
+                foreach (var e in localEngines) {
+                    var c = CreateConnection(e.Name, e.InstallPath, string.Empty, isUserCreated: false);
+                    connections[new Uri(e.InstallPath, UriKind.Absolute)] = c;
+                }
+            } else {
+                // Remove missing engines and add engines missing from saved connections
+                // Set 'is used created' to false if path points to locally found interpreter
+                foreach (var kvp in connections.ToList()) {
```
AlexanderSher: `connections.Where(c => !c.IsRemote).ToList()`

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -153,25 +152,71 @@ internal class ConnectionManager : IConnectionManager {
...
+                        }
+                        if (!valid) {
+                            connections.Remove(kvp.Key);
+                        }
+                    }
+                }
+
+                // Add newly installed engines
+                foreach (var e in localEngines) {
+                    if (!connections.Values.Any(x => x.Path.TrimTrailingSlash().EqualsIgnoreCase(e.InstallPath.TrimTrailingSlash()))) {
```
AlexanderSher: Why `TryGetValue` doesn't work?

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -60,5 +61,21 @@ public partial class ConnectionManagerControl : UserControl {
...
+            }
+        }
+
+        private void Connection_MouseDoubleClick(object sender, MouseButtonEventArgs e) {
+            Model?.ConnectAsync(GetConnection(e)).DoNotWait();
+            e.Handled = true;
+        }
+
+        private bool SuppressClose(object sender) {
+            var cvm = (sender as ListBoxItem)?.DataContext as IConnectionViewModel;
```
AlexanderSher: >(sender as ListBoxItem)?.DataContext as IConnectionViewModel; [](start = 22, length = 61)

GetConnection(e)

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -60,5 +61,21 @@ public partial class ConnectionManagerControl : UserControl {
...
+                Model?.ConnectAsync(GetConnection(e)).DoNotWait();
+            }
+        }
+
+        private void Connection_MouseDoubleClick(object sender, MouseButtonEventArgs e) {
+            Model?.ConnectAsync(GetConnection(e)).DoNotWait();
+            e.Handled = true;
+        }
+
+        private bool SuppressClose(object sender) {
```
AlexanderSher: Is it used?

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -24,8 +28,9 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
             UpdateCalculated();
         }
 
-        public ConnectionViewModel(IConnection connection) {
+        public ConnectionViewModel(IConnection connection, IConnectionManagerViewModel cmvm) {
             _connection = connection;
+            _cmvm = cmvm;
```
AlexanderSher: Field isn't used

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
```diff
@@ -126,116 +125,65 @@
...
+                                   Visibility="{Binding Path=IsRemote, Converter={x:Static rwpf:Converters.FalseIsCollapsed}}"/>
                         <TextBlock Grid.Column="4" Margin="0,1,3,2" Text="{Binding Path=Path, Mode=OneWay}"
-                                   TextTrimming="CharacterEllipsis" TextWrapping="NoWrap"/>
+                                   TextTrimming="CharacterEllipsis" TextWrapping="NoWrap"
+                                   Visibility="{Binding Path=IsRemote, Converter={x:Static rwpf:Converters.FalseIsCollapsed}}"/>
+
+                        <!-- connect button -->
+                        <Button x:Name="ButtonConnect" Grid.Column="5" Style="{StaticResource ToolWindowButtonStyle}" Click="ButtonConnect_Click" 
+                                Visibility="{Binding Path=IsConnected, Converter={x:Static rwpf:Converters.TrueIsCollapsed}}"
+                                Padding="0,0,8,0">
```
AlexanderSher: Is it intentional that `ButtonConnect` is separated from the other 2 buttons?

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
```diff
@@ -126,116 +125,65 @@
...
+                        <!-- edit/properties button -->
+                        <Button x:Name="ButtonEdit" Grid.Column="6" Style="{StaticResource ToolWindowButtonStyle}" Click="ButtonEdit_Click"
+                                ToolTip="{x:Static components:Resources.ConnectionManager_EditTooltip}">
+                            <Image Source="{x:Static rwpf:ImageSources.Property}"/>
                         </Button>
 
                         <!-- delete button -->
-                        <Button x:Name="ButtonDelete" Grid.Column="6" Style="{StaticResource ToolWindowButtonStyle}" Click="ButtonDelete_Click"
-                                ToolTip="{x:Static components:Resources.ConnectionManager_DeleteTooltip}" Visibility="Collapsed">
+                        <Button x:Name="ButtonDelete" Grid.Column="7" Style="{StaticResource ToolWindowButtonStyle}" Click="ButtonDelete_Click"
```
AlexanderSher: Shouldn't we just hide this button?

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
```diff
@@ -126,116 +125,65 @@
...
-                        <!-- Collapsed -->
-                        <Rectangle Fill="{StaticResource IconArrowRigth16}" Height="16" Width="16" />
-
-                        <!-- Connect button -->
+                        <!-- Connection icon -->
                         <ContentPresenter DataContext="{Binding}" Grid.Column="1" 
                                           ContentTemplate="{StaticResource ConnectButtonDataTemplate}" />
 
-                        <!-- Name & path -->
+                        <!-- Name, path, command line  -->
```
AlexanderSher: I don't see command line field

src/R/Wpf/Impl/Converters.cs
MikhailArkhipov: Fade icon out so it looks disabled

src/Common/Core/Impl/Extensions/IOExtensions.cs
MikhailArkhipov: OK

src/R/Components/Impl/ConnectionManager/Implementation/LocalConnections.cs
MikhailArkhipov: Looks like accidental add. Removed.

src/R/Components/Impl/ConnectionManager/IConnection.cs
MikhailArkhipov: To reuse implementation `ConnectionInfo` and also `IConnection` looks like live connection while `info`  is it's persistent part. Was there a reason they were separate?

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
MikhailArkhipov: Yes. Less probability of accidental clicks on other buttons.

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
MikhailArkhipov: Not anymore

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: This fixes the issue when user had zero connections and no local R. Allows to switch away from the null/stub connector (`NullRHostConnector`). I am open to alternatives, I want to improve the class anyway so it throw more descriptive exception than just `RHostDisconnectedException`

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
MikhailArkhipov: I considered it, but UI looks a bit odd with gaps on some entries. I'd like to gather some feedback.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: Check should be ignoring case and trailing slash existence. Isn't `TryGetValue` using default type comparison for values?

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
MikhailArkhipov: Done

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
```diff
@@ -65,6 +65,8 @@ internal class VsRInteractiveWorkflowProvider : IRInteractiveWorkflowProvider, I
             return _instanceLazy.Value;
         }
 
+        public IRInteractiveWorkflow Current => (_instanceLazy != null && _instanceLazy.IsValueCreated) ? _instanceLazy.Value : null;
```
AlexanderSher: Let's call it `Active`

src/R/Components/Impl/ConnectionManager/IConnection.cs
AlexanderSher: They had different fields originally

---
In reply to: [77669938](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77669938) [](ancestors = 77669938)

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -148,10 +144,10 @@ internal sealed class RToolsSettingsImplementation : IRToolsSettings {
         }
 
         private async Task SetSessionCodePage() {
-            var sessions = _workflowProvider.GetOrCreate().RSessions.GetSessions();
+            var sessions = GetRSessions();
             var cp = RToolsSettings.Current.RCodePage;
- 
-            foreach (var s in sessions.Where(s => s.IsHostRunning)) {
+
+            foreach (var s in GetRSessions().Where(s => s.IsHostRunning)) {
```
AlexanderSher: we already have list of sessions

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: ok, I've added a `Name` to the `IRHostConnector` in my changes, so I will change this code after merge

---
In reply to: [77671347](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77671347) [](ancestors = 77671347)

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -55,16 +57,24 @@ internal class ConnectionManager : IConnectionManager {
...
             _userConnections = new ConcurrentDictionary<Uri, IConnection>(userConnections);
 
             UpdateRecentConnections();
             SwitchBrokerToLastConnection();
         }
 
+        private void OnIdle(object sender, EventArgs e) {
+            _shell.Idle -= OnIdle;
+
+            _disposableBag.Add(_statusBar.AddItem(new ConnectionStatusBar {
```
AlexanderSher: This should be part of the `VsStatusBar` implementation. It already subscribes for `_dteEvents.OnStartupComplete`, so we may add a list of pending items that will be added to the status bar when it is ready.

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
AlexanderSher: Then let's use `Margin`, cause `Padding` may not offset the background, which is also considered part of the button

---
In reply to: [77670185](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77670185) [](ancestors = 77670185)

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: We can add a comparison method to the `IOExtensions` and then change dictionary comparer:
```{csharp}
public static CompareDirectoryPaths(this string path, string otherPath) {
    var pathLength = GetTrailingSlashTrimmedPathLength(path);
    var otherPathLength = GetTrailingSlashTrimmedPathLength(otherPath);
      
    return pathLength == otherPathLength && (pathLength == 0 || string.Compare (path, 0, otherPath, 0, otherPathLength, true, CultureInfo.InvariantCulture) == 0);
}

private static int GetTrailingSlashTrimmedPathLength(string path) {
    if (string.IsNullOrEmpty(path)) {
        return 0;
    } 
    if (path.EndsWithOrdinal("\\") || path.EndsWithOrdinal("/"))) {
        return path.Length - 1;
    }
    return path.Length;
}
```

---
In reply to: [77673244](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77673244) [](ancestors = 77673244)

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
AlexanderSher: Gaps? That's weird. Ok, let's leave it as it is

---
In reply to: [77672665](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77672665) [](ancestors = 77672665)

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: It is, the problem was with `DispatchOnUIThread` that causes message pump. I can plain call `AddItem` if you think ctor is only called on the main thread.

AlexanderSher: If the problem is in the `DispatchOnUIThread`  implementation, shouldn't we fix it in the `VsAppShell`? Maybe as a separate task/bug.

---
In reply to: [77699578](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77699578) [](ancestors = 77699578)

MikhailArkhipov: `DispatchOnUIThread` works as expected really, similar to `Dispatcher`. It is safe in 99% but caller does need to be aware that this is effectively a `PostMessage`

AlexanderSher: ok

---
In reply to: [77704011](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77704011) [](ancestors = 77704011)

src/Common/Core/Impl/Extensions/StringExtensions.cs
```diff
@@ -25,6 +25,9 @@ public static class StringExtensions {
         public static bool EndsWithOrdinal(this string s, string suffix) {
             return s.EndsWith(suffix, StringComparison.Ordinal);
         }
+        public static bool EndsWithOrdinal(this string s, char ch) {
```
AlexanderSher: Just `EndsWith` - we don't care about culture for a single char

src/Host/Client/Impl/Host/NullRHostConnector.cs
```diff
@@ -2,13 +2,16 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Common.Core;
 
 namespace Microsoft.R.Host.Client.Host {
     internal sealed class NullRHostConnector : IRHostConnector {
-        private static Task<RHost> Result { get; } = TaskUtilities.CreateCanceled<RHost>(new RHostDisconnectedException());
+        private static Task<RHost> Result { get; } = TaskUtilities.CreateCanceled<RHost>(
+            new RHostDisconnectedException(
+                string.Format(CultureInfo.InvariantCulture, Resources.NoConnectionsAvailable, Environment.NewLine, Environment.NewLine)));
```
AlexanderSher: Resources doesn't support line breaks?

MikhailArkhipov: Nope. When submitting test to MessageBox have to do this. `\n` comes as `\n' literal

AlexanderSher: We have `xml:space="preserve"` - so it should just work if you type the message on several lines.

---
In reply to: [77705742](https://github.com/Microsoft/RTVS/pull/2246#discussion_r77705742) [](ancestors = 77705742)

MikhailArkhipov: Lemme give it a shot. Didn't work well when I needed to space message in the ShowMessage

MikhailArkhipov: Ah, the problem is that it is resource editor that doesn't allow formatted text.

AlexanderSher:
:+1:

PR #2251, opened 2016-09-06
Prevent slow intellisense and hang in RMD publish

MikhailArkhipov:
When host is disconnect / broker connector is loopback

### DIFF
--- a/src/R/Support/Impl/Help/Packages/PackageIndex.cs
+++ b/src/R/Support/Impl/Help/Packages/PackageIndex.cs
@@ -158,14 +158,16 @@ namespace Microsoft.R.Support.Help.Packages {
         }
 
         private async Task<IPackageInfo> TryAddNewPackageAsync(string packageName) {
-            var packages = await GetPackagesAsync();
-            var package = packages.FirstOrDefault(p => p.Package.EqualsOrdinal(packageName));
-            if (package != null) {
-                var p = new PackageInfo(_host, package.Package, package.Description, package.Version);
-                await p.LoadFunctionsIndexAsync();
-                _packages[packageName] = p;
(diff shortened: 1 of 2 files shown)


src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -158,14 +158,16 @@ public sealed class PackageIndex : IPackageIndex {
...
         private async Task<IPackageInfo> TryAddNewPackageAsync(string packageName) {
-            var packages = await GetPackagesAsync();
-            var package = packages.FirstOrDefault(p => p.Package.EqualsOrdinal(packageName));
-            if (package != null) {
-                var p = new PackageInfo(_host, package.Package, package.Description, package.Version);
-                await p.LoadFunctionsIndexAsync();
-                _packages[packageName] = p;
-                _functionIndex.RegisterPackageFunctions(p);
-                return p;
+            if (_host.Session.IsHostRunning) {
```
AlexanderSher: It's better to use `try/catch`
```{csharp}
try {
    var packages = await GetPackagesAsync();
    var package = packages.FirstOrDefault(p => p.Package.EqualsOrdinal(packageName));
    if (package != null) {
        var p = new PackageInfo(_host, package.Package, package.Description, package.Version);
        await p.LoadFunctionsIndexAsync();
        _packages[packageName] = p;
        _functionIndex.RegisterPackageFunctions(p);
        return p;
    }
    return null;
}
```

MikhailArkhipov: OK

AlexanderSher:
:+1:

PR #2254, opened 2016-09-07
Fix enumeration of interpreters for non-local scenarios:

int19h:
- any interpreters that were manually specified in broker config are added to the list if they verify successfully;
- all auto-detected interpreters are also added to the list, using GUIDs as identifiers;
- if interpreter ID is not specified in an incoming connection, the first interpreter from the list is used.

Add interpreter names (for use in any future UI that would list all available interpreters for the broker when configuring connection).


int19h:
@MikhailArkhipov

MikhailArkhipov:
:+1:

PR #2260, opened 2016-09-07
Fix for environment variables for the logged on user

karthiknadig:
Fixes a bug where the rhost process does not receive the user environment variables (%USERPROFILE%, %USERNAME%, etc) when it is started with a different user identity than the broker process. #2234

### DIFF
--- a/src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
+++ b/src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
@@ -40,17 +40,17 @@ namespace Microsoft.R.Host.Broker.Interpreters {
 
         private IEnumerable<Interpreter> GetInterpreters() {
             if (_options.AutoDetect) {
-                _logger.LogTrace("Auto-detecting R ...");
+                _logger.LogTrace(Resources.Trace_AutoDetectingR);
 
                 var engines = new RInstallation().GetCompatibleEngines();
                 if (engines.Any()) {
                     foreach (var e in engines) {
                         var detected = new Interpreter(this, Guid.NewGuid().ToString(), e.Name, e.InstallPath, e.BinPath, e.Version);
-                        _logger.LogTrace($"R {detected.Version} detected at \"{detected.Path}\".");
+                        _logger.LogTrace(string.Format(Resources.Trace_DetectedR, detected.Version, detected.Path));
                         yield return detected;
                     }
                 } else {
-                    _logger.LogWarning("No R interpreters auto-detected.");
+                    _logger.LogWarning(Resources.Warning_NoRInterpreters);
                 }
             }
 
@@ -66,7 +66,7 @@ namespace Microsoft.R.Host.Broker.Interpreters {
                     }
                 }
 
-                _logger.LogError($"Failed to retrieve R installation data for interpreter \"{id}\" at \"{options.BasePath}\"");
+                _logger.LogError(string.Format(Resources.Error_FailedRInstallationData, id, options.BasePath));
             }
         }
     }
--- a/src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
+++ b/src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
@@ -46,7 +46,7 @@ namespace Microsoft.R.Host.Broker.Interpreters {
                 if (engines.Any()) {
                     foreach (var e in engines) {
                         var detected = new Interpreter(this, Guid.NewGuid().ToString(), e.Name, e.InstallPath, e.BinPath, e.Version);
-                        _logger.LogTrace(string.Format(Resources.Trace_DetectedR, detected.Version, detected.Path));
+                        _logger.LogTrace(Resources.Trace_DetectedR, detected.Version, detected.Path);
                         yield return detected;
                     }
                 } else {
@@ -66,7 +66,7 @@ namespace Microsoft.R.Host.Broker.Interpreters {
                     }
                 }
 
-                _logger.LogError(string.Format(Resources.Error_FailedRInstallationData, id, options.BasePath));
+                _logger.LogError(Resources.Error_FailedRInstallationData, id, options.BasePath);
             }
         }
     }
--- a/src/Host/Broker/Impl/NativeMethods.cs
+++ b/src/Host/Broker/Impl/NativeMethods.cs
@@ -213,7 +213,7 @@ namespace Microsoft.R.Host.Broker {
             out IntPtr phToken);
 
         [DllImport("credui.dll", CharSet = CharSet.Unicode)]
-        public static extern int CredUIParseUserName(
+        public static extern uint CredUIParseUserName(
                 string userName,
                 StringBuilder user,
                 int userMaxChars,
@@ -233,7 +233,5 @@ namespace Microsoft.R.Host.Broker {
             StringBuilder pszProfilePath,
             ref uint cchProfilePath);
 
-        [DllImport("kernel32.dll")]
-        public static extern uint GetLastError();
     }
 }
--- a/src/Host/Broker/Impl/NativeMethods.cs
+++ b/src/Host/Broker/Impl/NativeMethods.cs
@@ -83,6 +83,22 @@ namespace Microsoft.R.Host.Broker {
         LOGON32_PROVIDER_WINNT50 = 3
     }
 
+    [Flags]
+    public enum CreationFlags {
+        CREATE_SUSPENDED = 0x00000004,
+        CREATE_NEW_CONSOLE = 0x00000010,
+        CREATE_NEW_PROCESS_GROUP = 0x00000200,
+        CREATE_UNICODE_ENVIRONMENT = 0x00000400,
+        CREATE_SEPARATE_WOW_VDM = 0x00000800,
+        CREATE_DEFAULT_ERROR_MODE = 0x04000000,
+    }
+
+    [Flags]
+    public enum LogonFlags {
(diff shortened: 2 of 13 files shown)


src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -69,41 +69,49 @@ public class Session {
...
-                Arguments = $"--rhost-name \"{Id}\" {CommandLineArguments}",
+                Arguments = arguments,
                 RedirectStandardInput = true,
                 RedirectStandardOutput = true,
-                RedirectStandardError = true
+                RedirectStandardError = true,
+                LoadUserProfile = true
             };
 
+            if (WindowsIdentity.GetCurrent().User != ((WindowsIdentity)User).User) {
```
int19h: Does `((WindowsIdentity)User)` cast work in local scenarios when `--security:secret` is used? i.e. is it still a `WindowsIdentity` then?

karthiknadig: no this will not. I will update.

src/Host/Broker/Impl/NativeMethods.cs
```diff
@@ -232,5 +232,8 @@ internal static unsafe class NativeMethods {
             IntPtr hToken,
             StringBuilder pszProfilePath,
             ref uint cchProfilePath);
+
+        [DllImport("kernel32.dll")]
+        public static extern uint GetLastError();
```
int19h: This will not work reliably with P/Invoke (because it may invoke other Win32 APIs that do `SetLastError` as part of marshaling). You need to ensure that all functions have `[... SetLastError=true ...]`, and then use `Marshal.GetLastWin32Error`.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -58,24 +58,37 @@ public class SecurityManager {
             IntPtr token;
             WindowsIdentity winIdentity = null;
             string profilePath = "";
+
+            _logger.LogInformation($"Attempting login for user: {context.Username}");
```
int19h: Non-error logging here should probably use `LogTrace` instead of `LogInformation`. ASP.NET itself will log messages about authentication starting/failing at "info" level, and this is more detailed representation of that (and more than admins would normally need by default, probably), so it should go to the next verbosity level.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -58,24 +58,37 @@ public class SecurityManager {
...
+                _logger.LogInformation($"Attempting profile creation for user: {context.Username}");
                 uint error = NativeMethods.CreateProfile(winIdentity.User.Value, user.ToString(), profileDir, size);
                 // 0x800700b7 - Profile already exists.
                 if (error != 0 && error != 0x800700b7) {
+                    _logger.LogError($"Profile creation failed for user [{context.Username}] with error: {error.ToString("X")}");
                     return null;
+                } else if (error == 0x800700b7) {
+                    _logger.LogInformation($"Profile already exists for user: {context.Username}. Continuing with existing profile.");
+                } else {
+                    _logger.LogInformation($"Profile successfully created for user: {context.Username}");
```
int19h: This is one exception to the above - since this represents completion of an operation, and not just "I'm about to do something", this deserves to be `LogInformation`.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -58,24 +58,37 @@ public class SecurityManager {
...
+                _logger.LogInformation($"Login succeeded for user: {context.Username}");
                 winIdentity = new WindowsIdentity(token);
                 StringBuilder profileDir = new StringBuilder(NativeMethods.MAX_PATH);
                 uint size = (uint)profileDir.Capacity;
+
+                _logger.LogInformation($"Attempting profile creation for user: {context.Username}");
                 uint error = NativeMethods.CreateProfile(winIdentity.User.Value, user.ToString(), profileDir, size);
                 // 0x800700b7 - Profile already exists.
                 if (error != 0 && error != 0x800700b7) {
+                    _logger.LogError($"Profile creation failed for user [{context.Username}] with error: {error.ToString("X")}");
```
int19h: The messages should be consistent with respect to square brackets around username (either present everywhere, or absent everywhere).

Also, maybe single or double quotes would be a better fit? Or is [username] some sort of standard convention for Win32 usernames?

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -69,15 +69,17 @@ public class Session {
             CommandLineArguments = commandLineArguments;
         }
 
-        public void StartHost(SecureString password, string profilePath, ILogger outputLogger, ILogger messageLogger) {
+        public void StartHost(ILogger logger, SecureString password, string profilePath, ILogger outputLogger, ILogger messageLogger) {
```
int19h: It's better to pass this logger to constructor of `Session` instead, and save it in a field. The reason why the other two loggers here are passed to `StartHost` is because the stuff they log corresponds to one particular host process (its error output and its message pipe end, respectively). So the lifetime of those log writers also corresponds to that of the host, or at least the caller can make it so by passing new instances every time.

Login errors, however, pertain to the user and session, not the host per se (since they happen before the host process is even started). So the logger where they go belongs to the `Session` instance, and shared by other logic in this class.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -90,7 +92,9 @@ public class Session {
                 LoadUserProfile = true
             };
 
-            if (WindowsIdentity.GetCurrent().User != ((WindowsIdentity)User).User) {
+            var useridentity = User as WindowsIdentity;
+            if (useridentity != null && WindowsIdentity.GetCurrent().User != useridentity.User) {
+                logger.LogInformation($"Creating user environment variables for [{User.Name}] profile: {profilePath}");
                 // if broker and rhost are run as different users.
                 psi.EnvironmentVariables["USERNAME"] = username.ToString();
```
int19h: Might be worth logging the computed values as well, in case something goes awry, so that we can quickly check what they were in the logs.

Also, same thing about verbosity levels as above applies throughout this method - most everything that does not report an error should be `LogTrace`, except logging that indicate start/end of important operations, e.g. "Starting host process for session {Id} of user {Name}" at the beginning, and "Host process started successfully for session {Id} of user {Name}" after `_process.Start()`.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -129,6 +134,7 @@ public class Session {
                 _pipe = null;
             };
 
+            logger.LogInformation($"Starting RHost process: {rhostExePath} {arguments}");
```
int19h: Worth mentioning user and session ID here as well. Keep in mind that the log is written as a flat file, so with concurrent incoming requests, the messages will be interleaved. We'll need enough data to figure out which message belongs to which user+session (or just user, if it's something that's not really session-specific, like parsing username or setting up environment).

src/Host/Broker/Impl/Sessions/Session.cs
MikhailArkhipov: Please see #2269

src/Host/Broker/Impl/Security/SecurityManager.cs
MikhailArkhipov: #2269. Also, logging data should be in resources for localization

MikhailArkhipov: I can take #2269 for load balancing.

src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
```diff
@@ -40,17 +40,17 @@ public class InterpreterManager {
...
             if (_options.AutoDetect) {
-                _logger.LogTrace("Auto-detecting R ...");
+                _logger.LogTrace(Resources.Trace_AutoDetectingR);
 
                 var engines = new RInstallation().GetCompatibleEngines();
                 if (engines.Any()) {
                     foreach (var e in engines) {
                         var detected = new Interpreter(this, Guid.NewGuid().ToString(), e.Name, e.InstallPath, e.BinPath, e.Version);
-                        _logger.LogTrace($"R {detected.Version} detected at \"{detected.Path}\".");
+                        _logger.LogTrace(string.Format(Resources.Trace_DetectedR, detected.Version, detected.Path));
```
int19h: You don't need to use `string.Format` - they provide `params` overloads for `Log...` that do that for you. They have one other advantage, in that they also retain the format arguments as event data, so an advanced logger can log the values separately from the formatted message, for latter filtering on the values.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -61,25 +61,27 @@ public class Session {
             }
         }
 
-        internal Session(SessionManager manager, IIdentity user, string id, Interpreter interpreter, string commandLineArguments) {
+        internal Session(SessionManager manager, IIdentity user, string id, Interpreter interpreter, string commandLineArguments, ILogger<Session> sessionLogger) {
```
int19h: Just `ILogger`. The `<T>` thing has no semantic meaning, it's just a trick to make dependency injection create different loggers for you automatically. So if the component doesn't obtain its logger from dependency injection, it should just use `ILogger`.

int19h:
:+1: except for those two things above.

int19h:
:+1:

PR #2267, opened 2016-09-07
Connection UI fixes and tooltips

MikhailArkhipov:


src/Host/Process
```diff
@@ -1 +1 @@
-Subproject commit 991257326b5c0238881000c2160d1f7eebb20d8e
+Subproject commit ce944d91aff9aef525687c9febcf033b5cc33876
```
MikhailArkhipov: Fixing...

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
```diff
@@ -8,6 +8,10 @@
              xmlns:designTime="clr-namespace:Microsoft.R.Components.ConnectionManager.Implementation.View.DesignTime"
              xmlns:viewModel="clr-namespace:Microsoft.R.Components.ConnectionManager.ViewModel"
              xmlns:controls="clr-namespace:Microsoft.R.Wpf.Controls;assembly=Microsoft.R.Wpf"
+             xmlns:imaging="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.Imaging"
```
AlexanderSher: R.Components shouldn't reference any VS-specific assemblies

MikhailArkhipov: These are no shell dependent. Components already reference Text.* and CoreUtility. Same layer as R.Editor.

src/R/Interpreters/Impl/Implementation/RInterpreterInfo.cs
```diff
@@ -63,36 +63,28 @@ public sealed class RInterpreterInfo : IRInterpreterInfo {
...
                 }
-            } catch(IOException ioex) {
-                ex = ioex;
-            } catch (ArgumentException aex) {
-                ex = aex;
-
-            } catch(UnauthorizedAccessException uaex) {
-                ex = uaex;
-            }
+            } catch (IOException ioex) { ex = ioex; } catch (ArgumentException aex) { ex = aex; } catch (UnauthorizedAccessException uaex) { ex = uaex; }
```
AlexanderSher: maybe it will look better:
```{csharp}
} catch (Exception ex) when (ex is IOException || ex is ArgumentException || ex is UnauthorizedAccessException)
    coreShell?.ShowErrorMessage(
                    string.Format(CultureInfo.InvariantCulture, Resources.Error_ExceptionAccessingPath, InstallPath, ex.Message));
}
```
Up to you

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
AlexanderSher: Ok, got it. Microsoft.VisualStudio.Imaging references `stdole`, which is in `C:\Program Files (x86)\Common Files\Microsoft Shared\MSEnv\PublicAssemblies\stdole.dll` - is it ok?

---
In reply to: [77923956](https://github.com/Microsoft/RTVS/pull/2267#discussion_r77923956) [](ancestors = 77923956)

MikhailArkhipov: Not entirely sure, but probably OK as it is generic OLE stuff

src/R/Interpreters/Impl/Implementation/RInterpreterInfo.cs
MikhailArkhipov: Yeah, I like yours better

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
MikhailArkhipov: Since we are not building completely separate app, not referencing package-level things like *Interop and *Shell should be fine since tests will work without VS. If we were to build a separate app, we could look deeper.

src/R/Interpreters/Impl/Implementation/RInterpreterInfo.cs
```diff
@@ -55,44 +55,36 @@ public sealed class RInterpreterInfo : IRInterpreterInfo {
...
-            } catch (ArgumentException aex) {
-                ex = aex;
-
-            } catch(UnauthorizedAccessException uaex) {
-                ex = uaex;
-            }
+            } catch (Exception ex) when (ex is IOException || ex is ArgumentException || ex is UnauthorizedAccessException) { exception = ex; }
 
-            if(ex != null) {
+            if (exception != null) {
```
AlexanderSher: You don't need a local variable - `coreShell?.ShowErrorMessage` can be called directly from the catch block

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -100,6 +101,29 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
             set { SetProperty(ref _isConnected, value); }
         }
 
+        /// <summary>
+        /// Tooltip when hovered over connection name
+        /// </summary>
+        public string Information {
```
AlexanderSher: Have you verified that this property doesn't have to raise `PropertyChanged`?

AlexanderSher: Also, please add it to the interface

---
In reply to: [77930046](https://github.com/Microsoft/RTVS/pull/2267#discussion_r77930046) [](ancestors = 77930046)

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -100,6 +101,29 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
...
+            get {
+                if (IsRemote) {
+                    Uri uri;
+                    Uri.TryCreate(Path, UriKind.Absolute, out uri);
+
+                    return string.Format(CultureInfo.InvariantCulture, Resources.ConnectionManager_InformationTooltipFormatRemote,
+                        IsActive ? Resources.ConnectionManager_Connected : Resources.ConnectionManager_Disconnected,
+                        uri != null ? uri.Host : Resources.ConnectionManager_Unknown,
+                        uri != null ? uri.Port.ToString() : Resources.ConnectionManager_Default,
+                        !string.IsNullOrWhiteSpace(RCommandLineArguments) ? RCommandLineArguments : Resources.ConnectionManager_None);
```
AlexanderSher: This line is repeated in both cases - we can save it to a variable

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -40,8 +40,17 @@ public partial class ConnectionManagerControl : UserControl {
         }
 
         private void ButtonEdit_Click(object sender, RoutedEventArgs e) {
-            Model?.Edit(GetConnection(e));
-            ScrollEditedIntoView();
+            if (Model != null) {
+                var currentConnection = Model.EditedConnection;
+                if (Model.IsEditingNew || currentConnection != null) {
+                    Model.CancelEdit();
+                }
```
AlexanderSher: Isn't it covered in `ConnectionManagerViewModel.TryStartEditing`?

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml
```diff
@@ -30,9 +34,34 @@
...
+                    <DataTrigger Binding="{Binding Path=IsRemote}" Value="True">
+                        <Setter Property="ToolTip" Value="{x:Static components:Resources.ConnectionManager_EditRemoteTooltip}" />
+                    </DataTrigger>
+                </Style.Triggers>
+            </Style>
+
             <DataTemplate x:Key="EditConnectionDataTemplate" DataType="{x:Type viewModel:IConnectionViewModel}">
                 <AdornerDecorator>
-                    <Grid>
+                    <Grid Margin="0,6,0,8">
```
AlexanderSher: Do we need the upper margin?

MikhailArkhipov: When editing it better separates the editing block from connection entries above and below.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
MikhailArkhipov: OK. I think I might rename it to Tooltip since the purpose is formatted content. If we want to display information elsewhere, it will probably be built differently, not as a string. And if this is just for tooltip then it doesnt need changed event.

AlexanderSher: Ok

---
In reply to: [77935370](https://github.com/Microsoft/RTVS/pull/2267#discussion_r77935370) [](ancestors = 77935370)

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
MikhailArkhipov: Let me check. The request was to close the panel (and cancel editing) when user clicks the button again.

MikhailArkhipov: Updated `TryStartEditing` instead

AlexanderSher:
:+1:

PR #2268, opened 2016-09-08
Transactional switching between connections.

AlexanderSher:
Change doesn't include UI for the TestConnection - it will submitted in a separate PR after merge with #2267


src/Host/Client/Impl/Definitions/IRSessionProvider.cs
```diff
@@ -9,6 +9,7 @@
 namespace Microsoft.R.Host.Client {
     public interface IRSessionProvider : IDisposable {
         Uri BrokerUri { get; }
+        string BrokerName { get; }
```
MikhailArkhipov: Should we perhaps have IBroker and Uri and Name on it?

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -30,19 +30,22 @@ internal abstract class RHostConnector : IRHostConnector {
...
+
         public abstract Uri BrokerUri { get; }
 
         public bool IsRemote => !BrokerUri.IsFile;
 
-        protected RHostConnector(string interpreterId) {
+        protected RHostConnector(string name, string interpreterId) {
+            Name = name;
             _interpreterId = interpreterId;
-            _log = new LinesLog(FileLogWriter.InTempFolder("Microsoft.R.Host.BrokerConnector"));
```
MikhailArkhipov: nameof(type) or typeof().Name?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -73,9 +74,9 @@ internal sealed class RSession : IRSession, IRCallbacks {
             CanceledBeginInteractionTask = TaskUtilities.CreateCanceled<IRSessionInteraction>(new RHostDisconnectedException());
         }
 
-        public RSession(int id, IRHostConnector brokerConnector, Action onDispose) {
+        public RSession(int id, IRHostConnector hostConnector, Action onDispose) {
             Id = id;
-            BrokerConnector = brokerConnector;
+            RHostConnector = hostConnector;
             _onDispose = onDispose;
```
MikhailArkhipov: Connector sounds like proxy pattern, maybe we should name it as such at some point. Not urgent.

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -14,11 +14,24 @@
...
+        private readonly SemaphoreSlim _connectorSwitchLock = new SemaphoreSlim(1, 1);
+
+        private IRHostConnector _hostConnector;
         private int _sessionCounter;
 
-        private IRHostBrokerConnector BrokerConnector { get; } = new RHostBrokerConnector();
+        public bool IsRemote => _hostConnector.IsRemote;
+        public string Name => _hostConnector.Name;
+        public Uri BrokerUri => _hostConnector.BrokerUri;
+        public string BrokerName => _hostConnector.Name;
```
MikhailArkhipov: This combo of Name/Uri/IsRemote starts appearing often, perhaps an interface would be handy

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -61,14 +74,109 @@ public class RSessionProvider : IRSessionProvider, IRHostConnector {
...
+
+                var sessions = _sessions.Values.ToList();
+
+                if (sessions.Any()) {
+                    try {
+                        _callback.WriteConsole(Resources.RSessionProvider_StartConnectingToWorkspaceFormat.FormatInvariant(sessions.Count));
+                        await Task.WhenAll(sessions.Select(StartSwitchingBrokerAsync));
+                        _callback.WriteConsole(Resources.RSessionProvider_RestartingSessionsFormat.FormatInvariant(sessions.Count));
+                        await Task.WhenAll(sessions.Select(s => s.CompleteSwitchingBrokerAsync()));
+                    } catch (Exception) {
```
MikhailArkhipov: Log it?

src/Host/Client/Impl/Host/RHostConnector.cs
```diff
@@ -166,21 +173,23 @@ internal abstract class RHostConnector : IRHostConnector {
                 } catch (UnauthorizedAccessException) {
                     isValidCredentials = false;
                     continue;
-                } catch (Exception ex) when (ex is InvalidOperationException || ex is WebException || ex is ProtocolViolationException) {
+                } catch (Exception ex)
```
MikhailArkhipov: Probably need to log this

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: Should I add logging directly to the `WriteConsole`?

---
In reply to: [78040235](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78040235) [](ancestors = 78040235)

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionProviderCallback.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.R.Components.InteractiveWorkflow.Implementation {
+    public sealed class RSessionProviderCallback : IRSessionProviderCallback {
+        private readonly ICoreShell _coreShell;
+        private readonly Lazy<IRInteractiveWorkflow> _workflowLazy;
+
+        public async Task<IntPtr> GetApplicationWindowHandleAsync() {
+            await _coreShell.SwitchToMainThreadAsync();
+            return _coreShell.ApplicationWindowHandle;
+        }
```
MikhailArkhipov: Since window handle never changes we could change it to property and then prefetch and cache in VsAppShell. Won't have to be async then.

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
AlexanderSher: Ok, I'll rename IRHostConnector interface into IRBroker

---
In reply to: [78037959](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78037959) [](ancestors = 78037959)

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: I think we'll need separate connection log so user can look into it and send it to IT if there are problems. Other logs may be switched off (we'll need to provide a switch in RTM) or contain PI data not for viewing outside of the authorized group.

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
MikhailArkhipov: Maybe just IBroker - can implementation be used in Python at some point?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionProviderCallback.cs
AlexanderSher: It has to be async cause `GetApplicationWindowHandleAsync` may be called before VS is loaded. `SwitchToMainThreadAsync` ensures that UI thread is available so that `ApplicationWindowHandle` is correct

---
In reply to: [78041142](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78041142) [](ancestors = 78041142)

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
AlexanderSher: Ok. Or maybe IBrokerClient, cause it isn't a broker itself

---
In reply to: [78041694](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78041694) [](ancestors = 78041694)

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: Well, `RSessionProviderCallback.WriteConsole` is used only be `RSessionProvider`. 

---
In reply to: [78041545](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78041545) [](ancestors = 78041545)

MikhailArkhipov: Filed separate item https://github.com/Microsoft/RTVS/issues/2269.

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
MikhailArkhipov: Client sounds fine.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionProviderCallback.cs
MikhailArkhipov: OK

MikhailArkhipov:
:+1:

int19h:
:+1:

AlexanderSher:
@int19h @MikhailArkhipov Guys, please take a look at the changes since yesterday CR. I have found a whole bunch of crashes and hangs related to the concurrent calls of ConnectAsync and TrySwitchBrokerAsync, so I had to fix them before merging this PR

src/Common/Core/Impl/Threading/AsyncCountdownEvent.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public Task WaitAsync() => _mre.WaitAsync();
+
+        public void Signal() {
+            if (_count <= 0) {
+                throw new InvalidOperationException();
+            }
+
+            var count = Interlocked.Decrement(ref _count);
+            if (count < 0) {
```
MikhailArkhipov: Since above if checks for 0, do you expect concurrent Signal() may decrement counter between the check above and this decrement? Maybe it is safer to protect the entire method body?

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
```diff
@@ -11,39 +11,95 @@ namespace Microsoft.Common.Core.Threading {
     /// All other callers will wait until <see cref="Release"/> is called and then will get <see langword="true" />
     /// </summary>
     public class BinaryAsyncLock {
-        private TaskCompletionSource<bool> _tcs;
+        private static readonly Task<bool> FalseTask = Task.FromResult(false);
+        private static readonly Task<bool> TrueTask = Task.FromResult(true);
+        private readonly Queue<TaskCompletionSource<bool>> _queue = new Queue<TaskCompletionSource<bool>>();
```
MikhailArkhipov: ConcurrentQueue didn't work?

AlexanderSher: Too much has to be done as an atomic operation. The main problem is synchronizing `_isCompleted` with `_queue` operations. For example, concurrent calls of `Release` and `Reset` can bring the lock into state when it is considered released, but some of the tasks from queue are completed with `false` result

---
In reply to: [78201950](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78201950) [](ancestors = 78201950)

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -166,21 +174,23 @@ internal abstract class RHostConnector : IRHostConnector {
                 } catch (UnauthorizedAccessException) {
                     isValidCredentials = false;
                     continue;
-                } catch (Exception ex) when (ex is InvalidOperationException || ex is WebException || ex is ProtocolViolationException) {
+                } catch (Exception ex)
+                    when (ex is InvalidOperationException || ex is WebException || ex is ProtocolViolationException) {
                     throw new RHostDisconnectedException("HTTP error while connecting to session pipe: " + ex.Message, ex);
```
MikhailArkhipov: Error should be localizable

src/Common/Core/Impl/Threading/BinaryAsyncLock.cs
AlexanderSher: To minimize the impact, I've left only lightweight operations under locks (creation of `new TaskCompletionSource<bool>()`, reading `Task.IsCompleted` and queue operations)

---
In reply to: [78202661](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78202661) [](ancestors = 78202661,78201950)

src/Common/Core/Impl/Threading/AsyncCountdownEvent.cs
AlexanderSher: Not sure that I understand the question, but there two checks are actually a protection. If 2 signals are called concurrently when `_count == 1`, after decrement one of them will fail on `count < 0` check

---
In reply to: [78201558](https://github.com/Microsoft/RTVS/pull/2268#discussion_r78201558) [](ancestors = 78201558)

MikhailArkhipov:
👍

src/Common/Core/Impl/Threading/AsyncCountdownEvent.cs
MikhailArkhipov: So concurrent calls to Signal () are expected. OK.

PR #2273, opened 2016-09-08
Remote setup MSI

MikhailArkhipov:
Bundle is coming next in RTVS-pr


MikhailArkhipov:
Since this is mostly just setup authoring, merging as I need to try and build this in the lab.

Issue #2281, opened 2016-09-09
Switching from local to remote connection disconnects


int19h:
RTVS launched with the default local session, then told to switch to remote:
```
Switching R Workspace completed
Start switching to the 'pmin-vm-w7-vs14': http://pmin-vm-w7-vs14:5000/
Connecting sessions (2) to the new R Workspace...
R Host process is stopped. Click Reset to start a new one.Restarting sessions (2)...
> 1+2
Interactive Window is disconnected from R Session.
Try connecting to a remote machine in the Workspaces window or
click Reset to restart local R interpreter.Starting R Session...
```
To fix this, I had to Reset - that reconnected and worked.

Also, the message is misleading: "click Reset to restart local R interpreter" - it actually restarts the active connection, local or remote.

MikhailArkhipov:
Probably same as #2217

AlexanderSher:
Ok, I have a repro for this one. It happens when you double-click on the connection icon - it initiates connection to the same broker with the same parameters which isn't handled.
That kills the RHost and may leave one of the locks in the invalid state.

Issue #2217, opened 2016-08-25
Unable to recover and switch to local


MikhailArkhipov:
Killed VS via Debug | Terminate. Started anew. REPL no longer works.

![image](https://cloud.githubusercontent.com/assets/12820357/17989467/535dfa38-6ae2-11e6-8b04-1b1f6eda606f.png)

MikhailArkhipov:
#2214

MikhailArkhipov:
Still broken. Switching back throws

```
            if (Interlocked.CompareExchange(ref _initializationTcs, new TaskCompletionSourceEx<object>(), null) != null) {
                throw new InvalidOperationException("Another instance of RHost is running for this RSession. Stop it before starting new one.");
            }
```

need to dig around more.

MikhailArkhipov:
Hopefully gets fixed by the transactions

MikhailArkhipov:
However, this reproes only after 'Test connection' to a invalid remote. Attempt to switch to local fails, but actually the new setting is saved so restarting VS recovers.

AlexanderSher:
I've fixed a batch of issues related to the R host switching, but can you give exact steps for the last repro?

MikhailArkhipov:
1. connect to local
2. Create broken remote connection such as enter 'foo' in path.
3. Save and try connecting

Now you are broken and can't connect back to local (connect button does nothing):

![image](https://cloud.githubusercontent.com/assets/12820357/18321192/cce31240-74e1-11e6-8916-5b11012a765d.png)

MikhailArkhipov:
Or rather, it doesn't work first time. if you click few times more it finally connects

PR #2293, opened 2016-09-09
Fix couple of bugs

AlexanderSher:
#2217: Unable to recover and switch to local
#2281: Switching from local to remote connection disconnects


int19h:
:+1:

Issue #2106, opened 2016-07-20
Redirect HTTP traffic: Shiny apps


karthiknadig:
The shiny package by default uses loop-back interface. However, the package can be configured to generate and serve non-localhost URL using shiny.host and shiny.port options.

karthiknadig:
Got this working. Need to do some refactoring for stability and debugging.

PR #2294, opened 2016-09-12
Remote Shiny

karthiknadig:
Implements feature #2106

### DIFF
--- a/src/Common/Core/Impl/Extensions/ArraySegmentExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/ArraySegmentExtensions.cs
@@ -17,5 +17,15 @@ namespace Microsoft.Common.Core {
             }
             return data;
         }
+
+        public static async Task<byte[]> ToByteArrayAsync(this ArraySegment<byte> source) {
+            byte[] data = null;
+            using (MemoryStream ms = new MemoryStream()) {
+                await ms.WriteAsync(source.Array, source.Offset, source.Count);
+                await ms.FlushAsync();
+                data = ms.ToArray();
+            }
+            return data;
+        }
     }
 }
--- a/src/Common/Core/Impl/Extensions/ArraySegmentExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/ArraySegmentExtensions.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core {
+    public static class ArraySegmentExtensions {
+        public static async Task<byte[]> ToByteArrayAsync(this ArraySegment<byte> source, int count) {
+            byte[] data = null;
+            using (MemoryStream ms = new MemoryStream()) {
+                await ms.WriteAsync(source.Array, source.Offset, count);
+                await ms.FlushAsync();
+                data = ms.ToArray();
+            }
+            return data;
+        }
+    }
+}
--- a/src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
+++ b/src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
@@ -9,33 +9,86 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Common.Core;
 using Microsoft.R.Host.Protocol;
+using Microsoft.AspNetCore.WebSockets.Protocol;
+using System.Net.WebSockets;
+using System.Threading;
 
 namespace Microsoft.R.Host.Broker.RemoteUri {
     public class RemoteUriHelper {
+
+        private static async Task DoWebSocketReceiveSendAsync(WebSocket receiver, WebSocket sender, CancellationToken ct) {
+            if (receiver == null || receiver == null) {
+                return;
+            }
+
+            ArraySegment<byte> receiveBuffer = WebSocket.CreateServerBuffer(65335);
+            while (receiver.State == WebSocketState.Open && sender.State == WebSocketState.Open) {
+                if (ct.IsCancellationRequested) {
+                    Task.WhenAll(
+                        receiver.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", CancellationToken.None),
+                        sender.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", CancellationToken.None))
+                        .DoNotWait();
+                    return;
+                }
+
+                WebSocketReceiveResult result = await receiver.ReceiveAsync(receiveBuffer, ct);
+
+                byte[] data = null;
+                using (MemoryStream ms = new MemoryStream()) {
+                    await ms.WriteAsync(receiveBuffer.Array, receiveBuffer.Offset, result.Count);
+                    await ms.FlushAsync();
+                    data = ms.ToArray();
+                }
+
+                ArraySegment<byte> sendBuffer = new ArraySegment<byte>(data);
+                await sender.SendAsync(sendBuffer, result.MessageType, result.EndOfMessage, ct);
+
+                if (result.MessageType == WebSocketMessageType.Close) {
+                    await receiver.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", ct);
+                }
+            }
+        }
+
+        private static Task DoWebSocketWorkAsync(WebSocket localWebSocket, WebSocket remoteWebSocket, CancellationToken ct) {
+            return Task.WhenAll(
+                DoWebSocketReceiveSendAsync(localWebSocket, remoteWebSocket, ct),
+                DoWebSocketReceiveSendAsync(remoteWebSocket, localWebSocket, ct));
+        }
+
         public static async Task HandlerAsync(HttpContext context) {
             var url = context.Request.Headers[CustomHttpHeaders.RTVSRequestedURL];
 
-            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url);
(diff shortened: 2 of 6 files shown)


AlexanderSher:
>             if (headers.ContainsKey("Connection")) {

`TryGetValue`

---
Refers to: src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs:133 in 96061f5. [](commit_id = 96061f5f7f937b6bfdfe4f7b660bd0dd955110c0, deletion_comment = False)

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -6,43 +6,104 @@
...
 
+        private async Task DoWebSocketReceiveSendAsync(WebSocket receiver, WebSocket sender, CancellationToken ct) {
+            if (receiver == null || receiver == null) {
+                return;
+            }
+
+            ArraySegment<byte> receiveBuffer = WebSocket.CreateServerBuffer(65335);
+            while (receiver.State == WebSocketState.Open && sender.State == WebSocketState.Open) {
+                if (ct.IsCancellationRequested) {
+                    Task.WhenAll(
```
AlexanderSher: What do you want to achieve by calling `DoNotWait` for `WhenAll`?

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
+                UriBuilder ub = new UriBuilder(url) { Scheme = "ws" };
+                clientWebsocket = new ClientWebSocket();
+                await clientWebsocket.ConnectAsync(ub.Uri, CancellationToken.None);
+                var serverWebSocket = await context.WebSockets.AcceptWebSocketAsync(clientWebsocket.SubProtocol);
+                await DoWebSocketWorkAsync(serverWebSocket, clientWebsocket, CancellationToken.None);
+            } else {
+                SetRequestHeaders(request, context.Request.Headers);
+
+                if (context.Request.ContentLength > 0) {
+                    using (Stream reqStream = await request.GetRequestStreamAsync()) {
```
MikhailArkhipov: These 3 lines look like useful extension

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
+                    Task.WhenAll(
+                        receiver.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", CancellationToken.None),
+                        sender.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", CancellationToken.None))
+                        .DoNotWait();
+                    return;
+                }
+
+                WebSocketReceiveResult result = await receiver.ReceiveAsync(receiveBuffer, ct);
+
+                byte[] data = null;
```
MikhailArkhipov: looks like extension method, something like ToByArray()

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Common.Core;
 using Microsoft.R.Host.Protocol;
 
 namespace Microsoft.R.Host.Broker.RemoteUri {
     public class RemoteUriHelper {
+
+        private static async Task DoWebSocketReceiveSendAsync(WebSocket receiver, WebSocket sender, CancellationToken ct) {
+            if (receiver == null || receiver == null) {
```
int19h: Duplicate check. Did you mean `sender`?

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
+                return;
+            }
+
+            ArraySegment<byte> receiveBuffer = WebSocket.CreateServerBuffer(65335);
```
int19h: I'm not entirely sure what this method does, but I am pretty sure that it's not really necessary - you can just `new` up an array and wrap it in an array segment. I'm worried that going through the method might result in a buffer shared with something else, and potential races.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
+            while (receiver.State == WebSocketState.Open && sender.State == WebSocketState.Open) {
+                if (ct.IsCancellationRequested) {
+                    Task.WhenAll(
+                        receiver.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", CancellationToken.None),
```
int19h: There's no point in using `Task.WhenAll` here, since you're not actually waiting on it. On the other hand, if `CloseAsync` might throw, you need to `SilenceException` those.

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
karthiknadig: WhenAll is not needed. I will remote it.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
+                        sender.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", CancellationToken.None))
+                        .DoNotWait();
+                    return;
+                }
+
+                WebSocketReceiveResult result = await receiver.ReceiveAsync(receiveBuffer, ct);
+
+                byte[] data = null;
+                using (MemoryStream ms = new MemoryStream()) {
```
int19h: This shouldn't be necessary. `ArraySegment` provides a view into a slice of an array. You can just create a new `ArraySegment` with the desired offset/count, and pass that to `SendAsync`.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -5,37 +5,88 @@
...
+                    await ms.WriteAsync(receiveBuffer.Array, receiveBuffer.Offset, result.Count);
+                    await ms.FlushAsync();
+                    data = ms.ToArray();
+                }
+
+                ArraySegment<byte> sendBuffer = new ArraySegment<byte>(data);
+                await sender.SendAsync(sendBuffer, result.MessageType, result.EndOfMessage, ct);
+
+                if (result.MessageType == WebSocketMessageType.Close) {
+                    await receiver.CloseAsync(WebSocketCloseStatus.NormalClosure, "Normal closure", ct);
```
int19h: I believe this is supposed to be an empty string normally - message is for errors and such.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
int19h: It should be very uncommon to work with request stream directly - normally you'd have a controller and let MVC take care of JSON serialization.

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -6,43 +6,104 @@
...
+
+            if (!context.Request.IsWebSocketRequest) {
+                SetRequestHeaders(request, context.Request.Headers, localBaseUrl, remoteBaseUrl);
+            }
 
             // Add RTVS headers
             request.Headers.Add(CustomHttpHeaders.RTVSRequestedURL, GetRemoteUrl(context.Request.Url, remoteBaseUrl));
 
-            if(context.Request.InputStream.CanSeek && context.Request.InputStream.Length > 0) {
+            if (context.Request.InputStream.CanSeek && context.Request.InputStream.Length > 0) {
```
int19h: Kinda surprised that `CanSeek` is true here - I would normally expect it to not be for a request, since it's just reading from a socket ultimately.

Why is it needed? It doesn't look like there's any seeking below...

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
karthiknadig: yes. that should be sender.

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
karthiknadig: InputStream.Length requires Seek

int19h: Come to think of it, is `Length` needed, either? The read-write loop should do the right thing even if it's zero, and without the check, it'll still work on non-seekable streams (assuming there are some circumstances in which they may appear).

src/Common/Core/Impl/Extensions/ArraySegmentExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core {
+    public static class ArraySegmentExtensions {
+        public static async Task<byte[]> ToByteArrayAsync(this ArraySegment<byte> source, int count) {
```
int19h: `ArraySegment` has its own count - if this is a generic helper now, it should probably offer an overload that defaults `count` to that.

int19h:
:+1:

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
karthiknadig: It adds complications. For example, during "GET" if i request for the stream it will throw an exception.

PR #2295, opened 2016-09-12
Allow RSession to be restarted automatically after broker switch even if it failed to start with previous broker

AlexanderSher:


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -289,17 +289,20 @@ internal sealed class RSession : IRSession, IRCallbacks {
             var host = _host;
             var hostRunTask = _hostRunTask;
 
-            // Detach RHost from RSession
-            host.DetachCallback();
+            // host may be null if previous attempts to start it have failed
+            if (host != null) { 
+                // Detach RHost from RSession
+                host?.DetachCallback();
```
MikhailArkhipov: already checked for null

int19h:
:+1:

MikhailArkhipov:
👍

PR #2296, opened 2016-09-12
Microsoft R Client detection

MikhailArkhipov:
- Look into new MRC locations under R\R-Core
- Separate MRC from SQL from general list
- Detect MRC under R-Core
- Remove duplicates since MRC registers same path under multiple registry keys
- Format names consistently
- Add test for dupes

Also fixes null ref when accessing dte.MainWindow too early. Since some commands now want status bar access, menu command creation has to be done on idle when IDE startup completes.


src/Package/Impl/Commands/R/InstallRClientCommand.cs
```diff
@@ -20,7 +19,7 @@ internal sealed class InstallRClientCommand : PackageCommand {
         }
 
         protected override void Handle() {
-            var path = MicrosoftRClient.GetRClientPath();
+            var path = SqlRClientInstallation.GetRClientPath();
```
AlexanderSher: They are changing the name again?

src/Package/Impl/Packages/BasePackage.cs
```diff
@@ -57,20 +58,22 @@ public abstract class BasePackage<TLanguageService> : VisualStudio.Shell.Package
                 RegisterEditorFactory(editorFactory);
             }
 
-            var menuCommandService = (IMenuCommandService)GetService(typeof(IMenuCommandService));
-            foreach (var commmand in CreateMenuCommands()) {
-                menuCommandService.AddCommand(commmand);
-            }
+            IdleTimeAction.Create(() => {
```
AlexanderSher: Why do we need this?

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -64,6 +64,12 @@ internal class ConnectionManager : IConnectionManager {
             _userConnections = new ConcurrentDictionary<Uri, IConnection>(userConnections);
 
             UpdateRecentConnections();
+
+            _shell.Idle += OnIdle;
+        }
+
+        private void OnIdle(object sender, EventArgs e) {
```
AlexanderSher: `_shell.SwitchToMainThreadAsync` should handle this. Do you have a repro?

src/R/Interpreters/Impl/Implementation/SqlRClientInstallation.cs
```diff
@@ -0,0 +1,76 @@
...
+                        if (!string.IsNullOrEmpty(path) && path.Contains(_rServer)) {
+                            mrsInstalled = true;
+                        }
+                    }
+                }
+            } catch (Exception) { }
+
+            // If yes, check 32-bit registry for R engine installed by the R Server.
+            // TODO: remove this when MRS starts writing 64-bit keys.
+            if (mrsInstalled) {
```
AlexanderSher: Maybe extract this whole `if` into a separate method?

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: As in every F5 I do it hits null ref exception accessing DTE.MainWindow. This is b/c connection manager is up and commands get created by the package and then want interactive workflow which create connection manager which wants to populate status bar.

MikhailArkhipov: BTW, main thread does exist and can be reentered when DTE is still not ready. Perhaps instead of using DTE get dialog owner from IVsUiShell ?

AlexanderSher: Well, `DTEEvents.OnStartupComplete` happens only after status bar container is created and it is on main thread. 
I've just tried to repro it and it worked as expected. Do I need your changes to repro the issue?

---
In reply to: [78463049](https://github.com/Microsoft/RTVS/pull/2296#discussion_r78463049) [](ancestors = 78463049)

MikhailArkhipov: Advise to OnStartupComplete happens when GetRootVisual return null and that null ref happens in GetRootVisual (handled, but annoying - breaks into debugger all the time)

src/Package/Impl/Commands/R/InstallRClientCommand.cs
MikhailArkhipov: No, I just wanted to shorten the name from MicrosoftSqlRClientInstallation :-) And yes, there is one that comes with SQL (maybe I should add 'SQL' to its name in the connections) and one installed separately.

src/R/Interpreters/Impl/Implementation/SqlRClientInstallation.cs
MikhailArkhipov: Done

src/Package/Impl/Packages/BasePackage.cs
MikhailArkhipov: B/c of DTE.MainWindow access from commands (via interactive workflow -> connection manager)

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: It looks like you have setup debugger to stop on NRE (which isn't a default). But we can check `Microsoft.VisualStudio.PlatformUI.Shell.ViewManager.Instance.WindowProfile` for null instead. 

---
In reply to: [78464734](https://github.com/Microsoft/RTVS/pull/2296#discussion_r78464734) [](ancestors = 78464734)

MikhailArkhipov: Is it a documented thing?

AlexanderSher: No, it just public. The other option is to create a lightweight package that will always load and use it to verify if VS startup is completed or not. I've sent a PR

---
In reply to: [78469414](https://github.com/Microsoft/RTVS/pull/2296#discussion_r78469414) [](ancestors = 78469414)

MikhailArkhipov:
Changed DTE to IVSUIShell

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: Fixed via IVsUIShell

AlexanderSher: This may not be enough. It is possible that root controls aren't added yet, so TryAddItemsControlToVisualRoot won't find a parent panel.

---
In reply to: [78587125](https://github.com/Microsoft/RTVS/pull/2296#discussion_r78587125) [](ancestors = 78587125)

MikhailArkhipov: Then idle should work since it is fired by OLE infrastructure when app is fully initialize (comes from MSO code, formerly from Office). Idle is simpler than a separate package anyway.

AlexanderSher: Ok, then I'll merge your changes and add an idle verification inside VsStatusBar itself

---
In reply to: [78588775](https://github.com/Microsoft/RTVS/pull/2296#discussion_r78588775) [](ancestors = 78588775)

AlexanderSher:
:+1:

Issue #2291, opened 2016-09-09
Crash after switching to a different remote


huguesv:
In plot manager:
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentException
   at Newtonsoft.Json.Linq.JToken.op_Explicit(Newtonsoft.Json.Linq.JToken)
   at Newtonsoft.Json.Linq.JToken.ToObject(System.Type)
   at Newtonsoft.Json.Linq.JToken.ToObject[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]()
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions+<EvaluateAsync>d__1`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Components.Plots.Implementation.RPlotManager+<RefreshDeviceNum>d__59.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Components.Plots.Implementation.RPlotManager+<RSessionMutatedAsync>d__58.MoveNext()
   at Microsoft.Common.Core.TaskExtensions+<>c.<DoNotWaitThreadContinuation>b__6_0(System.Object)
   at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
[...]

PR #2299, opened 2016-09-13
Prevent json deserialization error if device id is not found.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/2291


int19h:
It feels like it might be clearer to return `NULL` (which should translate to JSON `null`), and use a nullable integer on the other side for deserialization.

huguesv:
Done!

int19h:
:+1:

PR #2301, opened 2016-09-13
Fix initial size of plot device which needs to be correctly from wpf …

huguesv:
…size to pixels.

Fixes https://github.com/Microsoft/RTVS/issues/2298


karthiknadig:
:shipit:

Issue #2297, opened 2016-09-12
RSessions are leaking


AlexanderSher:
- Open Package Manager, switch several times between tabs, install/uninstall packages
- Switch connection
```
Start switching to the 'asher-lt': http://192.168.1.72:5000/
Connecting sessions (10) to the new R Workspace...
```

PR #2305, opened 2016-09-13
Fix bug #2297

AlexanderSher:


src/Package/Impl/StatusBar/VsStatusBar.cs
```diff
@@ -21,49 +21,32 @@ internal class VsStatusBar : IStatusBar {
...
-        private EnvDTE.DTEEvents _dteEvents;
 
         [ImportingConstructor]
         public VsStatusBar(IApplicationShell shell) {
             _shell = shell;
         }
 
-        private Visual GetRootVisual() {
+        private void UpdateRootVisual() {
             var dte = _shell.GetGlobalService<EnvDTE.DTE>();
```
MikhailArkhipov: I believe this is no longer needed, IVsUIShell.DialogOwnerWindow works just fine (it does return top level HWND)

src/Package/Impl/Packages/RtvsStartupPackage.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Runtime.InteropServices;
+using System.Threading;
+using EnvDTE;
+using Microsoft.VisualStudio.R.Packages.R;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.R.Package {
+    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [Guid(RGuidList.RtvsStartupPackageGuidString)]
+    [InstalledProductRegistration("#7002", "#7003", RtvsProductInfo.VersionString, IconResourceID = 400)]
```
MikhailArkhipov: VS Performance will hunt us down for autoloads...

PR #2307, opened 2016-09-13
Several fixes

### LINKED ISSUE #2297: RSessions are leaking

AlexanderSher:
-Reuse existing extension method
- Fix #2297: RSessions are leaking
- Fix VsStatusBar so it tries to add items to visual root on idle if VS visual tree isn't ready


MikhailArkhipov:
👍

Issue #2310, opened 2016-09-13
Connection manager tries to pass remote URL as local for installation check


MikhailArkhipov:
When last used connection was remote, at start CM passes URL down to `RInterpreterInfo` yielding argument exception in 
``
        public IFileVersionInfo GetVersionInfo(string path) {
            var fvi = System.Diagnostics.FileVersionInfo.GetVersionInfo(path);
            return new FileVersionInfo(fvi.FileMajorPart, fvi.FileMinorPart);
        }
```
at

```
 	Microsoft.R.Common.Core.dll!Microsoft.Common.Core.IO.FileSystem.GetVersionInfo(string path) Line 51	C#
 	Microsoft.R.Interpreters.dll!Microsoft.R.Interpreters.RInterpreterInfo.GetRVersionFromBinary(Microsoft.Common.Core.IO.IFileSystem fs, string basePath) Line 90	C#
 	Microsoft.R.Interpreters.dll!Microsoft.R.Interpreters.RInterpreterInfo.DetermineVersion(Microsoft.Common.Core.IO.IFileSystem fs) Line 133	C#
 	Microsoft.R.Interpreters.dll!Microsoft.R.Interpreters.RInterpreterInfo.RInterpreterInfo(string name, string path, Microsoft.Common.Core.IO.IFileSystem fs) Line 39	C#
>	Microsoft.R.Components.dll!Microsoft.R.Components.ConnectionManager.Implementation.ConnectionManager.IsValidLocalConnection(string name, string path) Line 217	C#
 	Microsoft.R.Components.dll!Microsoft.R.Components.ConnectionManager.Implementation.ConnectionManager.CreateConnectionList() Line 192	C#
 	Microsoft.R.Components.dll!Microsoft.R.Components.ConnectionManager.Implementation.ConnectionManager.ConnectionManager(Microsoft.R.Components.StatusBar.IStatusBar statusBar, Microsoft.R.Components.Settings.IRSettings settings, Microsoft.R.Components.InteractiveWorkflow.IRInteractiveWorkflow interactiveWorkflow) Line 63	C#
 	Microsoft.R.Components.dll!Microsoft.R.Components.ConnectionManager.Implementation.ConnectionManagerProvider.CreateConnectionManager(Microsoft.R.Components.InteractiveWorkflow.IRInteractiveWorkflow interactiveWorkflow) Line 22	C#
[...]

Issue #2284, opened 2016-09-09
Remoteness should be displayed visibly at all times


smortaz:
for example on tool window names:

Remote - PavelRemote:5000

R Pkg Manager - Fatbox:4000

etc.

int19h:
Users likely won't care about anything other than hostname for this purpose, so we can skip the port (technically it's possible to have two different brokers on the same hostname, but realistically this isn't likely). Or better yet, use connection name.

Also, "X - Y" format is used for document windows in MDI (i.e. "foo.txt - Notepad"), so we should probably use something else. E.g. "R Package Manager (Remote1)".

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5dea54595b7ca024c580e38dd8e1ca76d9d1071d

PR #2311, opened 2016-09-13
#2284, #2310

MikhailArkhipov:
#2284 Remoteness should be displayed visibly at all times
#2310 Connection manager tries to pass remote URL as local


src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -214,8 +214,11 @@ internal class ConnectionManager : IConnectionManager {
 
         private bool IsValidLocalConnection(string name, string path) {
             try {
-                var info = new RInterpreterInfo(name, path);
-                return info.VerifyInstallation();
+                Uri uri;
+                if (Uri.TryCreate(path, UriKind.Absolute, out uri) && uri.IsFile) {
```
AlexanderSher: So if this method returns `false` for remote connection, `LastActiveConnection` will be set to null and user will be switched from remote connection to the first available local. Is it a desired behavior?

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
```diff
@@ -14,39 +15,67 @@
...
 using Microsoft.VisualStudio.Utilities;
+using static System.FormattableString;
 
 namespace Microsoft.VisualStudio.R.Package.Repl {
     [Export(typeof(IInteractiveWindowComponentContainerFactory))]
     internal class VsRInteractiveWindowComponentContainerFactory : IInteractiveWindowComponentContainerFactory {
         private readonly Lazy<IVsInteractiveWindowFactory> _vsInteractiveWindowFactoryLazy;
         private readonly IContentTypeRegistryService _contentTypeRegistryService;
+        private IRInteractiveWorkflow _interactiveWorkflow;
+        private IVsWindowFrame _frame;
```
int19h: Is the factory supposed to be mapped 1:1 to the entities that it is creating? It looks like you can only call `Create` once on this if you want it to work correctly.

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
```diff
@@ -14,39 +15,67 @@
...
             _contentTypeRegistryService = contentTypeRegistryService;
+
         }
 
-        public IInteractiveWindowVisualComponent Create(int instanceId, IInteractiveEvaluator evaluator) {
+        public IInteractiveWindowVisualComponent Create(int instanceId, IInteractiveEvaluator evaluator, IRInteractiveWorkflow interactiveWorkflow) {
             VsAppShell.Current.AssertIsOnMainThread();
 
+            _interactiveWorkflow = interactiveWorkflow;
+            _interactiveWorkflow.RSessions.BrokerChanged += OnBrokerChanged;
```
AlexanderSher: I think we can just save the reference to `interactiveWorkflow.RSessions`

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
```diff
@@ -14,39 +15,67 @@
...
+
+            UpdateWindowTitle();
 
             var componentContainer = new VisualComponentToolWindowAdapter<IInteractiveWindowVisualComponent>(toolWindow);
             var component = new RInteractiveWindowVisualComponent(vsWindow.InteractiveWindow, componentContainer);
             componentContainer.Component = component;
             return component;
         }
+
+        private void OnBrokerChanged(object sender, EventArgs e) {
```
AlexanderSher: I'd move the handler to the `RInteractiveWindowVisualComponent` itself, and unsubscribe it when visual component is disposed.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: It is called `IsValidLocalConnection` after all... It didn't work for remote connections anyways. Currently CM does not restore remote connection - which should probably be filed as a separate bug.

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
MikhailArkhipov: Probably not (at least not eventually with multiple REPLs). Let me see what I can do.

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
```diff
@@ -14,39 +15,67 @@
...
+            UpdateWindowTitle();
+        }
+
+        private void UpdateWindowTitle() {
+            var broker = _interactiveWorkflow.RSessions.Broker;
+            string title;
+            if (broker != null) {
+                title = broker.IsRemote
+                    ? Invariant($"{Resources.ReplWindowName} - {broker.Name} ({broker.Uri.ToString().TrimTrailingSlash()})")
```
int19h: This makes it look like an MDI parent window, or VS top level window itself (the "document - app" form). I think we need some other format to clearly disambiguate. `@` is traditionally used to indicate a network context - as in "Package Manager @ Remote1" etc.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: Or should we just remove `IsValidLocalConnection` check for the MRU connection?

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
MikhailArkhipov: It is not something unseen before... 

image

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: Let's file another bug then

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
MikhailArkhipov: OK, separated frame operations and moved to the component

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
MikhailArkhipov: Done

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: I fixed it, it now restores connection. But there is no indication in the UI that something is going on for ~5-10 seconds especially when REPL is closed.

AlexanderSher: I'm not sure how good it is to show progress bar right on top of the VS that has just started. We can have an indication on a status bar.

AlexanderSher:
:+1:

int19h:
:+1:

Issue #2324, opened 2016-09-13
RTools Options page not working


huguesv:
RTools -> Options

Result:

```
An error occurred loading this property page
```

![image](https://cloud.githubusercontent.com/assets/1696845/18495184/af505e60-79d0-11e6-845b-d25007921a1d.png)

MikhailArkhipov:
Not sure if related but despite I have MRC and R 3.3.1

```
Exception thrown: 'Microsoft.R.Host.Client.Host.RHostDisconnectedException' in mscorlib.dll
Exception thrown: 'Microsoft.R.Host.Client.Host.RHostDisconnectedException' in mscorlib.dll
Exception thrown: 'Microsoft.R.Host.Client.Host.RHostDisconnectedException' in Microsoft.R.Host.Client.dll
Exception thrown: 'Microsoft.R.Host.Client.Host.RHostDisconnectedException' in mscorlib.dll
Exception thrown: 'Microsoft.R.Host.Client.Host.RHostDisconnectedException' in mscorlib.dll
Exception thrown: 'Microsoft.R.Host.Client.Host.RHostDisconnectedException' in mscorlib.dll
There is no local R detected and no remote connection established.
You can try re-connecting to the remote machine, 
use another connection or install local R.
```

and

```
Exception thrown: 'Microsoft.R.Host.Client.REvaluationException' in Microsoft.R.Host.Client.dll
Exception thrown: 'Microsoft.R.Host.Client.REvaluationException' in mscorlib.dll
```

MikhailArkhipov:
Fixed

Issue #2316, opened 2016-09-13
Need separate installer for RHost for remote


MikhailArkhipov:
So it doesn't place host executable under 'Visual Studio'. Will be chained to the remote bundle.

MikhailArkhipov:
Implemented

Issue #2315, opened 2016-09-13
Remote services installer - can't load rtvs package


huguesv:
Needs to be installed by the remote installer?

```
Switching R Workspace completed
Error in base::loadNamespace("rtvs", lib.loc = ".") : 
  there is no package called ‘rtvs’
> 
```

MikhailArkhipov:
Yes it is...

MikhailArkhipov:
Also #2316

MikhailArkhipov:
Done

PR #2328, opened 2016-09-14
#2315, #2316, #2324

MikhailArkhipov:
RTools Options page not working #2324
Need separate installer for RHost for remote #2316
Remote services installer - can't load rtvs package #2315

Bundle build in RTVS-Pr coming next


MikhailArkhipov:
Also couple of fixes: caught null ref on await and we need to check against local connections only.

PR #2334, opened 2016-09-15
Fix #2333: Do not display console window for local broker

### LINKED ISSUE #2333: Do not display console window for local broker
`RTVS_HOST_CONSOLE` environment variable is supposed to control whether the window is displayed or not, but the code is currently broken and does the reverse of what it's supposed to do - the console is shown if variable is _absent_, and hidden if it's present. It should be the other way around.

While we're at it, the variable should probably get a different name, since it's used for broker now. And proper handling of true/false values.

int19h:
Rename RTVS_HOST_CONSOLE env variable to RTVS_SHOW_CONSOLE, and fix the logic to properly respect its value - default should be hidden.


MikhailArkhipov:
👍

Issue #2191, opened 2016-08-23
Split large blobs into chunks


karthiknadig:
Currently blobs are sent at once no matter the size. This is not ideal for massive files. We need to break massive content into chunks for transfer.

MikhailArkhipov:
CreateBlob_DisconnectedDuringCreate test is disabled b/c of OOM

PR #2340, opened 2016-09-16
Blob Stream implementation and updated tests.

karthiknadig:
Addresses  #2191


src/Common/Core/Impl/IO/FileSystem.cs
```diff
@@ -31,6 +31,9 @@ public sealed class FileSystem : IFileSystem {
         
         public void FileWriteAllBytes(string path, byte[] bytes) => File.WriteAllBytes(path, bytes);
 
+        public Stream CreateFile(string path) => File.Create(path);
+        public Stream FileOpenRead(string path) => File.OpenRead(path);
```
int19h: To avoid having tons of overloads here for `OpenWrite` etc, let's wrap `File.Open` instead - it allows to specify the mode via enum.

src/Host/Client/Impl/Blobs/DataTransferSession.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            if (doCleanUp) {
+                _cleanup.Add(blob);
+            }
+
+            return data;
+        }
+
+        public void Dispose() {
+            _blobService.DestroyBlobsAsync(_cleanup.Select(b => b.Id).ToArray()).DoNotWait();
```
AlexanderSher: Do you need to store blobs or just their ids?

src/Host/Client/Impl/Host/RHost.BlobReadRequest.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.R.Host.Client {
+    partial class RHost {
+        private class BlobReadRequest : Request<byte[]> {
+            private BlobReadRequest(RHost host, Message message, CancellationToken cancellationToken)
+                : base(host, message, cancellationToken) {
+            }
+
+            public static Task<BlobReadRequest> ReadAllAsync(RHost host, ulong blobId, CancellationToken cancellationToken) {
+                return ReadAsync(host, blobId, 0, -1, cancellationToken);
```
int19h: `long.Max`?

src/Host/Client/Impl/Blobs/RBlobStream.cs
```diff
@@ -0,0 +1,123 @@
...
+            _isDisposed = false;
+        }
+
+        private long _position;
+        public override long Position {
+            get {
+                return _position;
+            }
+            set {
+                Seek(value, SeekOrigin.Begin);
```
AlexanderSher: Seek is not implemented.

src/Host/Client/Impl/Blobs/DataTransferSession.cs
karthiknadig: Just the ids.

src/R/Components/Impl/Plots/Implementation/RPlotManagerProvider.cs
```diff
@@ -2,14 +2,15 @@
...
 using Microsoft.R.Components.InteractiveWorkflow;
 using Microsoft.R.Components.Settings;
 
 namespace Microsoft.R.Components.Plots.Implementation {
     [Export(typeof(IRPlotManagerProvider))]
     internal class RPlotManagerProvider : IRPlotManagerProvider {
-        public IRPlotManager CreatePlotManager(IRSettings settings, IRInteractiveWorkflow interactiveWorkflow) {
-            return new RPlotManager(settings, interactiveWorkflow, () => { });
+        public IRPlotManager CreatePlotManager(IRSettings settings, IRInteractiveWorkflow interactiveWorkflow, IFileSystem fileSystem) {
+            return new RPlotManager(settings, interactiveWorkflow, () => { }, fileSystem);
```
AlexanderSher: Does anyone use the third parameter?

src/Host/Client/Impl/Blobs/DataTransferSession.cs
AlexanderSher: Probably store them instead?

---
In reply to: [79244694](https://github.com/Microsoft/RTVS/pull/2340#discussion_r79244694) [](ancestors = 79244694)

src/Host/Client/Impl/Blobs/RBlobStream.cs
karthiknadig: This is intentional. No one should be setting Position (except in this class) as this stream does not support it.

AlexanderSher: Then we need better exception

---
In reply to: [79244905](https://github.com/Microsoft/RTVS/pull/2340#discussion_r79244905) [](ancestors = 79244905)

src/Host/Client/Impl/Host/RHost.BlobReadRequest.cs
int19h: Actually never mind that, since it won't fit into double on the wire.

karthiknadig: I prefer a strong indicator. Seeing a -1 in logs or while debugging makes the intent clear. Or a separate command for ReadAll.

int19h: How about making it nullable instead for clarity? So `null` means "read to end", and numbers mean size. And pass that `null` through JSON, and handle it on the C++ side.

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -37,7 +41,12 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
         private volatile Task _runTask;
         private volatile Task<REvaluationResult> _cancelEvaluationAfterRunTask;
         private volatile Task<ulong> _cancelCreateBlobAfterRunTask;
-        private volatile Task<byte[]> _cancelGetBlobAfterRunTask;
+        private volatile Task<byte[]> _cancelBlobReadAllAfterRunTask;
+        private volatile Task<byte[]> _cancelBlobReadAfterRunTask;
```
AlexanderSher: You don't need to have so many of them - they all are identical except for the type argument of the task. So you need to add only one more.
Btw, why `CreateBlobAsync` returns `ulong`, while other methods return `long`?

src/R/Components/Impl/Plots/Implementation/RPlotManagerProvider.cs
karthiknadig: Does no look like it. @huguesv any input on this?

src/Host/Client/Impl/Host/RHost.cs
karthiknadig: ulong is the blob id. the other methods return size of the content. .Net uses long to represent sizes in streams.

src/Host/Client/Impl/Blobs/RBlobStream.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+
+        public override void Flush() {
```
AlexanderSher: >Flush [](start = 29, length = 5)

What about `FlushAsync`?

src/Host/Client/Impl/Host/RHost.cs
int19h: `CreateBlobAsync` returns blob ID. I think other methods return the number of bytes read/written.

src/Host/Client/Impl/Blobs/RBlobStream.cs
karthiknadig: `Async `version of `Write`/`Read`/`Flush `call the Sync version internally. So, explicit implementation of `FlushAsync` is not needed.

Issue #2335, opened 2016-09-15
Switching or connecting should be bringing REPL up


MikhailArkhipov:
Otherwise it is totally unclear what is going on. REPL does come up but only when connection is established.

smortaz:
+1

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/03a29d0008f298846bfd06091e155f0dd2967367

Issue #2264, opened 2016-09-07
Improve error returns from broker


MikhailArkhipov:
#2262, @int19h  ```
We do, however, need to improve reporting from the broker for when a session is requested with non-existing interpreter ID (including blank ID when no interpreters are available). Right now we just throw, which gets reported to the client as HTTP 500 error with no body, so it's not possible to distinguish this particular error from any other failure on the broker side.

Probably the easiest way to do this is to add an extra HTTP X-header to the response, that would contain our own error code, if it's an anticipated error (and not a general failure). The client can then check that header if it receives an error response, and display the appropriate error messages.

Also, in many cases, we can just use more specific HTTP status codes instead of 500 (when an exception is unhandled, it's always reported as HTTP 500). This catalog will be helpful:

http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

For the situation when an entity (e.g. an interpreter in this case) is requested, but is not found, it seems that HTTP 422 might be the appropriate status code.
```

Or we could use something from unassigned ranges like 512-599.
428 Precondition Required or 417 Expectation Failed sound reasonable.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/03a29d0008f298846bfd06091e155f0dd2967367

PR #2341, opened 2016-09-16
#2264, #2335 + Session Info output

MikhailArkhipov:
#2335 Switching or connecting should be bringing REPL up
#2264 Improve error returns from broker


src/Host/Protocol/Impl/CustomHttpError.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Protocol {
+    public enum CustomHttpError {
```
int19h: We should not be using custom HTTP error codes - this is contrary to REST. We should definitely not be using codes from 5xx range, these are for _unexpected_ server errors. And 4xx range, which is the right one here, is too crowded already, and does not permit custom code assignment anyway.

In most cases, we should be reporting HTTP error 400 for cases like these, and add our own error code somewhere (either a custom header, or as part of error message JSON).

src/Host/Broker/Impl/About/AboutController.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public AboutController(InterpreterManager interpManager) {
+            _interpManager = interpManager;
+        }
+
+        [HttpGet]
+        public AboutHost Get() {
+            var a = new AboutHost();
+
+            a.Version = Assembly.GetExecutingAssembly().GetName().Version;
```
int19h: Can you put all this logic in a separate non-HTTP-aware class, and call it from here? I'd like controllers to be kept separate from the rest of it to simplify testing.

src/Host/Broker/Impl/Resources.resx
```diff
@@ -141,12 +141,18 @@
   <data name="Error_LogOnFailed" xml:space="preserve">
     <value>Log on failed for user: {0}</value>
   </data>
+  <data name="Error_NoRInterpreters" xml:space="preserve">
```
MikhailArkhipov: Probably not needed in broker. Will remove.

src/Host/Broker/Impl/About/AboutController.cs
```diff
@@ -0,0 +1,57 @@
...
+                foreach (var mo in search.Get()) {
+                    a.TotalVirtualMemory = GetValue(mo, "TotalVirtualMemorySize");
+                    a.FreeVirtualMemory = GetValue(mo, "FreeVirtualMemory");
+                    a.TotalPhysicalMemory = GetValue(mo, "TotalVisibleMemorySize");
+                    a.FreePhysicalMemory = GetValue(mo, "FreePhysicalMemory");
+                    break;
+                }
+            }
+
+            a.Interpreters = new RInstallation().GetCompatibleEngines().Select(e => e.Name).ToArray();
```
int19h: Do we need this? Configured interpreters are already exposed via `InterpretersController`.

src/Host/Broker/Impl/About/AboutController.cs
```diff
@@ -0,0 +1,57 @@
...
+            return a;
+        }
+
+        private long GetValue(ManagementBaseObject mo, string name) {
```
int19h: Since this does some math inside, it looks more like `GetSizeInGB` ?

src/Host/Broker/Impl/Sessions/SessionsController.cs
```diff
@@ -40,12 +44,18 @@ public class SessionsController : Controller {
...
+                if (interp == null) {
+                    return Task.FromResult(CustomErrorResult.Create(CustomHttpError.InterpreterNotFound));
+                }
+            } else {
+                interp = _interpManager.Interpreters.First();
+            }
 
             var session = _sessionManager.CreateSession(User.Identity, id, interp, securePassword, profilePath, request.CommandLineArguments);
-            return Task.FromResult(session.Info);
+            return Task.FromResult(new ObjectResult(session.Info) as IActionResult);
```
int19h: Nit: `FromResult<IActionResult>` (to avoid a runtime cast, and enforce static verification).

src/Host/Protocol/Impl/CustomHttpError.cs
int19h: My suggestion would be a custom `RTVS-Broker-Error` HTTP header, and we might as well use strings there rather than numbers to make it easier to see what's going on (and convert those strings to/from enums on either side with the usual `Parse` and `ToString`).

MikhailArkhipov: I didn't want to use strings since broker locale may be different from user machine locale(s). Each user can have different locale set, so numbers work better.

MikhailArkhipov: BTW nothing prevents us from inventing our own range, this is a not a public server anyway...

src/Host/Broker/Impl/About/AboutController.cs
MikhailArkhipov: We can reuse it, but technically interpreter manager may filter the list differently. but OK.

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -138,13 +142,15 @@ public class RSessionProvider : IRSessionProvider {
                         foreach (var session in sessions) {
                             session.CancelSwitchingBroker();
                         }
+                        BrokerChangeFailed?.Invoke(this, EventArgs.Empty);
```
AlexanderSher: This event isn't used

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
```diff
@@ -74,10 +79,20 @@ internal class VsRInteractiveWorkflowProvider : IRInteractiveWorkflowProvider, I
                                                     _plotsProvider, _activeTextViewTracker, _debuggerModeTracker, 
                                                     _shell, settings, _wss, () => DisposeInstance(sessionProvider));
             _disposableBag.Add(workflow);
+
+            sessionProvider.BrokerChanging += OnBrokerChanging;
             return workflow;
         }
 
+        private void OnBrokerChanging(object sender, EventArgs e) {
```
AlexanderSher: I think there is a side effect that focus will be changed each time user changes a broker.
You can use `workflow.ActiveWindow.Container.Show(focus: false, immediate: false);` and move this handler inside workflow

src/Host/Broker/Impl/About/AboutController.cs
int19h: They are two different things, strictly speaking. This one returns engines that are detected (but which may not be configured for use by this broker). `InterpretersControllers` returns the engines that were configured for use by the broker, which may or may not include auto-detected ones, and may also include some extra manually configured ones.

So the question is whether we do need the list of detected interpreters, if the broker wasn't actually configured to use them. If we do, this is useful. I'm just not sure whether we do.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -195,6 +199,13 @@ internal abstract class BrokerClient : IBrokerClient {
             return new RHost(name, callbacks, transport, cts);
         }
 
+        private async Task GetHostInformationAsync() {
```
AlexanderSher: Please add `cancellationToken`

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
```diff
@@ -39,5 +41,10 @@ public interface IRSessionProvider : IDisposable {
         /// <param name="name">Name of the broker. Will be displayed in REPL.</param>
         /// <param name="path">Either a local path to the R binary or a URL to the broker.</param>
         Task<bool> TrySwitchBrokerAsync(string name, string path = null);
+
+        /// <summary>
+        /// Displays broker machine and process information
+        /// </summary>
+        void DisplayBrokerInformation();
```
AlexanderSher: `PrintBrokerInformation` maybe? Cause it doesn't really display anything - it prints the information to the callback

src/Host/Protocol/Impl/CustomHttpError.cs
int19h: By strings I don't mean localizable user-readable text. I mean sending a header like this:
```
RTVS-Broker-API-Error: InterpreterNotFound
```
and parsing it on the other side as such (`Enum.Parse`), instead of using numbers. Since the protocol is already text-based, there doesn't seem to be much value in using numeric values here.

As far as inventing our own range, there are still problems with custom codes. For one thing, future HTTP spec updates may add new codes in that range (it happened before), and ASP.NET Core might start using them for something else entirely; or `HttpClient` might add special handling for them. Basically, we have no guarantee that the rest of the stack won't stomp on any custom values we use here. Sticking to what the protocol does and doesn't guarantee ensures our future ability to use standard HTTP and REST frameworks.

Also, while the API is not really public yet, it is quite likely to be in the future. People might want to write e.g. watchdog monitoring scripts and such - think about Azure management APIs. So, while matters such as API stability are not an issue, we should try to keep it a clean RESTful API following the spec and best practices.

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -46,7 +47,11 @@ public class WebService {
             var content = new StringContent(requestBody, Encoding.UTF8, "application/json");
             var response = EnsureSuccessStatusCode(await HttpClient.PutAsync(uri, content));
             var responseBody = await response.Content.ReadAsStringAsync();
-            return JsonConvert.DeserializeObject<TResponse>(responseBody);
+            try {
+                return JsonConvert.DeserializeObject<TResponse>(responseBody);
+            } catch(JsonSerializationException ex) {
+                throw new ProtocolViolationException(ex.Message);
```
AlexanderSher: Do we have any handlers for this exception?

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
MikhailArkhipov: It is the intent. If user is connecting, they most probably want REPL to come up and become active.

src/Host/Client/Impl/Definitions/IRSessionProvider.cs
MikhailArkhipov: OK

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
AlexanderSher: But why it should steal focus?

---
In reply to: [79254918](https://github.com/Microsoft/RTVS/pull/2341#discussion_r79254918) [](ancestors = 79254918)

AlexanderSher: But why it should steal focus?

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: I know - but we can add handler to display UI for example. I just didn't want to fire 'Changing' and then nothing if it fails.

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
MikhailArkhipov: The problem here is that workflow.ActiveWindow is null and we have to the create it, effectively duplicating the code in the show REPL command. We can alternatively add a GetOrCreateWindow to workflow but that seems to be a bit too much...

MikhailArkhipov: Or pass a parameter to the command?

src/Host/Client/Impl/Host/BrokerClient.cs
MikhailArkhipov: GetStringAsync doesn't have ct overload, I'd have to do regular GetAsync then

MikhailArkhipov: Don't we just get cancellation exception when we call HttpClient.CancelAllRequests ?

src/Host/Client/Impl/BrokerServices/WebService.cs
MikhailArkhipov: Yes, ConnectToBrokerAsync

src/Host/Client/Impl/Session/RSessionProvider.cs

---
In reply to: [79260010](https://github.com/Microsoft/RTVS/pull/2341#discussion_r79260010) [](ancestors = 79260010)

src/Host/Broker/Impl/About/AboutController.cs
MikhailArkhipov: Actually, I remember now. In the interpreter manager some interpreters don't have names (empty strings). RInstallation names are user-readable. Would be nice to unify of course.

src/Host/Broker/Impl/About/AboutController.cs
MikhailArkhipov: OK

src/Host/Client/Impl/Host/BrokerClient.cs
AlexanderSher: Yes, that's a good idea. You can register a handler and call `CancelPendingRequests`. Problem may happen only if we have more than 1 request

---
In reply to: [79260938](https://github.com/Microsoft/RTVS/pull/2341#discussion_r79260938) [](ancestors = 79260938)

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
AlexanderSher: Ok, we have 5 places where we call `GetOrCreateVisualComponent`, and in 4 of them we then show it. So I think we can just add a couple of parameters `(bool showWindow = false, bool focus = false)` to the `GetOrCreateVisualComponent` and move code that shows the window right there.

---
In reply to: [79260440](https://github.com/Microsoft/RTVS/pull/2341#discussion_r79260440) [](ancestors = 79260440)

src/Host/Protocol/Impl/CustomHttpError.cs
MikhailArkhipov: OK, working on it

src/Host/Broker/Impl/About/AboutController.cs
int19h: I think that was back when they only had IDs. Those were user-specified for custom interpreters, and empty string for the one and only auto-detected one.

With all the recent changes, interpreters have both IDs (again, user-specified for custom, and GUIDs for auto-detected) and names (user-specified for custom, obtained from `RInstallation` for auto-detected). That shows up in JSON returned from `/interpreters`, e.g. here's mine:
```json
[
  {
    "id": "c2a5ee84-c6af-4830-8630-475fe0f59e5f",
    "name": "R 3.2.1",
    "path": "C:\\Program Files\\R\\R-3.2.1",
    "binPath": "C:\\Program Files\\R\\R-3.2.1\\bin\\x64",
    "version": "3.2.1"
  },
  {
    "id": "4e275ceb-3c90-4262-83c2-d94bd6d57644",
    "name": "R 3.2.2",
    "path": "C:\\Program Files\\R\\R-3.2.2",
    "binPath": "C:\\Program Files\\R\\R-3.2.2\\bin\\x64",
    "version": "3.2.2"
  },
  {
    "id": "bed80ba6-674d-4aaa-bedd-a7c1a73749f4",
    "name": "R 3.2.3",
    "path": "C:\\Program Files\\Microsoft\\MRO\\R-3.2.3",
    "binPath": "C:\\Program Files\\Microsoft\\MRO\\R-3.2.3\\bin\\x64",
    "version": "3.2.3"
  },
  {
    "id": "c568f648-b68d-46e5-9952-c882ca31bb48",
    "name": "R 3.2.5",
    "path": "C:\\Program Files\\R\\R-3.2.5",
    "binPath": "C:\\Program Files\\R\\R-3.2.5\\bin\\x64",
    "version": "3.2.5"
  }
]
```

int19h: BTW, if `RInstallation` knows which ones are MRO/MRS and which ones are not, we might want to also expose it in `InterpreterInfo`. I totally forgot about it, but it may come in handy for branding (e.g. show different icons in Workspaces).

MikhailArkhipov: OK, I'll check it again. It was outputting empty string for me on a first try, but I might not have been synced to the tip back then.

MikhailArkhipov: I am getting proper name on remote, but empty name on local - but I may need to reset settings probably.

int19h: On local, there should only be one interpreter anyway, and that one is manually configured (via command line switches passed by RTVS), so its name doesn't really matter. If you want it to have a name for reporting purposes, you can add the switch to specify it in `LocalBrokerClient`. Should be something like this:
```
--R:interpreters:{InterpreterId}:name "Foo"
```

MikhailArkhipov: It does know MRO/MRC - it comes as part of the name.

src/Host/Broker/Impl/ApiErrorResult.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.R.Host.Protocol;
+
+namespace Microsoft.R.Host.Broker {
+    internal sealed class ApiErrorResult : ObjectResult {
+        public ApiErrorResult(HttpResponse response, BrokerApiError error) : base(error) {
+            StatusCode = StatusCodes.Status500InternalServerError;
```
int19h: Please, no HTTP 500. It's just not the right thing to use for REST API errors. HTTP 400 is the generic catch-all for "server was able to handle your request, but the request is invalid for whatever reason". HTTP 5xx is for "server was unable to handle your request, because it crashed due to a bug or other unforeseen circumstances". 4xx is always a client-induced error, and 5xx is always a server-induced error that client has no control over. So any exception that is caused by invalid input received from the client should be 4xx (usually 400).

src/Host/Client/Impl/Host/LocalBrokerClient.cs
```diff
@@ -100,7 +100,7 @@ public LocalBrokerClient(string name, string rHome, string rhostDirectory = null
                     process = Process.Start(psi);
                     process.EnableRaisingEvents = true;
 
-                    var cts = new CancellationTokenSource(10000);
+                    var cts = new CancellationTokenSource(100000);
```
int19h: This is an awfully long timeout for a local broker. Do we really want to wait for almost 2 minutes before reporting failure to start?

src/Host/Broker/Impl/ApiErrorResult.cs
int19h: Here's a decent reference that we can use as a guideline for our own errors.

https://msdn.microsoft.com/en-us/library/azure/dd179357.aspx?f=255&MSPPError=-2147217396

MikhailArkhipov: OK. I used to be 500 that's why I put it back :-)

MikhailArkhipov: I'll use Status412PreconditionFailed, 400 'Bad request' doesn't quite sound right.

src/Host/Client/Impl/Host/LocalBrokerClient.cs
MikhailArkhipov: It is removed, it was debug only temp change

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -23,6 +26,19 @@ public class WebService {
                 throw new UnauthorizedAccessException();
             }
 
+            IEnumerable<string> values;
+            if(response.Headers.TryGetValues(CustomHttpHeaders.RTVSApiError, out values)) {
+                var s = values.FirstOrDefault();
+                if (s != null) {
+                    BrokerApiError apiError;
+                    if (Enum.TryParse(s, out apiError)) {
+                        throw new RHostDisconnectedException(apiError);
```
int19h: This shouldn't be throwing `RHostDisconnectedException`, since this endpoint has nothing to do with host - it's just enumerating sessions. We probably need some kind of generic `BrokerApiException` that wraps `BrokerApiError`, and throw that, or else define custom exception types for every code (but since we can use `catch ... when ...`, I don't see much benefit there). 

`BrokerClient` should then catch that and wrap it into `RHostDisconnectedException` in `CreateBrokerSessionAsync`, like it already does for `HttpRequestException`.

src/Host/Broker/Impl/ApiErrorResult.cs
int19h: Yeah, ASP.NET uses 500 if you let an exception escape the controller. It's because we're technically not supposed to do that ever, so it's considered an internal server bug :)

412 is actually for a very specific kind of precondition failure - it's only when the client specified a precondition HTTP header (`If-Match`), and the server couldn't honor it. It's not supposed to be used for generic kind of precondition errors.

Note that the RFC 2616 definition of 400 ("malformed syntax") is outdated. The current definitions are in this one:
https://tools.ietf.org/html/rfc7231#section-6.5.1

> The 400 (Bad Request) status code indicates that the server cannot or will not process the request due to _something that is perceived to be a client error_ (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

Hence why it was adopted as a generic error code for this sort of stuff in REST. I think the only others that are used with any regularity are 403 and 404, but they only apply in very narrow circumstances.

src/Host/Broker/Impl/About/AboutController.cs
int19h: Right. So maybe add `InterpreterInfo.IsMicrosoftR` or something like that? So that VS could find out without having to parse the name.

int19h: Up to you though. I don't know if we really need this or not, just figured it might be useful later.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -133,8 +139,8 @@ internal abstract class BrokerClient : IBrokerClient {
                 } catch (UnauthorizedAccessException) {
                     isValidCredentials = false;
                     continue;
-                } catch (HttpRequestException ex) {
-                    throw new RHostDisconnectedException("HTTP error while creating session: " + ex.Message, ex);
+                } catch (BrokerApiErrorException apiex) {
+                    throw new RHostDisconnectedException(apiex.ApiError);
```
int19h: Might as well remove the status code from `RHostDisconnectedException` (since in many cases it doesn't have one), and just use `apiex` as the inner exception, I think. If anyone actually needs to find out why disconnect happened, they'll need to inspect `InnerException` in most cases anyway.

src/Host/Broker/Impl/About/AboutController.cs
MikhailArkhipov: I supplied the name from the client side. It gets complicated with MRO which is only distinguishable by its installation path... I am yet to cross that bridge with MRO detection

src/Host/Client/Impl/Host/BrokerClient.cs
MikhailArkhipov: I am afraid to break things here. There are few places than expect RHostDisconnectedException to contain the proper message.

int19h: That's kinda weird, because there are a bunch of places where `RHostDisconnectedException` wraps something else already (e.g. in `RHost.SendAsync`, it wraps `MessageTransportException`), and I'm about to add a few more with session lifetime handling. If we have any place that actually tries to parse its `Message`, it's already broken.

int19h: If you mean that it's expected to contain _some_ message, then we can just reuse `apiex.Message`.

MikhailArkhipov: Yeah, there are many places that create default but fair amount that supply custom message which then sometimes displayed in UI. There are 3 places that provide both message AND inner exception. This should be untangled and simplified, but will take some effort to straighten it out.

Basically, it is safe to provide the message. Alternatively we could supply both message and the inner exception. Supplying just the inner part may break code that expects readable message for the UI.

int19h: message + inner is what I had in mind

PR #2343, opened 2016-09-19
Fix #1878: Hovering over names in Variable Explorer shows value in tooltip

### LINKED ISSUE #1878: Hovering over names in Variable Explorer shows value in tooltip
Hovering over variable _name_ in VE produces a tooltip that displays variable _value_. This is confusing and inconsistent with Locals. It should use the column that is actually under the cursor - this would also provide a way to view excessively long variable names.

It also shows the value tooltip when hovering over the icon. It should probably show the type/class instead, since that's what the icon depends on (and then maybe hide the class/type columns by default, to provide more space for values?).

int19h:
Fix tooltips: icon shows typeof and class, name shows name, value shows value.

Hide typeof and class columns to free up more space for values.


MikhailArkhipov:
👍

karthiknadig:
:shipit:

PR #2378, opened 2016-09-21
Fix 2349

### LINKED ISSUE #2349: Plots: Null reference exception when dropping (drag/drop) on unassigned device
In RPlotDeviceControl.xaml.cs, in UserControl_DragEnter, `Model?.Device` is null:

`var targetDeviceId = Model?.Device.DeviceId;`

We should allow drop when null, a new graphics device will get created on drop.

huguesv:
Fixes #2349 Plots: Null reference exception when dropping (drag/drop) on unassigned device 

When a plot window doesn't have a graphics device, it creates one as necessary before issuing a copy plot command.  That was all set up to work, but a late change during code review broke the drag move handler code, which prevents that part of the feature from working.


huguesv:
Will need to be backported to 0.5 as well.

AlexanderSher:
👍

PR #2382, opened 2016-09-21
Fixes #2347 protocol errors are transmitted back to RTVS

### LINKED ISSUE #2347: Remote shiny/help crashes RTVS while handling favicon.ico requests

karthiknadig:
Previously favicon.ico requests would cause 404 from RHost which is normal. But the remoting code did not send this server error response back to RTVS/Browser.

### DIFF
--- a/src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
+++ b/src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
@@ -58,11 +58,11 @@ namespace Microsoft.R.Host.Client.BrokerServices {
                 }
             } finally {
                 if (response != null) {
-                    if (context.Request.IsWebSocketRequest) {
+                    if (context.Request.IsWebSocketRequest && response.StatusCode == HttpStatusCode.SwitchingProtocols) {
                         Stream respStream = response.GetResponseStream();
                         string subProtocol = response.Headers[Constants.Headers.SecWebSocketProtocol];
-                        var remoteWebSocket = CommonWebSocket.CreateClientWebSocket(respStream, subProtocol, TimeSpan.FromMinutes(10), 65335, true);
-                        var websocketContext = await context.AcceptWebSocketAsync(subProtocol, 65335, TimeSpan.FromMinutes(10));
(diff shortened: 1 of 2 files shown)


src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -47,21 +47,33 @@ public class RemoteUriWebService : IRemoteUriWebService {
...
+                    response = wex.Response as HttpWebResponse;
+                } else {
+                    throw wex;
+                }
+            } finally {
+                if (response != null) {
+                    if (context.Request.IsWebSocketRequest) {
+                        Stream respStream = response.GetResponseStream();
+                        string subProtocol = response.Headers[Constants.Headers.SecWebSocketProtocol];
+                        var remoteWebSocket = CommonWebSocket.CreateClientWebSocket(respStream, subProtocol, TimeSpan.FromMinutes(10), 65335, true);
```
int19h: Can you use named arguments here, so that it's clear what the number and the boolean correspond to?

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -47,21 +47,33 @@ public class RemoteUriWebService : IRemoteUriWebService {
...
+                response = (HttpWebResponse)await request.GetResponseAsync();
+            } catch (WebException wex) {
+                if(wex.Status == WebExceptionStatus.ProtocolError) {
+                    response = wex.Response as HttpWebResponse;
+                } else {
+                    throw wex;
+                }
+            } finally {
+                if (response != null) {
+                    if (context.Request.IsWebSocketRequest) {
```
int19h: This might do the wrong thing when you have a websocket request and an error response. A successful websocket request _must_ have HTTP response 101 (switching protocols). If it's any other error, the response is not a websocket stream (e.g. because the server to which the request was made doesn't support websockets), but rather just a generic HTTP response. So if you call `AcceptWebSocketAsync` on that, it'll break.

int19h:
👍

PR #2384, opened 2016-09-22
Several bug fixes

### LINKED ISSUE #2351: Reconnect button doesn't work (neither in CM nor in status bar)

### LINKED ISSUE #2352: Various remote status don't agree as to where I'm connected
build 2016-09-19.1

DDHUVALO073 is my local (client) machine
DDHUVALO083 is the remote machine

I started VS, R interactive wasn't open.  I opened Workspaces, updated my remote machine name/server, hit save, then connect.

R session says I'm connected to local machine (DDHUVALO073).  Status bar says I'm connected to local as well.  Connection manager and REPL thinks I'm connected to server (DDHUVALO083).

image

### LINKED ISSUE #2354: Connection progress doesn't dismiss when it can't connect
The progress bar dialog stays there forever, despite the connection error in REPL.

image

### LINKED ISSUE #2362: Failed connection should have different message when there is no fallback
1. Connect to a remote machine
1. Close VS
1. Open VS
1. Open Connection Manager. There should be no active connection
1. Try to connect to any connection that will fail

```
Connecting to the new R Workspace failed, restoring connection to 
'' : http://localhost/
```

### LINKED ISSUE #2372: Don't restore connection when new connection failed
- Connecting to remote machine make take time
- User may not necessarily want this
-  It is confusing: we restore local and if user missed the message, repl shows R header and user may have an impression that switch succeeded.

Just output:

'Connection failed. Open Workspaces window and select local R or create another remote connection'.

AlexanderSher:
- Fix #2372: Don't restore connection when new connection failed
- Fix #2362: Failed connection should have different message when there is no fallback
- Fix #2354: Connection progress doesn't dismiss when it can't connect
- Fix #2352: Various remote status don't agree as to where I'm connected
- Fix #2351: Reconnect button doesn't work (neither in CM nor in status bar)

### DIFF
--- a/src/Host/Client/Impl/Session/RSessionProvider.cs
+++ b/src/Host/Client/Impl/Session/RSessionProvider.cs
@@ -19,6 +19,7 @@ namespace Microsoft.R.Host.Client.Session {
         private readonly ConcurrentDictionary<Guid, RSession> _sessions = new ConcurrentDictionary<Guid, RSession>();
         private readonly DisposeToken _disposeToken = DisposeToken.Create<RSessionProvider>();
         private readonly BinaryAsyncLock _brokerSwitchLock = new BinaryAsyncLock();
+        private readonly BinaryAsyncLock _brokerDisconnectedLock = new BinaryAsyncLock();
         private readonly AsyncCountdownEvent _connectCde = new AsyncCountdownEvent(0);
 
         private readonly BrokerClientProxy _brokerProxy;
@@ -100,6 +101,7 @@ namespace Microsoft.R.Host.Client.Session {
 
         private void RSessionOnConnected(object sender, RConnectedEventArgs e) {
             OnBrokerConnected();
+            _brokerDisconnectedLock.ResetAsync().DoNotWait();
         }
 
         private void RSessionOnDisconnected(object sender, EventArgs e) {
@@ -107,10 +109,13 @@ namespace Microsoft.R.Host.Client.Session {
         }
 
         private async Task RSessionOnDisconnectedAsync() {
+            var token = await _brokerDisconnectedLock.WaitAsync();
             try {
-                // We don't want to show that connection is broken just because one of the sessions has been disconnected. Ping broker.
-                await _brokerProxy.PingAsync();
+                // We don't want to show that connection is broken just because one of the sessions has been disconnected. Need to test connection
+                await TestBrokerConnectionWithRHost(_brokerProxy, default(CancellationToken));
+                token.Reset();
             } catch (RHostDisconnectedException) {
+                token.Set();
                 OnBrokerDisconnected();
             }
         }
@@ -138,17 +143,21 @@ namespace Microsoft.R.Host.Client.Session {
             }
 
             try {
-                var callbacks = new NullRCallbacks();
-                var rhost = await brokerClient.ConnectAsync(nameof(TestBrokerConnectionAsync), callbacks, cancellationToken: cancellationToken);
-                try {
-                    var rhostRunTask = rhost.Run(cancellationToken);
-                    callbacks.SetReadConsoleInput("q()\n");
-                    await rhostRunTask;
-                } finally {
-                    rhost.Dispose();
-                }
+                await TestBrokerConnectionWithRHost(brokerClient, cancellationToken);
             } finally {
                 brokerClient.Dispose();
+            }
+        }
+
+        private static async Task TestBrokerConnectionWithRHost(IBrokerClient brokerClient, CancellationToken cancellationToken) {
+            var callbacks = new NullRCallbacks();
+            var rhost = await brokerClient.ConnectAsync(nameof(TestBrokerConnectionAsync), callbacks, cancellationToken: cancellationToken);
+            try {
+                var rhostRunTask = rhost.Run(cancellationToken);
+                callbacks.SetReadConsoleInput("q()\n");
+                await rhostRunTask;
+            } finally {
+                rhost.Dispose();
             }
         }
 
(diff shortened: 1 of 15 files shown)


src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -172,31 +187,22 @@ public class RSessionProvider : IRSessionProvider {
...
-                    }
-                }
-
-                if (!switchingFromNull) {
-                    _callback.WriteConsole(Resources.RSessionProvider_SwitchingRWorkspaceCompleted);
-                }
+                await SwitchBrokerAsync(cancellationToken, oldBroker);
+                oldBroker.Dispose();
                 PrintBrokerInformation();
+            } catch(Exception) {
```
int19h: Do we really want to swallow absolutely everything here?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -206,37 +212,109 @@ public class RSessionProvider : IRSessionProvider {
...
+                var transactions = startTransactionTasks.Select(t => t.Result).ToList();
+
+                _callback.WriteConsole(Resources.RSessionProvider_StartConnectingToWorkspaceFormat.FormatInvariant(transactions.Count));
+                await Task.WhenAll(transactions.Select(t => ConnectToNewBrokerAsync(t, cts)));
+
+                _callback.WriteConsole(Resources.RSessionProvider_RestartingSessionsFormat.FormatInvariant(transactions.Count));
+                OnBrokerDisconnected();
+                await Task.WhenAll(transactions.Select(t => CompleteSwitchingBrokerAsync(t, oldBroker, cts)));
+            } finally {
                 foreach (var task in startTransactionTasks.Where(t => t.Status == TaskStatus.RanToCompletion)) {
```
int19h: This seems like a potential race against task completion. Maybe drop `Where`, and in loop body, do `Dispose` inside `ContinueWith`?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -206,37 +212,109 @@ public class RSessionProvider : IRSessionProvider {
...
+                // All sessions should participate in reconnect. If any of it didn't start, cancel the rest.
+                var startTransactionTasks = sessions.Select(s => s.StartReconnectingAsync(cts.Token)).ToList();
+
+                try {
+                    await Task.WhenAll(startTransactionTasks);
+                    var transactions = startTransactionTasks.Select(t => t.Result).ToList();
+
+                    await Task.WhenAll(transactions.Select(s => ReconnectSessionAsync(s, cts)));
+                } finally {
+                    foreach (var task in startTransactionTasks.Where(t => t.Status == TaskStatus.RanToCompletion)) {
```
int19h: And same thing here.

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -206,37 +212,109 @@ public class RSessionProvider : IRSessionProvider {
...
             // All sessions should participate in switch. If any of it didn't start, cancel the rest.
-            var startTransactionTasks = sessions.Select(s => s.StartSwitchingBrokerAsync(cancellationToken)).ToList();
+            var startTransactionTasks = sessions.Select(s => s.StartSwitchingBrokerAsync(cts.Token)).ToList();
+
             try {
                 await Task.WhenAll(startTransactionTasks);
-            } catch (OperationCanceledException) {
+                var transactions = startTransactionTasks.Select(t => t.Result).ToList();
+
+                _callback.WriteConsole(Resources.RSessionProvider_StartConnectingToWorkspaceFormat.FormatInvariant(transactions.Count));
```
MikhailArkhipov: You can remove 'starting' and success messages. There is a progress dialog and we print session info (which will be happening on success). Less clutter in REPL

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: There is no race.
After tasks are acquired, we do `await Task.WhenAll(startTransactionTasks);`, which will ensure that all tasks are completed.

---
In reply to: [80102485](https://github.com/Microsoft/RTVS/pull/2384#discussion_r80102485) [](ancestors = 80102485)

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: Yeah, probably I should should rethrow exceptions that aren't OCE

---
In reply to: [80101432](https://github.com/Microsoft/RTVS/pull/2384#discussion_r80101432) [](ancestors = 80101432)

src/Host/Client/Impl/Session/RSessionProvider.cs

---
In reply to: [80108224](https://github.com/Microsoft/RTVS/pull/2384#discussion_r80108224) [](ancestors = 80108224)

MikhailArkhipov:
I have more changes around here, namely reporting actual errors rather than 500 and 404, I will merge and submit PR.

MikhailArkhipov:
👍

src/Host/Client/Impl/Session/RSessionProvider.cs
int19h: Then `Where` shouldn't be necessary?

int19h:
👍

AlexanderSher:
Do you also print them to the REPL?

---
In reply to: [248991540](https://github.com/Microsoft/RTVS/pull/2384#issuecomment-248991540) [](ancestors = 248991540)

MikhailArkhipov:
Yes I do. I consolidated error printing in one place since we sometimes print several times.

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: Without `Where`, `task.Result` will throw exception on the tasks that have failed.

---
In reply to: [80110049](https://github.com/Microsoft/RTVS/pull/2384#discussion_r80110049) [](ancestors = 80110049)

MikhailArkhipov:
Merge?

PR #2390, opened 2016-09-22
Extracted `WhenAllCancelOnFailure`

AlexanderSher:
This pattern is used in `RSessionProvider` several times - when any individual session fails on any step of switch or reconnection transaction, there is no reason for other sessions to continue, so we cancel them.


MikhailArkhipov:
👍

Issue #2381, opened 2016-09-21
When host could not start CM still shows connection as valid


MikhailArkhipov:
I renamed `zlibd.dll` to `zlibd.dll~` then started REPL

![image](https://cloud.githubusercontent.com/assets/12820357/18726227/535c46fe-7ff8-11e6-84a0-7f729d3bf1eb.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/47a5094efbcf5ae54100e2f719fb938fd0c7afbd

Issue #2371, opened 2016-09-20
"matrix" doesn't show up in completion list on MRO


int19h:
`matrix` function is not showing up in the completion list when typing "matrix" in the REPL. The completion list is there, but it only shows a single entry "Matrix" (note capitalization), corresponding to a library with the same name. The function is there and can be called, though.

int19h:
It appears that we have some issue with case sensitivity in completion lists in general. Do this:
```r
aaa <- 123
AAA <- 456
```
Only "aaa" shows up in the list, not "AAA". Now do:
```r
Aaa <- 789
```
Now only "Aaa" shows in the list.

MikhailArkhipov:
Regression after #2165

MikhailArkhipov:
Change ported

Issue #2355, opened 2016-09-20
Include C++ runtime library in remote setup


MikhailArkhipov:

MikhailArkhipov:
Interesting. We do have /MTd (not /MDd which would be multithreaded DLL...)

MikhailArkhipov:
I bet it is some third party libraries that want CRT as DLL.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/47a5094efbcf5ae54100e2f719fb938fd0c7afbd

Issue #2353, opened 2016-09-20
Broker returns 404 Not Found


huguesv:
build 2016-09-19.1

I used the R Remote Services installer, on Win 10 clean VM (without VS).
I ran the broker in an admin command prompt.

Here's the REPL output:
```
Couldn't connect session to 
'ddhuvalo083' : http://ddhuvalo083:5000/ 
because of the following error: HTTP error while creating session: The remote server returned an error: (404) Not Found.
Connecting to the new R Workspace failed, restoring connection to 
'Microsoft R Client' : C:\Program Files\Microsoft\R Client\R_SERVER\
```

Here's the broker output:
```
C:\Program Files\R Remote Service for Visual Studio\0.6>microsoft.r.host.broker --server.urls http://0.0.0.0:5000
trce: Microsoft.R.Host.Broker.Interpreters.InterpreterManager[0]
      Auto-detecting R ...
trce: Microsoft.R.Host.Broker.Interpreters.InterpreterManager[0]
      R 3.3.1 detected at \"C:\Program Files\R\R-3.3.1\"
info: Microsoft.R.Host.Broker.Interpreters.InterpreterManager[0]
      1 interpreters configured:
      'f96a8d60-eb96-4da2-a773-d79ca3d0ec61': 3.3.1 at "C:\Program Files\R\R-3.3.1"
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[3]
      Hosting starting
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[4]
      Hosting started
Hosting environment: Production
Content root path: C:\Program Files\R Remote Service for Visual Studio\0.6
Now listening on: http://0.0.0.0:5000
Application started. Press Ctrl+C to shut down.
dbug: Microsoft.AspNetCore.Server.Kestrel[1]
      Connection id "0HKV1IBQI36HB" started.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 PUT http://ddhuvalo083:5000/sessions/IntelliSense application/json; charset=utf-8 46
dbug: Microsoft.AspNetCore.Builder.RouterMiddleware[1]
      Request did not match any routes.
info: Odachi.AspNetCore.Authentication.Basic.BasicMiddleware[7]
      Basic was not authenticated. Failure message: No authorization header.
info: Odachi.AspNetCore.Authentication.Basic.BasicMiddleware[12]
[...]

PR #2393, opened 2016-09-22
Bugs fixes, extended error return from broker, error cleanup

MikhailArkhipov:
#2371 "matrix" doesn't show up in completion list on MRO
#2353 Broker returns 404 Not Found
#2381 When host could not start CM still shows connection as valid
#2355 Include C++ runtime library in remote setup

also fixes missing constants in completion such as TRUE, NA since before they used to be fetched from HTML docs and now have to be added manually.

Cleans up messaging in REPL, added some line breaks for readability. Adds returning extended errors from broker so we can better inform user why host didn't start.


src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -160,6 +161,18 @@ public class Session {
             HostToClientWorker(_process.StandardOutput.BaseStream, hostEnd).DoNotWait();
         }
 
+        private void StartSession() {
+            _process.Start();
+            _process.WaitForExit(250);
+            if (_process.HasExited && _process.ExitCode != 0) {
+                var message = ErrorCodeConverter.MessageFromErrorCode(_process.ExitCode);
```
int19h: `Process.ExitCode` is the errorlevel of the process, e.g. what you pass to `exit(...)`, and meaning of those is defined by the process itself. For R host, the user can cause it to return an arbitrary errorlevel - `quit(status = ...)`. If we always try to translate that to a Win32 error, we'll be generating spurious messages.

src/Host/Broker/Impl/Sessions/SessionsController.cs
```diff
@@ -54,15 +56,20 @@ public class SessionsController : Controller {
                 interp = _interpManager.Interpreters.First();
             }
 
-            var session = _sessionManager.CreateSession(User.Identity, id, interp, securePassword, profilePath, request.CommandLineArguments);
-            return Task.FromResult<IActionResult>(new ObjectResult(session.Info));
+            try {
+                var session = _sessionManager.CreateSession(User.Identity, id, interp, securePassword, profilePath, request.CommandLineArguments);
+                return Task.FromResult<IActionResult>(new ObjectResult(session.Info));
+            } catch (Exception ex) {
```
int19h: This should only catch and forward exceptions that are expected for process start - we don't want to report internal code bugs in the broker in that manner (usually not valuable, and is a security risk).

src/Host/Broker/Impl/Sessions/SessionsController.cs
```diff
@@ -54,15 +56,20 @@ public class SessionsController : Controller {
...
+                return Task.FromResult<IActionResult>(new ApiErrorResult(HttpContext.Response, BrokerApiError.UnableToStartRHost, ex));
+            }
         }
 
         [HttpGet("{id}/pipe")]
         public IActionResult GetPipe(string id) {
             var session = _sessionManager.GetSession(User.Identity, id);
-            if (session?.Process?.HasExited ?? true) {
-                return NotFound();
+            Exception ex = (session?.Process != null && session.Process.HasExited) ? new Win32Exception(session.Process.ExitCode) : null;
```
int19h: This isn't right - this shouldn't be reporting "unable to start RHost" when connecting to a pipe for a session that doesn't exist, because this endpoint is not responsible for creating sessions. 

So if a session is not in the list, this should still be 404. And there shouldn't be any case where the process has exited, but the session is still in the list, since we remove it in `Process.Exited`.

src/SetupRemote/Product.wxs
```diff
@@ -26,6 +26,12 @@
     <DirectoryRef Id="INSTALLFOLDER">
       <Component Id="ProductAssemblies" Guid="{9C717F8F-4A97-4FA5-8031-5A180467754D}" Win64="yes">
         <?include "$(var.SourceDir)\Setup\CommonAssemblies.wxi" ?>
+        <!-- For remote we must include C++ runtime as VS may not be installed -->
+        <?ifdef Debug?>
+        <File Source="$(var.LibDir)\vcruntime140d.dll" />
```
int19h: This is the debug version of the VC++ runtime. We should only be including this one for debug builds, and the release one for release builds.

Unless we're using a debug version of some library (zlib?) in our release builds. In which case we should switch to release version of that, since there's likely a significant perf difference.

src/Host/Broker/Impl/ApiErrorResult.cs
```diff
@@ -1,20 +1,24 @@
...
 
+using System;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.R.Host.Protocol;
 
 namespace Microsoft.R.Host.Broker {
     internal sealed class ApiErrorResult : ObjectResult {
-        public ApiErrorResult(HttpResponse response, BrokerApiError error) : base(error) {
+        public ApiErrorResult(HttpResponse response, BrokerApiError error, Exception ex = null) : base(error) {
```
int19h: While you're here, can you also fix `ApiErrorResult` to not take `HttpResponse` as parameter? It's supposed to get one in its `ExecuteAsync` method, and modify it there, not in constructor. You can override it, call `base`, and then add the header.

src/Host/Broker/Impl/Sessions/Session.cs
MikhailArkhipov: I know. However, OS failures come in the same code. Ex, missing dll comes as C0000135 (NTSTATUS == module missing). I think user terminating host with odd error codes is less common.  We could theoretically use CreateProcess and separate exit code from actual last error, but that is quite a bit more work...

src/Host/Broker/Impl/Sessions/SessionsController.cs
MikhailArkhipov: In this cases 404 is not informative. The goal is to give user more information what happened. For example broken installation (comes as 'module is missing'. This looks more useful than 'host start failed, http error 404'.

src/SetupRemote/Product.wxs
MikhailArkhipov: It does appear under ?ifdef Debug so it is only for debug build

src/Host/Broker/Impl/Sessions/Session.cs
int19h: User errors should always be positive. By the example you give, it sounds like OS errors would always be negative? If so, perhaps we could use that to differentiate?

int19h: Here's something that I found:

> By convention, processes that exit cleanly return either 0, 1, or a Win32 error code (which are all small positive numbers) so an exit code with the top bit set usually means a status code, typically because of an unhandled exception.

Also, in ntstatus.h, you can see that the "error" category starts with `0xC0000001L` (`STATUS_UNSUCCESSFUL`). So I think it's safe to ignore everything that's less than that.

src/Host/Broker/Impl/Sessions/SessionsController.cs
int19h: And with a wait that you've added there, won't that already be the case actually?

src/SetupRemote/Product.wxs
int19h: Oh, nevermind. Somehow I read the other one as "...d" as well.

src/Host/Broker/Impl/Sessions/Session.cs
MikhailArkhipov: There are two kinds or errors: NTSTATUS which may be positive and Win32 (GetLastError) that are positive. Not sure if we ever get Win32 one as exit code. From NTSTATUS, those that we are interested in are indeed negative. I built a list of about 20 error code that probably matter.

MikhailArkhipov: I am looking at https://msdn.microsoft.com/en-us/library/cc704588.aspx. I guess just filtering for negatives may be sufficient.

src/Host/Broker/Impl/Sessions/SessionsController.cs
MikhailArkhipov: Limit to Win32Exceptions? User may still want more information for bug reports - logging may be disabled in some cases.

src/Host/Broker/Impl/Sessions/SessionsController.cs
MikhailArkhipov: I guess it is now handled in the process start anyway.

src/Host/Broker/Impl/Sessions/SessionsController.cs
MikhailArkhipov: Basically this all mostly to give user something they either understand or at least report to use when logs are not available and then are unable to send feedback.

MikhailArkhipov: `*report to us`

src/Host/Broker/Impl/ApiErrorResult.cs
MikhailArkhipov: Done

MikhailArkhipov:
@AlexanderSher please have a look as well.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -160,6 +161,18 @@ public class Session {
             HostToClientWorker(_process.StandardOutput.BaseStream, hostEnd).DoNotWait();
         }
 
+        private void StartSession() {
+            _process.Start();
+            _process.WaitForExit(250);
```
AlexanderSher: Where does this number come from?

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -32,7 +32,8 @@ public class WebService {
                 if (s != null) {
                     BrokerApiError apiError;
                     if (Enum.TryParse(s, out apiError)) {
-                        throw new BrokerApiErrorException(apiError);
+                        response.Headers.TryGetValues(CustomHttpHeaders.RTVSBrokerException, out values);
```
AlexanderSher: What if this method returns `false`?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -197,8 +197,12 @@ public class RSessionProvider : IRSessionProvider {
                 BrokerChanging?.Invoke(this, EventArgs.Empty);
                 await SwitchBrokerAsync(cancellationToken, oldBroker);
                 oldBroker.Dispose();
-                PrintBrokerInformation();
-            } catch(Exception ex) {
+
+                if (brokerClient.IsRemote) {
+                    _callback.WriteConsole(Environment.NewLine + Resources.Connected + Environment.NewLine);
+                    PrintBrokerInformation();
```
AlexanderSher: Why you don't want to print information about local broker?

src/Host/Broker/Impl/Sessions/Session.cs
MikhailArkhipov: Out of the blue... 1000 seemed to long for a delay...

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: Generally user knows local machine parameters, so less clutter. They can request it via Session | Host Information command.

src/Languages/Core/Impl/Utility/ListHelpers.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Languages.Core.Utility {
     public static class ListHelpers {
-        public static void RemoveDuplicates<T>(this List<T> list) where T : IComparable<T> {
+        public static void RemoveDuplicates<T>(this List<T> list, Comparison<T> comparison = null) where T : IComparable<T> {
```
AlexanderSher: `Comparison<T>` doesn't require T to be `IComparable<T>`

src/Host/Client/Impl/BrokerServices/WebService.cs
MikhailArkhipov: values will be null and ctor below will take that into account.

src/Languages/Core/Impl/Utility/ListHelpers.cs
MikhailArkhipov: But if it is null then comparable is used, as before.

AlexanderSher: So the overload with `Comparison<T> comparison` can be called from the original method:
```{csharp}
public static void RemoveDuplicates<T>(this List<T> list) where T : IComparable<T> => list.RemoveDuplicates((i1, i2) => i1.CompareTo(i2));
```

---
In reply to: [80277594](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80277594) [](ancestors = 80277594)

src/Host/Client/Impl/BrokerServices/WebService.cs
AlexanderSher: Ok

---
In reply to: [80277735](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80277735) [](ancestors = 80277735)

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -275,25 +279,18 @@ public class RSessionProvider : IRSessionProvider {
                     await Task.WhenAll(transactions.Select(t => t.AcquireLockAsync(cancellationToken)));
                     await Task.WhenAll(transactions.Select(t => t.ReconnectAsync(cancellationToken)));
                 } catch (OperationCanceledException ex) when (!(ex is RHostDisconnectedException)) {
-                    _callback.WriteConsole(Resources.RSessionProvider_SwitchingWorkspaceCanceled.FormatInvariant(_brokerProxy.Name, GetUriString(_brokerProxy)));
```
AlexanderSher: So you don't want to print anything if user has canceled a switch?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -275,25 +279,18 @@ public class RSessionProvider : IRSessionProvider {
...
-                    _callback.WriteConsole(Resources.RSessionProvider_SwitchingWorkspaceFailed.FormatInvariant(_brokerProxy.Name, GetUriString(_brokerProxy), ex.Message));
+                    _callback.WriteConsole(Resources.RSessionProvider_ConnectionFailed.FormatInvariant(ex.Message));
                     throw;
-                } finally {
-                    foreach (var transaction in transactions) {
-                        transaction.Dispose();
-                    }
                 }
-            } else {
-                await TestBrokerConnectionWithRHost(_brokerProxy, cancellationToken);
```
AlexanderSher: Have you removed this intentionally?

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: Nope. They know they canceled it :-)

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: No, must be a merge thing, there were bunch of conflicts. Let me put it back.

src/Host/Client/Impl/Session/RSessionProvider.cs

---
In reply to: [80278538](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80278538) [](ancestors = 80278538)

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -50,9 +50,11 @@ internal class ConnectionManager : IConnectionManager {
 
             _disposableBag = DisposableBag.Create<ConnectionManager>()
                 .Add(_statusBarViewModel)
-                .Add(() => _sessionProvider.BrokerStateChanged -= BrokerStateChanged);
+                .Add(() => interactiveWorkflow.RSession.Connected -= RSession_Connected)
+                .Add(() => interactiveWorkflow.RSession.Disconnected -= RSession_Disconnected);
 
-            _sessionProvider.BrokerStateChanged += BrokerStateChanged;
+            interactiveWorkflow.RSession.Connected += RSession_Connected;
```
AlexanderSher: This is wrong - user will get notification that connection is broken every time RHost for Interactive Window RSession dies.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -247,27 +263,11 @@ internal class ConnectionManager : IConnectionManager {
                 }
             }
 
-            if (!string.IsNullOrEmpty(connectionInfo?.Path) && await TrySwitchBrokerAsync(connectionInfo)) {
-                return;
-            }
-
-            var connection = RecentConnections.FirstOrDefault(c => !c.IsRemote);
```
AlexanderSher: So we don't try to get user connected to anything if we failed to connect him to the last active connection?

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: The problem is that indication is confusing. It shows everything green but host is down. However, green can be treated as 'connected' only. I am wondering if user will get confused why things are not working especially when connected to remote.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -78,10 +78,10 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
             } catch (RHostBinaryMissingException) {
                 await _coreShell.ShowErrorMessageAsync(Resources.Error_Microsoft_R_Host_Missing);
                 return ExecutionResult.Failure;
-            } catch (RHostDisconnectedException) {
-                // We don't 
+            } catch (RHostDisconnectedException rhdex) {
                 return ExecutionResult.Success;
-            } catch (Exception) {
+            } catch (Exception ex) {
+                CurrentWindow.WriteError(Environment.NewLine + ex.Message);
```
AlexanderSher: There is a dilemma with what should be returned - `ExecutionResult.Success` or `ExecutionResult.Failure`. `ExecutionResult.Failure` freezes Interactive Window, forcing user to click Reset. But if connection is lost, Reset won't help. On the other hand, if connection is restored, there is no reason to freeze Interactive Window connection anymore.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: We can leave it as before and see what users will say.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: Correct. They may or may not want local started and may or may not want us to try last remote since it may be down or they may be offline. In fact, they may have REPL up since last session but not intend to use R.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: Yeah, we can add one more icon to show that not only connection is on, put session is actually working

---
In reply to: [80280152](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80280152) [](ancestors = 80280152)

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: Ok

---
In reply to: [80280455](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80280455) [](ancestors = 80280455)

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
MikhailArkhipov: I noticed. I initially tried Failure but it looked weird without prompt. It was not immediately clear to click reset.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
MikhailArkhipov: Or we can add yellow state, i.e. connected but broken

AlexanderSher: Sure!

---
In reply to: [80280797](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80280797) [](ancestors = 80280797)

AlexanderSher: But let's do it as a separate change

---
In reply to: [80280959](https://github.com/Microsoft/RTVS/pull/2393#discussion_r80280959) [](ancestors = 80280959,80280797)

MikhailArkhipov: I'll file a new item

src/R/Editor/Impl/Completion/Providers/KeywordCompletionProvider.cs
```diff
@@ -36,7 +37,9 @@ public class KeywordCompletionProvider : IRCompletionListProvider, IRHelpSearchT
                 var infoSource = _snippetInformationSource?.InformationSource;
                 ImageSource keyWordGlyph = GlyphService.GetGlyph(StandardGlyphGroup.GlyphKeyword, StandardGlyphItem.GlyphItemPublic, _shell);
 
-                foreach (string keyword in Keywords.KeywordList) {
+                // Union with constants like TRUE and other common things
+                var keywords = Keywords.KeywordList.Union(Logicals.LogicalsList).Union(Constants.ConstantsList);
```
AlexanderSher: Should it be `Union` (i.e. search and remove duplicates) or `Concat`?

MikhailArkhipov: Concat is fine.

AlexanderSher:
:+1:

Issue #2375, opened 2016-09-20
Null reference exception when plot window docked with documents and hidden


huguesv:
Dock the plot window in document well.
Open an .R file such that the plot window document is no longer visible.
Issue a plot command.

Note: if you split the document well into 2 horizontal or vertical parts, with the .R file in one, plot in the other, it will work fine.

```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Exception in RHost run loop:
System.NullReferenceException: Object reference not set to an instance of an object.

   at Microsoft.R.Components.Plots.Implementation.WpfUnitsConversion.ToPixels(Visual visual, Size wpfSize) in C:\Projects\huguesv\RTVS\src\R\Components\Impl\Plots\Implementation\WpfUnitsConversion.cs:line 16

   at Microsoft.R.Components.Plots.Implementation.View.RPlotDeviceControl.GetPixelSizeAndResolution(Size wpfSize) in C:\Projects\huguesv\RTVS\src\R\Components\Impl\Plots\Implementation\View\RPlotDeviceControl.xaml.cs:line 48

   at Microsoft.R.Components.Plots.Implementation.View.RPlotDeviceControl.GetPlotWindowProperties() in C:\Projects\huguesv\RTVS\src\R\Components\Impl\Plots\Implementation\View\RPlotDeviceControl.xaml.cs:line 32

   at Microsoft.R.Components.Plots.Implementation.RPlotDeviceVisualComponent.GetDeviceProperties() in C:\Projects\huguesv\RTVS\src\R\Components\Impl\Plots\Implementation\RPlotDeviceVisualComponent.cs:line 126

   at Microsoft.R.Components.Plots.Implementation......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------

```

then you get:
```
R Host process is stopped. Click Reset to start a new one.Interactive Window is disconnected from R Session.
Try connecting to a remote machine in the Workspaces window or
click Reset to restart local R interpreter.
> 
```

huguesv:
`var source = PresentationSource.FromVisual(visual);` returns `null`, `visual` input is not `null`. 
So next line crashes:
```
return (Size)source.CompositionTarget.TransformToDevice.Transform((Vector)wpfSize);
```

jflam:
We can't ship 0.5 with this bug. It's way too easy to hit.

MikhailArkhipov:
Yes, we should.

PR #2399, opened 2016-09-23
Prevent crash when plot window is docked as document window and hidden

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/2375

### DIFF
--- a/src/R/Components/Impl/Plots/Implementation/NonScaledImage.cs
+++ b/src/R/Components/Impl/Plots/Implementation/NonScaledImage.cs
@@ -20,7 +20,7 @@ namespace Microsoft.R.Components.Plots.Implementation {
                 // We prevent that scaling by calculating a size based on
                 // pixel size and dpi setting.
                 Size bitmapSize = new Size(bmp.PixelWidth, bmp.PixelHeight);
-                return WpfUnitsConversion.FromPixels(this, bitmapSize);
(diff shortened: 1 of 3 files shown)


src/R/Components/Impl/Plots/Implementation/NonScaledImage.cs
```diff
@@ -20,7 +20,7 @@ internal class NonScaledImage : Image {
                 // We prevent that scaling by calculating a size based on
                 // pixel size and dpi setting.
                 Size bitmapSize = new Size(bmp.PixelWidth, bmp.PixelHeight);
-                return WpfUnitsConversion.FromPixels(this, bitmapSize);
+                return WpfUnitsConversion.FromPixels(PresentationSource.FromVisual(this), bitmapSize);
```
MikhailArkhipov: So, according to below, 
```
var source = PresentationSource.FromVisual(Content as Visual);
if (source != null) {
```
PresentationSource.FromVisual can still return null?

MikhailArkhipov: Or it never does that for `this` ?

huguesv: It's never null here.

MikhailArkhipov:
👍

MikhailArkhipov:
Let me know when merged to 0_5, i can start the build

Issue #2374, opened 2016-09-20
Repositioning the plot toolwindow with drag & drop caused VS hang


huguesv:
It was stuck in the special transient UI that VS has for helping you re-arrange tool windows.

Happened on Shahrokh's machine. I put logs and dump here:
\\pytools\r\Bugs\plotwindowstuckdragdrop

MikhailArkhipov:
@huguesv - anything we need to take to 0.5 from here?

huguesv:
Not easy to repro, but possible.  I kept moving the toolwindow around, eventually it happened.  It looks like it's the drag & drop of plot themselves that interferes with the VS drag & drop of tool windows.

If I break into the debugger while it's stuck, I got:
```
 	[External Code]	
>	Microsoft.R.Components.dll!Microsoft.R.Components.Plots.Implementation.View.RPlotDeviceControl.Image_MouseMove(object sender, System.Windows.Input.MouseEventArgs e) Line 130	C#
 	[External Code]	
```

Stuck into this call:
```
DragDrop.DoDragDrop(this, obj, DragDropEffects.Copy | DragDropEffects.Move);
```

huguesv:
![image](https://cloud.githubusercontent.com/assets/1696845/18798859/55edb97e-8189-11e6-9a24-96f45d81d92f.png)

MikhailArkhipov:
I'd switch off just my code, enabled native and point symbols to `\\ddrps\symbols` and `\\symbols\symbols` - it is probably stuck somewhere in thread transitions

huguesv:
I'm inclined to say we take the fix I have for 0.5, linked above.  Once you know how to repro, it's actually pretty easy:
- Left click inside plot window.
- Left click on the "R Plot" tool window label and start dragging **up** such that you'll send MouseMove events to plot window.
- Plot window incorrectly thinks you started dragging a plot, and now you are multi tasking drag & drop :)

PR #2402, opened 2016-09-23
Fix drag drop of plot windows

huguesv:
It was being too aggressive, starting drag & drop operations when it shouldn't.  The most problematic of which is when it was starting a drag & drop of plot while VS was in toolwindow drag & drop mode.

Fixes https://github.com/Microsoft/RTVS/issues/2374

### DIFF
--- a/src/R/Components/Impl/Plots/Implementation/View/DragSurface.cs
+++ b/src/R/Components/Impl/Plots/Implementation/View/DragSurface.cs
@@ -7,15 +7,25 @@ using System.Windows.Input;
 
 namespace Microsoft.R.Components.Plots.Implementation.View {
     internal class DragSurface {
-        private Point _lastMousedownPosition;
+        private Point? _lastMousedownPosition;
 
         public void MouseDown(MouseEventArgs e) {
             _lastMousedownPosition = e.GetPosition(null);
         }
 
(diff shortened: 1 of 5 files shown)


src/R/Components/Impl/Plots/Implementation/View/DragSurface.cs
```diff
@@ -7,15 +7,25 @@
...
         }
 
+        public void MouseMove(MouseEventArgs e) {
+            if (e.MouseDevice.LeftButton == MouseButtonState.Released) {
+                _lastMousedownPosition = null;
+            }
+        }
+
+        public void MouseLeave(MouseEventArgs e) {
+            _lastMousedownPosition = null;
```
MikhailArkhipov: These event are generally not very reliable. You may or may not get mouse leave. At least in Win32 that was the case. Usually events get lost of mouse moves to another window. Ex: press down, drag quickly, release over another window. 'leave' and 'release' may never come. I remember I used to check on mouse enter and mouse move if button was still down and reset if it was not.

but this may or may not be happening these days.

MikhailArkhipov:
👍

MikhailArkhipov:
@jflam - do you want this in 0.5?

Issue #2403, opened 2016-09-23
When switching between connections, plots of previous connection remain


huguesv:
Right now we're cleanup up in the RSession.Disconnected event, so maybe we're not getting that event anymore when we switch between remote/local?  I suppose I can do it in RSession.Connected instead.

Note that this is not the same as https://github.com/Microsoft/RTVS/issues/2314, which is about supporting returning to an existing session.

MikhailArkhipov:
As happens to Help in #2321. Not sure about package manager though. @AlexanderSher - does it have facilities to refresh on new connection?

MikhailArkhipov:
Actually we should only clear when connecting to another session. Disconnecting and reconnecting to the same session should keep everything intact. Network reliability or the fact that they have to leave WiFi coverage should not be destroying user work.

@AlexanderSher  - is there a good even for this?

AlexanderSher:
If you mean switching from one broker to another - `BrokerChanged` will suffice.

huguesv:
Is there also a way to distinguish between lost connection + reconnect, versus the user hit Reset in the REPL?

Because we need to clear plots if they hit Reset, but we don't want to if they lose connection and reconnect.

PR #2404, opened 2016-09-23
Clear plot manager state in both connected and disconnected events.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/2403


MikhailArkhipov:
Actually we should *only* clear when connecting to another session. Disconnecting and reconnecting to the same session should keep everything intact. Network reliability or the fact that they have to leave WiFi coverage should not be destroying user work.

huguesv:
I'll close this and revisit it once the session reconnecting functionality is in.

PR #2406, opened 2016-09-26
Replace logging for large blobs with blob length and hash. 

### LINKED ISSUE #2394: CreateBlob_DisconnectedDuringCreate fails with OOM when run in group
Run all tests

image

karthiknadig:
Fixes #2394

### DIFF
--- a/src/Host/Protocol/Impl/Message.cs
+++ b/src/Host/Protocol/Impl/Message.cs
@@ -5,13 +5,14 @@ using System;
 using System.Text;
 using System.Diagnostics;
 using System.IO;
-using System.Runtime;
+using System.Security.Cryptography;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static System.FormattableString;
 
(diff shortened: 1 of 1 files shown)


MikhailArkhipov:
For huge blobs, what is the performance impact when calculating hashes? Just making sure logging is lightweight...

src/Host/Protocol/Impl/Message.cs
```diff
@@ -92,7 +93,14 @@ public class Message {
                 result += $":#{RequestId}#";
             }
 
-            var blobString = BitConverter.ToString(Blob);
+            string blobString = string.Empty;
```
int19h: This doesn't need the initializer, since it's assigned in both branches below.

int19h:
This is VS-side logging, not broker-side, so I wouldn't anticipate a big impact - it only needs to hash these for one user.

MD5 is very lightweight - around 300-400 megabytes/second on your typical laptop or desktop.

int19h:
👍

PR #2408, opened 2016-09-28
Logging changes

MikhailArkhipov:


src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -84,5 +84,15 @@ public static class IOExtensions {
             }
             return path.Length;
         }
+
+        public static string EnsureTrailingSlash(this string path) {
+            if (path != null &&  path.Length > 0) {
```
AlexanderSher: `!string.IsNullOrEmpty`
Also, do we want to check for alternate separator?

src/Common/Core/Impl/Logging/Logger.cs
```diff
@@ -0,0 +1,88 @@
...
+        private readonly IActionLogWriter _writer;
+        private readonly BinaryAsyncLock _initializationLock = new BinaryAsyncLock();
+
+        public void Dispose() {
+            foreach (var log in _logs) {
+                (log as IDisposable)?.Dispose();
+            }
+        }
+
+        internal Logger(string appName, ILoggingPermissions permissions, IActionLogWriter writer) {
```
AlexanderSher: So we have non-disposable logs?

src/Common/Core/Impl/Logging/Implementation/LoggingPermissions.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// Rules:
+    ///     a. Feedback and logging is permitted if telemetry is enabled. If telemetry is off, 
+    ///        admin can selectively enable feedback and logging in registry under HKLM.
+    ///     b. If telemetry is on, default logging level is 'Normal'. This allows recording of all events
+    ///        but security and privacy related data such as cross-process traffic or commands and data 
+    ///        in the interactive window.
+    ///     c. Admin can limit level of logging and disable feedback sending even if app telemetry is on.
+    /// </remarks>
+    [Export(typeof(ILoggingPermissions))]
+    internal sealed class LoggingPermissions : ILoggingPermissions {
```
AlexanderSher: Do we have any exports in Common.Core?

src/Common/Core/Impl/Services/ServiceBag.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Common.Core.Services {
+    public sealed class ServiceBag: IServiceBag {
```
AlexanderSher: What's the difference between `IServiceBag` and `ICoreShell`?

src/Common/Core/Impl/Shell/IApplicationConstants.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core.Shell {
+    /// <summary>
+    /// Defines application constants such as locale, registry key names, etc.
+    /// Implemented by the host application. Imported via MEF.
+    /// </summary>
+    public interface IApplicationConstants {
```
AlexanderSher: MEF or `ICoreShell` property?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -35,9 +39,10 @@ public class RSessionProvider : IRSessionProvider {
         public event EventHandler BrokerChanged;
         public event EventHandler<BrokerStateChangedEventArgs> BrokerStateChanged;
 
-        public RSessionProvider(IRSessionProviderCallback callback = null) {
+        public RSessionProvider(IActionLog log, IRSessionProviderCallback callback = null) {
```
AlexanderSher: Can we create a `NullActionLog` and make it a default value for `RSessionProvider`?

AlexanderSher: That will minimize amount of changes

---
In reply to: [80987120](https://github.com/Microsoft/RTVS/pull/2408#discussion_r80987120) [](ancestors = 80987120)

src/Common/Core/Impl/Extensions/IOExtensions.cs
MikhailArkhipov: It is interesting question. What if path is `a/b/c`, should we detect and use proper terminator?

MikhailArkhipov: I guess we can do that... If there is weird mix we'll just default to `\`

src/Common/Core/Impl/Logging/Logger.cs
MikhailArkhipov: They are disposed in ILoggingServices when package closes (in VsAppShell).

src/Common/Core/Impl/Logging/Implementation/LoggingPermissions.cs
MikhailArkhipov: It has been marked as MEF container in the manifest - I think since we had IFileSystem exported back then

src/Common/Core/Impl/OS/ProcessServices.cs
```diff
@@ -5,24 +5,12 @@
...
-
-        public void Start(ProcessStartInfo psi) {
-            Process.Start(psi);
+        public Process Start(ProcessStartInfo psi) {
+            return Process.Start(psi);
         }
 
-        public void Start(string path) {
-            Process.Start(path);
+        public Process Start(string path) {
```
MikhailArkhipov: It is huge and derives from Component... which derives from MarshalByRefObject and so on. I think no

src/Common/Core/Impl/Services/ServiceBag.cs
MikhailArkhipov: Basically in the next change I want to add ICoreServices that will hold IProcessServices, IFileSystem, IRegistry, IActionLog and ICoreShell (and more wrappers we care to add) so we can pass around fewer parameters. It is exported as ICoreServices.

MikhailArkhipov: We could, of course, add them all to the ICoreShell?

src/Common/Core/Impl/Shell/IApplicationConstants.cs
MikhailArkhipov: ICoreShell is getting bigger.... Almost 20 methods/properties. I want to split it into constants, shell related methods (show message box, progress, idle, app lifetime events) and basic/core services.

src/Common/Core/Impl/Services/ServiceBag.cs
AlexanderSher: It's a good question what should be the main entry point. `ICoreShell` is already imported through MEF. If we are adding `ICoreServices`, I'd rather make explicit properties for all the required types (`FileSystem`, `Registry`, `Log`, `Telemetry`) and add a `ICoreShell.CoreServices` property. 
`ICoreShell` itself requires a split - it contains methods that are related to the UI thread that should not be available in Common and RHost.Client, but that is a separate task.

---
In reply to: [80990180](https://github.com/Microsoft/RTVS/pull/2408#discussion_r80990180) [](ancestors = 80990180)

src/Common/Core/Impl/Shell/IApplicationConstants.cs
AlexanderSher: We already have `ApplicationWindowHandle` that I think can be moved into this interface (cause it won't change till the end of the process).

---
In reply to: [80990779](https://github.com/Microsoft/RTVS/pull/2408#discussion_r80990779) [](ancestors = 80990779)

src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -84,5 +84,16 @@ public static class IOExtensions {
             }
             return path.Length;
         }
+
+        public static string EnsureTrailingSlash(this string path) {
+            if (!string.IsNullOrEmpty(path)) {
+                char slash = path.IndexOf(Path.AltDirectorySeparatorChar) >= 0 ? Path.AltDirectorySeparatorChar : Path.DirectorySeparatorChar;
+                if (path[path.Length - 1] != slash) {
```
AlexanderSher: I think here we still need to check for both normal and alt, otherwise we may end up with something like `C:\b/a\/`

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
```diff
@@ -72,7 +72,8 @@ internal sealed class RProjectLoadHooks {
             , IRToolsSettings toolsSettings
             , IThreadHandling threadHandling
             , ISurveyNewsService surveyNews
-            , [Import(AllowDefault = true)] IProjectItemDependencyProvider dependencyProvider) {
+            , [Import(AllowDefault = true)] IProjectItemDependencyProvider dependencyProvider
+            , ICoreShell coreShell) {
```
AlexanderSher: if you import `ICoreShell`, can you please replace other usages of `VsAppShell.Current`?

MikhailArkhipov: Yes, except one `GetGlobalService`. I can either a) import AppShell instead or b) push `GetGlobalService` down to core.

src/Common/Core/Impl/Extensions/IOExtensions.cs
MikhailArkhipov: Frankly I wouldn't care much about mixes. Currently if there is `/` we get `/` otherwise `\`. If there is a mix, which one to use?

MikhailArkhipov: Or you mean if it ends with either just leave it alone? OK.

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: I'll do in the next round where this class will be receiving `ICoreServices`

src/Common/Core/Impl/Shell/IApplicationConstants.cs
MikhailArkhipov: Yes, I will move app window handle. I have further changes in line.

src/Common/Core/Impl/Extensions/IOExtensions.cs
AlexanderSher: Yes.

---
In reply to: [80995728](https://github.com/Microsoft/RTVS/pull/2408#discussion_r80995728) [](ancestors = 80995728)

src/Package/Impl/ProjectSystem/RProjectLoadHooks.cs
AlexanderSher: Let's just don't do anything with that `GetGlobalService` usage for now.

---
In reply to: [80995237](https://github.com/Microsoft/RTVS/pull/2408#discussion_r80995237) [](ancestors = 80995237)

MikhailArkhipov:
Merged with other changes regarding services and telemetry

src/Common/Core/Impl/Logging/Implementation/LoggingServices.cs
```diff
@@ -0,0 +1,34 @@
...
+            Permissions = permissions;
+            _appConstants = appConstants;
+        }
+
+        public ILoggingPermissions Permissions { get; }
+
+        public IActionLog GetOrCreateLog(string appName) {
+            if (_instance == null) {
+                var instance = new Logger(_appConstants.ApplicationName, Permissions, writer: null);
+                Interlocked.Exchange(ref _instance, instance);
```
AlexanderSher: You need to use `CompareExchange` in this case

src/Common/Core/Impl/Services/CoreServices.cs
```diff
@@ -0,0 +1,53 @@
...
+    public sealed class CoreServices : ICoreServices {
+        private readonly IApplicationConstants _appConstants;
+        private readonly ILoggingServices _loggingServices;
+        private IActionLog _log;
+
+        [ImportingConstructor]
+        public CoreServices(
+              IApplicationConstants appConstants
+            , ITelemetryService telemetry
+            , ILoggingServices loggingServices
```
AlexanderSher: Do we really need to import `ILoggingServices`? From the implementation, it looks quite simple

src/Common/Core/Impl/Services/ServiceBag.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Common.Core.Services {
+    public class ServiceBag: IServiceBag {
```
AlexanderSher: I can't find any usages of this class. Do you still need it?

src/Common/Core/Impl/Services/CoreServices.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.ComponentModel.Composition;
+using Microsoft.Common.Core.IO;
+using Microsoft.Common.Core.Logging;
+using Microsoft.Common.Core.OS;
+using Microsoft.Common.Core.Shell;
+using Microsoft.Common.Core.Telemetry;
+
+namespace Microsoft.Common.Core.Services {
+    [Export(typeof(ICoreServices))]
```
AlexanderSher: Do you import it anywhere separately from `ICoreShell`?

MikhailArkhipov: Some code may want it - they may just need basics without shell stuff.

src/Common/Core/Test/Shell/TestCoreServices.cs
```diff
@@ -0,0 +1,37 @@
...
+                null,
+                Substitute.For<IActionLog>(),
+                fs ?? Substitute.For<IFileSystem>(),
+                registry ?? Substitute.For<IRegistry>(),
+                ps ?? Substitute.For<IProcessServices>());
+        }
+
+        public static ICoreServices CreateReal() {
+            return new CoreServices(
+                Substitute.For<IApplicationConstants>(),
```
AlexanderSher: Why you don't use `TestAppConstants` here?

src/Common/Core/Impl/Services/ServiceBag.cs
MikhailArkhipov: Not really. We can probably remove it for now.

src/Common/Core/Impl/Services/CoreServices.cs
AlexanderSher: Ok

---
In reply to: [81219300](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81219300) [](ancestors = 81219300)

src/Common/Core/Impl/Services/CoreServices.cs
MikhailArkhipov: The trick is to delay-create log. AppShell is initialized before settings are loaded, but log needs to be created when permissions (partially come from settings) get known. So currently it is imported by VsAppShell which sets the setting. Then actual log is created when someone calls Log on CoreServices.

src/Common/Core/Impl/Threading/IMainThread.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.Common.Core.Threading {
+    public interface IMainThread {
```
AlexanderSher: Do you need this in tests?

src/Common/Core/Impl/Services/CoreServices.cs
AlexanderSher: Got it

---
In reply to: [81220028](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81220028) [](ancestors = 81220028)

src/Common/Core/Test/Shell/TestCoreServices.cs
MikhailArkhipov: Good point

src/Common/Core/Impl/Threading/IMainThread.cs
MikhailArkhipov: It used to be in Components.Wpf, but looks generic enough to be in Core

src/Common/Core/Impl/Services/CoreServices.cs
AlexanderSher: Ok, not entirely.

`LoggingServices` requires `IApplicationConstants` and `ILoggingPermissions`. The last one is implemented by `LoggingPermissions` that require `IApplicationConstants`, `ITelemetryService` and `IRegistry`, so all inputs are available in the `CoreServices` constructor, so you can create an parameter with `[Import(AllowDefault = true)]` and create an instance of `LoggingServices` without exporting. It will also guarantee that both `LoggingPermissions` and `CoreServices` use the same instance of `IRegistry`

---
In reply to: [81220421](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81220421) [](ancestors = 81220421,81220028)

MikhailArkhipov: Ah, yes. Although I didn't want to expose  ILoggingServices via CoreServices (it currently only exposes actual log)

src/Common/Core/Impl/Threading/IMainThread.cs
AlexanderSher: That's a good question. Core is referenced by Broker, which has no access to the UI and to concept of Main Thread. Host.Client also should have no idea about UI. We may eventually have to add assembly, something like `Common.Wpf` or `Common.UI` or `Core.UI` to store such things, and also move several methods from ICoreShell there as well.

---
In reply to: [81221364](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81221364) [](ancestors = 81221364)

src/Common/Core/Test/Shell/TestCoreShell.cs
```diff
@@ -8,13 +8,14 @@
...
+using Microsoft.Common.Core.Settings;
 using Microsoft.Common.Core.Shell;
-using Microsoft.Common.Core.Telemetry;
-using Microsoft.Common.Wpf.Threading;
+using Microsoft.Common.Core.Threading;
 using Microsoft.UnitTests.Core.Threading;
+using NSubstitute;
 
-namespace Microsoft.R.Components.Test.Fakes.Shell {
+namespace Microsoft.Common.Core.Test.Shell {
```
AlexanderSher: I'd prefer to have all fakes in a separate folder, otherwise it may be hard to find them, or misuse instead of stubs and visa versa

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -68,7 +69,8 @@ public class SessionManager {
                 password,
                 profilePath,
                 _loggingOptions.LogHostOutput ? _hostOutputLogger : null,
-                _loggingOptions.LogPackets ? _messageLogger : null);
+                _loggingOptions.LogPackets ? _messageLogger : null,
+                _loggingOptions.LogPackets ? LogVerbosity.Traffic : LogVerbosity.Minimal);
```
AlexanderSher: Is it possible that `LogHostOutput` is true, but `LogPackets` is false?

MikhailArkhipov: I think yes. Packets is actual data, but host also outputs basic stuff

AlexanderSher: Do we need a special verbosity for such case?

---
In reply to: [81224865](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81224865) [](ancestors = 81224865)

src/Common/Core/Test/Shell/TestCoreShell.cs
MikhailArkhipov: OK, moved

src/Host/Broker/Impl/Sessions/SessionManager.cs
MikhailArkhipov: There are 3 level and when logging something code can specify which min verbosity the message appears at. This is different from general concept of verbosity as in msbuild, for example. Regular concept is trace -> detailed verbosity, warning -> regular, error -> minimal. However, we also need to take into account privacy and security so level/severity is different from verbosity. 'Traffic' means that log entry may contain traffic data and if traffic logging is off, traffic-related errors won't get logged.

int19h: `LogHostOutput` is for host process stderr. This includes the packets, which the host logs, but it shows host's view of what it gets.

`LogPackets` is for the packets as they flow through the broker. 

So yes, these two are completely independent, but both involve traffic.

AlexanderSher: If they both include traffic, shouldn't we set verbosity to `Traffic` when any of those properties are true?

---
In reply to: [81228732](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81228732) [](ancestors = 81228732)

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -28,29 +28,24 @@ internal abstract class BrokerClient : IBrokerClient {
             TimeSpan.FromSeconds(5);
 #endif
 
-        protected DisposableBag DisposableBag { get; } = DisposableBag.Create<BrokerClient>();
-
-        private readonly LinesLog _log;
         private readonly string _interpreterId;
         private AboutHost _aboutHost;
 
+        protected DisposableBag DisposableBag { get; } = DisposableBag.Create<BrokerClient>();
```
AlexanderSher: Do you use this property anywhere?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -150,18 +168,45 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
...
-            _disposed = true;
-
-            _debuggerModeTracker.EnterBreakMode -= DebuggerEnteredBreakMode;
-            _debuggerModeTracker.LeaveBreakMode -= DebuggerLeftBreakMode;
-            _activeTextViewTracker.LastActiveTextViewChanged -= LastActiveTextViewChanged;
-            RSession.Disconnected -= RSessionDisconnected;
-            Operations.Dispose();
-            Connections.Dispose();
+            _disposableBag.TryDispose();
             _onDispose();
```
AlexanderSher: You can add `_onDispose` to the bag as well

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -150,18 +168,45 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
...
+            } else if (e.PropertyName == nameof(IRSettings.RCodePage)) {
+                SetSessionCodePage().DoNotWait();
+            } 
+        }
+
+        private async Task SetMirrorToSession() {
+            string mirrorName = _settings.CranMirror;
+            string mirrorUrl = CranMirrorList.UrlFromName(mirrorName);
+
+            foreach (var s in RunningSessions) {
```
AlexanderSher: Session may start or stop at any time. I'd rather iterate through all sessions - we catch all the exceptions anyway.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -150,18 +168,45 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
...
+                try {
+                    using (var eval = await s.BeginEvaluationAsync()) {
+                        await eval.SetVsCranSelectionAsync(mirrorUrl);
```
AlexanderSher: >SetVsCranSelectionAsync [](start = 35, length = 23)

I think you can call `SetVsCranSelectionAsync` on a session itself, without `BeginEvaluationAsync`

src/R/Components/Impl/Settings/IRSettings.cs
```diff
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.ComponentModel;
+using Microsoft.Common.Core.Logging;
 using Microsoft.R.Components.ConnectionManager;
 
 namespace Microsoft.R.Components.Settings {
-    public interface IRSettings {
+    public interface IRSettings: INotifyPropertyChanged {
```
AlexanderSher: This is not a view model, so I'd rather use our own event instead of `INotifyPropertyChanged`

src/R/Components/Impl/History/Implementation/RHistoryProvider.cs
```diff
@@ -24,7 +24,12 @@ internal class RHistoryProvider : IRHistoryProvider {
         private readonly Dictionary<ITextBuffer, IRHistory> _histories;
 
         [ImportingConstructor]
-        public RHistoryProvider(ITextBufferFactoryService textBufferFactory, IContentTypeRegistryService contentTypeRegistryService, IEditorOperationsFactoryService editorOperationsFactory, IRtfBuilderService rtfBuilderService, ITextSearchService2 textSearchService, IRSettings settings) {
+        public RHistoryProvider(ITextBufferFactoryService textBufferFactory 
+            , IContentTypeRegistryService contentTypeRegistryService
+            , IEditorOperationsFactoryService editorOperationsFactory
+            , IRtfBuilderService rtfBuilderService
```
AlexanderSher: Is it just formatting?

src/Package/Test/Feedback/FeedbackCommandTest.cs
```diff
@@ -0,0 +1,55 @@
...
+            TestStatus(cmd);
+        }
+
+        [Test]
+        public void SendSmile() {
+            var cmd = new SendSmileCommand(_lp, _services);
+            TestStatus(cmd);
+        }
+
+        private void TestStatus(OleMenuCommand cmd) {
```
AlexanderSher: We have `MenuCommandAssertions`, which takes care of updating status. It is used in `HelpOnCurrentTest.HelpTest`.

src/Package/Impl/Shell/VsAppShell.cs
```diff
@@ -47,9 +50,29 @@ namespace Microsoft.VisualStudio.R.Package.Shell {
...
+            , IRSettings settings
+            , ICoreServices coreServices
+            , IApplicationConstants appConstants) {
+
+            _telemetryService = telemetryService;
+            _coreServices = coreServices;
+            _appConstants = appConstants;
+
+            _settings = settings;
+            _settings.PropertyChanged += OnSettingsChanged;
```
AlexanderSher: Maybe add another event (`SettingsInitialized` or something like that) that will be called only once?

src/Package/Impl/Packages/R/PackageCommands.cs
```diff
@@ -48,19 +50,21 @@ internal static class PackageCommands {
             var wbs = exportProvider.GetExportedValue<IWebBrowserServices>();
             var pcsp = exportProvider.GetExportedValue<IProjectConfigurationSettingsProvider>();
             var dbcs = exportProvider.GetExportedValue<IDbConnectionService>();
+            var logPerms = exportProvider.GetExportedValue<ILoggingPermissions>();
+            var ps = new ProcessServices();
```
AlexanderSher: `ProcessServices` is available in `appShell.Services`

src/R/Components/Impl/Settings/IRSettings.cs
MikhailArkhipov: it is just System.ComponentModel, pretty basic stuff, it doesn't have much to do with WPF. 

---
In reply to: [81233986](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81233986) [](ancestors = 81233986)

src/Host/Client/Impl/Host/BrokerClient.cs
MikhailArkhipov: I believe derived classes used it

---
In reply to: [81231658](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81231658) [](ancestors = 81231658)

src/R/Components/Impl/History/Implementation/RHistoryProvider.cs
MikhailArkhipov: Yes

---
In reply to: [81234076](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81234076) [](ancestors = 81234076)

src/Package/Impl/Shell/VsAppShell.cs
MikhailArkhipov: We don't know if they are getting loaded, LoadSettings basically just instantiates the page and then calls 'set' for every property.

---
In reply to: [81236099](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81236099) [](ancestors = 81236099)

src/Common/Core/Impl/Threading/IMainThread.cs
MikhailArkhipov: Well in Windows there is always 'main thread' as long as there is a window - which is almost always true. 

---
In reply to: [81223138](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81223138) [](ancestors = 81223138,81221364)

AlexanderSher: Even in console application and service?

---
In reply to: [81241433](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81241433) [](ancestors = 81241433,81223138,81221364)

src/Host/Client/Impl/Host/BrokerClient.cs
AlexanderSher: yeah, found it

---
In reply to: [81240024](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81240024) [](ancestors = 81240024,81231658)

src/R/Components/Impl/Settings/IRSettings.cs
AlexanderSher: Ok

---
In reply to: [81239440](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81239440) [](ancestors = 81239440,81233986)

src/Package/Impl/Shell/VsAppShell.cs
AlexanderSher: Then how do we know if it was set for the first time?

---
In reply to: [81240939](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81240939) [](ancestors = 81240939,81236099)

src/Common/Core/Impl/Threading/IMainThread.cs
MikhailArkhipov: Console app, yes - it has a window (the thread that runs the message pump). It depends what you are going to call. You do can mark code as STA or even have to have to do it if your app consumes some COM object marked as STA. 

---
In reply to: [81241676](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81241676) [](ancestors = 81241676,81241433,81223138,81221364)

src/Package/Impl/Shell/VsAppShell.cs
MikhailArkhipov: This is main shell so it is reasonable to assume that it's ctor is called before everything else.

---
In reply to: [81242220](https://github.com/Microsoft/RTVS/pull/2408#discussion_r81242220) [](ancestors = 81242220,81240939,81236099)

AlexanderSher:
:shipit:

MikhailArkhipov:
@int19h - do you want to merge something before this so you won't have to resolve conflicts and rather have me do it :-) ?

int19h:
No worries, I'll do the merge on my end. Still working on getting that test pass complete anyway... And this doesn't look like it's going to have a lot of conflicts with my changes.

Issue #2387, opened 2016-09-22
Session lifetime, dehydration and rehydration - broker/VS


int19h:
Sub-item of https://github.com/Microsoft/RTVS/issues/2176:

> The broker's role is mainly in converting configuration options to command line arguments for newly spawned processes, and recording the shutdown mode (whether state was saved or not), to provide this information in the list of sessions, and to provide the option to restore state for eligible sessions when they're started next.

int19h:
Partially implemented in https://github.com/Microsoft/RTVS/pull/2413

PR #2413, opened 2016-09-30
Session management and reconnection (RTVS)

int19h:
When connecting to a host, query broker for any existing session with the same ID, and use it when available.

Add basic version checking for rtvs package when (re)connecting, to ensure that a compatible package version is used.

Fix REPL Reset not hard-killing the host process if q() doesn't worl.

Fix session not properly recorded as terminated when host process goes down.

Fix cancellation handling when recording requests without responses to replay during a future reconnect.

Fix websocket tunnel lifetime issues:
- close tunnel when host process goes down;
- use cooperative closure, so that the client is immediately aware of it;
- properly close the tunnel from VS end in all termination scenarios.

Refactor HTTP credentials handling to make it reusable for different endpoints.

### DIFF
--- a/src/Host/Broker/Impl/Sessions/Session.cs
+++ b/src/Host/Broker/Impl/Sessions/Session.cs
@@ -22,10 +22,10 @@ namespace Microsoft.R.Host.Broker.Sessions {
     public class Session {
         private const string RHostExe = "Microsoft.R.Host.exe";
 
-        private static readonly byte[] _endMessage;
         private readonly ILogger _sessionLogger;
         private Process _process;
-        private volatile MessagePipe _pipe;
+        private MessagePipe _pipe;
+        private volatile IMessagePipeEnd _hostEnd;
 
         public SessionManager Manager { get; }
 
@@ -40,6 +40,21 @@ namespace Microsoft.R.Host.Broker.Sessions {
 
         public string CommandLineArguments { get; }
 
+        private volatile SessionState _state;
+
+        public SessionState State {
+            get {
+                return _state;
+            }
+            set {
+                var oldState = _state;
+                _state = value;
+                StateChanged?.Invoke(this, new SessionStateChangedEventArgs(oldState, value));
+            }
+        }
+
+        public event EventHandler<SessionStateChangedEventArgs> StateChanged;
+
         public Process Process => _process;
 
         public SessionInfo Info => new SessionInfo {
@@ -48,31 +63,22 @@ namespace Microsoft.R.Host.Broker.Sessions {
             CommandLineArguments = CommandLineArguments,
         };
 
-        static Session() {
-            using (var stream = new MemoryStream()) {
-                using (var writer = new BinaryWriter(stream, Encoding.UTF8, leaveOpen: true)) {
-                    writer.Write(ulong.MaxValue - 1);
-                    writer.Write(0UL);
-                    writer.Write("!End".ToCharArray());
-                    writer.Write((byte)0);
-                    writer.Write("[]".ToCharArray());
-                    writer.Write((byte)0);
-                }
-
-                _endMessage = stream.ToArray();
-            }
-        }
-
-        internal Session(SessionManager manager, IIdentity user, string id, Interpreter interpreter, string commandLineArguments, ILogger sessionLogger) {
+        internal Session(SessionManager manager, IIdentity user, string id, Interpreter interpreter, string commandLineArguments, ILogger sessionLogger, ILogger messageLogger) {
             Manager = manager;
             Interpreter = interpreter;
             User = user;
             Id = id;
             CommandLineArguments = commandLineArguments;
             _sessionLogger = sessionLogger;
+
+            _pipe = new MessagePipe(messageLogger);
         }
 
-        public void StartHost(SecureString password, string profilePath, ILogger outputLogger, ILogger messageLogger, LogVerbosity verbosity) {
+        public void StartHost(SecureString password, string profilePath, ILogger outputLogger, LogVerbosity verbosity) {
+            if (_hostEnd != null) {
+                throw new InvalidOperationException("Host process is already running");
+            }
+
             string brokerPath = Path.GetDirectoryName(typeof(Program).Assembly.GetAssemblyPath());
             string rhostExePath = Path.Combine(brokerPath, RHostExe);
             string arguments = $"--rhost-name \"{Id}\" --rhost-log-verbosity {(int)verbosity} {CommandLineArguments}";
@@ -146,7 +152,9 @@ namespace Microsoft.R.Host.Broker.Sessions {
             };
 
             _process.Exited += delegate {
-                _pipe = null;
+                _hostEnd?.Dispose();
+                _hostEnd = null;
+                State = SessionState.Terminated;
             };
 
             _sessionLogger.LogInformation(Resources.Info_StartingRHost, Id, User.Name, rhostExePath, arguments);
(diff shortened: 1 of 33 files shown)


int19h:
Corresponding R-Host PR: https://github.com/Microsoft/R-Host/pull/91

int19h:
Note that this includes a bunch of stuff to support session rehydration (https://github.com/Microsoft/RTVS/issues/2387) in the future, but none of it is active right now. This is for the basic "reconnect to a live session" scenario. I'd like to keep the rest of the code in, though, for future work on more advanced reconnect scenarios.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -53,26 +69,32 @@ public class Session {
...
                     writer.Write(0UL);
-                    writer.Write("!End".ToCharArray());
+                    writer.Write("!Shutdown".ToCharArray());
                     writer.Write((byte)0);
-                    writer.Write("[]".ToCharArray());
+                    writer.Write("[false]".ToCharArray());
                     writer.Write((byte)0);
                 }
 
                 _endMessage = stream.ToArray();
```
int19h: This is actually unused now. I will remove it completely.

src/Host/Broker/Impl/Sessions/SessionState.cs
```diff
@@ -1,9 +1,10 @@
...
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System;
-
-namespace Microsoft.R.Host.Broker.Pipes {
-    public interface IOwnedMessagePipeEnd : IMessagePipeEnd, IDisposable {
+namespace Microsoft.R.Host.Broker.Sessions {
+    public enum SessionState {
+        Running,
+        Dormant,
```
int19h: Never used currently. Will be used for sessions that have saved state in the future.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -108,9 +110,29 @@ internal abstract class BrokerClient : IBrokerClient {
         public virtual async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000, CancellationToken cancellationToken = new CancellationToken()) {
             DisposableBag.ThrowIfDisposed();
             await TaskUtilities.SwitchToBackgroundThread();
+
             try {
-                await CreateBrokerSessionAsync(name, rCommandLineArguments, cancellationToken);
-                var webSocket = await ConnectToBrokerAsync(name, cancellationToken);
+                bool sessionExists = await IsSessionRunningAsync(name);
```
int19h: This is the one line that will need to change to completely disable the reconnect feature, if we decide to not do it for this milestone. Everything else can remain in.

int19h:
There are no tests for any of this currently. We need to discuss what needs to be tested first, and how to split that work up.

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -15,71 +15,123 @@
...
-        public WebService(HttpClient httpClient) {
+        public WebService(HttpClient httpClient, ICredentialsProvider credentialsProvider) {
             HttpClient = httpClient;
+            _credentialsProvider = credentialsProvider;
         }
 
         private static HttpResponseMessage EnsureSuccessStatusCode(HttpResponseMessage response) {
-            if (response.StatusCode == HttpStatusCode.Unauthorized || response.StatusCode == HttpStatusCode.Forbidden) {
-                throw new UnauthorizedAccessException();
+            bool disposeResponse = true;
```
AlexanderSher: I think you can do it without `disposeResponse`:
```{csharp}
try {
  * * *
} catch (Exception) {
    response.Dispose();
    throw;
}
```

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -15,71 +15,123 @@
...
+                    if (isValidCredentials != null) {
+                        _credentialsProvider.OnCredentialsValidated(isValidCredentials.Value);
                     }
                 }
             }
+        }
 
-            return response.EnsureSuccessStatusCode();
+        private async Task RepeatUntilAuthenticatedAsync(Func<Task> action) {
+            await RepeatUntilAuthenticatedAsync(async () => {
```
AlexanderSher: you don't need this await

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -15,71 +15,123 @@
...
+                await action();
+                return false;
+            });
         }
 
         public async Task<TResponse> HttpGetAsync<TResponse>(Uri uri, CancellationToken cancellationToken = default(CancellationToken)) {
-            var resonse = await HttpClient.GetAsync(uri, cancellationToken);
-            return JsonConvert.DeserializeObject<TResponse>(await resonse.Content.ReadAsStringAsync());
+            using (var response = await RepeatUntilAuthenticatedAsync(async () => EnsureSuccessStatusCode(await HttpClient.GetAsync(uri, cancellationToken)))) {
+                return JsonConvert.DeserializeObject<TResponse>(await response.Content.ReadAsStringAsync());
```
AlexanderSher: Can we get `UnauthorizedAccessException` from `ReadAsStringAsync`?

src/Host/Client/Impl/BrokerServices/WebService.cs
```diff
@@ -15,71 +15,123 @@
...
-            var responseBody = await response.Content.ReadAsStringAsync();
-            try {
-                return JsonConvert.DeserializeObject<TResponse>(responseBody);
-            } catch(JsonSerializationException ex) {
-                throw new ProtocolViolationException(ex.Message);
+
+            using (var response = await RepeatUntilAuthenticatedAsync(async () => {
+                var content = new StringContent(requestBody, Encoding.UTF8, "application/json");
+                return EnsureSuccessStatusCode(await HttpClient.PutAsync(uri, content, cancellationToken));
+            })) {
```
AlexanderSher: Open brace?

src/Host/Client/Impl/BrokerServices/WebService.cs
int19h: No, by the time you get to it, the HTTP response status code and headers have already been checked. This is just reading the body of the response.

src/Host/Client/Impl/BrokerServices/WebService.cs
int19h: Indeed. This is leftover from when I tried to use `return` for something, I think. But it should never return a response if there was an error, so `catch` will be better.

src/Host/Client/Impl/BrokerServices/WebService.cs
int19h: Not sure what the comment refers to? `using` needs a brace here (and response needs to be disposed - we haven't actually been doing this properly before, which might account for some of memory leaks). 

I guess I could declare and initialize `response` outside of `using`, and then just do `using (response) { ... }`. Would this be preferable readability-wise?

AlexanderSher: Well, probably extracting the body of the `RepeatUntilAuthenticatedAsync` argument will make it easier

---
In reply to: [81414506](https://github.com/Microsoft/RTVS/pull/2413#discussion_r81414506) [](ancestors = 81414506)

int19h:
Addressed feedback and resolved merge conflicts.

int19h:
Ping?

AlexanderSher:
👍

Issue #2415, opened 2016-09-30
Switching to offline remote connection doesn't produce error


MikhailArkhipov:
```
            var response = EnsureSuccessStatusCode(await HttpClient.PutAsync(uri, content, cancellationToken));
```

throws TaskCanceledException (rather than 404) which we handle and the offline connection gets lighted green which actually previous connection remains active.

```
            } catch (Exception ex) {
                _brokerProxy.Set(oldBroker);
                brokerClient.Dispose();
                BrokerChangeFailed?.Invoke(this, EventArgs.Empty);
                if (ex is OperationCanceledException || ex is RHostBrokerBinaryMissingException) { // RHostDisconnectedException is derived from OperationCanceledException
                    return false;
                }
                throw;
```

gets executed in `TrySwitchBrokerAsync`

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7235095f9d8517e3b869ea588d95700b8ac8fdf3

Issue #2395, opened 2016-09-23
Connection diagnostics is unreliable


MikhailArkhipov:
Over VPN, broker is not running: 

- RD session does not have problems
- ping indicates some issues, 25% loss
- RTVS says host is offline

We should improve diagnostics and messaging since RDP appears to be fine. We should be able to say machine is actually available. When I start the broker, we do can establish the connection.

So we should clearly separate 'machine is offline' from 'broker is not running'

![image](https://cloud.githubusercontent.com/assets/12820357/18791159/71a514d2-8166-11e6-928f-7c6b3739fa99.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7235095f9d8517e3b869ea588d95700b8ac8fdf3

Issue #2321, opened 2016-09-13
Help window should reset when switching sessions


int19h:
If help for something is brought up in a remote session, and then session is switched to local, the Help window continues to show the original topic from remote session. Furthermore, any links in the window will also open pages from remote session. This is observable when remote and local R versions differ (help pages include R version in the footer). 

Help should listen to session connected/disconnected events, same as other toolwindows, and reset itself accordingly.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7235095f9d8517e3b869ea588d95700b8ac8fdf3

Issue #2025, opened 2016-06-29
Auto-formatting add extra space at line break


Marchen:
When I hit ENTER between `something,` and `family` of the following code,
```R
glm(
	y ~ x, data = something,family = Gamma)
```

Auto-formatting puts extra space before `family`.
```R
glm(
	y ~ x, data = something,
	 family = Gamma)
```

If I undo, there is a step between the two:
```R
glm(
	y ~ x, data = something, family = Gamma)
```

MikhailArkhipov:
This is fairly normal in auto-format cases as it sometimes yields two undo steps. It is same as in C#, for example. The reason is that first editor introduces line break and then calls for formatting. VS does have compound undo actions where multiple changes can be wrapped in a single undo, but in this particular case it is hard to determine when to open such action since particular keystroke may or may not be followed by additional action.

Marchen:
Thanks for the explanation. I understood undo steps of auto-format in VS. But my point was that maybe because of order of sequential actions of auto-format (hit RETURN -> insert a space -> put line break and adjust indent), an extra space was put on the beginning of new line. In the example above, indent of the line start with `y ~ x` consists of a TAB, but that of the line start with `family` consists of a TAB and SPACE.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7235095f9d8517e3b869ea588d95700b8ac8fdf3

PR #2416, opened 2016-10-01
#2025, #2321, #2395, #2415

MikhailArkhipov:
#2025 Auto-formatting add extra space at line break
#2321 Help window should reset when switching sessions
#2415 Switching to offline remote connection doesn't produce error
#2395 Connection diagnostics is unreliable

### DIFF
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -89,11 +89,16 @@ namespace Microsoft.R.Host.Client.Host {
 
         public async Task PingAsync() {
             if (HttpClient != null) {
-                await CheckMachineOnlineAsync();
+                var isOnline = await CheckMachineOnlineAsync();
+                // Just in case ping was disable for security reasons, try connecting to the broker anyway.
                 try {
                     (await HttpClient.PostAsync("/ping", new StringContent(""))).EnsureSuccessStatusCode();
                 } catch (HttpRequestException ex) {
-                    throw new RHostDisconnectedException(Resources.Error_BrokerNotRunning, ex);
+                    if (isOnline) {
+                        throw new RHostDisconnectedException(Resources.Error_BrokerNotRunning, ex);
+                    } else {
+                        throw new RHostDisconnectedException(Resources.Error_HostNotResponding, ex);
+                    }
                 }
             }
         }
@@ -215,14 +220,13 @@ namespace Microsoft.R.Host.Client.Host {
             return url;
         }
 
-        private async Task CheckMachineOnlineAsync() {
+        private async Task<bool> CheckMachineOnlineAsync() {
             if (!Uri.IsFile) {
                 var ping = new Ping();
                 var reply = await ping.SendPingAsync(Uri.Host, 5000);
-                if (reply.Status != IPStatus.Success) {
-                    throw new RHostDisconnectedException(Resources.Error_HostNotResponding);
-                }
+                return reply.Status == IPStatus.Success;
             }
+            return true;
         }
     }
 }
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -214,10 +214,12 @@ namespace Microsoft.R.Host.Client.Host {
         }
 
         private async Task CheckMachineOnlineAsync() {
-            var ping = new Ping();
-            var reply = await ping.SendPingAsync(Uri.Host, 5000);
-            if (reply.Status != IPStatus.Success) {
-                throw new RHostDisconnectedException(Resources.Error_HostNotResponding);
+            if (!Uri.IsFile) {
+                var ping = new Ping();
+                var reply = await ping.SendPingAsync(Uri.Host, 5000);
+                if (reply.Status != IPStatus.Success) {
+                    throw new RHostDisconnectedException(Resources.Error_HostNotResponding);
+                }
             }
         }
     }
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -87,6 +87,7 @@ namespace Microsoft.R.Host.Client.Host {
         protected abstract void OnCredentialsValidated(bool isValid);
 
         public async Task PingAsync() {
+            await CheckMachineOnlineAsync();
             try {
                 (await HttpClient.PostAsync("/ping", new StringContent(""))).EnsureSuccessStatusCode();
             } catch (HttpRequestException ex) {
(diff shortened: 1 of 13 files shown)


src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -210,5 +214,15 @@ internal abstract class BrokerClient : IBrokerClient {
         public virtual string HandleUrl(string url, CancellationToken ct) {
             return url;
         }
+
+        private async Task CheckMachineOnlineAsync() {
+            if (!Uri.IsFile) {
+                var ping = new Ping();
+                var reply = await ping.SendPingAsync(Uri.Host, 5000);
```
AlexanderSher: ICMP Echo Request may be disabled for security reasons. I don't think we can rely on it outside the corporate network. Even if ping fails, we need to check if we can connect to the socket.

src/Host/Client/Impl/Host/BrokerClientProxy.cs
```diff
@@ -26,10 +26,10 @@ internal sealed class BrokerClientProxy : IBrokerClient {
             broker.Dispose();
         }
 
-        public string Name => _broker.Name;
-        public bool IsRemote => _broker.IsRemote;
-        public Uri Uri => _broker.Uri;
-        public AboutHost AboutHost => _broker.AboutHost;
+        public string Name => _broker?.Name;
```
AlexanderSher: Broker should never be null until it is disposed. So do we have calls to broker after its disposal?

MikhailArkhipov: I hit hull ref here

MikhailArkhipov: Specifically in 

```
        private void OnSessionConnectionStateChange(bool connected) {
            IsConnected = connected;
            UpdateActiveConnection();
            ConnectionStateChanged?.Invoke(this, new ConnectionEventArgs(IsConnected, ActiveConnection));
        }
```

which technically may come, i think, after broker was removed

src/Host/Client/Impl/Host/BrokerClient.cs
MikhailArkhipov: Makes sense. I'll change it so it returns bool and then we decide which exception to throw

src/Host/Client/Impl/Host/BrokerClientProxy.cs
AlexanderSher: Yes, that's true. In this case, we need to change `Dispose()` method:
```{csharp}
        public void Dispose() {
            var broker = Interlocked.Exchange(ref _broker, new NullBrokerClient());
            broker.Dispose();
        }
```

---
In reply to: [81584799](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81584799) [](ancestors = 81584799)

AlexanderSher: Sure

---
In reply to: [81586637](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81586637) [](ancestors = 81586637,81585142,81584799)

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -87,10 +88,18 @@ internal abstract class BrokerClient : IBrokerClient {
         protected abstract void OnCredentialsValidated(bool isValid);
 
         public async Task PingAsync() {
-            try {
-                (await HttpClient.PostAsync("/ping", new StringContent(""))).EnsureSuccessStatusCode();
-            } catch (HttpRequestException ex) {
-                throw new RHostDisconnectedException(Resources.Error_HostNotResponding.FormatInvariant(ex.Message), ex);
+            if (HttpClient != null) {
+                var isOnline = await CheckMachineOnlineAsync();
```
AlexanderSher: I thin you can add the same code to the `CreateBrokerSessionAsync` as well (and in this case you don't need to call `Ping` explicitly in `TrySwitchBrokerAsync`)

AlexanderSher: `ConnectAsync` is called not only when user switches session, but also when he tries to reconnect to the current one

---
In reply to: [81587164](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81587164) [](ancestors = 81587164)

MikhailArkhipov: PingAsync is a bit of a perf hit so I didn't want to put it in the high traffic area. But sure.

---
In reply to: [81587401](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81587401) [](ancestors = 81587401,81587164)

MikhailArkhipov: I guess only for Remote client?

---
In reply to: [81588299](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81588299) [](ancestors = 81588299,81587401,81587164)

AlexanderSher: You can limit it to the call of `CheckMachineOnlineAsync()` and call it inside exception block

---
In reply to: [81588299](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81588299) [](ancestors = 81588299,81587401,81587164)

AlexanderSher: Sure, we don't need to ping local client

---
In reply to: [81588525](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81588525) [](ancestors = 81588525,81588299,81587401,81587164)

MikhailArkhipov: Actually it throws its own PingException, need to handle that...

---
In reply to: [81588682](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81588682) [](ancestors = 81588682,81588299,81587401,81587164)

MikhailArkhipov: So putting ping it in ConnectAsync unfortunately gets this handled twice (main and intellisense sessions) and user gets two error messages


---
In reply to: [81589945](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81589945) [](ancestors = 81589945,81588682,81588299,81587401,81587164)

AlexanderSher: Does user get 2 messages when he switches connection or when he resets the REPL?

---
In reply to: [81606225](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81606225) [](ancestors = 81606225,81589945,81588682,81588299,81587401,81587164)

AlexanderSher: `WhenAllCancelOnFailure` should help against it - it throws only one exception

---
In reply to: [81607967](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81607967) [](ancestors = 81607967,81606225,81589945,81588682,81588299,81587401,81587164)

AlexanderSher: And actually you can check the cancellationToken to avoid pinging twice.

---
In reply to: [81608330](https://github.com/Microsoft/RTVS/pull/2416#discussion_r81608330) [](ancestors = 81608330,81607967,81606225,81589945,81588682,81588299,81587401,81587164)

AlexanderSher:
:shipit:

PR #2421, opened 2016-10-04
Connection switching and reconnection syncronization improvements + several bug fixes

### LINKED ISSUE #2183: Interactive Window RSession freezes after restart with PM open
- Start VS with Package Manager open
- Connect to remote broker
- Kill RHost on broker machine
- Click "Reset" on the REPL
- While connecting, switch back and forth between "Installed" and "Loaded" tabs

As a result, PM shows the list of installed packages (which means that connection to the broker is alive), but shows that loaded packages are unavailable.

### LINKED ISSUE #2397: Unable to reconnect to local when remote goes offline
Connect to remote
Shutdown remote broker
RTVS indicates connection is down
Try connecting to local R

- No effect. This is b/c connection managed remembers the last connection and attempts to reconnect. to it rather than switching to the new one. The connection that is being passed down is indeed local

image

### LINKED ISSUE #2401: Fail to connect to remote machine, then couldn't connect to anything, including local
Tried to connect to ddhuvalo083:5000, which didn't respond, progress bar dialog stayed up for 3-5 mins.
That's a machine that has an older build of remote services, which was lacking the VC CRT.
I hit cancel.
Now when I click connect for another remote or for local, nothing happens, as if I didn't click at all.

AlexanderSher:
- Fix #2401: Fail to connect to remote machine, then couldn't connect to anything, including local
- Fix #2397: Unable to reconnect to local when remote goes offline
- Fix #2183: Interactive Window RSession freezes after restart with PM open

`ConnectAsync` is called during switch, which should not allow any other calls of `ConnectAsync`, so it requires reader/writer lock (or a pair of semaphores) that support reentrancy and cancellation. Since `async/await` doesn't provider support reentrancy, I had to add a `ReentrancyToken` that can be passed from one lock to another.


MikhailArkhipov:
👍

int19h:
:+1:

PR #2432, opened 2016-10-04
Monitor process and UserProfile service

karthiknadig:
In this PR:
Monitoring process: Microsoft.R.Host.Monitor.exe is a WPF based application that manages restart and launch credentials for the Microsoft.R.Host.Broker.Process.
UserProfile Service: Microsoft.R.Host.UserProfile.exe is a Windows Service that runs a LocalSystem and allows Broker to create Windows User profile for logged on Users.
Added monitor and service to Dev14 remote setup.


src/Host/Broker/Impl/Pipes/ProcessHelpers.cs
```diff
@@ -10,6 +10,7 @@
 
 namespace Microsoft.R.Host.Broker.Pipes {
     internal static class ProcessHelpers {
+        /*
```
int19h: Do we still want to keep this code around, or are we sure we're not going to need this at this point? If the latter, might as well remove it...

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -48,6 +49,34 @@ public class SecurityManager {
             return new ClaimsPrincipal(identity);
         }
 
+        private uint CreateProfile(string username, string domain, string password,out StringBuilder profilePath, out bool profileExists) {
+            uint error = 0;
+            profilePath = new StringBuilder();
+            profileExists = false;
+            using (NamedPipeClientStream client = new NamedPipeClientStream("RUserCreatorPipe"))
```
int19h: Let's make it some kind of namespaced identifier (e.g. match the name of the assembly), or even use a GUID, just to make sure there's no collisions.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -48,6 +49,34 @@ public class SecurityManager {
...
+            using (BinaryReader reader = new BinaryReader(client))
+            using (BinaryWriter writer = new BinaryWriter(client)) {
+                try {
+                    client.Connect();
+
+                    writer.Write(username);
+                    writer.Write(domain);
+                    writer.Write(password);
+
+                    client.WaitForPipeDrain();
```
int19h: I'm curious why this is needed. It would seem that the next read should block on receiving the reply anyway?

On the other hand, you might want to explicitly call `Flush` on the writer, since it might have its own buffer.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -48,6 +49,34 @@ public class SecurityManager {
             return new ClaimsPrincipal(identity);
         }
 
+        private uint CreateProfile(string username, string domain, string password,out StringBuilder profilePath, out bool profileExists) {
```
int19h: This method is synchronous and blocking on all pipe operations. If the profile creation service hangs or goes down with the pipe still open, any request that hits this is going to get stuck, and will take down one worker thread with it.

I think it might be worthwhile to redo this in async, and use a cancellation token with a timeout on all operations. 

The catch is that binary readers and writers don't support async. So instead of using them, it might be better to aggregate the entirety of the request in a single object, then serialize the object to a string or blob, and dump that into the pipe with a single `WriteAsync` - and similarly read and process it from the other end.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,90 @@
...
+        public static int AutoRestartCount {
+            get {
+                return Properties.Settings.Default.AutoRestartCount;
+            }
+            set {
+                Properties.Settings.Default.AutoRestartCount = value;
+            }
+        }
+
+        private static int _autoRestartCount = 0;
```
int19h: It looks like this isn't the backing field for `AutoRestartCount`. If so, it shouldn't have the same name, to avoid confusion.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        static BrokerManager() {
+            AutoRestart = true;
+        }
+
+        public static Task StartBrokerInstanceAsync() {
```
int19h: Since this won't start the broker if one is already running, it should probably be called something like `EnsureBrokerIsRunningAsync`, or `CreateOrAttachToBrokerAsync`.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,90 @@
...
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Monitor {
+    public class BrokerManager {
+        private const string RHostBroker = "Microsoft.R.Host.Broker";
+        private static string RHostBrokerExe = $"{RHostBroker}.exe";
+        private static string RHostBrokerconfig = $"{RHostBroker}.Config.json";
```
int19h: Nit: casing.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,90 @@
...
+                    }
+                    
+                    _brokerProcess = new Process() { StartInfo = psi };
+
+                    _brokerProcess.Start();
+                }
+
+                _brokerProcess.EnableRaisingEvents = true;
+                MainWindow.SetStatusText($"Broker Process running ... {_brokerProcess.Id}");
+                _brokerProcess.Exited += _brokerProcess_Exited;
```
int19h: Since this event handler is subscribed after the process starts running, I think there's a race here if the broker crashes immediately on startup (e.g. due to a missing dependency) before the handler is there.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,90 @@
...
+                AutoRestart = true;
+            });
+        }
+
+        public static void ResetAutoStart() {
+            _autoRestartCount = 0;
+        }
+
+
+        private static async void _brokerProcess_Exited(object sender, EventArgs e) {
```
int19h: If `StartBrokerInstanceAsync` throws, the whole monitor process will go down because of uncaught exception on a background thread - probably not what we want. 

Generally speaking, it's better to avoid `async void` altogether, and always return tasks. Then, where the task really should be fire-and-forget, an explicit `DoNotWait()` makes the intent clear, and `SilenceException()` can be used to swallow stuff that is expected to happen in normal conditions.

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,388 @@
...
+
+        public static Task<bool> IsBrokerUserCredentialSavedAsync() {
+            return Task.Run(() => {
+                IntPtr credPtr;
+                bool credStatus = false;
+                if (CredRead(BrokerUserCredName, CredType.GENERIC, 0, out credPtr)) {
+                    credStatus = true;
+                } else {
+                    int error = Marshal.GetLastWin32Error();
+                    // ERROR_NOT_FOUND : Credentials were not found
```
int19h: Worth putting it into `NativeMethods` as a named constant?

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,388 @@
...
+                            domain,
+                            ref domainLen,
+                            password,
+                            ref passwordLen)) {
+                            throw new Win32Exception(Marshal.GetLastWin32Error());
+                        }
+
+                        IntPtr token = IntPtr.Zero;
+                        try {
+                            var usernameBldr = new StringBuilder(257);
```
int19h: `CRED_MAX_USERNAME_LENGTH`?

int19h: (or maybe it was a different constant... but either way, there should be a named constant).

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,388 @@
...
+                            if (token != IntPtr.Zero) {
+                                CloseHandle(token);
+                            }
+                        }
+
+                        break;
+                    }
+
+                    // CredUIPromptForWindowsCredentials failed to load some component
+                    if (ret != 0) {
```
int19h: Should this handle the return error codes for all the other `Cred*` functions here? E.g. if `CredUIParseUserName`, we just throw with the error code provided, without a custom error message - is that right?

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,388 @@
...
+        public static Task RemoveCredentialsAsync() {
+            return Task.Run(() => {
+                CredDelete(BrokerUserCredName, CredType.GENERIC, 0);
+            });
+        }
+
+
+        public static Task SetCredentialsOnProcess(ProcessStartInfo psi) {
+            return Task.Run(()=> {
+                IntPtr cred;
```
int19h: `CredFree` for this one is missing.

src/Host/Monitor/Impl/MainWindow.xaml.cs
```diff
@@ -0,0 +1,77 @@
...
+    public partial class MainWindow : Window {
+        public static MainWindow _currentWindow;
+        public MainWindow() {
+            InitializeComponent();
+            Loaded += MainWindow_Loaded;
+            _currentWindow = this;
+            Closing += MainWindow_Closing;
+        }
+
+        private async void MainWindow_Closing(object sender, System.ComponentModel.CancelEventArgs e) {
```
int19h: Same thing for all the `async void` methods here - what happens if any of the code inside throws?

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        public static Task RemoveCredentialsAsync() {
+            return Task.Run(() => {
+                CredDelete(BrokerUserCredName, CredType.GENERIC, 0);
+            });
+        }
+
+
+        public static Task SetCredentialsOnProcess(ProcessStartInfo psi) {
```
int19h: `...Async`

src/Host/UserProfile/Impl/RUserProfileService.cs
```diff
@@ -0,0 +1,64 @@
...
+            InitializeComponent();
+        }
+
+        private ManualResetEvent _workerdone;
+        private CancellationTokenSource _cts;
+        protected override void OnStart(string[] args) {
+            base.OnStart(args);
+            _cts = new CancellationTokenSource();
+            _workerdone = new ManualResetEvent(false);
+            Task.WhenAny(CreateProfileWorkerAsync(_cts.Token));
```
int19h: There's only a single task here, so you shouldn't need `WhenAny`.

src/Host/UserProfile/Impl/RUserProfileService.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        protected override void OnStop() {
+            base.OnStop();
+            _cts.Cancel();
+            _workerdone.WaitOne(TimeSpan.FromSeconds(10));
+        }
+
+        async Task CreateProfileWorkerAsync(CancellationToken ct) {
```
int19h: Same thing about async as on the other end of the pipe. I think it's worse here, actually, because it's more likely that broker will crash (it's more complex), it doesn't have the monitor process, and it doesn't own the server end of the pipe (so you won't notice that it's gone). So if broker crashes between writes for username/domain/password, or before it could read the result, this will get stuck in the middle of the loop, and won't respond to service stop requests.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -49,32 +51,35 @@ public class SecurityManager {
...
         }
 
-        private uint CreateProfile(string username, string domain, string password,out StringBuilder profilePath, out bool profileExists) {
-            uint error = 0;
-            profilePath = new StringBuilder();
-            profileExists = false;
-            using (NamedPipeClientStream client = new NamedPipeClientStream("RUserCreatorPipe"))
-            using (BinaryReader reader = new BinaryReader(client))
-            using (BinaryWriter writer = new BinaryWriter(client)) {
+        private async Task<JArray> CreateProfileAsync(string username, string domain, string password, CancellationToken ct) {
```
int19h: Given that all inputs are separate strongly typed values, the return should also be that, rather than `JArray` - i.e. this method should take responsibility both for serialization of the request, and for deserialization of the result.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -96,10 +101,25 @@ public class SecurityManager {
...
                 bool profileExists = false;
+#if DEBUG
+                CancellationTokenSource cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
+#else
+                CancellationTokenSource cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));
+#endif
+
                 _logger.LogTrace(Resources.Trace_UserProfileCreation, context.Username);
-                error = CreateProfile(user.ToString(), domain.ToString(), context.Password, out profileDir, out profileExists);
+                JArray result = CreateProfileAsync(user.ToString(), domain.ToString(), context.Password, cts.Token).GetAwaiter().GetResult();
```
int19h: Rather than blocking waiting for result, let's make this method async - it's caller is already async.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -49,32 +51,35 @@ public class SecurityManager {
...
+        private async Task<JArray> CreateProfileAsync(string username, string domain, string password, CancellationToken ct) {
+            using (NamedPipeClientStream client = new NamedPipeClientStream("Microsoft.R.Host.UserProfile.Creator{b101cc2d-156e-472e-8d98-b9d999a93c7a}")) { 
                 try {
-                    client.Connect();
+                    await client.ConnectAsync(ct);
 
-                    writer.Write(username);
-                    writer.Write(domain);
-                    writer.Write(password);
+                    JArray dataArray = new JArray();
```
int19h: Might be easier and more convenient to just shove all pieces in a struct and serialize that as a single object; and same thing for the response. Basically, the same way we do e.g. `SessionCreateRequest`.

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -172,6 +175,10 @@ public class CredentialManager {
                     psi.Domain = domainBldr.ToString();
                     psi.Password = pass;
                 }
+
+                if (cred != IntPtr.Zero) {
+                    CredFree(cred);
```
int19h: This should be inside `finally`, since some code paths above might not get here (e.g. if `CredUIParseUserName` fails, we throw).

src/Host/Monitor/Impl/MainWindow.xaml.cs
```diff
@@ -13,53 +16,52 @@ public partial class MainWindow : Window {
...
+                    } else {
+                        await CredentialManager.GetAndSaveCredentialsFromUserAsync();
+                    }
                 } else {
-                    await CredentialManager.GetAndSaveCredentialsFromUserAsync();
+                    await BrokerManager.CreateOrAttachToBrokerInstanceAsync();
                 }
-            } else {
-                await BrokerManager.StartBrokerInstanceAsync();
+            } catch (Exception ex) {
```
int19h: Can you add `when (!ex.IsCriticalException())` here? We don't want to swallow NRE and that kind of stuff.

Also, is there anything that will log the exception before this point? If not, we should log here, to diagnose issues.

src/Host/Monitor/Impl/MainWindow.xaml.cs
```diff
@@ -13,53 +16,52 @@ public partial class MainWindow : Window {
...
-        private async void RemoveBrokerUserBtn_Click(object sender, RoutedEventArgs e) {
-            await CredentialManager.RemoveCredentialsAsync();
+        private void RemoveBrokerUserBtn_Click(object sender, RoutedEventArgs e) {
+            CredentialManager.RemoveCredentials();
         }
 
-        public static async void SetStatusText(string message) {
-            await _currentWindow.Dispatcher.InvokeAsync(() => {
+        public static void SetStatusText(string message) {
+            _currentWindow.Dispatcher.Invoke(() => {
```
int19h: If `SetStatusText` is only ever called from event handlers and tasks that they spawn, I think this isn't necessary, since `await` will return to the UI context.

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -82,69 +67,87 @@ public class CredentialManager {
...
             });
         }
 
+        public static void RemoveCredentials() {
+            CredDelete(BrokerUserCredName, CredType.GENERIC, 0);
+        }
 
-        public static Task SetCredentialsOnProcess(ProcessStartInfo psi) {
+        public static Task SetCredentialsOnProcessAsync(ProcessStartInfo psi) {
             return Task.Run(()=> {
```
int19h: Is the purpose of `Task.Run` to force the code to run on background thread and drop any context, even if the caller is on UI thread? If so, `await TaskUtilities.SwitchToBackgroundThread()` is the preferred way to do so.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -47,29 +50,27 @@ public class BrokerManager {
...
             _autoRestartCount = 0;
         }
 
-
-        private static async void _brokerProcess_Exited(object sender, EventArgs e) {
-            if (AutoRestart && ++_autoRestartCount <= AutoRestartCount) {
-                await StartBrokerInstanceAsync();
+        private static void ProcessExited(object sender, EventArgs e) {
+            if (AutoRestart && ++_autoRestartCount <= AutoRestartMaxCount) {
+                CreateOrAttachToBrokerInstanceAsync().DoNotWait();
```
int19h: I'm not sure which thread this event handler will run on, but probably it's a threadpool thread, so expected exceptions need to be suppressed accordingly.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -10,16 +11,16 @@ namespace Microsoft.R.Host.Monitor {
     public class BrokerManager {
```
int19h: Testing this will be very painful, because it's tightly coupled to MainWindow UI. I would recommend not accessing `MainWindow` directly anywhere, and instead providing a generic callback for status updates and such. The caller of `CreateOrAttachToBrokerInstanceAsync` would then be responsible for translating those to status bar updates etc; and in tests, we'd use a mock to check that progress is reported correctly.

int19h: Actually, so far as I can see, the only two places where UI is updated here is at the end of `CreateOrAttachToBrokerInstanceAsync` and `StopBrokerInstanceAsync`. So instead of doing the update here, just change it so that the caller does it once the task completes successfully.

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -82,69 +67,87 @@ public class CredentialManager {
...
-                            domain,
-                            ref domainLen,
-                            password,
-                            ref passwordLen)) {
-                            throw new Win32Exception(Marshal.GetLastWin32Error());
+                        if (!CredUnPackAuthenticationBuffer(0, credentialBuffer, credentialSize, username, ref usernameLen, domain, ref domainLen, password, ref passwordLen)) {
+                            // Do another attempt by showing user the CredUI with the errorCode set.
+                            errorCode = Marshal.GetLastWin32Error();
+                            if (credentialBuffer != IntPtr.Zero) {
+                                Marshal.FreeCoTaskMem(credentialBuffer);
```
int19h: MSDN says that some extra cleanup is necessary here as a matter of security:

> When you have finished using the credential BLOB, clear it from memory by calling the SecureZeroMemory function, and free it by calling the CoTaskMemFree function.

src/Host/UserProfile/Impl/RUserProfileService.cs
```diff
@@ -36,25 +39,52 @@ partial class RUserProfileService : ServiceBase {
...
+                            bytesRead = await server.ReadAsync(request, 0, request.Length, ct);
+                        }
+                        
+                        byte[] data = new byte[bytesRead];
+                        Array.Copy(request, data, bytesRead);
 
-                        writer.Write(result.Win32Result);
-                        writer.Write(result.ProfileExists);
-                        writer.Write(result.ProfilePath);
+                        string json = Encoding.Unicode.GetString(data);
```
int19h: Same thing here about overload with offset/length.

src/Host/UserProfile/Impl/RUserProfileService.cs
```diff
@@ -36,25 +39,52 @@ partial class RUserProfileService : ServiceBase {
...
+                        JArray dataArray = JArray.Parse(json);
```
int19h: This can throw if incoming JSON is invalid - needs to be handled alongside `IOException` below.

src/Host/UserProfile/Impl/RUserProfileService.cs
```diff
@@ -36,25 +39,52 @@ partial class RUserProfileService : ServiceBase {
...
+
+                        uint error = 13; // invalid data
+                        bool profileExists = false;
+                        string profilePath = string.Empty;
+
+                        if(dataArray.Count == 3) {
+                            string username = dataArray[0].Value<string>();
```
int19h: This can also throw for malformed input (but if you deserialize into a struct instead, then it also becomes an exception out of `Parse`).

src/Host/UserProfile/Impl/RUserProfileService.cs
int19h: Also, it can throw if input is not valid UTF-16 - needs to be handled.

karthiknadig:
Is there an alternate option for `server.WaitForPipeDrain()`, I want to remove that line given that it has the potential to block indefinitely. The API has no timeout, and it does not accept cancellation token. I am currently using that to avoid disposing the server pipe before client has had an opportunity to read. `Task.Delay` instead ?

int19h:
I think a zero-length read would work - it would block (waiting for the client to do a write), and ultimately return 0 when the pipe gets closed from the client end. It's not quite the same thing (you'd be waiting for client to close the pipe, not for client to read everything you wrote), but is probably good enough here.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+
+        private static int _autoRestartCount = 0;
+
+        static BrokerManager() {
+            AutoRestart = true;
+        }
+
+        public static Task<int> CreateOrAttachToBrokerInstanceAsync(ILogger logger = null) {
```
MikhailArkhipov: Do we want to protect from multiple calls? I.e. if task is already created, don't create another one.

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,102 @@
...
+                    psi.Arguments = $"--config \"{configFilePath}\"";
+                    psi.UseShellExecute = false;
+                    psi.CreateNoWindow = false;
+                    psi.WorkingDirectory = assemblyRoot;
+
+                    if (Properties.Settings.Default.UseDifferentBrokerUser) {
+                        await CredentialManager.SetCredentialsOnProcessAsync(psi, logger);
+                    }
+
+                    _brokerProcess = new Process() { StartInfo = psi };
```
MikhailArkhipov: Should be using or extending IProcessServices for testability. OK to file separate task.

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,418 @@
...
+                if (CredRead(BrokerUserCredName, CredType.GENERIC, 0, out credPtr)) {
+                    credStatus = true;
+                } else {
+                    int error = Marshal.GetLastWin32Error();
+                    if (error == Win32ErrorCodes.ERROR_NOT_FOUND) {
+                        credStatus = false;
+                        logger?.LogDebug(Resources.Info_DidNotFindSavedCredentails);
+                    }
+                }
+                if (credPtr != IntPtr.Zero) {
```
MikhailArkhipov: Should be in finally

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,418 @@
...
+using System.Threading.Tasks;
+using Microsoft.Common.Core;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.R.Host.Monitor {
+    public class CredentialManager {
+        private const string BrokerUserCredName = "RHostBrokerUserCred";
+
+        public static Task<bool> IsBrokerUserCredentialSavedAsync(ILogger logger = null) {
+            return Task.Run(() => {
```
MikhailArkhipov: Concurrent calls?

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,418 @@
...
+                // User clicked the cancel button.
+                if (ret == Win32ErrorCodes.ERROR_CANCELLED) {
+                    logger?.LogInformation(Resources.Info_CredUIDismissed);
+                    break;
+                } else if (ret == Win32ErrorCodes.ERROR_SUCCESS) {
+                    // User entered credentials
+                    logger?.LogInformation(Resources.Info_CredUICredsReceived);
+                    StringBuilder username = new StringBuilder(CRED_MAX_USERNAME_LENGTH);
+                    StringBuilder domain = new StringBuilder(CRED_MAX_USERNAME_LENGTH);
+                    StringBuilder password = new StringBuilder(CRED_MAX_USERNAME_LENGTH);
```
MikhailArkhipov: Should be using SecureString / SecureString storage ( see my incoming PR)

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,418 @@
...
+                        errorCode = Marshal.GetLastWin32Error();
+                        logger?.LogError(Resources.Error_CredentialUnpackingFailed, errorCode);
+                        if (credentialBuffer != IntPtr.Zero) {
+                            Marshal.ZeroFreeCoTaskMemUnicode(credentialBuffer);
+                        }
+                        continue;
+                    } else {
+                        logger?.LogInformation(Resources.Info_CredUnpacked);
+                        // Credential buffer is no longer needed
+                        if (credentialBuffer != IntPtr.Zero) {
```
MikhailArkhipov: finally { }

src/Host/UserProfileInterface/Impl/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
```
MikhailArkhipov: Should not have this. See other projects on a) AssemblyInfo and linked GlobalAssemblyInfo.cs - they allow build process to inject versions

src/Host/UserProfileInterface/Impl/Microsoft.R.Host.UserProfileInterface.csproj
```diff
@@ -0,0 +1,58 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
```
MikhailArkhipov: Make sure it imports versioning targets as other projects do. As well as as MicroBuild targets so it is buildable in the lab. Also needs field that enables signing. See other projects for how-to.

src/Host/UserProfileInterface/Impl/IRUserProfileCreatorResult.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
```
MikhailArkhipov: Prob doesnt need most of these

src/Host/UserProfile/Impl/ServiceLoggerProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public ServiceLoggerProvider(StreamWriter writer) {
+            _writer = writer;
+        }
+
+        private static string GetLogFileName() {
+            return Path.Combine(Path.GetTempPath(), $@"Microsoft.R.Host.UserProfile_{DateTime.Now:yyyyMdd_HHmmss}_pid{Process.GetCurrentProcess().Id}.log");
+        }
+
+        public ILogger CreateLogger(string categoryName) {
```
MikhailArkhipov: Logging activity should follow settings. Logging can be disabled.

src/Host/UserProfile/Impl/packages.config
```diff
@@ -0,0 +1,20 @@
...
+  <package id="Microsoft.Extensions.DependencyInjection.Abstractions" version="1.0.0" targetFramework="net46" />
+  <package id="Microsoft.Extensions.Logging" version="1.0.0" targetFramework="net46" />
+  <package id="Microsoft.Extensions.Logging.Abstractions" version="1.0.0" targetFramework="net46" />
+  <package id="Microsoft.Extensions.Logging.Debug" version="1.0.0" targetFramework="net46" />
+  <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net46" />
+  <package id="System.Collections" version="4.0.11" targetFramework="net46" />
+  <package id="System.Collections.Concurrent" version="4.0.12" targetFramework="net46" />
+  <package id="System.ComponentModel" version="4.0.1" targetFramework="net46" />
+  <package id="System.Diagnostics.Debug" version="4.0.11" targetFramework="net46" />
+  <package id="System.Globalization" version="4.0.11" targetFramework="net46" />
```
MikhailArkhipov: Should be using packages.json and global nuget cache.

src/Host/UserProfile/Impl/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
```
MikhailArkhipov: Please add assembly info and global assembly info link

src/Host/UserProfile/Impl/Microsoft.R.Host.UserProfile.csproj
```diff
@@ -0,0 +1,129 @@
...
+      <Name>Microsoft.Common.Core</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Protocol\Impl\Microsoft.R.Host.Protocol.csproj">
+      <Project>{C46E5F53-4CAF-4C65-B173-CA8140FB41E0}</Project>
+      <Name>Microsoft.R.Host.Protocol</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+       Other similar extension points exist, see Microsoft.Common.targets.
```
MikhailArkhipov: Lab build targets, versioning targets

src/Host/UserProfile/Impl/Microsoft.R.Host.UserProfile.csproj
```diff
@@ -0,0 +1,129 @@
...
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Extensions.DependencyInjection.Abstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, processorArchitecture=MSIL">
+      <HintPath>..\..\..\..\NugetPackages\Microsoft.Extensions.DependencyInjection.Abstractions.1.0.0\lib\netstandard1.0\Microsoft.Extensions.DependencyInjection.Abstractions.dll</HintPath>
```
MikhailArkhipov: Should be using packages.json and global nuget cache

src/Host/Monitor/Impl/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Windows;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Microsoft.R.Host.Monitor")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("Microsoft.R.Host.Monitor")]
```
MikhailArkhipov: Global assembly info

src/Host/Monitor/Impl/Microsoft.R.Host.Monitor.csproj
```diff
@@ -0,0 +1,178 @@
...
+      <Install>true</Install>
+    </BootstrapperPackage>
+    <BootstrapperPackage Include="Microsoft.Net.Framework.3.5.SP1">
+      <Visible>False</Visible>
+      <ProductName>.NET Framework 3.5 SP1</ProductName>
+      <Install>false</Install>
+    </BootstrapperPackage>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\Common\Core\Impl\Microsoft.Common.Core.csproj">
```
MikhailArkhipov: Lab targets, version targets, global nuget, signing settings

src/Host/UserProfile/Impl/ServiceLoggerProvider.cs
karthiknadig: Where is the logging setting stored? I am not able to find it.

src/Host/Monitor/Impl/CredentialManager.cs
karthiknadig: I created a work item for this. I will user the APIs you created after your code goes in. #2460

src/Host/UserProfile/Impl/ServiceLoggerProvider.cs
karthiknadig: #2477

---
In reply to: [82319599](https://github.com/Microsoft/RTVS/pull/2432#discussion_r82319599) [](ancestors = 82319599)

src/Package/Impl/source.extension.vsixmanifest
```diff
@@ -123,6 +123,8 @@
 
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Host.exe" />
     <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Host.Broker.exe" />
+    <Asset Type="Microsoft.VisualStudio.Assembly" Path="Microsoft.R.Host.Monitor.exe" />
```
MikhailArkhipov: If this is remote only you don't need them in VSIX.

karthiknadig: got it. Will remove.

src/Host/UserProfile/Impl/project.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+  "dependencies": {
+    "Microsoft.Extensions.Configuration.Json": "1.0.0",
+    "Microsoft.Extensions.Logging": "1.0.0",
```
MikhailArkhipov: "MicroBuild.Core": "0.2.0",

src/SetupRemote/Product.wxs
```diff
@@ -26,10 +26,18 @@
     <DirectoryRef Id="INSTALLFOLDER">
       <Component Id="ProductAssemblies" Guid="{9C717F8F-4A97-4FA5-8031-5A180467754D}" Win64="yes">
         <?include "$(var.SourceDir)\Setup\CommonAssemblies.wxi" ?>
+        <File Source="$(var.BinDir)\Microsoft.R.Host.Monitor.exe" />
       </Component>
     </DirectoryRef>
     <?include RtvsPackage.wxi ?>
-
+    <Component Id="RemoteRUserProfile" Directory="INSTALLFOLDER" Guid="{7C54EB44-130F-4D8A-A9C5-2EBB96F9F359}" Win64="yes">
+      <File Source="$(var.BinDir)\Microsoft.R.Host.UserProfile.exe" />
```
MikhailArkhipov: Do you need app.config?

src/Host/Protocol/Impl/RUserProfileCreateResponse.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
```
MikhailArkhipov: Legal header

// Copyright (c) Microsoft Corporation. All rights reserved.

src/Host/Protocol/Impl/RUserProfileCreateRequest.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
```
MikhailArkhipov: Legal header (and all other files)

src/Host/Protocol/Impl/RUserProfileCreateRequest.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Protocol {
+    public struct RUserProfileCreateRequest {
+        public string Username { get; set; }
+        public string Domain { get; set; }
+        public string Password { get; set; }
+
+        public static RUserProfileCreateRequest Create(string username, string domain, string password) {
+            return new RUserProfileCreateRequest() { Username = username, Domain = domain, Password = password };
```
MikhailArkhipov: If you want static Create enforced, you may want private constructor

src/Host/Monitor/Impl/project.json
```diff
@@ -0,0 +1,14 @@
+﻿{
+  "dependencies": {
+    "Microsoft.Extensions.Logging": "1.0.0",
```
MikhailArkhipov: "MicroBuild.Core": "0.2.0",

src/Host/Monitor/Impl/Microsoft.R.Host.Monitor.csproj
```diff
@@ -0,0 +1,185 @@
...
+  </PropertyGroup>
+<PropertyGroup Condition="'$(NuGetPackageRoot)' == ''">
+    <NuGetPackageRoot>$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <PlatformTarget>x64</PlatformTarget>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x64'">
+    <OutputPath>bin\x64\Release\</OutputPath>
```
MikhailArkhipov: Should not have bin under project, 

  <Import Project="..\..\..\R.Settings.targets" />
  <PropertyGroup>
    <BaseIntermediateOutputPath>$(ObjDirectory)</BaseIntermediateOutputPath>
    <BaseOutputPath>$(BinDirectory)</BaseOutputPath>
    <IntermediateOutputPath>$(BaseIntermediateOutputPath)\$(Configuration)\$(AssemblyName)\</IntermediateOutputPath>
    <OutputPath>$(BaseOutputPath)\$(Configuration)\</OutputPath>
  </PropertyGroup>

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,434 @@
...
+                try {
+                    if (CredRead(BrokerUserCredName, CredType.GENERIC, 0, out cred)) {
+                        Credential credential = new Credential();
+                        credential = (Credential)Marshal.PtrToStructure(cred, typeof(Credential));
+
+                        byte[] passwordBytes = new byte[credential.credentialBlobSize];
+                        Marshal.Copy(credential.credentialBlob, passwordBytes, 0, credential.credentialBlobSize);
+                        string passwordText = Encoding.Unicode.GetString(passwordBytes);
+
+                        var usernameBldr = new StringBuilder(257);
```
MikhailArkhipov: CREDUI_MAX_...

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,434 @@
...
+        public static void SetCredentialsOnProcess(ProcessStartInfo psi, ILogger logger = null) {
+            lock (_setCredentialsOnProcessLock) {
+                IntPtr cred = IntPtr.Zero;
+                try {
+                    if (CredRead(BrokerUserCredName, CredType.GENERIC, 0, out cred)) {
+                        Credential credential = new Credential();
+                        credential = (Credential)Marshal.PtrToStructure(cred, typeof(Credential));
+
+                        byte[] passwordBytes = new byte[credential.credentialBlobSize];
+                        Marshal.Copy(credential.credentialBlob, passwordBytes, 0, credential.credentialBlobSize);
```
MikhailArkhipov: SecurityUtilities.SecureStringFromNativeBuffer

src/Host/Monitor/Impl/CredentialManager.cs
```diff
@@ -0,0 +1,434 @@
...
+                        var usernameBldr = new StringBuilder(257);
+                        var domainBldr = new StringBuilder(257);
+
+                        uint error = CredUIParseUserName(credential.userName, usernameBldr, usernameBldr.Capacity, domainBldr, domainBldr.Capacity);
+                        if (error != 0) {
+                            logger?.LogError(Resources.Error_UserNameParsing, error);
+                            throw new Win32Exception((int)error);
+                        }
+
+                        SecureString pass = new SecureString();
```
MikhailArkhipov: ToSecureString(), but it is a last resort. Better send native straight to SecureString as above

src/Host/Monitor/Impl/BrokerManager.cs
```diff
@@ -0,0 +1,110 @@
...
+                Process[] processes = Process.GetProcessesByName(RHostBroker);
+                if (processes.Length > 0) {
+                    _brokerProcess = processes[0];
+                    _brokerProcess.EnableRaisingEvents = true;
+                    _brokerProcess.Exited += async (object sender, EventArgs e) => {
+                        await ProcessExitedAsync(logger);
+                    };
+                    logger?.LogInformation(Resources.Info_BrokerAlreadyRunning, _brokerProcess.Id);
+                } else {
+                    string assemblyRoot = Path.GetDirectoryName(AppDomain.CurrentDomain.BaseDirectory);
```
MikhailArkhipov: Assembly.GetExecutingAssembly().GetAssemblyPath();

src/Host/Monitor/Impl/Microsoft.R.Host.Monitor.csproj
karthiknadig: I copied the csproj from Broker. It seems like they don't create bin and obj folders under the project.

src/Host/Broker/Impl/Security/SecurityManager.cs
```diff
@@ -48,7 +52,29 @@ public class SecurityManager {
...
+                    string jsonResp = Encoding.Unicode.GetString(responseRaw, 0, bytesRead);
+                    return JsonConvert.DeserializeObject<RUserProfileCreateResponse>(jsonResp);
+                } catch (Exception ex) when (!ex.IsCriticalException()) {
+                    _logger.LogError(Resources.Error_ProfileCreationFailedIO, request.Username);
+                    return RUserProfileCreateResponse.Blank;
+                }
+            }
+        }
+
+        private async Task<ClaimsPrincipal> SignInUsingLogonAsync(BasicSignInContext context) {
```
int19h: If I remember correctly, it's an external type from the package that we use for HTTP Basic. Out of the box, it doesn't do secure strings.

We can fork it and make our own implementation, but it would still need to make a secure string out of a plain string that it gets from HTTP header dictionary, which makes it moot to try to conceal it.

This is probably best left for a separate PR that would deal with this problem end to end (i.e. switch to some other auth scheme that does not have password unencrypted at any point, not even in the headers).

src/Host/Monitor/Impl/CredentialManager.cs
karthiknadig: I have a work item for this. #2460

src/Host/Monitor/Impl/CredentialManager.cs
karthiknadig: #2460

src/Host/Monitor/Impl/Logging/MonitorLogger.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.Common.Core.Disposables;
+using Microsoft.Extensions.Logging;
+using System;
+using System.IO;
+
+namespace Microsoft.R.Host.Monitor.Logging {
+    internal sealed class MonitorLogger : ILogger, IDisposable {
```
int19h: Can we reuse the file logger from the broker? It seems like these two ought to be fundamentally the same.

src/Host/Broker/Impl/Security/SecurityManager.cs
MikhailArkhipov: Agreed. We can to review this later.

MikhailArkhipov:
OK. Needs some more work, but should be testable.  👍

src/Host/Monitor/Impl/Logging/MonitorLogger.cs
karthiknadig: I will extractl the file logging into a common library. #2477

src/SetupRemote/Product.wxs
karthiknadig: No

Issue #2410, opened 2016-09-29
Keyboard support to invoke context menu in plot device window


huguesv:
Right now it's mouse right-click only, can't invoke the context menu with keyboard.

MikhailArkhipov:
Shift+F10

huguesv:
I can't get any keyboard events in the plot device user control.  The control contains just an `Image` (the plot) or `Label` (please plot something). I tried various solutions with marking the user control Focusable, calling Focus() on it in Loaded event, but I've never seen any Key event (PreviewKeyDown, PreviewKeyUp, KeyDown, or KeyUp).
Apparently, in those cases you're supposed to attach to the Window's PreviewKeyX events, instead of user contorl. If I call `Window.GetWindow(this)`, it looks like I get the main VS window.  Good news is that I now get key events. Bad news is that I get them from anywhere inside VS. There has to be a better solution than this.

MikhailArkhipov:
I don't think we need this since all commands are available on toolbar

PR #2439, opened 2016-10-06
Plot window clear on session connect, context menu via keyboard

### LINKED ISSUE #2403: When switching between connections, plots of previous connection remain

huguesv:
Partial fix for #2410 (makes keyboard context menu work plot history window, but not plot device window)
Fixes #2403 
Also makes the context menu key work in variable explorer (used to be just shift-f10).

### DIFF
--- a/src/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
+++ b/src/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
@@ -17,7 +17,11 @@ namespace Microsoft.R.Components.Plots.Implementation.View {
         private IRPlotHistoryViewModel ViewModel => (IRPlotHistoryViewModel)DataContext;
         private DragSurface _dragSurface = new DragSurface();
 
-        public event EventHandler<MouseButtonEventArgs> ContextMenuRequested;
+        /// <summary>
+        /// Show context menu at the specified point, which is relative
+        /// to the visual component control (ie. this control).
+        /// </summary>
+        public event EventHandler<PointEventArgs> ContextMenuRequested;
 
 
         public RPlotHistoryControl() {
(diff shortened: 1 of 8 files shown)


src/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
```diff
@@ -33,6 +37,11 @@ public partial class RPlotHistoryControl : UserControl {
             if (e.Key == Key.Enter) {
                 var entry = (IRPlotHistoryEntryViewModel)(((ListViewItem)sender).Content);
                 ActivatePlot(entry);
+            } else if (e.Key == Key.Apps || (e.SystemKey == Key.F10 && (Keyboard.Modifiers & ModifierKeys.Shift) == ModifierKeys.Shift)) {
+                ListViewItem item = (ListViewItem)sender;
+                var point = item.TranslatePoint(new Point(1, 1), this);
```
MikhailArkhipov: No nulls?

MikhailArkhipov:
👍

src/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
huguesv: The event is attached to ListViewItem directly, it won't fire if it's not a ListViewItem or if null.
```
            <ListView.ItemContainerStyle>
                <Style TargetType="ListViewItem">
                    <EventSetter Event="MouseDoubleClick" Handler="item_MouseDoubleClick"/>
                    <EventSetter Event="KeyUp" Handler="item_KeyUp"/>
                </Style>
            </ListView.ItemContainerStyle>
```

PR #2440, opened 2016-10-06
Update inactive session icon so it doesn't look cut off.

### LINKED ISSUE #2414: R icons look chopped off when there is no connection
image

We need to make gray circle darker so background doesn't blend into white.

huguesv:
Fixes #2414 

Here's what it looks like in all 3 themes:
image


MikhailArkhipov:
👍

AlexanderSher:
👍

PR #2447, opened 2016-10-06
SSL/TLS

MikhailArkhipov:


src/Common/Core/Impl/Security/Credentials.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Net;
+using System.Security;
+
+namespace Microsoft.Common.Core.Security {
+    public sealed class Credentials : ICredentials {
+        // Although NetworkCredential does support SecureString, it still exposes
```
int19h: But if we create a `NetworkCredential` out of this eventually anyway, why would having thewrapper make a difference?

src/Common/Core/Impl/Security/SecurityUtilities.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+            var sec = new SecureString();
+            foreach (var ch in s) {
+                sec.AppendChar(ch);
+            }
+            return sec;
+        }
+
+        public static IntPtr CreateSecureStringBuffer(int length) {
+            var initial = new char[length + 1];
```
int19h: Unused.

src/Host/Broker/Impl/ApiErrorResult.cs
```diff
@@ -1,35 +1,25 @@
...
-            _message = message;
-         }
-
-        public override void ExecuteResult(ActionContext context) {
-            var headers = context.HttpContext.Response.Headers;
-            headers.Add(CustomHttpHeaders.RTVSApiError, new Extensions.Primitives.StringValues(_brokerApiError.ToString()));
-            if (!string.IsNullOrEmpty(_message)) {
-                headers.Add(CustomHttpHeaders.RTVSBrokerException, _message);
+            var headers = response.Headers;
+            headers.Add(CustomHttpHeaders.RTVSApiError, new Extensions.Primitives.StringValues(error.ToString()));
```
int19h: Why is this back in the constructor, rather than `ExecuteResult`? `IActionResult` should only access `HttpContext` in execute...

src/Host/Broker/Impl/Security/TlsConfiguration.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Https;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.R.Host.Protocol;
+
+namespace Microsoft.R.Host.Broker.Security {
+    internal sealed class TlsConfiguration {
+        private readonly ILogger _logger;
+
+        public TlsConfiguration(ILogger logger) {
```
int19h: This should probably take `SecurityOptions` in constructor, since it's going to be the same for all calls on it.

src/Common/Core/Impl/Security/Credentials.cs
MikhailArkhipov: Yes. However, password is visible as plain text in a debugger since debugger evaluates properties and when `Password` is evaulated, plain text is returned. This is debugger-only scenario but ppl on the Net recommend to hide it since one may expose password to a co-worked when pair-debugging. I.e. this is mostly for us :-)

MikhailArkhipov: NetworkCredential is created and passed over rather than cached. So it won't appear in debugger.

src/Common/Core/Impl/Security/SecurityUtilities.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace Microsoft.Common.Core.Security {
+    public static class SecurityUtilities {
+        public static IntPtr CreatePasswordBuffer() {
+            return CreateSecureStringBuffer(NativeMethods.CREDUI_MAX_PASSWORD_LENGTH);
+        }
+
+        public static SecureString ToSecureString(this string s) {
```
MikhailArkhipov: The only place I need it is with `Find(Claims.Password)` which returns plain text string :-(

src/Host/Broker/Impl/Security/TlsConfiguration.cs
```diff
@@ -0,0 +1,68 @@
...
+            _logger.LogInformation(Resources.Trace_CertificateIssuer, certificate.Issuer);
+            _logger.LogInformation(Resources.Trace_CertificateSubject, certificate.Subject);
+
+            return certificate;
+        }
+
+        private bool IsLocalConnection(IConfigurationRoot configuration) {
+            try {
+                Uri uri;
+                var url = configuration.GetValue<string>("server.urls", null);
```
int19h: Note that this can be more than one URL - it can be an array in the config JSON, or command line can specify multiple `--server.urls` switches. I'm actually not sure if `GetValue<string>()` will even work in that case, or what the resulting string will be.

src/Host/Broker/Impl/ApiErrorResult.cs
MikhailArkhipov: Because I found that ExecuteResult actually never gets called so we ended up again without diagnostics. Anything I can do to make it executed?

---
In reply to: [82292942](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82292942) [](ancestors = 82292942)

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -74,7 +77,7 @@ public class Session {
             _pipe = new MessagePipe(messageLogger);
         }
 
-        public void StartHost(SecureString password, string profilePath, ILogger outputLogger, LogVerbosity verbosity) {
+        public void StartHost(ClaimsPrincipal cp, string profilePath, ILogger outputLogger, LogVerbosity verbosity) {
```
int19h: If we're only using the principal to retrieve the password, is it worth passing the whole thing? This way we get the non-secure string flow through and persist around at least until this method returns...

int19h: Also, it's not clear how it should interact with `User` - i.e. what happens if the caller passes a principal for a user different from `IIdentity` that was passed to the session constructor.

MikhailArkhipov: The only reason it is here is to reduce lifetime of the plain text password in

cp.FindFirst(Claims.Password)?.Value.ToSecureString()




---
In reply to: [82294590](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82294590) [](ancestors = 82294590)

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -22,17 +23,18 @@ public class RemoteUriWebService : IRemoteUriWebService {
 
         private Uri PostUri { get; }
 
-        public async Task GetResponseAsync(HttpListenerContext context, string localBaseUrl, string remoteBaseUrl, CancellationToken ct) {
+        public async Task GetResponseAsync(HttpListenerContext context, string localBaseUrl, string remoteBaseUrl, string webSocketsScheme, CancellationToken ct) {
             string postUri = null;
             if (context.Request.IsWebSocketRequest) {
-                UriBuilder ub = new UriBuilder(PostUri) { Scheme = "ws" };
+                UriBuilder ub = new UriBuilder(PostUri) { Scheme = webSocketsScheme };
```
int19h: Shouldn't this just map "http" -> "ws" and "https" -> "wss"? Are there any cases where we want anything different?

If this is to handle user input with "http://" in Workspaces window, I think it would be better to validate that there, and auto-correct it (i.e. if they edit the URL and use "http://", replace it with "https://" right there). Seeing "https" would indicate that connection is secured to the user.

MikhailArkhipov: This is for ?help

---
In reply to: [82295312](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82295312) [](ancestors = 82295312)

src/Common/Core/Impl/Security/SecurityUtilities.cs
MikhailArkhipov: This code was actually taken from where the SS was getting created off the Claims.Password

---
In reply to: [82293962](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82293962) [](ancestors = 82293962)

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -212,5 +214,13 @@ public class RemoteUriWebService : IRemoteUriWebService {
                 httpListenerResponse.Headers.Add(pair.Key, pair.Value);
             }
         }
+
+        private bool ValidateCertificate(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors) {
+            if ((sslPolicyErrors & SslPolicyErrors.RemoteCertificateNotAvailable) != 0) {
+                return false;
+            }
+            // Accept other cases. Validation for untrusted certificates is done by the broker.
```
MikhailArkhipov: Wrong comment. I meant certificate validation is done at the time we connect to the broker. This service does not need to go to the same extent.

---
In reply to: [82295549](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82295549) [](ancestors = 82295549)

src/Host/Broker/Impl/Security/TlsConfiguration.cs
MikhailArkhipov: Do we actually do this? 

---
In reply to: [82294136](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82294136) [](ancestors = 82294136)

src/Common/Core/Impl/Security/SecurityUtilities.cs
int19h: Yes, this is the root of the problem. It's not even about `Find`, fundamentally. The problem is that when HTTP request comes in and is parsed by Kestrel, the result of that parse is a bunch of non-secure strings containing all the header data - and that will include the credentials. So it doesn't really matter how far we push `SecureString` in our code, at some point we're going to run into that boundary, and we don't control the code on the other side of it.

We will likely need a custom authentication method that uses some sort of digest based auth (like HTTP Digest, but using something more secure than MD5), so that the password is secure as it flows through the request pipeline not under our control. Then when we decrypt it, it goes into `SecureString` immediately. The nonce should also be cleared immediately as soon as password is decrypted.

I'm not sure how to make this play well with `ClaimsCredential`, though. I think all claims are plain strings.

src/Host/Broker/Impl/ApiErrorResult.cs
MikhailArkhipov: Let me try that...

---
In reply to: [82297097](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82297097) [](ancestors = 82297097)

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -31,51 +31,57 @@ internal abstract class BrokerClient : IBrokerClient {
...
             Log = log;
+
+            _applicationWindowHandle = applicationWindowHandle;
             _interpreterId = interpreterId;
-            _credentials = credentials;
         }
 
-        protected void CreateHttpClient(Uri baseAddress) {
-            HttpClientHandler = new HttpClientHandler {
+        protected virtual void CreateHttpClient(Uri baseAddress, ICredentials credentials) {
```
int19h: `credentials` seems to be unused?

src/Host/Client/Impl/Host/LocalBrokerClient.cs
```diff
@@ -24,14 +24,15 @@ internal sealed class LocalBrokerClient : BrokerClient {
...
-        private static readonly LocalCredentialsDecorator Credentials = new LocalCredentialsDecorator();
+        private static readonly LocalCredentialsDecorator _credentials = new LocalCredentialsDecorator();
 
         private readonly string _rhostDirectory;
         private readonly string _rHome;
         private readonly BinaryAsyncLock _connectLock = new BinaryAsyncLock();
         private readonly ICoreServices _services;
 
         private Process _brokerProcess;
+        private int _disposed;
```
int19h: If I remember correctly, `DisposableBag` implements this for you. @AlexanderSher ?

src/Host/Broker/Impl/ApiErrorResult.cs
MikhailArkhipov: Yep, that works

---
In reply to: [82297314](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82297314) [](ancestors = 82297314,82297097)

src/Host/Client/Impl/Host/LocalBrokerClient.cs
MikhailArkhipov: It was throwing for me in some cases when object was disposed by the time `Add` was called.

---
In reply to: [82297812](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82297812) [](ancestors = 82297812)

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
MikhailArkhipov: I guess we could auto-match

---
In reply to: [82295561](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82295561) [](ancestors = 82295561,82295312)

src/Host/Broker/Impl/Security/TlsConfiguration.cs
int19h: We don't do this when we launch the broker for local connections. But anyone can start it manually from command line with multiple `--server.urls`, and this code doesn't know how it was started.

src/Host/Client/Impl/Host/LocalBrokerClient.cs
AlexanderSher: I'll add `TryAdd` method

---
In reply to: [82298237](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82298237) [](ancestors = 82298237,82297812)

src/Host/Broker/Impl/Security/TlsConfiguration.cs
MikhailArkhipov: Well, is it something we care about? If yes, does it then return IEnumerable<string>?

---
In reply to: [82298678](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82298678) [](ancestors = 82298678)

src/Host/Client/Impl/Host/LocalBrokerClient.cs
```diff
@@ -24,14 +24,15 @@ internal sealed class LocalBrokerClient : BrokerClient {
         private const string InterpreterId = "local";
 
         private static readonly bool ShowConsole;
-        private static readonly LocalCredentialsDecorator Credentials = new LocalCredentialsDecorator();
+        private static readonly LocalCredentialsDecorator _credentials = new LocalCredentialsDecorator();
```
AlexanderSher: VS showed me that it must be `Credentials`. What is our convention for `private static readonly`?

src/Host/Client/Impl/Host/RemoteBrokerClient.cs
```diff
@@ -2,55 +2,76 @@
...
+            _credentials = new RemoteCredentialsDecorator(applicationWindowHandle, brokerUri);
+            _authority = new UriBuilder { Scheme = brokerUri.Scheme, Host = brokerUri.Host, Port = brokerUri.Port }.ToString();
             CreateHttpClient(brokerUri);
         }
 
         public override string HandleUrl(string url, CancellationToken ct) {
             return WebServer.CreateWebServer(url, HttpClient.BaseAddress.ToString(), ct);
         }
 
+        protected override ICredentialsDecorator Credentials => _credentials;
```
AlexanderSher: This property is absolutely identical in both `RemoteBrokerClient` and `LocalBrokerClient`. I'd rather make it a ctor parameter as it was before

MikhailArkhipov: Sure. It was needed since it used to be NetworkCredential but since you added the decorator, it is no longer needed.

---
In reply to: [82300188](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82300188) [](ancestors = 82300188)

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
karthiknadig: The PostUri points to an endpoint on the broker `http[s]://<broker>:<port>/remoteuri`. This should match whatever scheme that is used to connect with the broker.

src/Common/Core/Impl/Security/SecurityServices.cs
```diff
@@ -0,0 +1,49 @@
...
+                    flags |= CREDUI_FLAGS_ALWAYS_SHOW_UI;
+                }
+                var credui = new CREDUI_INFO {
+                    cbSize = Marshal.SizeOf(typeof(CREDUI_INFO)),
+                    hwndParent = applicationWindowHandle,
+                };
+
+                // For password, use native memory so it can be securely freed.
+                passwordStorage = SecurityUtilities.CreatePasswordBuffer();
+                int err = CredUIPromptForCredentials(ref credui, authority, IntPtr.Zero, 0,
```
karthiknadig: Please use CredUIPromptForWindowsCredentials

src/Host/Broker/Impl/Sessions/SessionsController.cs
```diff
@@ -35,16 +35,8 @@ public class SessionsController : Controller {
...
-            string password = User.FindFirst(Claims.Password)?.Value;
-            if (password != null) {
-                securePassword = new SecureString();
-                foreach (var ch in password) {
-                    securePassword.AppendChar(ch);
-                }
-            }
-
             string profilePath = User.FindFirst(Claims.RUserProfileDir)?.Value;
+            var password = User.FindFirst(Claims.Password)?.Value.ToSecureString();
```
karthiknadig: Cant this be stored in Cred Manager instead of having this floating around? we can CredRead as needed.

int19h: That's an interesting idea. We could generate GUIDs for resource name, and store those GUIDs as claims. That way, any request that _doesn't_ need the password (most of them) will never even access it, and those that do, can `CredRead` directly into a `SecureString` at the point of use, and immediately discard the string after. If I understand correctly, cred manager encrypts everything it stores with a user-specific key, so dumps etc won't contain any recoverable passwords. 

However, how do we properly clean up stored credentials, especially if broker crashes? Is there any way to set a relatively short expiration date on them?

int19h: ,,, although `SecureString` provides the same encryption guarantees. So we might as well just have our own static dictionary of IDs to `SecureString`, and use IDs for claims. That way, it  all gets cleaned up when process exits, and has no potential to spill over to other apps. Also, cred manager requires a logon session, it seems, which might be bad if we want to make broker a service in the future?

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
MikhailArkhipov: They are auto-matched now

---
In reply to: [82314287](https://github.com/Microsoft/RTVS/pull/2447#discussion_r82314287) [](ancestors = 82314287)

MikhailArkhipov:
Going to merge if nothing else needed. Cred management is a separate item.

karthiknadig:
:shipit:

Issue #2409, opened 2016-09-29
Create DB connection in a separate environment


MikhailArkhipov:
And perhaps other settings?

```
I’m not really sold on creating objects in the global environment though. 
That’s not supposed to happen with well-behaved packages, and the same 
principle should apply for IDEs. Eg what happens if the user already has 
an object called dbConnection? Does it just get overwritten silently?

I think it would be better to make a custom environment for settings, 
and attach it to the search path. That way anything that RTVS creates 
will still be visible, but not affect user-created objects.
```

Hong-Revo:
Thinking about it some more, you probably shouldn’t even attach the settings environment. Instead, refer to all settings objects with the namespace-qualified name, `settings:::dbConnection`.

The risk with attaching the environment is that you can change the meaning of user code that doesn’t have explicit namespace references. Not using namespaces is a bad habit among many R programmers (me included).

Issue #2420, opened 2016-10-04
Backslashes in the SQL connection string content need to be escaped


MikhailArkhipov:
`\\`

huguesv:
If in general it's up to the user to escape the backslash for a regular setting (when typing it in the settings UI), then we probably want to do the escaping in `AddDbConnectionCommand.Handle`, because the command is adding the setting on behalf of the user.
Once it's added, loading and saving settings don't unescape/escape the values, and SQL isn't treated specially.

huguesv:
I'm changing my mind.  It should be escaped/unescaped when it's serialized/deserialized in settings.R (for all types of settings).  This matches what C# does with settings files, and resx files, ie. it's not up to the user to escape the values entered in the UI.

With my original plan, I was ending having to escape/unescape the settings values in different places where the SQL string was retrieved/updated, and that was getting too far away from where the settings.R file handling (too easy to miss).

huguesv:
Fixed in https://github.com/Microsoft/RTVS/pull/2449

PR #2449, opened 2016-10-06
Store settings in an environment.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/2409

File looks like this now:
```
# Application settings file.
# File content was generated on 10/6/2016 3:09:07 PM.

settings <- new.env()

settings$name1 <- 'value1'

settings$name2 <- 'value2'

```

And it can import settings from v0.5 or older. Once you save, it will rewrite it in new format.

So if you source the settings file, you access them via `settings$name1`.  That's not exactly what is mentioned in the bug though, which says you should be able to access them via `settings:::name1`.  Not sure if that was a mistake of if there's a way to achieve that.


src/R/Components/Impl/Application/Configuration/IO/ConfigurationSettingsWriter.cs
```diff
@@ -64,16 +66,16 @@ public sealed class ConfigurationSettingsWriter : IConfigurationSettingsWriter {
         }
 
         private static string FormatValue(IConfigurationSetting s) {
```
MikhailArkhipov: While you are at it, could you fix ##2420? If category is SQL, we should do `\\`

MikhailArkhipov:
👍

huguesv:
Added a fix for #2420.

All values are escaped/unescaped when saving/writing to Settings.R, which means the UI is slightly annoying to use when newlines are present, but user can always edit those directly in Settings.R. The benefit is that the user doesn't need to worry about generating invalid strings, there's no need to know they are stored in R code as opposed XML or whatever.

FYI `FromRStringLiteral` was not used before. I made changes to that one and `ToRStringLiteral` so that you can round-trip values that have characters that need escaping.

src/Host/Client/Impl/Extensions/RStringExtensions.cs
```diff
@@ -35,7 +46,33 @@ public static class RStringExtensions {
             for (int i = 1; i < s.Length - 1; ++i) {
                 char c = s[i];
                 if (escape) {
-                    sb.Append(c);
+                    // https://stat.ethz.ch/R-manual/R-devel/library/base/html/Quotes.html
+                    switch (c) {
```
MikhailArkhipov: looks like
```
if(ch == oneof(...)) {
     append('\');
     append(ch)
}

**Comment by MikhailArkhipov** [comment]
👍

File: src/Host/Client/Impl/Extensions/RStringExtensions.cs (hunk 2 above)
Comment (huguesv) [inline src/Host/Client/Impl/Extensions/RStringExtensions.cs]: that would be producing the equivalent of `"\\t"` instead of the `"\t"` we need.

File: src/R/Components/Impl/Application/Configuration/IO/ConfigurationSettingsWriter.cs [hunk 3]
```diff
@@ -29,16 +30,18 @@ public sealed class ConfigurationSettingsWriter : IConfigurationSettingsWriter {
 
         /// <summary>
         /// Persists settings into R file. Settings are written
-        /// into R file as assignment statements such as 'name &lt;- value'.
+        /// into R file as assignment statements such as 'settings$name &lt;- value'.
         /// Value
         /// </summary>
         public void SaveSettings(IEnumerable<IConfigurationSetting> settings) {
             WriteHeader();
+            _writer.WriteLine("settings <- new.env()");
```
int19h: `new.env()` creates an environment that has the current environment (probably `.GlobalEnv` here) as its parent; so if an identifier is not found, it will be looked in that parent. 

I would suggest rewriting this like so:
```r
settings <- as.environment(list(
    name1 = value1,
    ...,
    nameN = valueN));
```

huguesv: I had considered this, but it was a much larger change to the parsing code, and I wasn't sure if it was beneficial.

One thing that's not shown in my example at the top is that each setting may have metadata associated with it, which is stored in comments right now,  So with the new format it would look like this:
```
settings <- as.environment(list(
    # [Category] SQL
    # [Description] Database connection string
    # [Editor] ConnectionStringEditor
    name1 = 'value1',

    # [Category] SQL
    # [Description] Database connection string
    # [Editor] ConnectionStringEditor
    name2 = 'value2'));
```

One downside with that format is we would lose the ability to have a setting that references another. We have 2 types of setting right now, string or expression.
For example, right now it's possible to have:
```
first_name = 'john'
last_name = 'doe'
full_name = first_name + last_name
```

int19h: Good point. How about just using `as.environment(list())` to get a blank new environment that's not connected to anything else, and then using the existing approach to fill it?

huguesv: Done!

src/Host/Client/Impl/Extensions/RStringExtensions.cs
```diff
@@ -17,7 +17,18 @@ public static class RStringExtensions {
                 return nullValue;
             }
 
-            return quote + s.Replace("\\", "\\\\").Replace("" + quote, "\\" + quote) + quote;
+            return quote + s.Replace("\\", "\\\\").Replace("" + quote, "\\" + quote).EscapeCharacterConstants() + quote;
+        }
+
+        private static string EscapeCharacterConstants(this string s) {
+            s = s.Replace("\n", "\\n")
```
int19h: This looks wrong. Consider this string as input :
```
\\n
```
After the first replace (in the method above) it becomes:
```
\\\\n
```
Then this replace will make it into
```
\\\\\n
```
And now when R parses it, it'll see `\\` `\\` `\n` - and `\n` becomes an unintended newline, instead of literal `\` followed by literal `n`.

I'm actually not clear why this is needed at all. The original code replaces all instances of `\` with `\\`, turning them into literal backslashes instead of the beginning of an escape sequence. So if input is `\n`, it becomes `\\n`, and then R will unescape `\\` and turn it back to `\n` when parsing, as intended. So there shouldn't be a need for any special handling for `\n` and other such escapes.

src/Host/Client/Impl/Extensions/RStringExtensions.cs
```diff
@@ -35,7 +46,33 @@ public static class RStringExtensions {
...
+                        case 'a':
+                            sb.Append("\a");
+                            break;
+                        case 'f':
+                            sb.Append("\f");
+                            break;
+                        case 'v':
+                            sb.Append("\v");
+                            break;
+                        default:
```
int19h: There's also the hexadecimal escape, e.g. "\x20" for space. Note that those can be single- or double-digit, and they're case-insensitive - `"\xA"` and `"\x0a"` are both valid.

int19h: Also octal - `"\0123"` etc. Up to 3 digits, each of which can be anything from 0 to 8 inclusive (I have no idea why 8 is allowed, but that's what the code says):

https://github.com/wch/r-source/blob/trunk/src/main/scan.c#L259

src/Host/Client/Impl/Extensions/RStringExtensions.cs
huguesv: In settings textbox, the string with length 2 `\\n` becomes the R string literal `"\\\n"`, and when going back, the first backslash indicates escape, the second one is added to C# string.  Third backslash indicates escape, then 'n' means we add newline char to the C# string, which means textbox now has same original value `\\n`.

Without the change, we can't round-trip starting from settings.R.  If there's the R literal `"\t"` (length 1) it becomes just `t` in the settings textbox, and then when written back to settings.r, just `"t" again.

int19h: To clarify, when I wrote `\\n` above, I meant those literal three characters typed by the user - not `\n` as in newline. In C# terms, `@"\\n"`. So far as I can see, the rest of the reasoning is valid if you start from that point.

If the user actually types a backslash followed by the newline, we need to convert the newline itself to `\n` - there's no dealing with backslashes in the input in that case, so it's a different thing.

huguesv:
Added parsing of hex and octal escape sequence.

src/Host/Client/Impl/Extensions/RStringExtensions.cs
```diff
@@ -69,9 +68,61 @@ public static class RStringExtensions {
...
+                            sb.Append(c);
+                            break;
+                        case '"':
                             sb.Append(c);
                             break;
+                        case '\'':
+                            sb.Append(c);
+                            break;
+                        default:
+                            if (c >= '0' && c <= '7') {
```
int19h: Should be `<= '8'` per R sources. The rest of it doesn't change (i.e. `\8` is the same as `\10`).

int19h:
👍

src/Host/Client/Impl/Extensions/RStringExtensions.cs
huguesv: 8 is not valid
```
> print('\8')
Error: '\8' is an unrecognized escape in character string starting "'\8"
```

PR #2465, opened 2016-10-07
#2425, #2427 (autosaving and restoring session state)

### LINKED ISSUE #2425: Automatically save host state when connection to client is lost
Host process should be notified by the broker when the client has unexpectedly dropped connection for any reason, and should save its state to an Rdata file in a known location, to permit loading it in a new session.

### LINKED ISSUE #2427: Prompt to reload saved session state
Per #2425, host processes will save state on unexpected disconnects. When reconnecting, the user should get a prompt to reload the state if present.

int19h:
Resolve #2425: Automatically save host state when connection to client is lost
Resolve #2427: Prompt to reload saved session state

Disable reconnecting to live sessions.

When client disconnects from a websocket connection to broker, have broker disconnect from the host, so that the latter can shut down.

When shutting down host due to disconnect, save workspace image.

When starting a new session, check for autosaved workspace image, and prompt user to reload it.

When VS is closed, gracefully terminate any running hosts instead of just disconnecting from them.

When switching from one connection to another, gracefully terminate the old host once the switch is successful.

Expose session state in /sessions endpoint in the broker.

### DIFF
--- a/src/Host/Broker/Impl/Sessions/Session.cs
+++ b/src/Host/Broker/Impl/Sessions/Session.cs
@@ -61,6 +61,7 @@ namespace Microsoft.R.Host.Broker.Sessions {
             Id = Id,
             InterpreterId = Interpreter.Id,
             CommandLineArguments = CommandLineArguments,
+            State = State,
         };
 
         internal Session(SessionManager manager, IIdentity user, string id, Interpreter interpreter, string commandLineArguments, ILogger sessionLogger, ILogger messageLogger) {
@@ -212,15 +213,23 @@ namespace Microsoft.R.Host.Broker.Sessions {
         }
 
         private async Task ClientToHostWorker(Stream stream, IMessagePipeEnd pipe) {
-            while (true) {
-                var message = await pipe.ReadAsync(Program.CancellationToken);
-                var sizeBuf = BitConverter.GetBytes(message.Length);
-                try {
-                    await stream.WriteAsync(sizeBuf, 0, sizeBuf.Length);
-                    await stream.WriteAsync(message, 0, message.Length);
-                    await stream.FlushAsync();
-                } catch (IOException) {
-                    break;
+            using (stream) {
+                while (true) {
+                    byte[] message;
+                    try {
+                        message = await pipe.ReadAsync(Program.CancellationToken);
+                    } catch (PipeDisconnectedException) {
+                        break;
+                    }
+
+                    var sizeBuf = BitConverter.GetBytes(message.Length);
+                    try {
+                        await stream.WriteAsync(sizeBuf, 0, sizeBuf.Length);
+                        await stream.WriteAsync(message, 0, message.Length);
+                        await stream.FlushAsync();
+                    } catch (IOException) {
+                        break;
+                    }
                 }
             }
         }
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -97,7 +97,7 @@ namespace Microsoft.R.Host.Client.Host {
             await TaskUtilities.SwitchToBackgroundThread();
 
             try {
-                bool sessionExists = await IsSessionRunningAsync(name, cancellationToken);
+                bool sessionExists = false; //await IsSessionRunningAsync(name, cancellationToken);
 
                 WebSocket webSocket;
                 while (true) {
--- a/src/Host/Client/Impl/Session/RSessionProvider.cs
+++ b/src/Host/Client/Impl/Session/RSessionProvider.cs
@@ -48,6 +48,7 @@ namespace Microsoft.R.Host.Client.Session {
             _disposeToken.ThrowIfDisposed();
             return _sessions.GetOrAdd(guid, id => CreateRSession(guid));
         }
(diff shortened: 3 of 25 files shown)


int19h:
Requires https://github.com/Microsoft/R-Host/pull/93

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -89,7 +89,7 @@ internal abstract class BrokerClient : IBrokerClient {
             await TaskUtilities.SwitchToBackgroundThread();
 
             try {
-                bool sessionExists = await IsSessionRunningAsync(name, cancellationToken);
+                bool sessionExists = false; //await IsSessionRunningAsync(name, cancellationToken);
```
karthiknadig: >//await IsSessionRunningAsync(name, cancellationToken); [](start = 44, length = 55)

Is this not needed?

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -241,4 +245,55 @@ fetch_file <- function(file_path) {
...
+	if (identical(res, 'YES')) {
+		# Use try instead of tryCatch, so that any errors are printed as usual.
+		loaded <- FALSE;
+		try({
+			load(autosave_filename, envir = .GlobalEnv);
+			loaded <- TRUE;
+		});
+
+		if (loaded) {
+			message(sprintf('Loaded workspace from autosaved image "%s".\n', autosave_filename));
```
karthiknadig: >	 [](start = 1, length = 1)

tab to space

src/Host/Client/Impl/Host/BrokerClient.cs
int19h: Yes, that is the intent (since we decided to not go with reconnection in RTM). Comment is just so that it's readily obvious how to revert this later, when we get to it.

src/Host/Client/Impl/rtvs/R/util.R
int19h: Looks like we have a mix of tabs and spaces all over our .R files, and also different indentation levels. I'll go over and reformat them all.

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -81,7 +82,14 @@ public class RSessionProvider : IRSessionProvider {
                 return;
             }
 
-            foreach (var session in _sessions.Values) {
+            var sessions = GetSessions();
+            var stopHostTasks = sessions.Select(session => session.StopHostAsync());
+            try {
+                Task.WhenAll(stopHostTasks).GetAwaiter().GetResult();
```
AlexanderSher: What is this change for?

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -207,15 +208,23 @@ public class Session {
         }
 
         private async Task ClientToHostWorker(Stream stream, IMessagePipeEnd pipe) {
```
AlexanderSher: >ClientToHostWorker [](start = 27, length = 18)

`ClientToHostWorker` doesn't create a stream, so I don't think it should dispose it either

src/Host/Client/Impl/Session/RSessionProvider.cs
int19h: To gracefully stop the hosts when we close VS, instead of merely disconnecting from them (disposing an `RHost` instance disconnects from the broker, but doesn't tell it to terminate the process).

AlexanderSher: In this case, shouldn't we schedule it to run in UI thread?

---
In reply to: [82472969](https://github.com/Microsoft/RTVS/pull/2465#discussion_r82472969) [](ancestors = 82472969)

int19h:
Rebased and ready for merge - any additional feedback?

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: In this case, shouldn't we schedule it to run in UI thread?

MikhailArkhipov:
👍

src/Host/Client/Impl/Session/RSessionProvider.cs
int19h: Would it make a difference? All it does is do a bunch of evals, mostly, which don't care which thread they're on. Or are you saying that VS will not actually wait for `Dispose` to complete?

int19h: If I understand correctly, the chain of `Dispose` calls that gets us here starts with the disposal of some MEF component - won't that necessarily be waited on by MEF?

int19h: I ran it under debugger, and `Dispose` is, in fact, called on UI thread when VS is closed.

AlexanderSher: MEF can call `Dispose` from any thread, but you're probably right that VS should wait until its MEF components are disposed.

AlexanderSher:
👍

PR #2476, opened 2016-10-10
Fix #2444: TestBrokerConnectionWithRHost unexpectedly throws OperationCanceledException

### LINKED ISSUE #2444: TestBrokerConnectionWithRHost unexpectedly throws OperationCanceledException
```{csharp}
        private async Task RSessionOnDisconnectedAsync() {
            var token = await _brokerDisconnectedLock.WaitAsync();
            try {
                if (!token.IsSet) {
                    // We don't want to show that connection is broken just because one of the sessions has been disconnected. Need to test connection
                    await TestBrokerConnectionWithRHost(_brokerProxy, default(CancellationToken), default(ReentrancyToken));
                    token.Reset();
                }
                token.Set();
                OnBrokerDisconnected();
            } catch (Exception) {
                token.Reset();
            }
        }
```

`TestBrokerConnectionWithRHost` gets default `CancellationToken`, so it shouldn't throw `OperationCanceledException`.

AlexanderSher:


MikhailArkhipov:
👍

PR #2479, opened 2016-10-10
Update references, fix Dev15 build

MikhailArkhipov:
Some (but not complete) fixes for plot integration tests. More to come. Also, separated SQL into a separate assembly so we don't need to ship DACfx assemblies as it is caused trouble. Separation also allows us to run with or without SQL Tools present (port from 0.5)

RHost is https://github.com/Microsoft/R-Host/pull/95


huguesv:
Do we block install if not on Update 3? If not, we should after this.

👍

MikhailArkhipov:
We do

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -83,11 +83,10 @@ public class RSessionProvider : IRSessionProvider {
             }
 
             var sessions = GetSessions();
-            var stopHostTasks = sessions.Select(session => session.StopHostAsync());
             try {
+                var stopHostTasks = sessions.Select(session => session.StopHostAsync());
```
AlexanderSher: What exceptions do we actually expect here?

src/Package/Test/Sql/CommandTest.cs
```diff
@@ -122,7 +126,10 @@ public class CommandTest {
             });
             _pss.GetSelectedProject<IVsHierarchy>().Returns(hier);
 
-            var cmd = new PublishSProcCommand(_appShell, _pss);
+            var servicesProvider = Substitute.For<IDacPackageServicesProvider>();
+            servicesProvider.GetDacPackageServices().Returns(Substitute.For<IDacPackageServices>());
```
AlexanderSher: Shouldn't `NSubstitute` provide a default return value automatically?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -60,37 +62,30 @@ internal sealed class RSessionCallback : IRSessionCallback {
...
 
             await _coreShell.SwitchToMainThreadAsync();
-
             return await workflow.Plots.DeviceCreatedAsync(deviceId);
         }
 
         public async Task PlotDeviceDestroy(Guid deviceId, CancellationToken ct) {
-            var provider = _coreShell.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>();
-            var workflow = provider.GetOrCreate();
+            var workflow = _workflowProvider.GetOrCreate();
```
AlexanderSher: If `_workflowProvider` is disposed already, it will throw `ObjectDisposedException`. Do we handle it anywhere?

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: Actually, not sure. This change was experimental, I'll undo it.

src/Package/Test/Sql/CommandTest.cs
MikhailArkhipov: If it does then sure.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
MikhailArkhipov: It is a good question. I mentioned that fixes are incomplete... I think we may want to cache workflow here so we won't be calling create when everything is already destroyed.

src/R/Sql/Impl/Publish/DacPackageServices.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.ComponentModel.Composition;
+using Microsoft.R.Components.Sql.Publish;
+
+namespace Microsoft.VisualStudio.R.Sql.Publish {
+    [Export(typeof(IDacPackageServices))]
```
AlexanderSher: If you export `IDacPackageServices`, why do you need a `IDacPackageServicesProvider`?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
AlexanderSher: Ok then, let's discuss it as a separate issue: #2481

src/R/Sql/Impl/Publish/DacPackageServices.cs
MikhailArkhipov: The goal is to be able to load RTVS package without SQL tools and associated assemblies. We now only reference DacFx. We can get away with just `IDacPackageServices`. MEF catalog will see it, but only load for reflection. But - then all code that tries and use the interface will have to check for the SQL tools presence before attempting to use the interface. I think 'GetExportedValueOfDefault` will still throw if export is found, but not loadable - will it? With the provider which always exists call checks for SSDT and then returns null.

AlexanderSher: `GetExportedValueOfDefault` throws if there is more than one export, and doesn't throw if there are no exports. If export is found, it there is no difference - exceptions in export loading logic aren't handled.
But, if we export a provider, we don't need to export `IDacPackageServices` themselves - `DacPackageServicesProvider.GetDacPackageServices` can explicitly create an instance.

src/Package/Impl/Sql/Services/DacPackageServicesProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal sealed class DacPackageServicesProvider: IDacPackageServicesProvider {
+        private readonly ICoreShell _coreShell;
+
+        [ImportingConstructor]
+        public DacPackageServicesProvider(ICoreShell coreShell) {
+            _coreShell = coreShell;
+        }
+
+        public IDacPackageServices GetDacPackageServices() {
+            if(SqlTools.CheckInstalled(_coreShell)) {
```
AlexanderSher: Do we need `SqlTools` to be a separate class? Maybe move the method into `DacPackageServicesProvider`?

src/Package/Impl/Sql/Commands/PublishSProcCommand.cs
```diff
@@ -50,8 +51,15 @@ internal sealed class PublishSProcCommand : ICommandGroupHandler {
 
         public bool TryHandleCommand(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut) {
             if (commandId == RPackageCommandId.icmdPublishSProc) {
-                if (SqlTools.CheckInstalled(_appShell)) {
+                if (_dacServicesProvider.GetDacPackageServices() != null) {
```
AlexanderSher: Do you need to make similar changes for the `PublishSProcOptionsCommand`?

src/Package/Impl/Sql/Services/DacPackageServicesProvider.cs
MikhailArkhipov: I thought about it and decided maybe some other code might need it, but I guess we can make it private for now and factor it out later if needed.

src/R/Components/Test/Plots/RPlotIntegrationTest.cs
```diff
@@ -25,6 +25,8 @@
 
 namespace Microsoft.R.Components.Test.Plots {
     [ExcludeFromCodeCoverage]
+    [Category.Plots]
+    [Collection(CollectionNames.NonParallel)]
```
AlexanderSher: The main reason to use `NonParallel` is lack of isolated context. For example, shared files that can be overwritten or deleted or global singletons that carry their state between tests but are expected to be reset to some basic state between tests. `Microsoft.R.Components.Test` shouldn't have such cases cause it has individual MEF containers for every test and AFAIK doesn't use shared external resources that can be changed by several tests.

src/R/Sql/Impl/Publish/DacPackageServices.cs
MikhailArkhipov: But wouldn't that create hard link even if code is not called and throw on assembly loading? If code in `DacPackageServicesProvider` did `new DacPackageServices()` and that class, in turn, had code `new DacPacBuilder()` and the `DacPacBuilder` implementation called `TSqlModel` which was not present since SSDT assemblies are not installed, wouldn't that prevent package from loading?

I.e. the point is to prevent loading of `Microsoft.VisualStudio.R.Sql` when SSDT is not installed.

AlexanderSher: I actually think `IDacPackageServices` shouldn't been exported, cause otherwise someone can write code that imports `IDacPackageServices` instead of `IDacPackageServicesProvider`.

AlexanderSher: No, I'm not suggesting calling `new DacPackageServices()`, I thought about loading assembly and create an instance via `Activator`:
```{csharp}
var asm = Assembly.Load("Microsoft.VisualStudio.R.Sql");
var type = asm.GetTypes().FirstOrDefault(t => typeof(IDacPackageServices).IsAssignableFrom(typeof(C1)));
var instance = (IDacPackageServices)Activator.CreateInstance(type);
```

MikhailArkhipov: Yeah, OK. I tried to avoid that thing :-)

AlexanderSher: If you want to do it through MEF, the right way is to load "Microsoft.VisualStudio.R.Sql.dll" into its own assembly catalog that doesn't pollute main VS MEF catalog with information about incorrectly loaded assembly. With only one type required, I'd prefer simple reflection.

AlexanderSher:
👍

PR #2482, opened 2016-10-11
Switched Credential UI back CredUIPromptForCredentials instead of CredUIPromptForWindowsCredentials

### LINKED ISSUE #2478: CredUIPromptForWindowsCredentials in common.core has incorrect PInvoke signature
Current:
            ref CREDUI_INFO pUiInfo,
            string pszTargetName,
            IntPtr Reserved,
            int dwAuthError,
            StringBuilder pszUserName,
            int ulUserNameMaxChars,
            IntPtr pszPassword,
            int ulPasswordMaxChars,
            ref bool pfSave,
            int dwFlags

Should be:
            ref CREDUI_INFO credInfo,
            int dwAuthError,
            ref uint authPackage,
            IntPtr InAuthBuffer,
            uint InAuthBufferSize,
            out IntPtr refOutAuthBuffer,
            out uint refOutAuthBufferSize,
            ref bool pfSavec,
            int dwFlags

karthiknadig:
CredUIPromptForWindowsCredentials  does not provide a delayed save feature. Until we have a similar blessed mechanism with the newer API it seems better to use the older one. Fixes #2478

### DIFF
--- a/src/Common/Core/Impl/NativeMethods.cs
+++ b/src/Common/Core/Impl/NativeMethods.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Common.Core {
         public const int CREDUI_FLAGS_KEEP_USERNAME = 0x100000;
 
         [DllImport("credui", CharSet = CharSet.Auto)]
-        public static extern int CredUIPromptForWindowsCredentials(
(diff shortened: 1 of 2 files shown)


int19h:
👍

PR #2486, opened 2016-10-11
Fix #2484: Reload .RData shows in local connection case

### LINKED ISSUE #2484: Reload .RData shows in local connection case
Run RTVS with local connection
Debug | Terminate All
Run again

image

int19h:
Change message so that it is applicable to situations where the disconnect was caused by VS and/or broker crashing or being forcibly terminated.


MikhailArkhipov:
👍

Issue #2480, opened 2016-10-10
Formatting: Spaces placed around hyphen used for numeric negation 


bordecal:
When a hyphen is used for numeric negation in an expression preceded by an opening parenthesis, it is wrapped in spaces as if it were being used for subtraction.  For example, the second line from the following original code

```R
x <- 5
y <- (-x + 2) / 3
```

will be reformatted as
```R
y <- ( - x + 2) / 3
```

This does not occur when the hyphen is not preceded by an opening parenthesis.  e.g.:

```R
y <- -x + 1
```

It does, however, happen even if the parentheses are used for a function call.  e.g.:

```R
y <- sqrt( - x + 1)
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f4755eedde9c2354cf6fdf3ca5e92ccafb40472d

Issue #2470, opened 2016-10-07
Formatting: comments are not indented same as statements around them


int19h:
Before:
```r
f <- function() {
  x <- 1;
  # foo
  x <- 2
}
```
After:
```r
f <- function() {
    x <- 1;
  # foo
    x <- 2
}
```

Curiously, this doesn't happen if the initial indentation is zero - if it is, then comment is indented same as the surrounding code.

MikhailArkhipov:
Comments were never formatted/indented, comment positioning is a bit of a matter of taste (I don't always agree with the way C# insists on) but we can add this.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f4755eedde9c2354cf6fdf3ca5e92ccafb40472d

PR #2489, opened 2016-10-11
#2480 #2470 and unfiled crash in remote ?help

MikhailArkhipov:
#2480 Spaces placed around hyphen used for numeric negation			
#2470 Formatting: comments are not indented same as statement

Fix crash on remote ?help

### DIFF
--- a/src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
+++ b/src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
@@ -26,10 +26,9 @@ namespace Microsoft.R.Host.Client.BrokerServices {
 
         public async Task GetResponseAsync(HttpListenerContext context, string localBaseUrl, string remoteBaseUrl, CancellationToken ct) {
             string postUri = null;
-            bool https = new Uri(remoteBaseUrl, UriKind.Absolute).IsHttps();
 
             if (context.Request.IsWebSocketRequest) {
-                UriBuilder ub = new UriBuilder(PostUri) { Scheme = https ? "wss" : "ws" };
+                UriBuilder ub = new UriBuilder(PostUri) { Scheme ="wss" };
                 postUri = ub.Uri.ToString();
             } else {
                 postUri = PostUri.ToString();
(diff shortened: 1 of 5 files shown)


src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -26,10 +26,9 @@ public class RemoteUriWebService : IRemoteUriWebService {
 
         public async Task GetResponseAsync(HttpListenerContext context, string localBaseUrl, string remoteBaseUrl, CancellationToken ct) {
             string postUri = null;
-            bool https = new Uri(remoteBaseUrl, UriKind.Absolute).IsHttps();
 
             if (context.Request.IsWebSocketRequest) {
-                UriBuilder ub = new UriBuilder(PostUri) { Scheme = https ? "wss" : "ws" };
```
int19h: Doesn't this regress local help? Or do we skip the URI service in that case?

MikhailArkhipov:
It is not used in the local case

src/R/Core/Impl/Formatting/RFormatter.cs
```diff
@@ -440,7 +440,10 @@ public class RFormatter {
         private void AppendOperator() {
             string text = _textProvider.GetText(_tokens.CurrentToken);
             if (_tokens.Position == 0 || TokenOperator.IsUnaryOperator(_tokens, _textProvider, TokenOperator.GetOperatorType(text), -1)) {
-                AppendToken(leadingSpace: true, trailingSpace: false);
+                // Determine if we need to insert leading space. We do in 'x + -1' but not in x + (-y + 1)
+                var precedingTokenType = _tokens.LookAhead(-1).TokenType;
+                bool leadingSpace = precedingTokenType == RTokenType.Operator;
```
int19h: How does this handle cases where the preceding operator is also unary? E.g. `+-x` or `~-x`.

Also, for `x + -1`, wouldn't both spaces just come from the `+`?

MikhailArkhipov:
We can technically walk back the chain until we find non-unary. Let me try that. As for spaces, in the text builder all operations are 'soft' so adding another space on top of existing one is no-op. Similarly, soft indent doesn't actually insert spaces/tabs until some non-ws sequence is inserted.

int19h:
I see. But so long as all binary operators always add spaces around them, it sounds like we should just never add a space here at all, regardless of what precedes it? So far as I can see, it should always produce the right result without having to do any checks.

MikhailArkhipov:
Good point

int19h:
👍

PR #2490, opened 2016-10-12
Fixes #2487 bunch of fixes

### LINKED ISSUE #2487: Remote installer issues
1. Installer needs a default broker config file.
2. Add monitor process to start-up
3. User service status signal APIs to let windows know that service is starting/stopping.

karthiknadig:
Added json config file for broker.
Added code to autostart monitor.
Adds status info for the user profile service.


src/Host/Broker/Impl/Microsoft.R.Host.Broker.Config.json
```diff
@@ -0,0 +1,3 @@
+﻿{
+  "server.urls": "https://0.0.0.0:5000"
```
int19h: Now that we're actually configuring the port out of the box, instead of letting the user do so, let's pick something better - 5000 is the default for all ASP.NET Core apps. 

https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt

src/Host/UserProfile/Impl/Program.cs
```diff
@@ -12,7 +12,7 @@ static class Program {
         static void Main() {
 #if DEBUG
             // Uncomment the line below to debug the Broker-Service
-            // Debugger.Launch();
+            Debugger.Launch();
```
int19h: This should stay commented by default.

src/Host/Broker/Impl/Microsoft.R.Host.Broker.csproj
```diff
@@ -173,6 +173,9 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>True</Visible>
     </Content>
+    <None Include="Microsoft.R.Host.Broker.Config.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
int19h: This should be "Newer" to avoid F5 always doing a rebuild.

int19h:
👍

PR #2492, opened 2016-10-12
Fix #2325: Switching sessions doesn't stop debugging

### LINKED ISSUE #2325: Switching sessions doesn't stop debugging
When R debugger is attached to the session, switching sessions (e.g. from local to remote) doesn't stop debugging. I haven't looked into what other problems it will cause down the line, but there will definitely be many because the state is essentially radically changed without debugger being aware (including Browse/non-Browse prompt and other aspects). We should force-detach debugger when switching.

int19h:
Stop debugging when broker switching is complete.


MikhailArkhipov:
👍

PR #2493, opened 2016-10-13
Broker as Network Service

karthiknadig:
This is for future. when we want to move broker to network service. I just wanted to have the code i had written to launch process as user to be saved somewhere.
This change does the following:
1. Allows broker to run as a service (using argument **--start.as service** remote scenario), by default runs as regular exe (local scenario).
2. Updated remote installer that installs **Broker** as a **NetworkService**
3. Monitoring process dependency **removed** entirely
4. Hardens the named pipe between UserProfile (Local System) and Broker (Network Service).


src/Host/Broker/Impl/ProjectInstaller.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.ComponentModel;
+using System.IO;
+
+namespace Microsoft.R.Host.Broker {
+    [RunInstaller(true)]
+    public partial class ProjectInstaller : System.Configuration.Install.Installer {
```
MikhailArkhipov: Prop don't need long name, 'using' is preferable

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -75,105 +75,150 @@ public class Session {
...
-                Arguments = arguments,
-                RedirectStandardInput = true,
-                RedirectStandardOutput = true,
-                RedirectStandardError = true,
-                LoadUserProfile = true
-            };
-            
+            var usernameBldr = new StringBuilder(NativeMethods.CREDUI_MAX_USERNAME_LENGTH + 1);
+            var domainBldr = new StringBuilder(NativeMethods.CREDUI_MAX_PASSWORD_LENGTH + 1);
+
```
MikhailArkhipov: Domain name length same as pwd length?

src/Host/Broker/Impl/Startup/Program.cs
```diff
@@ -1,124 +1,31 @@
...
         public static void Main(string[] args) {
             var configBuilder = new ConfigurationBuilder().AddCommandLine(args);
-            Configuration = configBuilder.Build();
+            var configuration = configBuilder.Build();
 
-            string configFile = Configuration["config"];
+            string startAs = configuration["start.as"];
+            string configFile = configuration["config"];
             if (configFile != null) {
                 configBuilder.AddJsonFile(configFile, optional: false);
```
MikhailArkhipov: Do we want try/catch if file doesnt exist, log something?

src/Host/Broker/Impl/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,97 @@
...
+                                string value = envVar.Substring(idx + 1);
+                                eb.Add(key, value);
+                            }
+                        }
+                        // unicode string + unicode null
+                        ptr += data.Length + 2;
+                    }
+                }
+            } finally {
+                DestroyEnvironmentBlock(env);
```
MikhailArkhipov: Probably only when it is not zero

src/Host/Broker/Impl/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,97 @@
...
+            get {
+                if (ContainsKey(key)) {
+                    return base[key];
+                } else {
+                    return null;
+                }
+            }
+            set {
+                if (ContainsKey(key)) {
+                    base[key] = value;
```
MikhailArkhipov: I think you can always do [key] = value, it doesnt have to exist

src/Host/Broker/Impl/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            Marshal.Copy(data, 0, ptr, data.Length);
+
+            length = data.Length;
+            return ptr;
+        }
+
+        public new string this[string key] {
+            get {
+                if (ContainsKey(key)) {
```
MikhailArkhipov: TryGetValue() so it won't have to do key checking twice

src/Host/Broker/Impl/Win32Process.cs
```diff
@@ -0,0 +1,229 @@
...
+                ref si,
+                out pi)) {
+                throw new Win32Exception(Marshal.GetLastWin32Error());
+            }
+
+            stdin = new FileStream(new SafeFileHandle(stdinWrite, true), FileAccess.Write, 0x1000, false);
+            stdout = new FileStream(new SafeFileHandle(stdoutRead, true), FileAccess.Read, 0x1000, false);
+            stderror = new FileStream(new SafeFileHandle(stderrorRead, true), FileAccess.Read, 0x1000, false);
+
+            // TODO: handle cleanup for process and thread
```
MikhailArkhipov: We prob need to destroy handles here if something fails

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -75,105 +75,150 @@ public class Session {
...
+                int pid  = Win32Process.StartProcessAsUser(useridentity, rhostExePath, arguments, Path.GetDirectoryName(rhostExePath), eb, out stdin, out stdout, out stderror);
 
-                psi.EnvironmentVariables["USERPROFILE"] = $"{psi.EnvironmentVariables["HOMEDRIVE"]}{psi.EnvironmentVariables["HOMEPATH"]}";
-                _sessionLogger.LogTrace(Resources.Trace_EnvironmentVariable, "USERPROFILE", psi.EnvironmentVariables["USERPROFILE"]);
+                _process = Process.GetProcessById(pid);
+                _process.EnableRaisingEvents = true;
 
-                psi.EnvironmentVariables["APPDATA"] = $"{psi.EnvironmentVariables["USERPROFILE"]}\\AppData\\Roaming";
-                _sessionLogger.LogTrace(Resources.Trace_EnvironmentVariable, "APPDATA", psi.EnvironmentVariables["APPDATA"]);
+                _process.Exited += delegate {
```
int19h: There is a race condition here now, since we start the process before subscribing the event. I'm not sure if `Process` does any magic to still raise the event in retrospect when a handler gets added, but most likely not. So this code might need to be factored into a helper method, and then you can call that helper both from here and from `if (HasExited)` below.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -75,105 +75,150 @@ public class Session {
...
+                }
 
-                psi.EnvironmentVariables["TEMP"] = $"{psi.EnvironmentVariables["LOCALAPPDATA"]}\\Temp";
-                _sessionLogger.LogTrace(Resources.Trace_EnvironmentVariable, "TEMP", psi.EnvironmentVariables["TEMP"]);
+                _sessionLogger.LogInformation(Resources.Info_StartedRHost, Id, User.Name);
+
+                HostToClientErrorWorker(stderror, _process.Id, (int processid, string errdata) => {
+                    outputLogger?.LogTrace(Resources.Trace_ErrorDataReceived, processid, errdata);
+                }).DoNotWait();
+            } else {
```
int19h: Can we unify both scenarios under the first branch? i.e. just use regular process creation Win32 function, but leave everything else the same? I'm worried that there will be subtly different behavior here due to differing implementations of pipes etc - the more can be unified, the less likely that is.

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -0,0 +1,131 @@
...
+                .AddConsole(LogLevel.Trace)
+                .AddProvider(new FileLoggerProvider(_startupOptions.Name));
+            _logger = _loggerFactory.CreateLogger<Program>();
+
+            if (_startupOptions.Name != null) {
+                _logger.LogInformation(Resources.Info_BrokerName, _startupOptions.Name);
+            }
+        }
+
+        public static void StartApp(IConfigurationRoot configuration) {
```
int19h: Argument appears to be unused?

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -0,0 +1,131 @@
...
+            var tlsConfig = new TlsConfiguration(_logger, _securityOptions);
+            var httpsOptions = tlsConfig.GetHttpsOptions(Configuration);
+            CreateWebHost(httpsOptions).Run();
+        }
+
+        public static void StartService(IConfigurationRoot configuration) {
```
int19h: Same here.

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -0,0 +1,131 @@
...
+#if DEBUG
+            // Uncomment line below to debug service
+            Debugger.Launch();
+#endif
+            ServiceBase.Run(new ServiceBase[] { new BrokerService() });
+        }
+
+        public static void CreateAndRunWebHostForService() {
```
int19h: This appears to be identical to `StartApp` above.

src/Host/Broker/Impl/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,97 @@
...
+                    byte[] data = Encoding.Unicode.GetBytes(envData);
+                    ms.Write(data, 0, data.Length);
+                    ms.Write(nulls, 0, nulls.Length);
+                }
+                ms.Write(nulls, 0, nulls.Length);
+                return ms.ToArray();
+            }
+        }
+
+        public IntPtr ToPtr() {
```
int19h: Let's call these `ToHGlobal`, so that it's clear what kind of memory management is necessary for the returned value.

src/Host/Broker/Impl/Win32EnvironmentBlock.cs
karthiknadig: Dictionary specifically throws KeyNotFoundException. Comment in the object viewer for the operator.
 If the specified key is not found, a get operation throws a System.Collections.Generic.KeyNotFoundException

int19h: That's only for the getter. So `x = d[k]` will throw if there is no `k`, but `d[k] = x` will work and add a new entry.

MikhailArkhipov: For the getter you can do `TryGetValue`

karthiknadig:
Codebase has diverged a lot. Will create a new PR when I have this ready.

PR #2494, opened 2016-10-13
Fix #2485: Progress bar shouldn't block message boxes

### LINKED ISSUE #2485: It is possible to initiate 'Connect' when 'Testing' is in progress
So the repro in the test connection case is

    Test connection
    Cred dialog comes up
    Enter creds
    Progress dialog appears

image

Canceling it hangs VS

```
 	mscorlib.dll!System.Threading.Tasks.Task.Wait(System.TimeSpan timeout)	Unknown
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.WaitSynchronouslyCore(System.Threading.Tasks.Task task) Line 313	C#
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.WaitSynchronously(System.Threading.Tasks.Task task) Line 290	C#
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTask.CompleteOnCurrentThread() Line 839	C#
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.Run(System.Func<System.Threading.Tasks.Task> asyncMethod) Line 412	C#
 	Microsoft.VisualStudio.R.Package.dll!Microsoft.VisualStudio.R.Package.Shell.VsProgressDialog.Show(System.Func<System.Threading.CancellationToken, System.Threading.Tasks.Task> method, string waitMessage, int delayToShowDialogMs) Line 21	C#

```

AlexanderSher:


src/Common/Core/Impl/Disposables/DisposeToken.cs
```diff
@@ -14,7 +14,7 @@ public sealed class DisposeToken {
             return new DisposeToken(Invariant($"{typeof(T).Name} instance is disposed"));
         }
 
-        public bool IsDisposed => _disposed == 0;
+        public bool IsDisposed => _disposed == 1;
```
MikhailArkhipov: While we are at it - in order to unblock tests, shall we at least apply a workaround in `DisconnectAsync` - only `await _runTask` when it is running (or do we actually care at all?) and same in `StopHostAsync`. Also, check disposed in `RSessionOnDisconnectedAsync` and not call `TestBrokerConnectionWithRHost`.

MikhailArkhipov:
👍

Issue #2481, opened 2016-10-11
IRSessionCallback concurrent access


AlexanderSher:
It is possible that `IRSessionCallback` methods are called when `IRInteractiveWorkflowProvider` is disposed or is going to start disposing (so that `workflow` can be acquired but can't be used). `IRInteractiveWorkflowProvider` can be disposed from any thread, so dispatching `IRSessionCallback` methods calls to the UI thread won't help.

MikhailArkhipov:
#2464

AlexanderSher:
Closing because of #2519, #2520, #2521

Issue #2445, opened 2016-10-06
About Host should report number of connected users


MikhailArkhipov:
Useful to gauge how heavily machine is being used.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/d438dc8cf54f4b468cdf5687d4de94cd7f754da2

Issue #2200, opened 2016-08-25
"Include In Project" gives "Sequence contains no elements" error


klausmh:
My project files tend to get hidden in the Solution Explorer. When selecting "Show All Files" and then "Include In Project", I get a "Sequence contains no elements" dialog box error.

MikhailArkhipov:
R project is not a traditional VS project (i.e. files are not listed in the project file so it is compatible with how R Studio works). Solution Explorer simply shows content of the folder. So 'Include in project' doesn't apply here. If some files are not visible, can you please give us more details - perhaps send us logs via R Tools -> Feedback -> Send Frown.

Thanks!

MikhailArkhipov:
Can also be one of #2108 #2060

MikhailArkhipov:
We have 'exclude from project' disabled/removed, need to do the same with this one.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/d438dc8cf54f4b468cdf5687d4de94cd7f754da2

PR #2501, opened 2016-10-13
#2200 #2445 and workarounds to unblock test runs

MikhailArkhipov:
#2445 About Host should report number of connected users
#2200 "Include In Project" gives "Sequence contains no elements" error
Test **workarounds** (#2481)
Dev15 reference fixes


src/Package/Impl/ProjectSystem/Commands/DisabledProjectCommand.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+using System.Collections.Immutable;
+using Microsoft.VisualStudio.ProjectSystem;
+#if VS14
+using Microsoft.VisualStudio.ProjectSystem.Designers;
+#endif
+
+namespace Microsoft.VisualStudio.R.Package.ProjectSystem.Commands {
+    internal abstract class DisabledProjectCommand : ICommandGroupHandler {
```
int19h: Can't quite figure out what the name for this means. Looking at the code, it's more like `ProjectNodeSelectionCommand`?

MikhailArkhipov: It is command that is disabled/removed from the project menu... since it doesnt apply

MikhailArkhipov: Maybe it is over flexible and should not have the node selector...

AlexanderSher:
What tests are these fixes addressing?

MikhailArkhipov:
PlotIntegration, HistoryIntegration

AlexanderSher:
I'm currently testing a fix for #2481, it allows `PlotIntegration` to pass.
To be clear - it is not just a test issue. There is a real deadlock that happens when plot (or any other request) tries access UI thread while RSessionProvider.Dispose is synchronously waits on UI thread for that request to complete.

MikhailArkhipov:
Yes I know, that is why it was called 'workaround'. If you are close, I'll delay the merge.

MikhailArkhipov:
Added friendly certificate message

src/Host/Monitor/Impl/MainWindow.xaml
```diff
@@ -1,21 +1,21 @@
...
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:Microsoft.R.Host.Monitor"
-        mc:Ignorable="d"
-        Title="R Host Monitor" Height="350" Width="525">
-    <Grid>
+        Title="R Remote Services Monitor" Height="350" Width="560" ResizeMode="NoResize"
+        Background="#FFFCFCFC">
```
int19h: If we're mucking around with background color, text color may also need to be adjusted, so that it doesn't get unreadable with some themes. Also, it needs to be tested with various accessibility aids enabled, esp. high contrast.

I think it might be easier to just leave it alone and use the standard Windows theme colors. There's no particular reason to give this special styling - it's not inside VS, so it doesn't need to follow VS theme, and for a standalone application, following the Windows theme is preferable.

MikhailArkhipov: It looks ugly with white background AND white status field.

MikhailArkhipov: Applied theme is Aero which is light anyway

int19h: Well, it's up to Windows to fix their theme :)

We could style the status field though. So long as it's both background and text color, so it's always legible.

int19h: If we explicitly apply light Aero somewhere (I don't see this in XAML), then we have accessibility issues with the app, because it doesn't respect high-contrast settings.

If we don't, then it's not a given that the theme is actually light. Windows has a "High Contrast Black" theme out of the box, that can be enabled in settings (Ease of Access | High Contrast).

MikhailArkhipov: Kathik says it is temporary anyway

MikhailArkhipov:
I added null check for now as well as not saving settings on construction

PR #2502, opened 2016-10-14
Fixes issue with server installation.

karthiknadig:
This change adds restart as post install requirement. Restarting makes it more reliable at starting a LocalSystem service.


int19h:
👍

PR #2516, opened 2016-10-15
Fix #2326: Crash after switching back to local

### LINKED ISSUE #2326: Crash after switching back to local
I was connected to remote, and clicked to connect to local Microsoft R Client.

Everything crashed:
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.R.Host.Client.REvaluationException
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions.ThrowOnError(System.String, Microsoft.R.Host.Client.REvaluationResult)
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions+<ExecuteAsync>d__3.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Host.Client.Session.RSession+<AfterHostStarted>d__96.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Host.Client.Session.RSession+<EvaluateAsync>d__78.MoveNext()

```

```
Application: Microsoft.R.Host.Broker.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ObjectDisposedException
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean ByRef)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(System.Runtime.InteropServices.SafeHandle, Boolean ByRef)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv+NativeMethods.uv_async_send(Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvAsyncHandle)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.async_send(Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvAsyncHandle)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.PostCloseHandle(System.Action`1<IntPtr>, IntPtr)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvHandle.ReleaseHandle()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvStreamHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
   at System.Runtime.InteropServices.SafeHandle.Finalize()

```

Logs in \pytools\r\Bugs\1606

int19h:
Improve exception handling when initializing the new session. For errors that will not cause other components to malfunciton later (e.g. setting codepage or mirror), print out the error, but continue. For errors that may cause major malfunctions, print out the detailed error and terminate the session immediately, but avoid VS crashing.

### DIFF
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -360,41 +360,67 @@ namespace Microsoft.R.Host.Client.Session {
         }
 
         private async Task AfterHostStarted(RSessionEvaluationSource evaluationSource, RHostStartupInfo startupInfo) {
-            using (var evaluation = await evaluationSource.Task) {
-                // Load RTVS R package before doing anything in R since the calls
-                // below calls may depend on functions exposed from the RTVS package
-                var libPath = IsRemote ? "." : Path.GetDirectoryName(Assembly.GetExecutingAssembly().GetAssemblyPath());
+            try {
+                using (var evaluation = await evaluationSource.Task) {
+                    // Load RTVS R package before doing anything in R since the calls
+                    // below calls may depend on functions exposed from the RTVS package
+                    var libPath = IsRemote ? "." : Path.GetDirectoryName(Assembly.GetExecutingAssembly().GetAssemblyPath());
 
-                await LoadRtvsPackage(evaluation, libPath);
+                    await LoadRtvsPackage(evaluation, libPath);
 
-                if (!IsRemote && startupInfo.WorkingDirectory != null) {
-                    await evaluation.SetWorkingDirectoryAsync(startupInfo.WorkingDirectory);
-                } else {
-                    await evaluation.SetDefaultWorkingDirectoryAsync();
-                }
+                    if (!IsRemote && startupInfo.WorkingDirectory != null) {
+                        await evaluation.SetWorkingDirectoryAsync(startupInfo.WorkingDirectory);
+                    } else {
+                        await evaluation.SetDefaultWorkingDirectoryAsync();
+                    }
 
-                var callback = _callback;
-                if (callback != null) {
-                    await evaluation.SetVsGraphicsDeviceAsync();
+                    var callback = _callback;
+                    if (callback != null) {
+                        await evaluation.SetVsGraphicsDeviceAsync();
 
-                    string mirrorUrl = callback.CranUrlFromName(startupInfo.CranMirrorName);
-                    await evaluation.SetVsCranSelectionAsync(mirrorUrl);
+                        string mirrorUrl = callback.CranUrlFromName(startupInfo.CranMirrorName);
 
-                    await evaluation.SetCodePageAsync(startupInfo.CodePage);
-                    await evaluation.SetROptionsAsync();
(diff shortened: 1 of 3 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -358,41 +358,67 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
+                        }
+
+                        try {
+                            await evaluation.SetCodePageAsync(startupInfo.CodePage);
+                        } catch (REvaluationException ex) {
+                            await WriteErrorAsync(Resources.Error_SessionInitializationCodePage, startupInfo.CodePage, ex.Message);
+                        }
+
+                        try {
+                            await evaluation.SetROptionsAsync();
```
AlexanderSher: Should we continue initialization if any of the steps has failed?

int19h: Some of those things aren't critical - say, if the codepage wasn't set properly, there might be bugs due to that (which is why it's still good to spell it out), but it doesn't really hinder any of our toolwindows from working properly. So I figured that it would be better to not block the user from using the product altogether for those errors - we should still get bug reports from the error messages that will be printed out. 

In contrast, failing to load the `rtvs` package is a fatal error. Some of these others, I wasn't sure about, and in all cases where I thought there was some chance of things breaking in a spectacular way down the line (i.e. some toolwindow not working or VS crashing), I didn't allow it to continue after those.

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -358,41 +358,67 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
+                        } catch (REvaluationException ex) {
+                            await WriteErrorAsync(Resources.Error_SessionInitializationOptions, ex.Message);
+                        }
+
+                        await evaluation.OverrideFunctionAsync("setwd", "base");
+                        await evaluation.SetFunctionRedirectionAsync();
+
+                        try {
```
MikhailArkhipov: Perhaps make some `ProtectedEval` or `IgnorableEval` extension with optional print of the message? I think there are a few other places that perform similar things.

MikhailArkhipov:
👍

src/Host/Client/Impl/Session/RSession.cs
int19h: I'll need to think about how to do that one right. It's easy to do for no-result evals, but most of evals outside of this place return values.

PR #2522, opened 2016-10-17
Fix couple of issues

### LINKED ISSUE #2426: Do not silently overwrite sessions that are in the process of saving their state
#2425 means that sessions can be in a "currently live, but saving state after connection was lost" state. There's a possibility of a race condition there whereby the user might try to reconnect when session is still in that state; currently, we just kill & restart the session, thus losing any data it might have been saving.

VS needs to recognize this new state, and prompt the user whether they want to reconnect right now (and lose the state), or wait until the state is fully saved.

### LINKED ISSUE #2519: Add Wait(Task) method that supports waiting on UI thread without blocking it
Lack of this method causes deadlocks when synchronous methods wait tasks on UI thread

AlexanderSher:
- Fix #2519: Add Wait(Task) method that supports waiting on UI thread without blocking it
- Fix #2426: Do not silently overwrite sessions that are in the process of saving their state
- Removed unit-test workaroung

### DIFF
--- a/src/Common/Core/Impl/Tasks/ITaskService.cs
+++ b/src/Common/Core/Impl/Tasks/ITaskService.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core.Tasks {
+    public interface ITaskService {
+        bool Wait(Task task, CancellationToken cancellationToken = default(CancellationToken), int ms = Timeout.Infinite);
+    }
+}
--- a/src/Host/Client/Impl/Host/BrokerClientProxy.cs
+++ b/src/Host/Client/Impl/Host/BrokerClientProxy.cs
@@ -34,11 +34,10 @@ namespace Microsoft.R.Host.Client.Host {
 
         public Task PingAsync() => _broker.PingAsync();
 
-        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000,
-            CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
+        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000, bool preserveSessionData = false, CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
 
             using (await _connectArwl.ReaderLockAsync(cancellationToken, reentrancyToken)) {
-                return await _broker.ConnectAsync(name, callbacks, rCommandLineArguments, timeout, cancellationToken);
+                return await _broker.ConnectAsync(name, callbacks, rCommandLineArguments, timeout, preserveSessionData, cancellationToken, reentrancyToken);
             }
         }
 
--- a/src/Host/Client/Impl/Host/BrokerClientProxy.cs
+++ b/src/Host/Client/Impl/Host/BrokerClientProxy.cs
@@ -34,10 +34,10 @@ namespace Microsoft.R.Host.Client.Host {
 
         public Task PingAsync() => _broker.PingAsync();
 
-        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000, bool preserveSessionData = false, CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
+        public async Task<RHost> ConnectAsync(BrokerConnectionInfo connectionInfo, CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
 
             using (await _connectArwl.ReaderLockAsync(cancellationToken, reentrancyToken)) {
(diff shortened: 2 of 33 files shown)


src/Common/Core/Impl/Tasks/ITaskService.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core.Tasks {
+    public interface ITaskService {
```
MikhailArkhipov: At some point we may move DispatchOnUIThread here

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -396,16 +396,16 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
                     await _transport.CloseAsync();
                     token.Set();
                 } catch (OperationCanceledException) {
+                    token.Set();
                 } catch (MessageTransportException) {
+                    token.Set();
                 } finally {
                     token.Reset();
```
MikhailArkhipov: So it try it does `token.Set();` and then immediate `Reset`?

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -396,16 +396,16 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
...
                 }
             }
 
             try {
-                if (_runTask.Status == TaskStatus.Running) {
```
MikhailArkhipov: I'd add AF for the awaitable state so if we do still hang, we'll know.

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: This operations are exclusive - each `token` can do either `Set` or `Reset` which immediately releases lock or passes it to the next awaiter. Any subsequent calls of `Set` or `Reset` have no effect.

src/Host/Client/Impl/Resources.resx
```diff
@@ -212,4 +212,8 @@ This should never happen with a production Remote R Services, so please check wi
 
 If you using a test Remote R Server with a self-signed certificate and are certain about the remote machine security, click OK, otherwise cancel the connection.</value>
   </data>
+  <data name="AbortRDataAutosave" xml:space="preserve">
+    <value>Previous R session terminated unexpectedly, and its global workspace is currently saving.
+Do you want to abort this operation and start current session immidiately?</value>
```
MikhailArkhipov: `currently being saved`. I think.

MikhailArkhipov:
👍

src/Host/Client/Impl/Host/BrokerClientProxy.cs
```diff
@@ -34,11 +34,10 @@ internal sealed class BrokerClientProxy : IBrokerClient {
 
         public Task PingAsync() => _broker.PingAsync();
 
-        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000,
-            CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
+        public async Task<RHost> ConnectAsync(string name, IRCallbacks callbacks, string rCommandLineArguments = null, int timeout = 3000, bool preserveSessionData = false, CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
```
AlexanderSher: Sure. Another thing that I'm considering is to use `AsyncLocal` instead of `ReentrancyToken`. That will simplify signature, the only problem is that `AsyncLocal` pollutes `ExecutionContext` which is kind of expensive operation (any write to the `AsyncLocal` creates a new copy of the dictionary that contains all values).

src/Host/Client/Impl/Host/RHost.cs
MikhailArkhipov: Sort of like `PulseEvent`?

src/Host/Client/Impl/Resources.resx
AlexanderSher: Ok

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: What do you want to test with `Assert.Fail`?

MikhailArkhipov: It hangs forever if state is `WaitingForActivation`

AlexanderSher: No, `WaitingForActivation` is a normal state for a task without action, which you get when you use `TaskCompletionSource` or get task from `async` method that contains custom awaiters. In this particular case, `_runTask` will be in `WaitingForActivation` state until `RunWorker` is completed, and then will be immediately switched to either `RanToCompletion` or  `Failed` or `Canceled`.

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: No. `BinaryAsyncLock` is kind of a mix of `ManualResetEvent`, `AutoResetEvent` and a boolean flag. If you call `Set`, all waiting threads are released and lock is set. It usually means that operation that has happened under lock is completed successfully and other threads don't need to repeat it. If you call `Reset`, only one waiting thread will be released and lock is not set. That means that old lock holder has failed to complete the operation, and the new lock holder need to make another attempt.
You can call `Set` or `Reset` multiple times, but only the first call matters. It is similar to `TaskCompletionSource.TrySetResult` and `TaskCompletionSource.TrySetCanceled` - you can't convert completed task into canceled and visa versa. To change `BinaryAsyncLock` state again, you either need to acquire another token or call `EnqueueReset`.

int19h:
👍

Issue #2499, opened 2016-10-13
"Connection test passed" label is too bright


int19h:
Bright green on white is not very readable. Probably needs to be dialed down to #008000 or something like that.

![image](https://cloud.githubusercontent.com/assets/1175142/19367485/446b852e-9150-11e6-9137-62d8fb4c9146.png)

MikhailArkhipov:
Need also test on dark theme

PR #2524, opened 2016-10-17
#2499 "Connection test passed" label is too bright

MikhailArkhipov:
Added color service. We may choose to put some common colors and brushes there from WpfOverrides and from editor standard classification as needed.


AlexanderSher:
VsWpfOverrides is WPF specific, and so the types that are used there. I'd like to avoid putting this information into commons, cause it is not just UI specific, but it is WPF-specific. 
For this particular bug, I'd rather add another brush to the `Microsoft.R.Wpf.Brushes` and set it in VsWpfOverrides. Colors and brushes aren't part of the specific ViewModel - they are part of the visual studio environment. So instead of putting theme-specific conditional logic into every view model, I'd prefer to have set of colors in one place.

MikhailArkhipov:
There is at least one place that needs to tell dark mode. In WpfOverrides we'll then have same colors. I don't want to wire them to editor colors as they can be changed by the user.

AlexanderSher:
AFAIK editor colors have nothing to do with VS themes: https://studiostyl.es/
So I don't want to put editor colors in VsWpfOverrides.

> There is at least one place that needs to tell dark mode.

Do you mean `HelpVisualComponent`?

MikhailArkhipov:
Yes. WpfOverrides will need that too I think, something like

```
            if (colorService.IsDarkTheme) {
                Brushes.SuccessMessageTextBrush = new SolidColorBrush(Color.FromArgb(0xFF, 0x49, 0x8B, 0x4E));
```

MikhailArkhipov:
BTW, we'll need colors that are not necessarily WPF specific. For REPL it was suggested to support [ANSI color escapes](http://ascii-table.com/ansi-escape-sequences.php).

AlexanderSher:
I thought REPL is using editor styling.

MikhailArkhipov:
It does, but request was to be able to specify color for the output in generic way, maybe even accessible from R code outputting to console.

AlexanderSher:
Well, that's a separate task. If we want to specify it from console, it would probably become part of the `IRSessionCallback`.
`VsWpfOverrides` needs information about theme, but it needs it only once. And CSS file is chosen only once. Both things can be set during package initialization and actually belong to the package layer.
We can rename `VsWpfOverrides` into `VsThemeOverrides` and add a string property that will have a path to that CSS file.

MikhailArkhipov:
That exposes a bit too much about inner implementation of the help component. It may or may not have CSS internally. BTW, it also might be useful to add `is high contrast`.

As for ANSI colors, we'll have to perform theme-based remapping since say, yellow, may work OK in dark theme but become barely visible in light. So code that specifies yellow color would run in console nicely even if user is in light or high contrast mode

AlexanderSher:
> That exposes a bit too much about inner implementation of the help component.

Agreed. But if we want an application level property, we already have `IApplicationConstants`.

As for the ANSI color remapping, it will be affected by the editor colors, and so may become quite complicated.

src/Common/Core/Impl/Shell/IColorService.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core.Shell {
+    public interface IColorService {
+        bool IsDarkTheme { get; }
```
AlexanderSher: I still think we should move it to the `IApplicationConstants` and have a property that returns enum. Something like `ApplicationTheme`.

MikhailArkhipov: Mkay

src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -72,7 +72,7 @@ public static class IOExtensions {
             var pathLength = GetTrailingSlashTrimmedPathLength(path);
             var otherPathLength = GetTrailingSlashTrimmedPathLength(otherPath);
 
-            return pathLength == otherPathLength && (pathLength == 0 || string.Compare(path, 0, otherPath, 0, otherPathLength, true, CultureInfo.InvariantCulture) == 0);
+            return pathLength == otherPathLength && (pathLength == 0 || string.Compare(path, 0, otherPath, 0, otherPathLength, StringComparison.OrdinalIgnoreCase) == 0);
```
AlexanderSher: `StringExtensions.EqualsIgnoreCase`?

AlexanderSher:
👍

PR #2529, opened 2016-10-18
Fix #2526: RSessionTest.StartStop fails

### LINKED ISSUE #2526: RSessionTest.StartStop fails
Looks like a product issue - test is quite simple

AlexanderSher:
- Added lock for StopHostAsync
- Added handing for session disposal to broker switching and reconnecting
- Added more delay time for tests that check concurrent access


int19h:
👍

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -262,34 +266,31 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
+        public IRSessionSwitchBrokerTransaction StartSwitchingBroker() => !_disposeToken.IsDisposed ? new BrokerTransaction(this) : null;
 
-        public IRSessionReconnectTransaction StartReconnecting() {
-            _disposeToken.ThrowIfDisposed();
-            return new BrokerTransaction(this);
-        }
+        public IRSessionReconnectTransaction StartReconnecting() => !_disposeToken.IsDisposed ? new BrokerTransaction(this) : null;
 
         public async Task StopHostAsync() {
             _disposeToken.ThrowIfDisposed();
```
MikhailArkhipov: As a matter of fact, it looks like any await can get the object disposed. Do we care?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -225,7 +234,7 @@ public class RSessionProvider : IRSessionProvider {
             var oldBroker = _brokerProxy.Set(brokerClient);
             try {
                 BrokerChanging?.Invoke(this, EventArgs.Empty);
-                await SwitchBrokerAsync(oldBroker, cancellationToken, lockToken.Reentrancy);
+                await SwitchBrokerAsync(cts.Token, lockToken.Reentrancy);
                 oldBroker.Dispose();
 
                 if (brokerClient.IsRemote) {
```
MikhailArkhipov: If (unlikely, but anyway) writing to console throws, we'll be setting back disposed oldBroker

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -262,34 +266,31 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
+                await StopHostAsync(BrokerClient, _startupInfo.Name, _host, _hostRunTask);
 
-            await StopHostAsync(BrokerClient, _startupInfo.Name, _host, _hostRunTask);
+                stopToken.Set();
+            } finally {
+                stopToken.Reset();
+            }
         }
 
         private static async Task StopHostAsync(IBrokerClient brokerClient, string hostName, RHost host, Task hostRunTask) {
```
MikhailArkhipov: This one can be called from 2 places without lock (`StopHostAsync` locks but `AfterHostStarted` - not sure)  - do we care?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -748,81 +754,90 @@ private class BrokerTransaction : IRSessionSwitchBrokerTransaction, IRSessionRec
             }
 
             public async Task CompleteSwitchingBrokerAsync(CancellationToken cancellationToken) {
+                if (_session._disposeToken.IsDisposed) {
```
MikhailArkhipov: Perhaps make a small method 'bool CheckPrerequisites()` that would check conditions like disposed, null `_lockToken` etc?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -748,81 +754,90 @@ private class BrokerTransaction : IRSessionSwitchBrokerTransaction, IRSessionRec
...
+                }
+
                 if (_lockToken == null) {
                     throw new InvalidOperationException($"{nameof(AcquireLockAsync)} must be called before {nameof(CompleteSwitchingBrokerAsync)}");
                 }
 
-                _session._disposeToken.ThrowIfDisposed();
-
-
                 if (_session._startupInfo == null) {
```
MikhailArkhipov: Could be in `CheckPrerequisites` as well

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -748,81 +754,90 @@ private class BrokerTransaction : IRSessionSwitchBrokerTransaction, IRSessionRec
...
                     return;
                 }
 
-                var host = _session._host;
-                // host may be null if previous attempts to start it have failed
-                if (host != null) {
-                    // Detach RHost from RSession
-                    host.DetachCallback();
+                try {
+                    var host = _session._host;
```
MikhailArkhipov: Fragment looks identical to one in the `CompleteSwitchingBrokerAsync`

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -182,6 +185,10 @@ public class RSessionProvider : IRSessionProvider {
         }
 
         public async Task<bool> TrySwitchBrokerAsync(string name, string path = null, CancellationToken cancellationToken = default(CancellationToken)) {
+            if (_disposeToken.IsDisposed) {
+                return false;
+            }
+
             await TaskUtilities.SwitchToBackgroundThread();
 
             var brokerClient = CreateBrokerClient(name, path);
```
MikhailArkhipov: `TrySwitchBrokerAsync` appears to be called from 2 places in the Connection Manager - technically we can end up with two concurrent brokerClient here

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -225,7 +234,7 @@ public class RSessionProvider : IRSessionProvider {
             var oldBroker = _brokerProxy.Set(brokerClient);
             try {
                 BrokerChanging?.Invoke(this, EventArgs.Empty);
-                await SwitchBrokerAsync(oldBroker, cancellationToken, lockToken.Reentrancy);
+                await SwitchBrokerAsync(cts.Token, lockToken.Reentrancy);
```
MikhailArkhipov: Looks like only BrokerClientProxy uses it

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -249,20 +259,18 @@ public class RSessionProvider : IRSessionProvider {
             return true;
         }
 
-        private async Task SwitchBrokerAsync(IBrokerClient oldBroker, CancellationToken cancellationToken, ReentrancyToken reentrancyToken) {
-            var sessions = _sessions.Values.ToList();
-            if (sessions.Any()) {
-                await SwitchSessionsAsync(sessions, cancellationToken, reentrancyToken);
+        private async Task SwitchBrokerAsync(CancellationToken cancellationToken, ReentrancyToken reentrancyToken) {
+            var transactions = _sessions.Values.Select(s => s.StartSwitchingBroker()).ExcludeDefault().ToList();
```
MikhailArkhipov: `IEnumerable<IRSession> GetSessions()` should have ToArray() since otherwise it is returning Values that can change.

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: That is a really good question what we have to do here. In a single-threaded environment, `ObjectDisposedException` should be thrown if someone tries to use an object that was already disposed.
But in our case, it is possible to call a method before object is disposed and it may become disposed while this method is awaiting a web response (so it isn't even in a stack trace).

I can see several options here:
- Add a lock that will make Dispose wait until some of the methods are completed. In this case, methods that are started will be able to complete. In this case, `ObjectDisposedException` will be thrown only if method is called after object is disposed, but no exception will happen if `Dispose` is called while method is running.
- Add a CancellationTokenSource that will request cancellation during dispose. In this case, `ObjectDisposedException` will be thrown only if method is called after object is disposed, but there will be `OperationCanceledException` if `Dispose` is called while method is running.
- Add more `ThrowIfDisposed` checks into methods. In this case, `ObjectDisposedException` can be thrown in any time and we need to handle it as a normal exception. Right now I'm thinking that it is the least preferable option.

@int19h , I'd like to hear your opinion as well.

---
In reply to: [83912025](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83912025) [](ancestors = 83912025)

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: `AfterHostStarted` also calls `StopHostAsync()`. This method is called without lock from `CompleteSwitchingBrokerAsync`, but it is called for the detached host, so we don't care.

---
In reply to: [83915213](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83915213) [](ancestors = 83915213)

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: yes, that's true.

---
In reply to: [83921016](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83921016) [](ancestors = 83921016)

src/Host/Client/Impl/Session/RSession.cs
MikhailArkhipov: Cancellation looks OK. However, how do we wait for the cancellation to actually complete and do we need to? We do have some code that, say, requests cancellation and then assumes it is done. But it may task some time to actually throw, right?

Cancellation before dispose also matches state machine - imagine requests come and get placed in a queue to get processed and there is a worker that processes them. So Start -> Dispose would be handled as `Start` -> `Cancel (if worker still running)` -> `Wait for Cancellation` -> `Dispose` or `Stop` (if already started) -> `Wait for Stopped` and then `Dispose`.

AlexanderSher: We can't guarantee that cancellation is completed, because we have no idea how many tasks are actually subscribed for that cancellation token and when do they check it. We also can't control the scheduler that may postpone one task in favor of another. But what we can be sure is that all tasks have received notification before `RequestCancellation` is completed, so the next time any code in the task will check for it, it will get an exception.

Ok then, I'll proceed with this option, and add this feature to `DisposableToken` and `DisposableBag`.

---
In reply to: [83955367](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83955367) [](ancestors = 83955367)

int19h: I think that `Dispose` as cancellation for any in-flight async operations makes sense. Making it wait on in-flight operations would be very surprising, since that's not normally how `Dispose` behaves; and some of those (e.g. evals) can be pretty long running, too, introducing undesired blocking.

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Done

---
In reply to: [83917898](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83917898) [](ancestors = 83917898)

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: In case of a switch to the background thread, if caller is on the main thread, continuation will be immediately executed in a thread pool. If it is on the background thread already, then no awaiting will happen at all.

It is different in case of UI thread though. We need to add `cancellationToken` to the `SwitchToMainThreadAsync`.

---
In reply to: [83911726](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83911726) [](ancestors = 83911726)

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: You mean these 2 lines:
```{csharp}
                        // Detach RHost from RSession
                        host.DetachCallback();

                        // Cancel all current requests (if any)
                        await _session.CancelAllAsync(cancellationToken);
```

---
In reply to: [83918298](https://github.com/Microsoft/RTVS/pull/2529#discussion_r83918298) [](ancestors = 83918298)

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Fixed

---
In reply to: [84113015](https://github.com/Microsoft/RTVS/pull/2529#discussion_r84113015) [](ancestors = 84113015,83911726)

MikhailArkhipov:
👍

PR #2542, opened 2016-10-19
#2532 and couple more fixes

### LINKED ISSUE #2532: Existence of Pandoc is checked locally
So in remote case if pandoc is missing *locally* we won't let user publish. On the other hand, if pandoc is missing on the server, user may get error messages in REPL.

MikhailArkhipov:
#2532 Existence of Pandoc is checked locally

- Fix endless requests to RData for non-existing functions (semi-hang debugger data tip)
- Fix issue when opening markdown without REPL up asserts b/c VsShell hasn't been initialized as R package is not loaded yet.

Improve remote host information printing:
- Move code to components
- Print after the R information is printed

### DIFF
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -85,9 +85,12 @@ namespace Microsoft.R.Host.Client.Host {
         }
 
         public async Task<AboutHost> GetHostInformationAsync(CancellationToken cancellationToken) {
-            var response = await HttpClient.GetAsync("/about", cancellationToken);
-            var s = await response.Content.ReadAsStringAsync();
-            return !string.IsNullOrEmpty(s) ? JsonConvert.DeserializeObject<AboutHost>(s) : AboutHost.Empty;
+            string result = null;
+            try {
+                var response = await HttpClient.GetAsync("/about", cancellationToken);
+                result = await response.Content.ReadAsStringAsync();
+            } catch (OperationCanceledException) { }
+            return !string.IsNullOrEmpty(result) ? JsonConvert.DeserializeObject<AboutHost>(result) : AboutHost.Empty;
         }
 
         public virtual async Task<RHost> ConnectAsync(BrokerConnectionInfo connectionInfo, CancellationToken cancellationToken = default(CancellationToken), ReentrancyToken reentrancyToken = default(ReentrancyToken)) {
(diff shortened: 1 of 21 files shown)


src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
```diff
@@ -27,9 +28,39 @@ public sealed class SessionInformationCommand : IAsyncCommand {
...
+
+        private void OnBrokerChanged(object sender, EventArgs e) {
+            if (_interactiveWorkflow.RSession.IsRemote) {
+                PrintBrokerInformationAsync(reportTelemetry: true).DoNotWait();
+            }
+        }
+
+        private async Task PrintBrokerInformationAsync(bool reportTelemetry = false) {
+            var a = await _interactiveWorkflow.RSessions.Broker.GetHostInformationAsync();
+            var console = _interactiveWorkflow.RSessions.Console;
```
AlexanderSher: On the level of `InteractiveWorkflow`, we can write directly to the interactive window - no need to reference it back from `RSession`

MikhailArkhipov: OK

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -297,3 +297,16 @@ enable_autosave <- function(delete_existing) {
...
     });
 }
+
+exists_on_path <- function(filename) {
+    folders <- strsplit(Sys.getenv('PATH'), ';')[[1]];
+    folders <- gsub("\\", "/", folders, fixed = TRUE);
+    folders <- sub("(\\/$)", "", folders);
+    folders <- paste0(folders, "/", filename);
+    for (f in folders) {
+        if (file.exists(f)) {
```
int19h: `file.exists` is vectorized, so you should be able to just do `any(file.exists(folders))`.

int19h:
:+1:

src/Host/Client/Impl/rtvs/R/util.R
MikhailArkhipov: I know but I didn't want to go through all paths and rather stop when found

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -87,6 +84,12 @@ internal abstract class BrokerClient : IBrokerClient {
             }
         }
 
+        public async Task<AboutHost> GetHostInformationAsync(CancellationToken cancellationToken) {
+            var response = await HttpClient.GetAsync("/about", cancellationToken);
```
AlexanderSher: Do we want to handle any exceptions here?

int19h: This method should probably be in a separate service class, actually.

MikhailArkhipov: Command that uses it will propagate exception to VS which typically handles and shows the error box.

MikhailArkhipov: But we can handle typical transport/canceled if you prefer

AlexanderSher:
👍

Issue #2477, opened 2016-10-10
Logging in monitor and user profile should be writing to system event log


karthiknadig:
Logging has common code between monitor, user profile and broker. Extract it out into a common library.

MikhailArkhipov:
Basic information, no need for it to be controlled by the VS/Broker options.

PR #2543, opened 2016-10-19
Logging from UserProfile 

karthiknadig:
Added event logging to UserProfile Service. Monitor will continue to log to a file, since it is a application that runs as a user. #2477

### DIFF
--- a/src/Host/UserProfile/Impl/ServiceLogger.cs
+++ b/src/Host/UserProfile/Impl/ServiceLogger.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using Microsoft.Common.Core.Disposables;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


src/Host/UserProfile/Impl/ServiceLogger.cs
```diff
@@ -28,23 +27,41 @@ internal sealed class ServiceLogger : ILogger, IDisposable {
...
-            }
-
-            string message = formatter(state, exception);
-            lock (writer) {
-                writer.WriteLine("[{0:u}] <{1}> ({2}):", DateTime.Now, _category, logLevel.ToString()[0]);
-                writer.WriteLine(message);
-
-                if (exception != null) {
-                    writer.WriteLine("Exception: " + exception);
+            lock (_logLock) {
```
karthiknadig: I preserved this lock from the writer. I don't think it is needed. Please confirm.

int19h: Since there are no shared objects here, it doesn't look like it's needed.

src/Host/UserProfile/Impl/ServiceLogger.cs
```diff
@@ -28,23 +27,41 @@ internal sealed class ServiceLogger : ILogger, IDisposable {
...
+                string message = formatter(state, exception);
+                using (EventLog eventLog = new EventLog("Application")) {
```
MikhailArkhipov: Would be nice to create our own node but I remember it required admin rights.

image

src/Host/UserProfile/Impl/ServiceLogger.cs
```diff
@@ -28,23 +27,41 @@ internal sealed class ServiceLogger : ILogger, IDisposable {
...
+                    EventLogEntryType logType = GetLogType(logLevel);
+                    eventLog.Source = "Application";
+                    string logMessage;
+                    if (exception != null) {
+                        logMessage = string.Format("[{0:u}] <{1}> ({2}), Message: {3}, Exception: {4}", DateTime.Now, _category, logLevel.ToString()[0], message, exception);
```
int19h: Since message and exception details are both likely to be long, and possibly multiline, I would suggest using newlines instead of commas here, and dropping "Message". So that the output would be something like this:
```
[10:11:12] <Component> (I):
Something bad happened

Exception: InvalidOperationException in .. .at ...
  at ...
  at ..
```

int19h:
:+1:

src/Host/UserProfile/Impl/ServiceLogger.cs
MikhailArkhipov: Should be formatting to current culture so time/date appear familiar to admin

src/Host/UserProfile/Impl/ServiceLogger.cs
```diff
@@ -28,23 +27,41 @@ internal sealed class ServiceLogger : ILogger, IDisposable {
...
+                    } else {
+                        logMessage = string.Format("[{0:u}] <{1}> ({2}), Message: {3}", DateTime.Now, _category, logLevel.ToString()[0], message);
```
MikhailArkhipov: Current culture

MikhailArkhipov:
👍

PR #2544, opened 2016-10-19
Broker on client windows allows only 1 user to create sessions.

### LINKED ISSUE #2330: Broker should not allow multiple clients on plain Windows
Windows Server should be required for shared environment.

karthiknadig:
Fixes #2330

### DIFF
--- a/src/Host/Broker/Impl/NativeMethods.cs
+++ b/src/Host/Broker/Impl/NativeMethods.cs
@@ -30,7 +30,49 @@ namespace Microsoft.R.Host.Broker {
         public const int CRED_MAX_CREDENTIAL_BLOB_SIZE = 512;
         public const int CREDUI_MAX_USERNAME_LENGTH = CRED_MAX_USERNAME_LENGTH;
         public const int CREDUI_MAX_PASSWORD_LENGTH = (CRED_MAX_CREDENTIAL_BLOB_SIZE / 2);
-        
+
+        // OS type
+        const int VER_NT_WORKSTATION = 0x0000001;
+        const int VER_NT_DOMAIN_CONTROLLER = 0x0000002;
+        const int VER_NT_SERVER = 0x0000003;
+
+        // Mask
+        const int VER_MINORVERSION = 0x0000001;
+        const int VER_MAJORVERSION = 0x0000002;
+        const int VER_BUILDNUMBER = 0x0000004;
+        const int VER_PLATFORMID = 0x0000008;
+        const int VER_SERVICEPACKMINOR = 0x0000010;
+        const int VER_SERVICEPACKMAJOR = 0x0000020;
+        const int VER_SUITENAME = 0x0000040;
+        const int VER_PRODUCT_TYPE = 0x0000080;
+
+        // conditions
+        const int VER_EQUAL = 1;
+        const int VER_GREATER = 2;
+        const int VER_GREATER_EQUAL = 3;
+        const int VER_LESS = 4;
+        const int VER_LESS_EQUAL = 5;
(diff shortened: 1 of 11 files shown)


src/Host/Broker/Impl/NativeMethods.cs
```diff
@@ -52,5 +94,19 @@ internal static unsafe class NativeMethods {
...
+        static extern ulong VerSetConditionMask(ulong dwlConditionMask, uint dwTypeBitMask, byte dwConditionMask);
+
+        [DllImport("kernel32.dll")]
+        static extern bool VerifyVersionInfo([In] ref OSVERSIONINFOEX lpVersionInfo, uint dwTypeMask, ulong dwlConditionMask);
+
+        public static bool IsWindowsServer() {
+            OSVERSIONINFOEX osVersionInfo = default(OSVERSIONINFOEX);
+            osVersionInfo.dwOSVersionInfoSize = (uint)Marshal.SizeOf(osVersionInfo);
+            osVersionInfo.wProductType = VER_NT_WORKSTATION;
+            ulong mask = VerSetConditionMask(0, VER_PRODUCT_TYPE, VER_EQUAL);
```
MikhailArkhipov: https://msdn.microsoft.com/en-us/library/bb773795(v=VS.85).aspx may be simpler to use

src/Host/Broker/Impl/Resources.resx
```diff
@@ -237,4 +237,7 @@
   <data name="Trace_UserProfileDirectory" xml:space="preserve">
     <value>User {0} profile directory: {1}</value>
   </data>
+  <data name="Exception_MaxAllowedUsers" xml:space="preserve">
+    <value>Broker has reached maximum allowed concurrent user limit.</value>
```
MikhailArkhipov: If this is user-visible I'd split it into 2 specific messages. `Windows Workstation only allows a singe user session' and 'R Remote Services computer has reached maximum number of concurrent users'

karthiknadig: Where should the second message come from? The first one comes from the session manager in broker.

MikhailArkhipov: I see that we throw ArgumentOutOfRange exception. If you try it, does it get propagated to the client via BrokerApiError? If so, then there should be error code that will map to the appropriate message in REPL. See `BrokerApiErrorException`

karthiknadig: I have made the change:
image

karthiknadig:
Ping.

src/Host/Broker/Impl/Resources.resx
```diff
@@ -237,4 +237,7 @@
   <data name="Trace_UserProfileDirectory" xml:space="preserve">
     <value>User {0} profile directory: {1}</value>
   </data>
+  <data name="Exception_MaxAllowedUsers" xml:space="preserve">
+    <value>Windows Workstation only allows a singe user session.</value>
```
huguesv: singe means monkey in french.

src/Host/Client/Impl/Resources.resx
```diff
@@ -188,4 +188,10 @@ If you using a test Remote R Server with a self-signed certificate and are certa
     <value>Previous R session terminated unexpectedly, and its global workspace is currently being saved.
 Do you want to abort this operation and start current session immidiately?</value>
   </data>
+  <data name="Error_MaxUsersAllowed" xml:space="preserve">
+    <value>Windows Workstation only allows a singe user session.</value>
```
huguesv: same typo

PR #2545, opened 2016-10-19
Fix #2535: Can't view list when inside a function and subsequent debugger hang

### LINKED ISSUE #2535: Can't view list when inside a function and subsequent debugger hang
```
z <- function(n) {
    x <- as.list(strsplit(Sys.getenv('PATH'), ';'));
    for (e in x) {
        y <- e;
    }
}
```

Set bp on 'for'
Attach debugger
Source
z("foo")

BP hit. In locals, expand `x`
Click on view button on the child entry

image

Bunch of invalid operation exceptions under debugger
 
image

Stop debugging. Note prompt is still `Browse[2]>`
Attach Debugger

'Debugger operation in progress...`

Try canceling. VS hangs.

int19h:
Provide an environment-independent expression to the grid viewer, such that it can be evaluated in any environment with the same result.

### DIFF
--- a/src/Debugger/Impl/AD7CustomViewer.cs
+++ b/src/Debugger/Impl/AD7CustomViewer.cs
@@ -19,7 +19,7 @@ namespace Microsoft.R.Debugger {
                 return VSConstants.E_FAIL;
             }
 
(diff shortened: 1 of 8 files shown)


MikhailArkhipov:
👍

Issue #2496, opened 2016-10-13
Automatically add schema and port when creating connection


int19h:
When adding a new connection in the Workspaces window, the schema is always going to be `https`, and the port is usually going to be the default 5444. We should allow users to type just the hostname in the textbox, and automatically add the schema and the port, probably in focus lost event (so as soon as user tabs away from the textbox or clicks on any button, it's added).

int19h:
It would also be nice to automatically fill the name textbox from host URL if it's empty, using hostname as the name.

MikhailArkhipov:
👍 

I think we should have name and port boxes, port pre-populated perhaps with 5444

AlexanderSher:
I'm not sure this is a good idea. I'm ok with adding https, but I don't think we should add a port - users may have any kind of public URL.

int19h:
The proposed scheme is as follows:

1. If user types what looks like a valid _absolute_ URL, we leave it alone. So if they do `http://foo`, we do *not* add the port, because the `http` schema implies port 80 per RFC.
2. If it's not a valid absolute URL, prepend `https://`, and see if that makes it a valid URL. If it's not a valid URL, leave the original input alone, but highlight it as error (the usual WPF decorator for those things).
3. If, after adding `https://`, it is a valid URL, but doesn't specify the port number, then set port to `5444`.

Some examples - user input on the left, post-validation textbox value on the right:

`https://foo:5444` -> `https://foo:5444` (no change)
`https://foo` -> `https://foo` (no change)
`http://foo` -> `http://foo` (no change)
`foo` -> `https://foo:5444`
`foo:5000` -> `https://foo:5000`
`username:password@foo` -> `https://username:password@foo:5444`

(Note that the latter case means that we shouldn't just be checking the URL for the colon to determine whether the port was specified or not, since colon is used in other cases, and doesn't necessarily indicate port - better to parse it after adding the schema, and check `Uri.IsDefaultPort`.)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ad355a956dcbea857f32db4fbd878884f3e44acf

Issue #2248, opened 2016-09-06
Parser should allow ellipsis past $


MikhailArkhipov:
RTVS currently treats ... as an invalid name for syntax highlighting:

![screen](https://cloud.githubusercontent.com/assets/12820357/18281766/fb897974-7412-11e6-9f16-fb6e4a279e81.png)

 But this is legal R code, for example I can run the above function like so:
```
> f(a, c=3)
    nargs missing.a missing.b    others 
        2         0         1         1
```

There are a few cases floating around of using match.case and ... like this, to manipulate function calls. You can see it in nnet::nnet.formula for instance, and I think it used to be present in stats::lm and stats::glm as well (to construct a call to model.frame from the supplied arguments).

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ad355a956dcbea857f32db4fbd878884f3e44acf

PR #2547, opened 2016-10-20
#2248, #2496

MikhailArkhipov:
#2496 Automatically add schema and port when creating connection + tests
#2248 Parser should allow ellipsis past $ + tests


src/R/Components/Impl/ConnectionManager/Implementation/Connection.cs
```diff
@@ -20,5 +30,43 @@ internal class Connection : ConnectionInfo, IConnection {
...
+
+        public override string UserProvidedPath {
+            get { return base.UserProvidedPath; }
+            set {
+                base.UserProvidedPath = value;
+                Path = ToCompletePath(value);
+            }
+        }
+
+        public static string ToCompletePath(string path) {
```
AlexanderSher: I think this should be part of `ConnectionViewModel`, and change should be visible to the user (that is at least what we have discussed online). I'd expect to change the URL when path input field loses focus.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionInfo.cs
```diff
@@ -6,15 +6,16 @@
 namespace Microsoft.R.Components.ConnectionManager.Implementation {
     public class ConnectionInfo : IConnectionInfo {
         public string Name { get; set; }
+        public virtual string UserProvidedPath { get; set; }
```
AlexanderSher: Why is it virtual?
And, btw, virtual properties are evil :)

src/R/Components/Impl/ConnectionManager/Implementation/Connection.cs
MikhailArkhipov: Originally I placed it there. However, it seemed more logical for it to be in connection considering persistence in settings (made old connections that only store path work fine).

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionInfo.cs
MikhailArkhipov: Connection overrides it. This one persists into settings, Connection also handles updating Path when UserProvidedPath is set. What's wrong with virtual properties?

src/R/Components/Impl/ConnectionManager/Implementation/Connection.cs
MikhailArkhipov: Generally speaking, Path should be {get;}. However, we have people with connections written already as 'Path'... This is why I don't like much json in settings.

MikhailArkhipov: Another possibility is to revert to 'Path' as 'user supplied' and add `GetConnectionUrl` method.

AlexanderSher: I don't think we should store this `short` path at all. It is just a shortcut that we provide during the input - similar to tab in VS code editor. Since the full path contains the whole short path, we don't limit user in editing it lately. But at least he will edit the real path that is used for connection.

MikhailArkhipov: Well user wants to see and edit what they have provided. I.e. if they said 'host' we should not be suddenly replacing it with 'https://host:5444'

MikhailArkhipov: Similar to, say, e-mail connections: user cares about e-mail and password but doesn't care about POP or IMAP details added internally.

MikhailArkhipov: Tooltip shows full URL btw. IMO in most cases user doesn't care about it - TMD.

MikhailArkhipov: I personally don't want to see 'auto-correction'. If I entered 'mymachine' this is what I want to see in UI. Internal details should be left to the code to handle. I don't want edit box to suddenly change to full URL when I click Save or restart VS.

int19h: The most obvious analogy here is the web browser address bar. If you type "github.com" there, it'll automatically append "https://", _and you will see the resulting URL_. I think this is a very similar case.

Also, in PTVS, we've been doing this for a very long time for remote debugging URLs - if you type "foo" as machine name in the Attach to Process dialog, it becomes "tcp://foo:5678" as soon as you tab away. I'm not aware of any complaints about this behavior.

int19h: The other option that I would be fine with, is treatment similar to how Chrome handles URLs, in that we do a _two-way_ conversion. In other words, we just say that the single canonical form for an URL with `https://` and default port number is to omit both. So if you type `foo`, we keep it as `foo` in UI, but treat it as `https://foo:5000` when connecting. But if you type `https://foo:5000`, we strip it down to just `foo`. For example, try typing "http://example.com:80" into Chrome address bar, and see what it does.

MikhailArkhipov: The latter is better IMO. Good example is mobile Safari. Type `CNN` in IE or Chrome and they take you to search. Ty CNN in Safari and it takes you there. It doesn't add `http` or anything - does add .com since you may want to be able to tell site.com from site.org.

Although I'd rather leave user data alone. Not everyone appreciate forced education ;-)

AlexanderSher: Ok, I'd like to get @jflam or @smortaz opinion on this.

MikhailArkhipov:
OK, made it UI only, correction when clicking Save

MikhailArkhipov:
Ping

AlexanderSher:
👍

Issue #2507, opened 2016-10-14
Remote: rtvs::: is not autocompleting


jflam:
When I want to fetch a remote file and I type "rtvs:::" I'm expecting to see command completion, but I don't. The command works fine, however and retrieves the file successfully.

rtvs:::fetch_file("foo.csv")

int19h:
It seems like there's some general problem with `::` and `:::` Intellisense. I don't get a list when I just type those in the REPL, and Ctrl+Space gives a list of everything available globally, rather than just package/namespace members.

MikhailArkhipov:
`utils::` and `RODBC::` work fine

![image](https://cloud.githubusercontent.com/assets/12820357/19448090/2378978c-9455-11e6-9724-e6c0d6388b8e.png)

![image](https://cloud.githubusercontent.com/assets/12820357/19448102/3650d1f8-9455-11e6-9963-c6be60196ad7.png)

It appears that 'rtvs' is not really a package since it does not appear in the list of loaded or installed packages and hence completion does not see it.

![image](https://cloud.githubusercontent.com/assets/12820357/19448162/76be32f8-9455-11e6-8f9d-8531b34bb478.png)

![image](https://cloud.githubusercontent.com/assets/12820357/19448216/bd9c3b84-9455-11e6-8ac6-7af2f130733e.png)

MikhailArkhipov:
It is 'loaded via a namespace (and not attached)`

MikhailArkhipov:
http://stackoverflow.com/questions/14988722/in-r-what-does-loaded-via-a-namespace-and-not-attached-mean

_It means the package (In this case R) can access the package functions/objects, **but the user can not** without explicitly loading the tools package_

int19h:
This is not correct - loaded but not attached simply means that it's not on the search path, and therefore its identifiers cannot be used without qualification. But `::` and `:::` should still work, because they work off the list of loaded namespaces, not off the search path. Note how they do work when you actually evaluate R code - you can actually execute `rtvs:::fetch_file`, it's only the code completion that's not working.

If our code completion uses `search()` for this, then we should fix it to use `loadedNamespaces()` instead to be consistent with R.

MikhailArkhipov:
It uses `_host.Session.InstalledPackagesAsync()` so currently only what appears in the package manager

int19h:
Ah, that makes sense. `rtvs` is not a properly installed package, and it won't show up there.

[...]

Issue #2500, opened 2016-10-13
REPL stuck spinning when trying to show help


int19h:
Connected to remote machine, did `?winDialog` in the REPL. The output in the REPL has "httpd server started", but the help window didn't show up, and REPL is stuck.

int19h:
Crash dump:

\\pytools\r\Bugs\2500

int19h:
Clicked "Reset" in REPL, it said "Stopping R Session", then after a few seconds I got this:

![image](https://cloud.githubusercontent.com/assets/1175142/19367928/527d9236-9152-11e6-8e6b-bbffd694c506.png)

int19h:
Clicked "Ignore" in the message box, the REPL prompt is now unblocked and I can enter expressions there, but trying to evaluate anything just gets it stuck again. I would expect it to print the usual "REPL is disconnected, click Reset" message.

MikhailArkhipov:
```
 	ntdll.dll!_ZwCreateFile@44() Line 777	Unknown
 	mswsock.dll!SockSocket(int AddressFamily, int SocketType, int Protocol, _GUID * pProviderId, unsigned int g, unsigned long dwFlags, unsigned long providerFlags, unsigned long serviceFlags1, unsigned long catId, _SOCKET_INFORMATION * * Socket) Line 1130	C
 	mswsock.dll!WSPSocket(int AddressFamily, int SocketType, int Protocol, _WSAPROTOCOL_INFOW * lpProtocolInfo, unsigned int g, unsigned long dwFlags, int * lpErrno) Line 286	C
 	ws2_32.dll!WSASocketW(int af, int type, int protocol, _WSAPROTOCOL_INFOW * lpProtocolInfo, unsigned int g, unsigned long dwFlags) Line 278	C++
 	ws2_32.dll!LookupAddressForName(const wchar_t * pNodeName, const wchar_t * pServiceName, int ProtocolFamily, int SocketType, int Protocol, unsigned short ServicePort, int Flags, int IpFlags, addrinfoW * * * ppPrev) Line 3701	C++
 	ws2_32.dll!GetAddrInfoW(const wchar_t * pNodeName, const wchar_t * pServiceName, const addrinfoW * pHints, addrinfoW * * ppResult) Line 4903	C++

...

 	System.dll!System.Net.Dns.TryGetAddrInfo(string name, System.Net.AddressInfoHints flags, out System.Net.IPHostEntry hostinfo) Line 1108	C#
[...]

PR #2557, opened 2016-10-24
#2500, #2507

MikhailArkhipov:
#2500 REPL stuck spinning when trying to show help
#2507 Remote: rtvs::: is not autocompleting

### DIFF
--- a/src/Host/Client/Impl/BrokerServices/WebServer.cs
+++ b/src/Host/Client/Impl/BrokerServices/WebServer.cs
@@ -11,41 +11,46 @@ using static System.FormattableString;
 
 namespace Microsoft.R.Host.Client.BrokerServices {
     public class WebServer {
+        private static Dictionary<int, WebServer> Servers { get; } = new Dictionary<int, WebServer>();
+
+        private readonly IRemoteUriWebService _remoteUriService;
+        private HttpListener _listener;
+
         public string LocalHost { get; }
-        public int LocalPort { get; }
+        public int LocalPort { get; private set; }
         public string RemoteHost { get; }
         public int RemotePort { get; }
 
-        private readonly HttpListener _listener;
-        private readonly IRemoteUriWebService _remoteUriService;
-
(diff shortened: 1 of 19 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -687,10 +687,12 @@ internal sealed class RSession : IRSession, IRCallbacks {
         /// <param name="url"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        Task IRCallbacks.WebBrowser(string url, CancellationToken cancellationToken) {
-            var newUrl = BrokerClient.HandleUrl(url, cancellationToken);
+        async Task IRCallbacks.WebBrowser(string url, CancellationToken cancellationToken) {
+            var newUrl = await BrokerClient.HandleUrlAsync(url, cancellationToken);
             var callback = _callback;
-            return callback != null ? callback.ShowHelpAsync(newUrl, cancellationToken) : Task.CompletedTask;
+            if (callback != null) {
```
AlexanderSher: We can check `_callback` before calling `HandleUrlAsync`

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -11,41 +11,51 @@
...
-                _listener.Prefixes.Add(Invariant($"http://{LocalHost}:{LocalPort}/"));
-                try {
-                    _listener.Start();
-                } catch (HttpListenerException) {
-                    _listener.Close();
-                    continue;
+        }
+
+        public Task InitializeAsync(CancellationToken ct) {
+            return Task.Run(() => {
```
AlexanderSher: Can you explain this change?

MikhailArkhipov: #2500. Original code did the loop in sync without option to cancel. From the dump it appears loop was stuck in HttpListener.AddPrefix.

AlexanderSher:
👍

Issue #2538, opened 2016-10-19
dev.new(), New Plot Window no longer activates the new device


huguesv:
This used to work consistently, but now it's only working sometimes.

Example:
```
plot(1)
dev.new()
plot(2)
```

Sometimes the second plot goes to the same window as the first, instead of going to the new window (which does come up, but isn't activated properly).

huguesv:
Updated title to indicate that New Plot Window command is also affected.

huguesv:
This was caused by a resize event which changed the active device as a side effect.

PR #2559, opened 2016-10-24
Several plot operations no longer change the active graphics device.

huguesv:
Had to adjust the exporting code, as it was relying on `selectplot` to change the active device, which it no longer does.
Add some tests to ensure active plot device doesn't get changed when it shouldn't.

Fixes https://github.com/Microsoft/RTVS/issues/2538
See also https://github.com/Microsoft/R-Host/pull/98


AlexanderSher:
👍

int19h:
👍

PR #2562, opened 2016-10-25
Remote files completion

MikhailArkhipov:


src/Common/Core/Impl/Collections/EnumerableExtensions.cs
```diff
@@ -113,5 +113,11 @@ public static class EnumerableExtensions {
                 }
             }
         }
+
+        public static void ForEach<T>(this IEnumerable<T> enumerable, Action<T> action) {
```
AlexanderSher: https://blogs.msdn.microsoft.com/ericlippert/2009/05/18/foreach-vs-foreach/
TL;DR: there should be no LINQ methods that cause only side effects

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -64,33 +69,78 @@ public class FilesCompletionProvider : IRCompletionListProvider {
...
         #endregion
 
+        private Task<List<RCompletion>> GetRemoteDirectoryItemsAsync(string directory) {
+            return Task.Run(async () => {
+                var session = _workflow.RSession;
+                var completions = new List<RCompletion>();
+
+                try {
+                    var rPath = directory.ToRPath();
+                    var files = await session.EvaluateAsync<JArray>(Invariant($"as.list(list.files(path = '{rPath}', include.dirs = FALSE))"), REvaluationKind.Normal);
```
huguesv: $"as.list(list.files(path = {rPath.ToRStringLiteral()}, include.dirs = FALSE))"

otherwise this won't work if path has a single quote char in it.

same below

src/Common/Core/Impl/Collections/EnumerableExtensions.cs
MikhailArkhipov: Meh. Eric was always opinionated. There is one on `List<T>` and we are using it. And there used to be one when Linq was an extension.

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
MikhailArkhipov: OK

src/R/Editor/Test/Completions/FileCompletionProviderTest.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        [Test]
+        public void LocalFiles() {
+            var workflow = Substitute.For<IRInteractiveWorkflow>();
+
+            var provider = new FilesCompletionProvider(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), workflow, _imagesProvider, _glyphService);
+            var entries = provider.GetEntries(null);
+            entries.Should().NotBeEmpty();
+            entries.Should().Contain(e => e.DisplayText == "MyLibrary");
```
huguesv: I don't see a "MyLibrary" under My Documents folder.
Maybe you should create a known directory structure under the test output folder, and look for what you've created.

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -64,33 +67,78 @@ public class FilesCompletionProvider : IRCompletionListProvider {
...
+        private Task<List<RCompletion>> GetRemoteDirectoryItemsAsync(string directory) {
+            return Task.Run(async () => {
+                var session = _workflow.RSession;
+                var completions = new List<RCompletion>();
+
+                try {
+                    var rPath = directory.ToRPath().ToRStringLiteral();
+                    var files = await session.EvaluateAsync<JArray>(Invariant($"as.list(list.files(path = '{rPath}', include.dirs = FALSE))"), REvaluationKind.Normal);
+                    var dirs = await session.EvaluateAsync<JArray>(Invariant($"as.list(list.dirs(path = '{rPath}', full.names = FALSE, recursive = FALSE))"), REvaluationKind.Normal);
+
```
huguesv: Now you need to remove the single-quote from the C# string:
$"as.list(list.files(path = {rPath}, include.dirs = FALSE))"

MikhailArkhipov: `ToRStringLiteral(quote: '\'')` since we do need it quoted right?

MikhailArkhipov: Default is double

src/R/Editor/Test/Completions/FileCompletionProviderTest.cs
MikhailArkhipov: Good point. Tried to pick something standard but I guess yes, need to pre-created.

src/Common/Core/Impl/Collections/EnumerableExtensions.cs
AlexanderSher: `List<T>` is a fixed-length read-write collection. Side effects for it is normal.
`IEnumerable<T>` is a state machine, not a collection of elements. 

`IEnumerable<T>.ForEach` instead of `foreach` sometimes means that LINQ sequence can be simplified.

For example:
```{csharp}
var list = new List<string>();
sequence1.ForEach(e => list.Add(e));
sequence2.ForEach(e => list.Add(e));
return list;
```

While it can be just:
```{csharp}
return sequence1.Concat(sequence2).ToList();
```

If `foreach` is really required in place, you can always call `sequence.AsList().ForEach()`

MikhailArkhipov: `AsList` allocates more memory. All the argument is really about readability and as Eric said 'philosophical objections'

AlexanderSher: But it isn't needed in your changeset, where can use either `AddRange` or `ToList` depending on the preferred style, but either of these options gives you better code.

Many of the design patterns are 'philosophical objections', yet I don't want to add an extension that doesn't give any advantage so far.

MikhailArkhipov: My preferred style is ForEach, but generally IDC

huguesv:
It's making me nervous that the test is writing and deleting from my documents folder, but other than that 👍

PR #2581, opened 2016-10-26
 Fuzztests for User Profile service namedpipe pipe

karthiknadig:
1. Added tests for the user profile service named pipe.
2. Added timed cancellations to service pipe.


src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,135 @@
...
+        // empty username domain password
+        [InlineData("{\"Username\": \"\", \"Domain\": \"\", \"Password\": \"\"}", "", "", "", true, false, false)]
+        // whitespace input string
+        [InlineData("                     ", null, null, null, true, false, false)]
+        public async Task CreateProfileTest(string input, string username, string domain, string password, bool isValidParse, bool isValidAccount, bool isExistingAccount) {
+            var creator = UserProfileCreatorMock.Create(username, domain, password, isValidParse, isValidAccount, isExistingAccount);
+            await CreateProfileTestRunnerAsync(creator, input, isValidParse, isValidAccount, isExistingAccount);
+        }
+
+        [Test]
```
int19h: We should probably have a distinct category for fuzz tests. Take a look at how other test classes use test category attributes.

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,135 @@
...
+        public async Task CreateProfileFuzzTest() {
+            string inner = "\"Username\": {0}, \"Domain\": {1}, \"Password\":{2}";
+            List<string> testResults = new List<string>();
```
int19h: Unused.

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,135 @@
...
+                    await CreateProfileTestRunnerAsync(creator, json, false, false, false, true);
+                } catch (IOException) {
+                    // expect pipe to fail. The client side pipe throws an IOException when the server side pipe 
+                    // closes due to IO error or attempt to access unauthorized memory.
+                }
+            }
+        }
+
+        private byte[] GenerateBytes() {
+            Random rd = new Random((int)DateTime.Now.Ticks);
```
int19h: You can just use the default constructor for `Random` - it automatically uses ticks or something similar.

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,135 @@
...
+        // Missing closing paren
+        [InlineData("{", null, null, null, false, false, false)]
+        // empty json object
+        [InlineData("{}", null, null, null, false, false, false)]
+        // No username, domain and password
+        [InlineData("{\"Username\":, \"Domain\":, \"Password\":}", null, null, null, false, false, false)]
+        // empty username domain password
+        [InlineData("{\"Username\": \"\", \"Domain\": \"\", \"Password\": \"\"}", "", "", "", true, false, false)]
+        // whitespace input string
+        [InlineData("                     ", null, null, null, true, false, false)]
```
int19h: Can you also add an explicit test case for zero-length input?

karthiknadig:
Ping

src/Common/Core/Impl/OS/IUserCredentials.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core.OS {
+    public interface IUserCredentials {
+        string Username { get; set; }
+        string Password { get; set; }
```
MikhailArkhipov: Scares me a bit. SecureString?

src/Host/Protocol/Impl/RUserProfileCreatorImpl.cs
```diff
@@ -0,0 +1,57 @@
...
+                } else {
+                    logger?.LogError(Resources.Error_UserProfileDirectoryWasNotFound, credentails.Domain, credentails.Username, Marshal.GetLastWin32Error());
+                    result = RUserProfileCreateResponse.Create((uint)Marshal.GetLastWin32Error(), profileExists, profileDir.ToString());
+                }
+            } else {
+                logger?.LogError(Resources.Error_UserLogonFailed, credentails.Domain, credentails.Username, Marshal.GetLastWin32Error());
+                result = RUserProfileCreateResponse.Create((uint)Marshal.GetLastWin32Error(), false, null);
+            }
+
+            if (token != IntPtr.Zero) {
```
MikhailArkhipov: Probably should be in try-finally

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,140 @@
...
+                    if (isValidParse) {
+                        result.Error.Should().Be((uint)(isValidAccount ? 0 : 13));
+                        result.ProfileExists.Should().Be(isExistingAccount);
+                    } else {
+                        result.Should().BeNull();
+                    }
+                }
+            }
+
+            testDone.Wait();
```
MikhailArkhipov: We have async tests, they use regular awaits and IAsyncLifetime. Otherwise if test hangs, it may block all subsequent tests

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,140 @@
...
+                byte[] responseRaw = new byte[1024];
+                var bytesRead = await client.ReadAsync(responseRaw, 0, responseRaw.Length, ct);
+                jsonResp = Encoding.Unicode.GetString(responseRaw, 0, bytesRead);
+            }
+            return JsonConvert.DeserializeObject<UserProfileResultMock>(jsonResp);
+        }
+
+        private async Task CreateProfileTestRunnerAsync(IUserProfileServices creator, string input, bool isValidParse, bool isValidAccount, bool isExistingAccount, int serverTimeOut, int clientTimeOut, bool isFuzzTest = false) {
+            ManualResetEventSlim testDone = new ManualResetEventSlim(false);
+            Task.Run(async () => {
```
MikhailArkhipov: I'd split method into fuzz part and regular part

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,140 @@
...
+                            f.ShouldNotThrow();
+                        } else {
+                            Func<Task> f = () => RUserProfileCreator.CreateProfileAsync(serverTimeOutms: serverTimeOut, clientTimeOutms: clientTimeOut, userProfileService: creator);
+                            await f.ShouldThrowAsync<Exception>();
+                        }
+                    }
+                } finally {
+                    testDone.Set();
+                }
+            }).DoNotWait();
```
MikhailArkhipov: Why we can't await it?

src/Host/Protocol/Test/Microsoft.R.Host.Protocol.Test.csproj
```diff
@@ -0,0 +1,91 @@
...
+  <ItemGroup>
+  </ItemGroup>
+  <ItemGroup>
+    <FilesToSign Include="$(OutputPath)\$(AssemblyName).dll">
+      <Authenticode>Microsoft</Authenticode>
+      <StrongName>StrongName</StrongName>
+    </FilesToSign>
+  </ItemGroup>
+  <Import Project="$(VSToolsPath)\TeamTest\Microsoft.TestTools.targets" Condition="Exists('$(VSToolsPath)\TeamTest\Microsoft.TestTools.targets')" />
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
MikhailArkhipov: MicroBuild targets and props

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
```diff
@@ -0,0 +1,140 @@
...
+        [InlineData("                     ", null, null, null, true, false, false)]
+        // empty string
+        [InlineData("", null, null, null, false, false, false)]
+        public async Task CreateProfileTest(string input, string username, string domain, string password, bool isValidParse, bool isValidAccount, bool isExistingAccount) {
+            var creator = UserProfileCreatorMock.Create(username, domain, password, isValidParse, isValidAccount, isExistingAccount);
+            await CreateProfileTestRunnerAsync(creator, input, isValidParse, isValidAccount, isExistingAccount, 500, 500);
+        }
+
+        [Test]
+        [Category.FuzzTest]
```
int19h: I suggested having a generic category for fuzz stuff, so that they could be excluded from test runs altogether (to get deterministic runs if we need to).

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
karthiknadig: See explanation below for this. we can put a time out here. Note, if this test ends without the server disconnecting, the next test will certainly fail. This wait is needed to ensure all threads are done.

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
karthiknadig: added

src/Host/Protocol/Test/Microsoft.R.Host.Protocol.Test.csproj
karthiknadig: I don't see that in other test projects. Should it be there?

src/Host/Protocol/Impl/RUserProfileCreator.cs
```diff
@@ -83,7 +83,7 @@ public class RUserProfileCreator {
 
             string json = Encoding.Unicode.GetString(requestRaw, 0, bytesRead);
 
-            var requestData = JsonConvert.DeserializeObject<RUserProfileCreateRequest>(json);
+            var requestData = JsonConvert.DeserializeObject<RUserProfileCreateRequest>(json, new SecureStringJsonConverter());
```
int19h: You can also stick `JsonConverterAttribute` onto the property in the class. Then you won't have to repeat the converter on every call to `Serialize` or `Deserialize`.

src/Host/Protocol/Test/UserProfileServicePipe/UserProfileServicePipeTest.cs
MikhailArkhipov: OK

src/Host/Protocol/Test/Microsoft.R.Host.Protocol.Test.csproj
MikhailArkhipov: Hm interesting. OK, `"MicroBuild.Core": "0.2.0",` in project.json I guess would be sufficient.

int19h:
👍

PR #2582, opened 2016-10-26
GetLibraryPathAsync updated to use remote environment to determine lib path for user

karthiknadig:


huguesv:
Do PackageManagerIntegrationTests still pass?

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -171,6 +171,11 @@ public static class RSessionEvaluationCommands {
             return evaluation.EvaluateAsync<JArray>(script, REvaluationKind.Normal);
         }
 
+        public static Task<JArray> UserLibraryPathsAsync(this IRExpressionEvaluator evaluation) {
+            var script = @"rtvs:::packages.userlibpaths()";
+            return evaluation.EvaluateAsync<JArray>(script, REvaluationKind.Normal);
```
int19h: I believe you can just do `EvaluateAsync<string[]>`, so long as all values in the array are actually guaranteed to be strings (or null).

int19h: ... which the code at the call site assumes to be the case, since it uses a cast that will fail if something is not actually a string.

int19h:
👍

huguesv:
👍

PR #2584, opened 2016-10-26
Fix #2572: User count returned when connecting to remote is incorrect.

### LINKED ISSUE #2572: User count returned when connecting to remote is incorrect.
It looks like the user count is derived from the number of keys in the users-to-session dictionary in broker. The dictionary clean-up after removing the last session for a user must also remove the key. Currently it leaves the key with a zero session list.

int19h:
Clean up user entries for users who have no remaining sessions.

Allow no SSL for remote in debug builds only, for ease of testing.


MikhailArkhipov:
👍

Issue #2503, opened 2016-10-14
Rmd code chunk incorrectly marked as syntax error


Hong-Revo:
If you have `{r eval=FALSE}`, with no comma after the `r`, it gets the squiggly red lines:
![image](https://cloud.githubusercontent.com/assets/11328890/19399273/7c3b9b18-929c-11e6-935a-13efb65283c8.png)

This is [valid syntax](https://www.rstudio.com/wp-content/uploads/2015/02/rmarkdown-cheatsheet.pdf) though. Using `{r, eval=FALSE}` passes the syntax checker.

MikhailArkhipov:
Side effect of the fact that the entire `{r expression}` is parsed as R (matches highlight). So what parser sees is `curly` - `identifier: r` - `identifier` - `operator =` and this sequence is missing operator between identifiers. R is embedded and does not know it is inside markdown hence the squiggly.

MikhailArkhipov:
Relates to #2377, might as well just fix both

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f981c81aef6dcd32be84d6f372e40e69670feeb1

Issue #2377, opened 2016-09-21
RMD occasionally loses colorization


MikhailArkhipov:
Probably projections tracking issue

```
I just wanted to ask if there is a way to get correct R syntax highlighting in R-chunks inside of Rmarkdown files.  When I open my RMarkdown files in Visual Studio 2015 upd3, using RTVS 0.4 I get only singe colored R-chunks with occasional “errors” due to apostrofs in the comments. See the screenshots below:
```

![rmd](https://cloud.githubusercontent.com/assets/12820357/18719221/51b17c9c-7fdb-11e6-89ca-dddb64a2a943.png)

MikhailArkhipov:
[TEST.Rmd.txt](https://github.com/Microsoft/RTVS/files/485491/TEST.Rmd.txt)

MikhailArkhipov:
Reproes when selecting both regions in RMD and then pasting RMD text that containes another region with missing starting triple-tick

MikhailArkhipov:
In this particular case the problem is in the `date` line. the closing `"` is lost so the remaining R code becomes string until the next double quote.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f981c81aef6dcd32be84d6f372e40e69670feeb1

PR #2585, opened 2016-10-27
#2377 #2503

MikhailArkhipov:
#2377 RMD occasionally loses colorization
#2503 Rmd code chunk incorrectly marked as syntax error 

Two fixes:
a. Properly recognize length of leading and trailing separator lengths in R blocks rather than assuming 1/3
b. Use contained language host (markdown) so R parser can skip over `R` in `{r x = y, ...}` rather than trying to add it as an identifier. Contained language host is the way embedded language communicates with the primary so primary can control some of the functionality such as indent level, formatting etc.


MikhailArkhipov:
Will merge for the BB

PR #2587, opened 2016-10-27
Build 0.5 for Dev15

MikhailArkhipov:
Changes:
- Interactive window now references Shell.15.0 thus many projects got project.15.0.json
- Checked in separate 14.0 and 15.0 interactive window assemblies
- VSIX manifest is different in Dev15 hence source.manifest.14/15.0.vsixmanifest and copy pre-build step
- Referencing Shell.15.0 prevents using PlatformUI in XAML (which was a bad idea anyway) so added intermediate PlatformDialogWindow and styling via WPF Overrides
- More WPF overrides for the grid and data viewer
- Updated 15.0 references to the latest and greatest RC
- Unzipped templates (SWIX prefers that)
- Changed Add New Item code to work with unzipped templates
- Changed 14.0 setup authoring to install unzipped templates

Separate RTVS-pr is coming soon


MikhailArkhipov:
Need to merge since we need build ASAP

PR #2588, opened 2016-10-27
Updating license and adding 3rd party notices text.

jflam:
Adding THIRDPARTYNOTICES.TXT which contains the license text for all 3rd Party OSS used in RTVS


int19h:
`libzip` and other stuff that's in R-Host should go into that repo - we already have a bunch of notices there, directly in LICENSE:

https://github.com/Microsoft/R-Host/blob/master/LICENSE

int19h:
@jflam, can you split the list?

These should be in R-Host repo:
* Boost: Boost Software License 1.0 [1]
* picojson: Simplified BSD License [2]
* libzip: New BSD License [3]
* zlib: zLib License [4]
* minhook: Simplified BSD License [5]

These should be in RTVS repo:
* libuv: MIT License [6]
* Newtonsoft.Json: MIT License [7]
* ASP.NET Core: MIT License [8]

All the build-time stuff should also be in RTVS repo.

Also, this list seems to be missing R itself?

MikhailArkhipov:
This PR is months old. Reopen if needed still.

PR #2591, opened 2016-10-28
Fix #2570: AsyncReaderWriterLock hangs on requesting WriterLock after cancellation

### LINKED ISSUE #2570: AsyncReaderWriterLock hangs on requesting WriterLock after cancellation
```{csharp}
            var cts = new CancellationTokenSource();

            var task1 = _arwl.WriterLockAsync(CancellationToken.None);
            var task2 = _arwl.WriterLockAsync(cts.Token);

            cts.Cancel();
            task1.Result.Dispose();
            var task3 = _arwl.WriterLockAsync(CancellationToken.None);

            task1.Should().BeRanToCompletion();
            task2.Should().BeCanceled();
            task3.Should().BeRanToCompletion();
```

AlexanderSher:
- Separated lock items queue from the rest of the logic
- Some new tests added

### DIFF
--- a/src/Common/Core/Impl/Microsoft.Common.Core.csproj
+++ b/src/Common/Core/Impl/Microsoft.Common.Core.csproj
@@ -144,6 +144,7 @@
     <Compile Include="Telemetry\TelemetryArea.cs" />
     <Compile Include="Telemetry\TelemetryServiceBase.cs" />
     <Compile Include="Threading\BinaryAsyncLock.cs" />
+    <Compile Include="Threading\AsyncReaderWriterLock.Queue.cs" />
     <Compile Include="Threading\DelayedAsyncAction.cs" />
     <Compile Include="Threading\IAsyncReaderWriterLockToken.cs" />
     <Compile Include="Threading\IMainThread.cs" />
--- a/src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
+++ b/src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
@@ -1,0 +1,239 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading;
+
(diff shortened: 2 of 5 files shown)


MikhailArkhipov:
Can you elaborate what is getting broken and what is queue is for? Synchronization is getting complicated... Perhaps there are simpler approaches? 

Maybe make a single thread or a task and put everything for establishing connection there? Reentrancy is not allowed while thread is active - user has to cancel first. New connections cannot be established until thread is canceled. connection info then would be a single structure or class and new one only returned when connection thread completes successfully (and connection data exchange happens in the main thread)? This is, for example, how background parsing and tree exchange is performed - and same for the async task list population/update. Events are only fired when update is done and only from UI thread.

AlexanderSher:
> Can you elaborate what is getting broken and what is queue is for?

Queue is for better debugging. Without it its quite hard to find out if lock is handing and why, cause there is no explicit list of pending tasks. Broken part was a missing case that is described in bug: if the last writer request was canceled and other writers are released, new writer request hanged.

> Maybe make a single thread or a task and put everything for establishing connection there? 

We can't stick to a single thread while we have awaiters. We can create a wrapping task for the switch, but that is pretty much what `await` does for us. Switch is already executed as one chunk and no other code can start a session or initiate another switch. But connection can die by itself when user clicked reconnect, host can die during switching, user can cancel switch while one of the sessions is killing its host, etc.

>  Reentrancy is not allowed while thread is active
Reentrancy is required because we need to distinguish somehow connections initiated by switch and connections initiated by any other code, and they should be exclusive - no other code can connect during switch and visa versa. Simple semaphore can't be used because switch needs all sessions to be connected to the new broker under one lock.

AlexanderSher:
#2597

MikhailArkhipov:
Thread can have awaitable statement inside it. Is something else can initiate connection switch except the user? Anyways - just from the type of issues we've been seeing it looks like there are points where state machine may end up in a transition state and it doesn't know how to get out. So point is to try and make activities run sequentially until either fail or succeed limiting user opportunities to intervene.

MikhailArkhipov:
👍

MikhailArkhipov:
#2570

int19h:
👍

PR #2605, opened 2016-10-31
Fix #2597: Simplify synchronization logic of session switching and initialization.

### LINKED ISSUE #2597: Simplify synchronization logic of session switching and initialization
Right now RSession deals with 3 different locks which cause different types of synchronization issues.
This needs to be simplified.

AlexanderSher:
3 locks are merged into one. Now switching and session initialization is controlled by the same lock. The only independent lock that is left in RSession is `_stopHostLock` which prevents `StopHostAsync` from running twice and used only in this method.


MikhailArkhipov:
👍

Issue #2601, opened 2016-10-28
Sort out settings load on package initialize


MikhailArkhipov:
VS does't like that

```
LegacySitePackage failed for package [RPackage]Source: 'Microsoft.VisualStudio.Shell.15.0'
Description: Object reference not set to an instance of an object.
System.NullReferenceException: Object reference not set to an instance of an object.
at Microsoft.VisualStudio.Shell.DialogPage.SetPropertyValue(PropertyDescriptor descriptor, Object automationObject, Object value)
at Microsoft.VisualStudio.Shell.DialogPage.LoadSettingFromStorage(PropertyDescriptor prop)
at Microsoft.VisualStudio.Shell.DialogPage.LoadSettingsFromStorage()
at Microsoft.VisualStudio.R.Package.Options.R.RToolsOptionsPage.LoadSettingsFromStorage()
at Microsoft.VisualStudio.R.Package.Options.R.RToolsOptionsPage.LoadSettings()
at Microsoft.VisualStudio.R.Packages.R.RPackage.Initialize()
at Microsoft.VisualStudio.Shell.Package.Microsoft.VisualStudio.Shell.Interop.IVsPackage.SetSite(IServiceProvider sp)
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0c596e001c0d28b61a1be02008ee7d6deb01d21e

Issue #2594, opened 2016-10-28
Occasionally there are multiple popups for certificate validation


MikhailArkhipov:
Unclear what is the exact repro, probably async/timing issue.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0c596e001c0d28b61a1be02008ee7d6deb01d21e

Issue #2556, opened 2016-10-24
Occasional ObjectDisposedException on web socket


MikhailArkhipov:
So far I have seen it in tests, but we may need to look if this is a product issue.

![image](https://cloud.githubusercontent.com/assets/12820357/19656752/05f81d64-99d6-11e6-96c2-4e7137ca3549.png)

MikhailArkhipov:
`DisconnectAsync` throws

```
>	Microsoft.R.Host.Client.dll!Microsoft.R.Host.Client.RHost.Dispose() Line 60	C#
 	Microsoft.R.Host.Client.dll!Microsoft.R.Host.Client.Session.RSession.Dispose() Line 114	C#
 	Microsoft.R.Host.Client.dll!Microsoft.R.Host.Client.Session.RSessionProvider.Dispose() Line 95	C#
 	Microsoft.R.Components.Test.dll!Microsoft.R.Components.Test.Fakes.InteractiveWindow.TestRInteractiveWorkflowProvider.DisposeInstance(Microsoft.R.Host.Client.IRSessionProvider sessionProvider) Line 98	C#
```

AlexanderSher:
Do you have it with the latest bits?

MikhailArkhipov:
It appears to be better now. However - I think the socket can get disposed elsewhere in the network stack. Perhaps we should still protect against it.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0c596e001c0d28b61a1be02008ee7d6deb01d21e

PR #2608, opened 2016-10-31
Bug fixes

### LINKED ISSUE #2549: Canceling untrusted certificate connection throw unhandled exception
```
An unhandled exception of type 'System.Net.Http.HttpRequestException' occurred in mscorlib.dll
Additional information: An error occurred while sending the request.
```

### LINKED ISSUE #2607: Remove setting storage from package property page
We should be using `IWritableSettingsStorage` and only use Tools | Options page for editing

MikhailArkhipov:
#2549 Canceling untrusted certificate connection throw unhandled exception 
 -  `HttpClient.GetAsync("/about", ...)` was not handling HTTP exceptions.

#2594 Occasionally there are multiple popups for certificate validation
 - remember outcome for certificate for the lifetime of the remote broker client

#2556 Occasional ObjectDisposedException on web socket
 - handle disposed exception on the socket since it can be disposed elsewhere

#2601 Sort out settings load on package initialize
- move setting loading off the package initialization path (we need to fix #2607 for real thing)


huguesv:
Please check the host submodule, it looks like you may be reverting to an older commit there.

MikhailArkhipov:
Sure. I thought I merged, but will double check

MikhailArkhipov:
I don't see that PR is updating the submodule

huguesv:
It looks different than it used to:
image

MikhailArkhipov:
Looks like just because my R-Host was on 0.5, master decided that it likes 0.5 and not R-Host master

src/Package/Impl/Packages/R/RPackage.cs
```diff
@@ -112,23 +112,27 @@ internal class RPackage : BasePackage<RLanguageService>, IRPackage {
             VsWpfOverrides.Apply();
             CranMirrorList.Download();
 
+            base.Initialize();
+
+            ProjectIconProvider.LoadProjectImages();
+            LogCleanup.DeleteLogsAsync(DiagnosticLogs.DaysToRetain);
+
+            IdleTimeAction.Create(CompleteInit, 20, this.GetType(), VsAppShell.Current);
```
AlexanderSher: CPS requests package load:
```{csharp}
            await _threadHandling.SwitchToUIThread();
            // Make sure R package is loaded
            VsAppShell.EnsurePackageLoaded(RGuidList.RPackageGuid);

            // Verify project is not on a network share and give warning if it is
            CheckRemoteDrive(_projectDirectory);

            _workflow = _workflowProvider.GetOrCreate();
```
Does it still work with this change?

MikhailArkhipov: The package will be loaded since initialization will be complete. But it depends what exactly project uses from the package. Appears to work in real life, but that's just one machine. I assume we are talking project reopen with workspace auto-load?

MikhailArkhipov: (1) is basically add 

RPackage.LoadSettings

and call it from the project adter loading the package

MikhailArkhipov:
Added IRPackage.LoadSettings and modified EnsurePackageLoaded

AlexanderSher:
Ok, let's try it
👍

PR #2614, opened 2016-11-01
Several fixes

AlexanderSher:
- Fix DisposableBag to throw ObjectDisposedException
- Fix TaskUtilities.WhenAllCancelOnFailure for the case when cancellationToken is canceled already
- Changed some of the unit-tests to throw exception instead of hanging


MikhailArkhipov:
👍

PR #2617, opened 2016-11-01
Avoid using DialogPage.Save/LoadSettings

MikhailArkhipov:
For 0.5 changes are not complete. I'd like to clean up names, add tests to the settings persistence as well as remove `Browsable(false)` attributes. This is to be done in master. For Dev15 we only want to avoid using dialog page for storage/loading. Codee may still call SaveSettings on the dialog page, but implementation is via VS settings manager.

#2618


src/Package/Impl/Options/R/RToolsOptionsPage.cs
```diff
@@ -291,18 +294,37 @@ public class RToolsOptionsPage : DialogPage {
...
             base.OnApply(e);
         }
+
+        private void PersistSettings() {
+            SaveSettingsToStorage();
+            _storage.Persist();
+        }
+
+        protected override void LoadSettingFromStorage(PropertyDescriptor prop) {
+            if (_storage.SettingExists(prop.Name)) {
```
AlexanderSher: So we intentionally don't call base method here?

src/Package/Impl/Shell/VsSettingsStorage.cs
```diff
@@ -0,0 +1,141 @@
...
+    [Export(typeof(ISettingsStorage))]
+    internal sealed class VsSettingsStorage : ISettingsStorage {
+        private const string _collectionPath = "R Tools";
+        private readonly Dictionary<string, object> _settings = new Dictionary<string, object>();
+        private readonly ShellSettingsManager _shellSettingsManager;
+        private readonly WritableSettingsStore _store;
+        private readonly object _lock = new object();
+
+        public VsSettingsStorage() {
+            _shellSettingsManager = new ShellSettingsManager(RPackage.Current);
```
AlexanderSher: `ShellSettingsManager` - is it visual studio type?

src/Package/Impl/Options/R/RToolsOptionsPage.cs
MikhailArkhipov: Yes. I don't want to mess with this code too in 0.5 - it already handles property names. What really should be done is to remove save/load calls from here, have own table of property names for each setting (or use approach similar to `BindableBase.SetProperty` with caller name attribute) and only go via setting storage.  So `RPackage.LoadSettings` won't be calling `RToolsOptionsPage.LoadSettings`

src/Package/Impl/Shell/VsSettingsStorage.cs
```diff
@@ -0,0 +1,141 @@
...
+            lock (_lock) {
+                _settings[name] = value;
+            }
+        }
+
+        public void Persist() {
+            lock (_lock) {
+                foreach (var s in _settings) {
+                    var t = s.Value.GetType();
+                    if (typeof(bool).IsAssignableFrom(t)) {
```
AlexanderSher: `s.Value is bool`?

src/Package/Impl/Shell/VsSettingsStorage.cs
```diff
@@ -0,0 +1,141 @@
...
+            Debug.Assert(_store != null);
+
+            if (!_store.CollectionExists(_collectionPath)) {
+                _store.CreateCollection(_collectionPath);
+            }
+        }
+
+        public bool SettingExists(string name) {
+            lock (_lock) {
+                if (_settings.ContainsKey(name)) {
```
AlexanderSher: Maybe just add a comment?

src/Package/Impl/Shell/VsSettingsStorage.cs
AlexanderSher: Ok

src/Package/Impl/Shell/VsSettingsStorage.cs
MikhailArkhipov: sure

src/Package/Impl/Shell/VsSettingsStorage.cs
```diff
@@ -67,21 +85,21 @@ internal sealed class VsSettingsStorage : ISettingsStorage {
...
 
         public void Persist() {
             lock (_lock) {
-                foreach (var s in _settings) {
+                foreach (var s in _settingsCache) {
                     var t = s.Value.GetType();
-                    if (typeof(bool).IsAssignableFrom(t)) {
+                    if (s.Value is bool) {
                         _store.SetBoolean(_collectionPath, s.Key, (bool)s.Value);
                     } else if (typeof(int).IsAssignableFrom(t) || t.IsEnum) {
```
AlexanderSher: `s.Value is int || s.Value is Enum`

AlexanderSher:
👍

src/Package/Impl/Shell/VsSettingsStorage.cs
AlexanderSher: `typeof(int).IsAssignableFrom(typeof(uint))` and `typeof(uint).IsAssignableFrom(typeof(int))` both return `false`.

MikhailArkhipov: I guess since compiler doesnt like such assignments either. :-) Will add separate clause in master.

PR #2620, opened 2016-11-02
Fix #2079: Memory leak: Every interaction or evaluation registers itself for host cancellation token and never unregisters

### LINKED ISSUE #2079: Memory leak: Every interaction or evaluation registers itself for host cancellation token and never unregisters
As a result, all evaluation and interaction objects are preserved until RHost is disposed

AlexanderSher:


src/Common/Core/Impl/Tasks/CancellationTokenUtilities.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Common.Core.Tasks {
+    public static class CancellationTokenUtilities {
+        public static void UnregisterOnCompletion(this CancellationTokenRegistration registration, Task task) {
+            task.ContinueWith(UnregisterCancellationToken, registration, default(CancellationToken), TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+        }
+
+        private static void UnregisterCancellationToken(Task task, object state) {
+            ((CancellationTokenRegistration)state).Dispose();
```
MikhailArkhipov: If we cast it anyway, why is it object?

AlexanderSher: `ContinueWith` accepts only `Action<Task,object>`

AlexanderSher:
@karthiknadig , I've made some changes in Host.Protocol code, can you please take a look?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -559,41 +559,42 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
-            while (!ct.IsCancellationRequested) {
-                try {
-                    if (await EvaluateAll(contexts, mutated, hostCancellationToken)) {
-                        // EvaluateAll has raised the event already, so reset the flag.
-                        mutated = false;
-                    } else if (mutated) {
-                        // EvaluateAll did not raise the event, but we have a pending mutate to inform about.
-                        OnMutated();
-                    }
+                        if (ct.IsCancellationRequested) {
```
MikhailArkhipov:
```
if(!ct.IsCancellationRequested) {
    var evaluationSource = await _pendingEvaluationSources.ReceiveAsync(ct);
    ...
}
```
and avoid extra returns

src/Host/Client/Impl/Session/RSessionEvaluation.cs
```diff
@@ -27,6 +28,7 @@ internal sealed class RSessionEvaluation : IRSessionEvaluation {
 
         public void Dispose() {
             _hostCancellationRegistration.Dispose();
+            _clientCancellationRegistration.Dispose();
```
MikhailArkhipov: Since they don't appear to be readonly, is there chance they may be null (i.e. ctor called but code that sets them got never executed)?

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Sure

src/Host/Client/Impl/Session/RSessionRequestSource.cs
```diff
@@ -11,21 +11,23 @@ namespace Microsoft.R.Host.Client.Session {
     internal sealed class RSessionRequestSource {
         private readonly TaskCompletionSourceEx<IRSessionInteraction> _createRequestTcs;
         private readonly TaskCompletionSourceEx<object> _responseTcs;
+        private CancellationTokenRegistration _cancellationTokenRegistration;
```
MikhailArkhipov: Appear to be readonly

src/Common/Core/Impl/Tasks/CancellationTokenUtilities.cs
MikhailArkhipov: OK

src/Host/Client/Impl/Session/RSessionEvaluation.cs
MikhailArkhipov: disposable structs huh. OK.

karthiknadig:
Protocol changes look good.

---
In reply to: [257934093](https://github.com/Microsoft/RTVS/pull/2620#issuecomment-257934093) [](ancestors = 257934093)

MikhailArkhipov:
👍

Issue #2579, opened 2016-10-26
Activate plot command rename


huguesv:
Activate Plot command should be Show Plot (so it doesn't imply that it activates the device, which it won't do anymore https://github.com/Microsoft/RTVS/pull/2559)

Issue #2540, opened 2016-10-19
Access to plot windows via main menu


huguesv:
Right now there's the legacy R Tools -> Windows -> Plots menu item, which is non-deterministic now that we have multiple plot windows.

Since there's a shortcut associated with it, it probably should be changed to go to the active plot window, instead of whichever random one it goes to now.

We should have a submenu with a dynamic list of all plot windows, including ones that VS may not have created yet.  Those that don't have a graphics device associated with them can have a fixed name for example "Unassigned", the ones that are associated will have "Device 3" for example.

PR #2629, opened 2016-11-03
Plot commands

huguesv:
Fix https://github.com/Microsoft/RTVS/issues/2540 Access to plot windows via main menu
Fix https://github.com/Microsoft/RTVS/issues/2579 Activate plot command rename

This adds menu items to navigate to any visible plot window, or any plot window with a graphics device (visible or not).

Also adds a command to hide all plot windows.

The command R Tools->Windows->Plots (CTRL-6) now behaves more predictably. It goes to the plot window with the active graphics device, instead of seemingly random plot window. If there isn't any active device, it opens the plot window which will have the active device with the next plot operation.


src/Package/Impl/Commands/R/ShowMainPlotWindowCommand.cs
```diff
@@ -0,0 +1,26 @@
...
+            _plotManager = workflow.Plots;
+        }
+
+        protected override void SetStatus() {
+            Enabled = true;
+        }
+
+        protected override void Handle() {
+            var component = _plotManager.GetOrCreateMainPlotVisualComponent();
+            component?.Container.Show(true, false);
```
AlexanderSher: Is it possible that `component == null`?

huguesv: If visual component for active device is not found, which would be a bug.  I debated whether or not to use a ? there.  I can remove it.

huguesv: I just realized I can move this command to Components project, since it no longer has VS dependencies, so I'll do that as well.

huguesv:
Any other comments? If not I'll get this in for tonight's build so we can try it out at BB tomorrow.

int19h:
👍

PR #2650, opened 2016-11-05
Two bug fixes + several test fixes

### LINKED ISSUE #2430: Restore connection if the last connection was local
So we match 0.5 and earlier

### LINKED ISSUE #2645: VS stuck while connecting to remote
Karthik says all my sessions are live, but VS is with the connecting popup.

Dump on pytools R bugs share folder 2645. Build is 2016-11-03.1.

AlexanderSher:
- Fix #2645: VS stuck while connecting to remote
- Fix #2430: Restore connection if the last connection was local
- Increased delays for some of the tests
- Split RHostPipeFuzzTest into iterations + IntRangeAttribute
- Added DisposableBag wrapper around CompositionContained in the TestExportProvider
- Added GetExportedValue overload that supports metadata filtering
- Removed some dead code
- Fix several other tests

### DIFF
--- a/src/R/Editor/Application.Test/Formatting/AutoFormatTest.cs
+++ b/src/R/Editor/Application.Test/Formatting/AutoFormatTest.cs
@@ -17,13 +17,17 @@ namespace Microsoft.R.Editor.Application.Test.Formatting {
     public class AutoFormatTest : IDisposable {
         private readonly IExportProvider _exportProvider;
         private readonly EditorHostMethodFixture _editorHost;
+        private readonly bool _autoFormat;
 
         public AutoFormatTest(REditorApplicationMefCatalogFixture catalogFixture, EditorHostMethodFixture editorHost) {
             _exportProvider = catalogFixture.CreateExportProvider();
             _editorHost = editorHost;
+            _autoFormat = REditorSettings.AutoFormat;
+            REditorSettings.AutoFormat = true;
         }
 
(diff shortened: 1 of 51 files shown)


src/R/Editor/Application.Test/Formatting/AutoFormatTest.cs
```diff
@@ -17,13 +17,17 @@ namespace Microsoft.R.Editor.Application.Test.Formatting {
     public class AutoFormatTest : IDisposable {
         private readonly IExportProvider _exportProvider;
         private readonly EditorHostMethodFixture _editorHost;
+        private readonly bool _autoFormat;
 
         public AutoFormatTest(REditorApplicationMefCatalogFixture catalogFixture, EditorHostMethodFixture editorHost) {
             _exportProvider = catalogFixture.CreateExportProvider();
             _editorHost = editorHost;
+            _autoFormat = REditorSettings.AutoFormat;
```
MikhailArkhipov: As long they are not parallel

MikhailArkhipov:
👍

src/R/Editor/Application.Test/Formatting/AutoFormatTest.cs
AlexanderSher: Sure, all tests for the code that rely on `REditorSettings` static properties should be non-parallel.

Issue #2553, opened 2016-10-21
Enable local documentation viewing from R Tools/Documentation menu items


jflam:
Today, we point exclusively to the online version of the documentation in these menus:

![image](https://cloud.githubusercontent.com/assets/8877/19605399/a9d7acb6-976e-11e6-92ab-5b4c115d62ab.png)

We have feedback from users that at a minimum, we need to have an option where the user can pick whether they want the local copies of the documentation or the online version of the documentation. The default for those commands should point to the local copies of the documentation.

If we don't implement the option, we should make those commands point to the local copies of the documentation.

PR #2652, opened 2016-11-07
Prefer opening local documentation for active connection if available

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/2553

Having an option for it seem overkill.

This will open the documentation from local disk when there's an active local connection, otherwise it will go online.

It would be possible to check all connections if there is no active connection, but that feels like it would be unpredictable behavior.


int19h:
👍

src/R/Components/Impl/Documentation/Commands/OpenDocumentationCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+        public CommandStatus Status {
+            get {
+                return CommandStatus.Supported | CommandStatus.Enabled;
+            }
+        }
+
+        public Task<CommandResult> InvokeAsync() {
+            var url = _onlineUrl;
+
+            if (!string.IsNullOrEmpty(_localRelativePath)) {
```
MikhailArkhipov: Technically local and remote can be different versions. Do we care?

huguesv: Exactly. Online version should be good enough for this relatively less common case.

Issue #2641, opened 2016-11-04
Indentation messed up for incomplete expressions


revodavid:
(See also #2577)

In an R script window, type:

```
x <- 1 +
```

and press ENTER. The cursor appears at the beginning of the next line. Indentation should be inserted. Even if you manually add indentation, RTVS insists on REMOVING indentation and eturning the cursor to the beginning of the next line.  Type ENTER `3` at the end of:

```
X <- 1 +
  2
```

and you get

```
x <- 1 +
2 +
3
```

RStudio does this correctly: typing `x <- 1 +` ENTER `2` leaves the follwing in the editor:

```
X <- 1 +
  2
```

There are many complex expressions in R that are NOT function blocks that require indentation to be readable, and RTVS messes up the indentation. Some examples:

magrittr pipes
```
Teams %>%
 filter(teamID %in% c("CHN", "CLE") & yearID > 1900) %>%
 mutate(Team = fct_drop(fct_recode(teamID, Cubs = "CHN", Indians = "CLE"))) %>%
 mutate(RunsPerGame = R / G) %>%
 mutate(HitsPerGame = HA / G) -> CubsInd.team
```

ggplot layers
```
p <- ggplot(data = CubsInd.team, aes(yearID,RunsPerGame)) +
    geom_point(aes(color = Team)) +
    geom_smooth(aes(color = Team), method = "loess") +
    xlab("Year") +
    ylab("Runs per game") +
    ggtitle("Runs per Game")
```

MikhailArkhipov:
This basically amounts to 

a). Providing proper smart indent in incomplete expressions.
b). Respecting user indentation inside multiline expressions.

Implemented (a). (b) actually exists but it looks like there is a but that causes #2469 and perhaps some other similar cases.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5e1a474e3a2e53241a2add081f0508da16b29ec1

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/698b187013402962ac7ff35ccdf9ae4d321a1869

PR #2653, opened 2016-11-07
#2641

MikhailArkhipov:
#2641 Indentation messed up for incomplete expressions 

Parse expression line we do in REPL in order to determine if it is complete. Based on the completeness drive smart indent and user indentation preservation.


src/R/Core/Impl/AST/Extensions/AstRootExtensions.cs
```diff
@@ -38,5 +39,18 @@ public static class AstRootExtensions {
             }
             return null;
         }
+
+        public static bool IsCompleteExpression(this AstRoot expressionAst) {
+            foreach (var error in expressionAst.Errors) {
+                if (error.ErrorType == ParseErrorType.CloseCurlyBraceExpected ||
+                    error.ErrorType == ParseErrorType.CloseBraceExpected ||
+                    error.ErrorType == ParseErrorType.CloseSquareBracketExpected ||
+                    error.ErrorType == ParseErrorType.FunctionBodyExpected ||
```
int19h: It seems that this should apply to all composite statements that have a mandatory body - e.g. `if`, `while` etc. Do we do that already? If so, this enum member should probably be renamed to `BodyExpected`.

MikhailArkhipov: This is code we used to use in REPL to show continuation. But no, it does not currently detect missing body on scope-less 'if' or similar.

MikhailArkhipov: #2655

src/R/Core/Impl/Formatting/RFormatter.cs
```diff
@@ -651,7 +653,9 @@ public class RFormatter {
             int end = _tokens.IsEndOfStream() ? _textProvider.Length : _tokens.CurrentToken.Start;
 
             string text = _textProvider.GetText(TextRange.FromBounds(start, end));
-            if (!preserveUserIndent && string.IsNullOrWhiteSpace(text)) {
+            bool lineBreakBeforeToken = _tokens.IsLineBreakAfter(_textProvider, _tokens.Position - 1);
+
+            if (lineBreakBeforeToken && !preserveUserIndent && string.IsNullOrWhiteSpace(text)) {
```
MikhailArkhipov: Yep

src/R/Core/Impl/Formatting/FormattingScope.cs
```diff
@@ -17,7 +17,8 @@ internal sealed class FormattingScope : IDisposable {
         private readonly TextBuilder _tb;
         private readonly int _previousIndentLevel;
 
-        public int CloseBracePosition { get; private set; } = -1;
+        public RToken OpenBraceToken { get; set; }
+        public int CloseBracePosition { get; } = -1;
```
MikhailArkhipov: It is set in ctor so I guess initial value is not needed

AlexanderSher:
👍

Issue #2636, opened 2016-11-03
Saved remote connection are not actually saved


MikhailArkhipov:
Create remote connection
Shut down VS
[I uninstalled R here but it prob doesnt matter]
Open VS

- No remote connections in the list

AlexanderSher:
Happens because settings get restored after `ConnectionManager` is instantiated

MikhailArkhipov:
This will be fixed by my upcoming PR for #2589

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0e32d5ffd6bfb6ac69a669b8238846cb564f77e3

MikhailArkhipov:
Actually still broken

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/1ee7574d76997eff0abad7bde52f4b2f3d1c6605

PR #2657, opened 2016-11-07
#2636 Saved remote connection are not actually saved

MikhailArkhipov:


src/Package/Impl/Packages/R/RPackage.cs
```diff
@@ -110,6 +111,9 @@ internal class RPackage : BasePackage<RLanguageService>, IRPackage {
                 return;
             }
 
+            var settings = VsAppShell.Current.ExportProvider.GetExportedValue<IRToolsSettings>() as IRPersistentSettings;
```
AlexanderSher: Can't we just call `GetExportedValue<IRPersistentSettings>()`?

MikhailArkhipov: If we want to export persistent, sure. I wanted all the code to deal with save/load randomly.

PR #2658, opened 2016-11-08
3 bug fixes

### LINKED ISSUE #2639: Terminating R and then using REPL produces incorrect error
Terminate R in remote (Session -> Terminate R)
Try 1+1

image

I am actually still connected. It is that host is down.

### LINKED ISSUE #2644: Package manager RHost session should use the name PackageManager
Currently it uses "IsolatedRHost" + session.ID

### LINKED ISSUE #2656: Reset doesn't work after RHost process is terminated
- Connect to local broker, open Interactive Window
- Terminate RHost.exe
- Click "Reset"

**Actual** RSession isn't started
**Expected** RSession is started

AlexanderSher:
- Fix #2639: Terminating R and then using REPL produces incorrect error
- Fix #2656: Reset doesn't work after RHost process is terminated
- Fix #2644: Package manager RHost session should use the name PackageManager

### DIFF
--- a/src/Host/Client/Impl/Resources.Designer.cs
+++ b/src/Host/Client/Impl/Resources.Designer.cs
@@ -241,8 +241,7 @@ namespace Microsoft.R.Host.Client {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Interactive Window is disconnected from R Session.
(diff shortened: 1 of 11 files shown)


MikhailArkhipov:
👍

PR #2662, opened 2016-11-08
Fix regression introduced by 9fca44e (Fix #2644: Package manager RHost session should use the name PackageManager)

### LINKED ISSUE #2644: Package manager RHost session should use the name PackageManager

AlexanderSher:
plus test


AlexanderSher:
@karthiknadig , can you please take a look?

karthiknadig:
:shipit:

PR #2663, opened 2016-11-08
Fuzz Tests for RTVS side message protocol

### LINKED ISSUE #2660: Fuzz Test RTVS side message protocol

karthiknadig:
Fixes #2660

### DIFF
--- a/src/Host/Protocol/Impl/Message.cs
+++ b/src/Host/Protocol/Impl/Message.cs
@@ -39,7 +39,7 @@ namespace Microsoft.R.Host.Protocol {
 
         private Message() { }
 
-        public static Message Create(byte[] data) {
+        public static Message Parse(byte[] data) {
             Message message = new Message();
             try {
                 int offset = 0;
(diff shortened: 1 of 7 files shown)


AlexanderSher:
👍

src/Host/Protocol/Impl/Message.cs
```diff
@@ -36,37 +37,42 @@ public class Message {
             Blob = blob ?? new byte[0];
         }
 
-        public Message(byte[] data) {
+        private Message() { }
+
+        public static Message Create(byte[] data) {
```
int19h: What is the benefit of having a factory method here rather than constructor?

karthiknadig: It was just a style choice. I did not want to put parsing code in the constructor.

int19h: That's a good point. But let's call it `Parse` then.

Issue #2659, opened 2016-11-08
Smart indent is set incorrectly when line ends with space


MikhailArkhipov:
... or otherwise being formatted. Try

```
function() {
    field <- |
        x
}
```
end with
```
function() {
    field <-
|
        x
}
```

works correctly when there is no space after <-. Most probably regression since when we started formatting after ENTER rather than before. Formatting probably re-positions caret improperly.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/698b187013402962ac7ff35ccdf9ae4d321a1869

Issue #2596, opened 2016-10-28
First time intellisense is slow


MikhailArkhipov:
in REPL specifically

MikhailArkhipov:
```
>	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.Providers.PackageFunctionCompletionProvider.GetPackageByName(string packageName) Line 174	C#
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.Providers.PackageFunctionCompletionProvider.GetAllFilePackages.AnonymousMethod__16_0(string packageName) Line 165	C#
 	[External Code]	
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.Providers.PackageFunctionCompletionProvider.GetAllFilePackages(Microsoft.R.Editor.Completion.RCompletionContext context) Line 164	C#
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.Providers.PackageFunctionCompletionProvider.GetPackages(Microsoft.R.Editor.Completion.RCompletionContext context) Line 104	C#
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.Providers.PackageFunctionCompletionProvider.GetEntries(Microsoft.R.Editor.Completion.RCompletionContext context) Line 62	C#
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.RCompletionSource.PopulateCompletionList(int position, Microsoft.VisualStudio.Language.Intellisense.ICompletionSession session, System.Collections.Generic.IList<Microsoft.VisualStudio.Language.Intellisense.CompletionSet> completionSets, Microsoft.R.Core.AST.AstRoot ast) Line 86	C#
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.RCompletionSource.AugmentCompletionSession(Microsoft.VisualStudio.Language.Intellisense.ICompletionSession session, System.Collections.Generic.IList<Microsoft.VisualStudio.Language.Intellisense.CompletionSet> completionSets) Line 57	C#
 	[External Code]	
 	Microsoft.Languages.Editor.dll!Microsoft.Languages.Editor.Completion.CompletionController.TriggerCompletion() Line 400	C#
 	Microsoft.Languages.Editor.dll!Microsoft.Languages.Editor.Completion.CompletionController.ShowCompletion(bool autoShownCompletion) Line 384	C#
[...]

PR #2677, opened 2016-11-09
Formatting and intellisense fixes

### LINKED ISSUE #2467: Formatting: if-expression on newline after <- is indented wrong
Original code:

``` r
function() {
    field <-
        if (missing(fields)) {
            function(x) TRUE;
        } else {
            function(x) x %in% fields
        }
}
```

After "Format Document":

``` r
function() {
    field <-
        if (missing(fields)) {
        function(x) TRUE;
    } else {
        function(x) x %in% fields
    }
}
```

MikhailArkhipov:
#2467 Formatting: if-expression on newline after <- is indented wrong 
#2641 Indentation messed up for incomplete expressions 
#2659 Smart indent is set incorrectly when line ends with space
#2596 First time intellisense is slow

- Added analysis of incomplete expressions to formatter and smart indenter to help with indent calculation in multiline expressions.
- Track indent level based on the line that contains keyword that defines scope. Align content and the closing brace accordingly.
- Move some of the formatter code to helper classes
- Move package index building to broker connected since there is no R when package starts

- Derive interactive window mock from ToolPane (otherwise code throws in tests)
- Fix completion tests that now need explicit info fetching for function description


src/Languages/Core/Impl/Formatting/IndentBuilder.cs
```diff
@@ -233,5 +233,17 @@ public IndentBuilder(IndentType indentType, int indentSize, int tabSize)
 
             return indent;
         }
+
+        /// <summary>
+        /// Determines indentation based on the leading whitespace in the current line.
+        /// </summary>
+        public static int GetLineIndentSize(TextBuilder tb, int position, int tabSize) {
```
int19h: This should probably check `position` for validity, and assert or throw if it's 0.

int19h: But you subtract 1 from it below. What happens if the result is -1, and that is passed to the indexer?

MikhailArkhipov: `i >= 0` should bypass the loop?

int19h: Ahh, subtle.

src/R/Editor/Impl/Completion/Providers/PackageFunctionCompletionProvider.cs
```diff
@@ -153,19 +156,23 @@ public class PackageFunctionCompletionProvider : IRCompletionListProvider, IRHel
...
 
             IEnumerable<string> loadedPackages = _loadedPackagesProvider?.GetPackageNames() ?? Enumerable.Empty<string>();
             IEnumerable<string> filePackageNames = context.AstRoot.GetFilePackageNames();
             IEnumerable<string> allPackageNames = PackageIndex.PreloadedPackages.Union(filePackageNames).Union(loadedPackages);
 
-            return allPackageNames
-                .Select(packageName => GetPackageByName(packageName))
+            var list = new List<IPackageInfo>();
+            foreach (var name in allPackageNames) {
+                var info = await _packageIndex.GetPackageInfoAsync(name);
```
int19h: Does `GetPackageInfoAsync` do any R evals or anything else that requires a roundtrip to the host? If so, we might want to make it a batch API, so that it can do it for all these packages in one go, to avoid unnecessary latency on every request.

int19h: Can we then at least fire off all the async requests at once, and `WaitAll` a combination of them, so that throughput is maximized? It seems like in the worst case, this will produce a latency that is a multiple of the underling network latency by the number of packages fetched - meaning seconds of time, which would be very noticeable.

MikhailArkhipov:
Made package fetching in batch. Generally it shouldn't be necessary since index is built and missing package means it was freshly installed or user mistyped its name so it is rarely more than one.

And fixed tests since some need function index pre-built and some were building it but didn't actually use it.

int19h:
👍

PR #2678, opened 2016-11-09
Two bug fixes

### LINKED ISSUE #2665: Opening REPL with fresh install gives confusing message
Open VS (no R windows)
Open REPL:

image

we should tell what to do or just pick something

### LINKED ISSUE #2673: IsolatedSession causes RHost leaks and creating RHost processes that never die
`RSessionProvider.BeginEvaluationAsync` takes a lot of time, on a remote connection it is much longer than fetching packages data.
We should assign a session to package manager and use it instead of temporary sessions.

AlexanderSher:
- Fix #2665: Opening REPL with fresh install gives confusing message
- Fix #2673: IsolatedSession causes RHost leaks and creating RHost processes that never die


MikhailArkhipov:
👍

PR #2686, opened 2016-11-10
Partial improvement to Reset REPL.

karthiknadig:
This change reduces the chances of doing Forced RHost Terminations. Previously we ended up doing forced termination almost every time. This added to the overall delay while doing reset.

### DIFF
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -317,7 +317,8 @@ namespace Microsoft.R.Host.Client.Session {
             // Try graceful shutdown with q() first.
             if (host != null) {
                 try {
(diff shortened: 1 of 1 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -317,7 +317,8 @@ internal sealed class RSession : IRSession, IRCallbacks {
             // Try graceful shutdown with q() first.
             if (host != null) {
                 try {
-                    await Task.WhenAny(hostRunTask, host.QuitAsync(), Task.Delay(500)).Unwrap();
+                    host.QuitAsync().DoNotWait();
```
int19h: Need to silence exceptions on this. I think it's safe to silence everything here.

int19h:
👍

PR #2689, opened 2016-11-10
Fix NRE on markdown open

MikhailArkhipov:
- Remove plain MD from our handlers. Let Web Extensions handle it.
- Harden VsAppShell so export provider and composition service can be dynamically initialized without R package load.
- Completion test fixes/improvements
- Add dynamic visibility flag to Feedback commands


src/Markdown/Editor/Impl/ContentType/MdContentTypeDefinition.cs
```diff
@@ -6,11 +6,9 @@
 
 namespace Microsoft.Markdown.Editor.ContentTypes {
     public sealed class MdContentTypeDefinition {
-        public const string LanguageName = "Markdown";
-        public const string ContentType = "Markdown";
-        public const string FileExtension1 = ".md";
-        public const string FileExtension2 = ".markdown";
-        public const string FileExtension3 = ".rmd";
```
karthiknadig: This is a breaking change, note this down for Release notes?

MikhailArkhipov: Not really. 99% people use RMD. We never fully supported plain markdown and generally people install Web Extensions to edit it. Dev14 U2+ also provides textmate grammar so MD is colorized.

PR #2692, opened 2016-11-14
Hardening named pipe to specifically allow access to only authenticated user.

karthiknadig:


karthiknadig:
Merging this since the change restricts to Authenticated Users which has always been the case for this service.

PR #2695, opened 2016-11-15
Fix #2575: F5 hangs when Remote + Reset R Interactive is on

### LINKED ISSUE #2575: F5 hangs when Remote + Reset R Interactive is on
Create new R Project
Right click project, Properties
Check Reset R Interactive on run
Check Transfer files on run
Connect to remote R, wait for REPL to be ready
Press F5 to run

Result: Debug is Starting progress bar dialog that doesn't go away.

If you don't check "Reset R Interactive on run", it works fine.

int19h:
Switch to main thread before taking the credentials lock, to avoid deadlocking when main thread is waiting on the currently executing task.

### DIFF
--- a/src/Common/Core/Impl/Security/SecurityService.cs
+++ b/src/Common/Core/Impl/Security/SecurityService.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Common.Core.Security {
         }
 
         public Task<Credentials> GetUserCredentialsAsync(string authority, bool invalidateStoredCredentials, CancellationToken cancellationToken = default(CancellationToken)) {
+            _coreShellLazy.Value.AssertIsOnMainThread();
+
             var showDialog = invalidateStoredCredentials;
             var credentials = new Credentials();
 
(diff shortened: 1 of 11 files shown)


src/Common/Core/Impl/Security/SecurityService.cs
```diff
@@ -22,7 +22,7 @@ internal class SecurityService : ISecurityService {
             _coreShellLazy = coreShellLazy;
         }
 
-        public async Task<Credentials> GetUserCredentialsAsync(string authority, bool invalidateStoredCredentials, CancellationToken cancellationToken = default(CancellationToken)) {
+        public Task<Credentials> GetUserCredentialsAsync(string authority, bool invalidateStoredCredentials, CancellationToken cancellationToken = default(CancellationToken)) {
```
AlexanderSher: Let's add an assert for UI thread

int19h: Added.

AlexanderSher:
👍

PR #2698, opened 2016-11-15
Broker detects if server port is available.

### LINKED ISSUE #2536: Broker hang sometimes races with the new instance of Broker

karthiknadig:
Fixes #2536

### DIFF
--- a/src/Host/Broker/Impl/Startup/Program.cs
+++ b/src/Host/Broker/Impl/Startup/Program.cs
@@ -57,7 +57,24 @@ namespace Microsoft.R.Host.Broker.Startup {
 
             var tlsConfig = new TlsConfiguration(_logger, _securityOptions);
             var httpsOptions = tlsConfig.GetHttpsOptions(Configuration);
-            CreateWebHost(httpsOptions).Run();
+
+            Uri uri = GetServerUri(Configuration);
+            try {
+                CreateWebHost(httpsOptions).Run();
+            } catch(AggregateException ex) when (ex.IsPortInUseException()) {
+                _logger.LogError(0, ex.InnerException, Resources.Error_ConfiguredPortNotAvailable, uri?.Port);
+            }
+        }
+
+        private static Uri GetServerUri(IConfigurationRoot configuration) {
+            try {
+                Uri uri;
+                var url = configuration.GetValue<string>("server.urls", null);
+                if (Uri.TryCreate(url, UriKind.Absolute, out uri) && uri.Port != 0) {
+                    return uri;
+                }
+            } catch (Exception) { }
+            return null;
         }
 
         public static IWebHost CreateWebHost(HttpsConnectionFilterOptions httpsOptions) {
--- a/src/Host/Broker/Impl/Extensions/ExceptionExtensions.cs
+++ b/src/Host/Broker/Impl/Extensions/ExceptionExtensions.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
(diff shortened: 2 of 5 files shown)


src/Host/Broker/Impl/Startup/Program.cs
```diff
@@ -57,7 +61,38 @@ public class Program {
 
             var tlsConfig = new TlsConfiguration(_logger, _securityOptions);
             var httpsOptions = tlsConfig.GetHttpsOptions(Configuration);
-            CreateWebHost(httpsOptions).Run();
+
+            Uri uri = GetServerUri(Configuration);
+            if (IsPortAvailable(uri?.Port)) {
```
int19h: This is still fundamentally a race condition, right? Since any other process can grab the port between the check and the creation. Can we somehow handle the error that happens during the creation if the port ends up being taken, to mitigate that corner case?

karthiknadig: This check was to handle the case where the port is occupied even before broker starts. I will update with mitigation for the corner case.

karthiknadig: @int19h  I am running into a issue where the logger is disposed before i can log the port failure, in the corner case. An exception in webHost.Run() (UvException with EADDRINUSE) results in logger getting disposed and by the time we get to logging the message there is no file logger present. Any suggestions on how to fix that?

int19h: Kinda weird, given that you create the logger. What disposes it?

karthiknadig: The web host disposes it before raising the exception.

```
Microsoft.R.Host.Broker.exe!Microsoft.R.Host.Broker.Logging.FileLogger.Dispose() Line 20	C#
Microsoft.R.Host.Broker.exe!Microsoft.R.Host.Broker.Logging.FileLoggerProvider.Dispose() Line 39 + 0xb bytes	C#
Microsoft.Extensions.Logging.dll!Microsoft.Extensions.Logging.LoggerFactory.Dispose() + 0x59 bytes	
Microsoft.Extensions.DependencyInjection.dll!Microsoft.Extensions.DependencyInjection.ServiceProvider.Dispose() + 0x19c bytes	
Microsoft.AspNetCore.Hosting.dll!Microsoft.AspNetCore.Hosting.Internal.WebHost.Dispose() + 0x4b bytes	
Microsoft.AspNetCore.Hosting.dll!Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(Microsoft.AspNetCore.Hosting.IWebHost host, System.Threading.CancellationToken token, string shutdownMessage) + 0x47b bytes	
Microsoft.AspNetCore.Hosting.dll!Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(Microsoft.AspNetCore.Hosting.IWebHost host) + 0xb8 bytes	
Microsoft.R.Host.Broker.exe!Microsoft.R.Host.Broker.Startup.Program.Main(string[] args) Line 68 + 0x1f bytes	C#

```

int19h: https://github.com/aspnet/Hosting/blob/2bddba8f9023c0d14332b90235c7dffe16e3e353/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L65

https://github.com/aspnet/Hosting/blob/2bddba8f9023c0d14332b90235c7dffe16e3e353/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L244

Looks like they're disposing the web host unconditionally when `Run` exits, and there are no hooks provided to intercept and log the exception before then. And the host, when it is disposed, disposes the service container with all its services.

So we can either rewrite `Run` (should be doable since it's just an extension method, but it's not as small as I'd like), or else just recreate the logger provider.

karthiknadig:
Ping.

int19h:
👍

PR #2709, opened 2016-11-17
Remove credentials when a connection is deleted.

### LINKED ISSUE #2699: Deleting a connection does not remove it from credential manager

karthiknadig:
Fixes #2699


karthiknadig:
This change specifically addresses the case when a remote connection is deleted from workspaces window, the backing credentials are also deleted. This however does NOT address the following issues:
1. Uninstall and re-installing VS in a different location  leaves behind credentials in the cred manager. 
2. Updating/changes to the credentials are not supported in the workspaces UI. This can be done by editing the entry in the credential manager. After this change is checked in removing and adding the connection allows you to change credentials.

src/Common/Core/Impl/OS/CredentialManagerUtil.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using static Microsoft.Common.Core.NativeMethods;
+
+namespace Microsoft.Common.Core.OS {
+    public class CredentialManagerUtil {
```
AlexanderSher: We have an `ISecurityService` already. I think it can go there directly.

int19h:
👍

PR #2710, opened 2016-11-17
Workspace fixes

### LINKED ISSUE #2632: Test connection does not work when supplying just the machine name
Since we only append protocol and port on Save

### LINKED ISSUE #2634: Delete connection should show confirmation dialog
Accidental clicking removes it and user may have problem recalling the machine name

### LINKED ISSUE #2635: Trailing space in the machine name make 'Save' disabled
Paste `KANADIG-RSRV ` (note trailing space). Save is disabled - it is confusing since it is unclear why as everything appears OK. Delete trailing space - Save is now enabled.

huguesv:
Fix #2632 Test connection does not work when supplying just the machine name
Fix #2634 Delete connection should show confirmation dialog
Fix #2635 Trailing space in the machine name make 'Save' disabled

### DIFF
--- a/src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
+++ b/src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
@@ -168,7 +168,7 @@ namespace Microsoft.R.Components.ConnectionManager.Implementation.ViewModel {
                 || !RCommandLineArguments.EqualsIgnoreCase(_connection?.RCommandLineArguments);
 
             Uri uri = null;
-            var isPathValid = Uri.TryCreate(GetCompletePath(), UriKind.Absolute, out uri);
+            var isPathValid = Uri.TryCreate(Path, UriKind.Absolute, out uri);
             if (string.IsNullOrEmpty(Name)) {
                 IsValid = false;
                 SaveButtonTooltip = Resources.ConnectionManager_ShouldHaveName;
@@ -183,10 +183,23 @@ namespace Microsoft.R.Components.ConnectionManager.Implementation.ViewModel {
             IsRemote = !(uri?.IsFile ?? true);
         }
 
-       public string GetCompletePath() {
+        public void PathEditCompleted() {
+            // After the user is done editing the path, we fix it up automatically
+            Path = GetCompletePath();
+
+            // Conveniently populate the name, if the user directly edited the path
+            // without specifying a name.
+            if (string.IsNullOrWhiteSpace(Name)) {
+                Name = Path;
+            }
+        }
+
+        internal string GetCompletePath() {
             // https://foo:5444 -> https://foo:5444 (no change)
-            // https://foo -> https://foo (no change)
-            // http://foo -> http://foo (no change)
+            // https://foo -> https://foo:443
+            // http://foo -> http://foo:80
+            // http://FOO -> http://foo:80
+            // http://FOO:80 -> http://foo:80
             // foo->https://foo:5444
 
             var path = Path?.Trim();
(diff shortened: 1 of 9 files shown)


src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -183,15 +183,29 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
...
         }
 
-       public string GetCompletePath() {
+        public void PathEditCompleted() {
+            // After the user is done editing the path, we fix it up automatically
+            Path = GetCompletePath();
+
+            // Conveniently populate the name, if the user directly edited the path
+            // without specifying a name.
+            if (string.IsNullOrWhiteSpace(Name)) {
```
int19h: One catch here is that the first time they leave focus, we'll auto-generate the name, but if they later update the URI the name will stay as is. I think this should rather check if the name is empty or looks like it was auto-generated (from the URI before the edit), and update it in the latter case also.

huguesv: Good idea, I'll change it.

huguesv: Oh wait, I don't know what the previous path was, it's already been updated.

int19h: Can the old path be stashed away somewhere?

int19h: Actually, maybe it's better to split this away from here, and instead handle it in TextChanged event (or whatever it's called), rather than on LostFocus. That way, you'll see the name update (if it was empty or auto-generated) directly as you're typing, which I think will look nicer, and it should also be easier to get the old property value in that event.

int19h: (but the URL fixup should still be on focus loss)

huguesv:
Making the name track whatever was typed was problematic, once the typed path was auto-completed.
Instead, I ended up changing it so that the name tracks the host name. So you can come in later and edit an old connection by changing the host name, and the name will match it if it was previously automatically generated.

int19h:
I think this is actually even better - hostname by itself is what most people will care about, and makes a more descriptive label.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -183,15 +187,56 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
...
+
+            var previousProposedName = GetProposedName(previousPath);
+            var currentProposedName = GetProposedName(currentPath);
+
+            // Avoid changing anything if the edit
+            // has no effect on the name (like changing the path's port)
+            if (previousProposedName != currentProposedName) {
+                // Check if the name was calculated from the previous path
+                var currentName = Name ?? string.Empty;
+                if (string.IsNullOrEmpty(currentName) || string.Compare(currentName, previousProposedName, StringComparison.CurrentCultureIgnoreCase) == 0) {
```
int19h: I believe we have `EqualsIgnoreCase` extension method on strings somewhere.

int19h: But also, does this need to ignore case? It feels like the only time when it'd be different is if the user manually edited it to make the case differ, in which case it might be better to not get in the way.

huguesv: I can go either way, I have tried both. It felt like it may be more annoying not to track the new name, then to lose a mostly insignificant case change.

int19h: I don't really have a strong opinion on it either way.

int19h:
👍

AlexanderSher:
👍

PR #2713, opened 2016-11-17
Fix #2550: Interrupt printing in REPL

### LINKED ISSUE #2550: Interrupt printing in REPL
Sometimes I'll do something stupid like trying to print a 1-million-row data frame. When this happens, the REPL becomes unresponsive, including the Interrupt R button. All I can do is wait until it's finished (it doesn't crash, mind you, which is good).

Would it be possible to interrupt R while it's in the middle of printing?

int19h:
Avoid using DispatchOnUIThread when printing text where possible, to prevent message queue saturation.

### DIFF
--- a/src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
+++ b/src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
@@ -248,8 +248,9 @@ namespace Microsoft.R.Components.InteractiveWorkflow.Implementation {
         /// Workaround for interactive window that does not currently support
         /// 'carriage return' i.e. writing into the same line
         /// </summary>
-        private string _lastMessage;
+        private volatile string _lastMessage;
         private int _lastPosition = -1;
+
         private void OutputBuffer_Changed(object sender, TextContentChangedEventArgs e) {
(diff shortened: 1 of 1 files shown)


src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -261,26 +262,32 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
...
+                    // written into the same place.
+                    _lastMessage = message.Substring(1);
+                    _lastPosition = _lastPosition < 0 ? CurrentWindow.OutputBuffer.CurrentSnapshot.Length : _lastPosition;
+                    message = "|"; // replacement placeholder so we can receive 'buffer changed' event
+
                     CurrentWindow.Write(message);
                     CurrentWindow.FlushOutput(); // Must flush so we do get 'buffer changed' immediately.
+                });
+            } else {
+                if (_lastMessage != null) {
```
AlexanderSher: `_lastMessage` has to be volatile

int19h: Fixed.

AlexanderSher:
👍

Issue #2708, opened 2016-11-17
Connection manager (workspaces) doesn't respect VS theme


huguesv:
![image](https://cloud.githubusercontent.com/assets/1696845/20402188/8fd907e0-acb0-11e6-9802-0a2d70dcbaa2.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/e083bf16d48cbca63550deeb70ac06407475089d

MikhailArkhipov:
Now light theme is not quite good.

![image](https://cloud.githubusercontent.com/assets/12820357/20542474/96d1753c-b0b6-11e6-9b9a-6ae52dd64eaf.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0c7df6d1552c3a36e431fad30fc338dd80acc2b9

huguesv:
Still bad in dark theme:
![image](https://cloud.githubusercontent.com/assets/1696845/20690727/1a9188c0-b581-11e6-87c1-cce66c1fa78b.png)

MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/20690772/6724e470-b581-11e6-8a76-3a73139ffe22.png)

MikhailArkhipov:
Looks like `Add` template is missing theme

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/e223045a9636ce7a070e28485d5ef629305c6b90

Issue #2707, opened 2016-11-17
Publish With Options - to Project not working


athafoud:
Steps to reproduce

1. Create an empty Database project
2. Create an R Project
3. Add a new SQL Stored Procedure Item 
4. Write some R code
5. From R Tools Menu -> Data -> Publish With Options
6. Select Project in Publish to and the Database you previously created in Target
7. In R code Replacement select "In stored procedure @script parameter"
8. Hit Publish

The "R" folder is created in the selected Database project, but it is empty

If in step 7, instead of "In stored procedure @script parameter", we select "In a separate table", the created "RCode.PostDeployment" is invalid. The content of "RCode.PostDeployment" are
`INSERT INTO RCodeTable`
the rest of the script is missing. The "CreateRCodeTable" is created as expected.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/e083bf16d48cbca63550deeb70ac06407475089d

Issue #2571, opened 2016-10-25
Colorization gets broken in REPL


MikhailArkhipov:
Might be issue with the ASCII progress bar code.

![image](https://cloud.githubusercontent.com/assets/12820357/19703796/02dab094-9aba-11e6-8b00-6b82d2424e66.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/e083bf16d48cbca63550deeb70ac06407475089d

Issue #2186, opened 2016-08-20
IntelliSense support for data.table objects


ksavin:
data.table is a popular package that extends data.frame and is capable of fast and complex data transformations.

Its syntax exposes column names when applying transformations within "[ ]", e.g:

```
library(data.table)
dt <- data.table(mtcars)

# find average gross horse powers for 6-cylinder cars grouping by number of forward gears
dt[cyl == 6, mean(hp), by = gear]
```

It would be great if auto-completion of column names within brackets is supported by IntelliSense.

smortaz:
+1

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/e083bf16d48cbca63550deeb70ac06407475089d

PR #2720, opened 2016-11-19
Provide host load indication and bug fixes

MikhailArkhipov:
Host measures load once in a few seconds and caches the result. Clients update meters periodically. This can be made configurable.

#2707 Publish With Options - to Project not working
- should receive list of files rather that erroneously looking into target project

- themed control template

#2571 Colorization gets broken in REPL 
- concurrency issues caused replacement at the wrong position.  BufferChanged must come right after replacement character is inserted so we always replace at the end of the buffer.

#2186 IntelliSense support for data.table objects


src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -248,49 +237,59 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
...
+                // Note: DispatchOnUIThread is expensive, and can saturate the message pump when there's a lot of output,
+                // making UI non-responsive. So avoid using it unless we need it - and we only need it for FlushOutput,
+                // and we only need it to handle CR.
+                if (message.Length > 1 && message[0] == '\r' && message[1] != '\n') {
+                    _coreShell.DispatchOnUIThread(() => {
+                        CurrentWindow.FlushOutput();
+                        // If message starts with CR we remember current output buffer
+                        // length so we can continue writing lines into the same spot.
+                        // See txtProgressBar in R.
+                        if (message.Length > 1 && message[0] == '\r' && message[1] != '\n') {
```
AlexanderSher: Looks like a merge conflict - this condition is checked twice

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -248,49 +237,59 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
...
         /// 'carriage return' i.e. writing into the same line
         /// </summary>
-        private volatile string _lastMessage;
-        private int _lastPosition = -1;
+        struct MessagePos {
+            public string Message;
+            public int Position;
+        }
 
+        private readonly Stack<MessagePos> _messageStack = new Stack<MessagePos>();
```
AlexanderSher: Why it is a stack and not a queue?

src/R/Components/Impl/Information/HostLoadIndicatorViewModel.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        private void OnTimer(object sender, ElapsedEventArgs e) {
+            if (!_disposed && !_busy) {
+                UpdateModelAsync().DoNotWait();
+            }
+        }
+
+        internal Task UpdateModelAsync() {
+            _busy = true;
```
AlexanderSher: What is this flag for?

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -66,9 +72,13 @@ internal class ConnectionManager : IConnectionManager {
...
         private async Task CompleteInitializationAsync() {
             await _shell.SwitchToMainThreadAsync();
-            _disposableBag.Add(_statusBar.AddItem(new ConnectionStatusBar {
-                DataContext = _statusBarViewModel
-            }));
+            _disposableBag
+                .Add(_statusBar.AddItem(new ConnectionStatusBar {
+                    DataContext = _statusBarViewModel
+                }))
+                .Add(_statusBar.AddItem(new HostLoadIndicator() {
```
AlexanderSher: Parentheses aren't required

src/Host/Broker/Impl/About/AboutController.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Microsoft.R.Host.Broker.About {
     [Authorize(Policy = Policies.RUser)]
-    [Route("/about")]
```
int19h: Why this change? We've used lowercase for HTTP paths everywhere else, and it is also the standard REST convention.

int19h: If you just want to distinguish different things, add subpoints under `/about`  (and maybe rename it to `/info`). So we can have `/info/host`, `/info/load` etc.

Although I'd make it `/status/load` etc instead. `/about` or `/info` should be for static stuff that doesn't change for a particular instance of broker, like version numbers or install locations.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
MikhailArkhipov: So in case we do get multiple changes they will be  applied in reverse order so positions won't change for change 1 after change 2 is applied.

src/R/Components/Impl/Information/HostLoadIndicatorViewModel.cs
MikhailArkhipov: So we ignore timer events while `GetHostInformationAsync` is sill being executed.

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
MikhailArkhipov: Yep. fixed.

MikhailArkhipov:
Adding picture

image

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
```diff
@@ -40,27 +41,29 @@ public sealed class SessionInformationCommand : IAsyncCommand {
         }
 
         private async Task PrintBrokerInformationAsync(bool reportTelemetry = false) {
-            var a = await _interactiveWorkflow.RSessions.Broker.GetHostInformationAsync();
-            var window = _interactiveWorkflow.ActiveWindow?.InteractiveWindow;
+            var a = await _interactiveWorkflow.RSessions.Broker.GetHostInformationAsync<AboutHost>();
+            if (a != null) {
```
MikhailArkhipov: OK

src/R/Components/Impl/Information/HostLoadIndicatorViewModel.cs
```diff
@@ -0,0 +1,89 @@
...
+            return _interactiveWorkflow.RSessions.Broker.GetHostInformationAsync<HostLoad>()
+                .ContinueWith((t) => {
```
MikhailArkhipov: OK

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
AlexanderSher: Does it mean that only one change may apply?

src/R/Components/Impl/Information/HostLoadIndicatorViewModel.cs
```diff
@@ -0,0 +1,92 @@
...
+        private void OnTimer(object sender, ElapsedEventArgs e) {
+            if (!_disposed && !_busy) {
+                UpdateModelAsync().DoNotWait();
+            }
+        }
+
+        internal async Task UpdateModelAsync() {
+            _busy = true;
+            try {
+                var result = await _interactiveWorkflow.RSessions.Broker.GetHostInformationAsync<HostLoad>();
```
AlexanderSher: Since we have this code that pings broker all the time, I'd rather move it into `RSessionProvider` and raise an event every time it receives a message. It will allow us to get rid of the expensive `RSessionProvider.RSessionOnDisconnectedAsync`

And you don't need a timer. `TaskScheduler` will do it for you:
```{csharp}
while (_isDisposed) {
    await Task.Delay(interval, cancellationToken);
    var info = await Broker.GetHostInformationAsync<HostLoad>(cancellationToken);
}
```

int19h:
👍

src/R/Components/Impl/Information/HostLoadIndicatorViewModel.cs
MikhailArkhipov: OK. It is a bit more work, so I am going to open separate item. I guess you want some event like HostStateChanged (live/dead) fired when load cannot be retrieved and we'd use it instead of testing connection. #2722

PR #2723, opened 2016-11-21
Fixes sourcing file on remote with debugger attached.

### LINKED ISSUE #2495: Can't source files in a remote session when debugger is attached
```
x <- 1
y <- 2
z <- 3
```

set bp somewhere
Attach debugger
Try sourcing file

image

karthiknadig:
Fixes #2495

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/util.R
+++ b/src/Host/Client/Impl/rtvs/R/util.R
@@ -253,6 +253,13 @@ save_to_project_folder <- function(blob_id, project_name, dest_dir) {
     invisible(call_embedded('save_to_project_folder', blob_id, project_name, path.expand(dest_dir), temp_dir));
 }
 
+save_to_temp_folder <- function (blob_id, file_name) {
+    temp_file <- paste0(tempdir(), '/', file_name);
+    unlink(temp_file);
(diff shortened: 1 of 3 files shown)


karthiknadig:
Ping

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -253,6 +253,13 @@ save_to_project_folder <- function(blob_id, project_name, dest_dir) {
     invisible(call_embedded('save_to_project_folder', blob_id, project_name, path.expand(dest_dir), temp_dir));
 }
 
+save_to_temp_folder <- function (blob_id, file_name) {
+    temp_file <- paste0(tempdir(), '/', file_name);
```
MikhailArkhipov: what happens if TEMP is not set?. Unlikely, but anyway. Also, I assume we know that the file name is unique.

MikhailArkhipov:
👍

src/Host/Client/Impl/rtvs/R/util.R
karthiknadig: It uses R User directory. If TMPDIR, TMP, TEMP, are checked first, if none of them are defined it uses R_USER

PR #2724, opened 2016-11-21
Test fixes

MikhailArkhipov:
SetDirectory* tests got broken since commands how check if host is running. Updated and added cases for remote.

PeekInternalFunction01 was failing b/c startupInfo.WorkingDirectory was empty and not null which caused `setwd` to fail at the host startup.

Editor test needs to be UI test. Was failing occasionally depending on the test execution thread,


MikhailArkhipov:
Will merge since this basically just text fixes.

PR #2725, opened 2016-11-21
Skip connecting while editing text boxes.

### LINKED ISSUE #2633: Double-click on machine name in 'edit connection' launches test connection
Create connection
Open it for editing
Double-click on machine name (say, you want to paste over selection)
Observed: Test connection starts
Expected: select machine name in the edit box

karthiknadig:
Fixes #2633

### DIFF
--- a/src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
+++ b/src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
@@ -15,10 +15,10 @@ namespace Microsoft.R.Components.ConnectionManager.Implementation.View {
     /// Interaction logic for ConnectionManagerControl.xaml
     /// </summary>
     public partial class ConnectionManagerControl : UserControl {
-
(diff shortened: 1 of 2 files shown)


src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -82,14 +82,37 @@ public partial class ConnectionManagerControl : UserControl {
         }
 
         private void Connection_MouseDoubleClick(object sender, MouseButtonEventArgs e) {
-            Model?.Connect(GetConnection(e));
+            if (!_editing) {
```
AlexanderSher: `(e.Source as ListBoxItem)?.IsSelected == false` should be enough.

karthiknadig: you are correct. That works.

MikhailArkhipov:
👍

PR #2727, opened 2016-11-22
Fix #2721: AF in cat(), argument type mismatch in protocol message

### LINKED ISSUE #2721: AF in cat(), argument type mismatch in protocol message
Reported as crash in 0.5, need to check what is going on there

```
x <- "2016-11-19T20:21:41.7452661"
cat(x)
```

in

```
            if (arg.Type != expectedType1 && arg.Type != expectedType2) {
                throw ProtocolError($"{name} must be {expectedType1} or {expectedType2}:", this);
```

arg.Type is `Date` while code expects `String`

int19h:
Work around JamesNK/Newtonsoft.Json#862 by explicitly requesting no "smart" date parsing during JSON deserialization.

### DIFF
--- a/src/Common/Core/Impl/Json/Json.cs
+++ b/src/Common/Core/Impl/Json/Json.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.IO;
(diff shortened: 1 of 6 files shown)


int19h:
Note that this pull request is for the `Release/0_5` branch. It will need to be ported to master separately.

MikhailArkhipov:
👍

Issue #2734, opened 2016-11-22
Create infrastructure for #2733


AlexanderSher:
- References between method fixtures
- `ConnectionInfoFixture` that starts "remote" broker locally if no URL is provided
- bat or ps file to run tests from command line

MikhailArkhipov:
So is it going in this week or is it going to be Sprint 9?

AlexanderSher:
Most of the work is done, so let's leave it in this sprint. I can update estimate to reflect the progress.

MikhailArkhipov:
It's OK. Updating estimate for progress then screws up velocity calculations :-(

MikhailArkhipov:
Do we still need this in S9? I'd rather avoid major code restructuring at this point in cycle.

AlexanderSher:
Those are mostly test code changes (product changes are in the pull request #2738 ).

MikhailArkhipov:
Move to the future? Doesn't look like it is going to fit into 1.0

AlexanderSher:
I've actually checked in this one yesterday.
The other one though can be moved to the next sprint.

Issue #2733, opened 2016-11-22
Add unit tests for remote connections


AlexanderSher:

MikhailArkhipov:
This thing is a year old. I guess won't happen...

Issue #2722, opened 2016-11-21
Use host load information retrieval to track connection state


MikhailArkhipov:
Per @AlexanderSher 

```
Since we have this code that pings broker all the time, I'd rather move 
it into RSessionProvider and raise an event every time it receives a message. 
It will allow us to get rid of the expensive 
RSessionProvider.RSessionOnDisconnectedAsync
```

PR #2738, opened 2016-11-22
Product changes for #2734: Create infrastructure for #2733

AlexanderSher:


src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -88,46 +87,50 @@ public class RSessionProvider : IRSessionProvider {
...
+            NotifyBrokerConnectedAsync().DoNotWait();
         }
 
         private void RSessionOnDisconnected(object sender, EventArgs e) {
-            RSessionOnDisconnectedAsync().DoNotWait();
+            NotifyBrokerDisconnectedAsync().DoNotWait();
         }
 
-        private async Task RSessionOnDisconnectedAsync() {
+        private async Task NotifyBrokerConnectedAsync() {
```
AlexanderSher: This change can be reverted completely since #2722 will solve the problem

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -127,60 +127,65 @@ internal abstract class BrokerClient : IBrokerClient {
...
-            try {
-                await sessions.PutAsync(name, new SessionCreateRequest {
-                    InterpreterId = _interpreterId,
-                    CommandLineArguments = rCommandLineArguments,
-                }, cancellationToken);
-            } catch (BrokerApiErrorException apiex) {
-                throw new RHostDisconnectedException(apiex);
-            }
+            var sessions = new SessionsWebService(HttpClient, _credentials, Log);
+            using (Log.Measure(LogVerbosity.Normal, $"Create broker session \"{name}\"")) {
```
MikhailArkhipov: Invariant

src/Common/Core/Impl/Logging/LogExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Common.Core.Disposables;
+
+namespace Microsoft.Common.Core.Logging {
+    public static class LogExtensions {
+        public static IDisposable Measure(this IActionLog log, LogVerbosity verbosity, string message) {
+            if (log.LogVerbosity < verbosity) {
+                return Disposable.Empty;
+            }
+
+            log.Write(verbosity, MessageCategory.General, $"{message} started");
```
MikhailArkhipov: Invariant or FxCop bites us

MikhailArkhipov: As well as all other places

src/Common/Core/Impl/Logging/LogExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+            return Disposable.Create(() => {
+                stopwatch.Stop();
+                log.Write(verbosity, MessageCategory.General, $"{message} completed in {stopwatch.ElapsedMilliseconds} ms.");
```
MikhailArkhipov: If we also want this to work for customers, we'd need to use ETL

PR #2746, opened 2016-11-22
Fixes issue with creating multiple workspaces with the same machine.

### LINKED ISSUE #2743: Not able create multiple entries with same machine name but different command line
Try creating the following:
Name: remote1
Machine: remote-server
Command line: --some-args1

Name: remote2
Machine: remote-server
Command line: --some-args2

karthiknadig:
Fixes #2743


MikhailArkhipov:
There are tests for the CM, it is quite testable, would be nice to add one for this case.

MikhailArkhipov:
👍

Issue #2747, opened 2016-11-23
Canceling test connection does not really cancel


MikhailArkhipov:
Start testing remote connection
Double-click on local

RTVS connects to local
Few seconds later certificate verification popup shows up

System.Net knows nothing about cancellation.

```
>	Microsoft.R.Host.Client.dll!Microsoft.R.Host.Client.Host.RemoteBrokerClient.ValidateCertificateHttpHandler(object sender, System.Security.Cryptography.X509Certificates.X509Certificate certificate, System.Security.Cryptography.X509Certificates.X509Chain chain, System.Net.Security.SslPolicyErrors sslPolicyErrors) Line 71	C#
 	System.dll!System.Net.ServerCertValidationCallback.Callback(object state)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)	Unknown
 	System.dll!System.Net.ServerCertValidationCallback.Invoke(object request, System.Security.Cryptography.X509Certificates.X509Certificate certificate, System.Security.Cryptography.X509Certificates.X509Chain chain, System.Net.Security.SslPolicyErrors sslPolicyErrors)	Unknown
 	System.dll!System.Net.ServicePoint.HandshakeDoneProcedure.CertValidationCallback(string hostName, System.Security.Cryptography.X509Certificates.X509Certificate certificate, System.Security.Cryptography.X509Certificates.X509Chain chain, System.Net.Security.SslPolicyErrors sslPolicyErrors)	Unknown
 	System.dll!System.Net.Security.SecureChannel.VerifyRemoteCertificate(System.Net.Security.RemoteCertValidationCallback remoteCertValidationCallback)	Unknown
 	System.dll!System.Net.Security.SslState.CompleteHandshake()	Unknown
 	System.dll!System.Net.Security.SslState.CheckCompletionBeforeNextReceive(System.Net.Security.ProtocolToken message, System.Net.AsyncProtocolRequest asyncRequest)	Unknown
[...]

Issue #2742, opened 2016-11-22
NRE in Connection Manager


MikhailArkhipov:
Open Workspaces with no connections active
Add remote
Click test connection
While connection is being tested click Cancel on editing connection (no on test)
Now click connect to local

Connection is null but IsConnected is true.

![image](https://cloud.githubusercontent.com/assets/12820357/20543556/1b2251ea-b0bb-11e6-816b-6026e76fbae2.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/0c7df6d1552c3a36e431fad30fc338dd80acc2b9

Issue #2740, opened 2016-11-22
Got 3 (three) certificate dialog when connecting to remote


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/20543066/64b2a910-b0b9-11e6-9473-99501347e7ca.png)

MikhailArkhipov:
Looks like UI thread reenters and we get second dialog on the same subject

```
 	Microsoft.VisualStudio.R.Package.dll!Microsoft.VisualStudio.R.Package.Shell.VsAppShell.ShowMessage(string message, Microsoft.Common.Core.Shell.MessageButtons buttons, Microsoft.Common.Core.Shell.MessageType messageType) Line 274	C#
 	Microsoft.R.Common.Core.dll!Microsoft.Common.Core.Security.SecurityService.ValidateX509CertificateAsync(System.Security.Cryptography.X509Certificates.X509Certificate certificate, string message, System.Threading.CancellationToken cancellationToken) Line 71	C#
 	[Resuming Async Method]	
 	[External Code]	
>	Microsoft.VisualStudio.R.Package.dll!Microsoft.VisualStudio.R.Package.Shell.VsAppShell.ShowMessage(string message, Microsoft.Common.Core.Shell.MessageButtons buttons, Microsoft.Common.Core.Shell.MessageType messageType) Line 274	C#
 	Microsoft.R.Common.Core.dll!Microsoft.Common.Core.Security.SecurityService.ValidateX509CertificateAsync(System.Security.Cryptography.X509Certificates.X509Certificate certificate, string message, System.Threading.CancellationToken cancellationToken) Line 71	C#
 	[Resuming Async Method]	
 	[External Code]	
 	Microsoft.VisualStudio.R.Package.dll!Microsoft.VisualStudio.R.Package.Shell.VsProgressDialog.Show(System.Func<System.Threading.CancellationToken, System.Threading.Tasks.Task> method, string waitMessage, int delayToShowDialogMs) Line 21	C#
 	Microsoft.R.Components.dll!Microsoft.R.Components.ConnectionManager.Implementation.ViewModel.ConnectionManagerViewModel.Connect(Microsoft.R.Components.ConnectionManager.ViewModel.IConnectionViewModel connection) Line 230	C#

```

AlexanderSher:
Amount of message boxes is equal to the amount of `RSession` instances

MikhailArkhipov:
[...]

PR #2748, opened 2016-11-23
#2708, #2740, #2747, #2742

MikhailArkhipov:
#2740 Got 3 (three) certificate dialog when connecting to remote
#2747 Canceling test connection does not really cancel
#2742 NRE in Connection Manager

### DIFF
--- a/src/Host/Client/Impl/Host/RemoteBrokerClient.cs
+++ b/src/Host/Client/Impl/Host/RemoteBrokerClient.cs
@@ -18,7 +18,10 @@ namespace Microsoft.R.Host.Client.Host {
     internal sealed class RemoteBrokerClient : BrokerClient {
         private readonly IConsole _console;
         private readonly ICoreServices _services;
-        private string  _certificateHash;
+        private readonly SemaphoreSlim _verificationSemaphore = new SemaphoreSlim(1, 1);
+
+        private string _certificateHash;
+        private bool? _certificateValidationResult;
 
         static RemoteBrokerClient() {
             ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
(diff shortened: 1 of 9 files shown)


src/Host/Client/Impl/Host/RemoteBrokerClient.cs
```diff
@@ -45,8 +50,12 @@ public RemoteBrokerClient(string name, Uri brokerUri, ICoreServices services, IC
                 ? new RHostDisconnectedException(Resources.Error_BrokerNotRunning, exception)
                 : await base.HandleHttpRequestExceptionAsync(exception);
         }
-        
+
         private bool ValidateCertificateHttpHandler(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors) {
+            if(_cancellationToken.IsCancellationRequested) {
```
AlexanderSher: Can you please try one more option?
`sender` should be of the `HttpWebRequest` type, and if it is, you should be able to check `HttpWebRequest.Status`

src/Host/Client/Impl/Host/RemoteBrokerClient.cs
```diff
@@ -55,26 +64,33 @@ public RemoteBrokerClient(string name, Uri brokerUri, ICoreServices services, IC
             if (sslPolicyErrors.HasFlag(SslPolicyErrors.RemoteCertificateNotAvailable)) {
                 Log.WriteAsync(LogVerbosity.Minimal, MessageCategory.Error, Resources.Error_NoBrokerCertificate).DoNotWait();
                 _console.Write(Resources.Error_NoBrokerCertificate);
-            } else {
-                var hashString = certificate.GetCertHashString();
-                bool accepted = true;
+                return false;
+            }
 
+            _verificationSemaphore.Wait();
```
AlexanderSher: I think simple `lock` should be enough here (since we don't have any async code).

src/Host/Client/Impl/Host/RemoteBrokerClient.cs
AlexanderSher: Ok, looks like it isn't possible to get information from `HttpWebRequest`. `Aborted` property is internal.

Issue #2753, opened 2016-11-23
Occasional deadlock in ValidateX509CertificateAsync


MikhailArkhipov:
`ValidateCertificateHttpHandler` wants to show the dialog, but `SwitchToMainThreadAsync` never completes. 

I think this might happen if first thread enters the lock and then second call to ValidateCertificateHttpHandler comes on the main thread which gets blocked on the lock and `SwitchToMainThreadAsync` deadlocks.

MikhailArkhipov:
May be mitigated by #2749

AlexanderSher:
Do you have a stack trace with deadlock?

MikhailArkhipov:
I've seen it twice so far. Trying to repro.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ef304d2a88ec47c2dc2ddd5b383415832e75c7b5

Issue #2750, opened 2016-11-23
Double-click on connection for switching no longer works


MikhailArkhipov:

MikhailArkhipov:
`((e.Source as ListBoxItem)?.IsSelected == false)` is not sufficient since item IS selected

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ef304d2a88ec47c2dc2ddd5b383415832e75c7b5

Issue #2735, opened 2016-11-22
Setup no longer displays version of the product


MikhailArkhipov:
Something got broken

![image](https://cloud.githubusercontent.com/assets/12820357/20542128/24ad4b12-b0b5-11e6-90d4-38e93ecdd6a8.png)

MikhailArkhipov:
Fallout of Dev15 related changes

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ef304d2a88ec47c2dc2ddd5b383415832e75c7b5

PR #2759, opened 2016-11-28
#2735, #2750, #2753

MikhailArkhipov:
#2735 Setup no longer displays version of the product
#2753 Occasional deadlock in ValidateX509CertificateAsync 
 - mostly workaround, rejecting calls on UI thread and added cancellation handling
#2750 Double-click on connection for switching no longer works 

Minor changes in session naming - consolidate session names in one place.

### DIFF
--- a/src/Host/Client/Impl/Definitions/GuidList.cs
+++ b/src/Host/Client/Impl/Definitions/GuidList.cs
@@ -6,5 +6,7 @@ using System;
 namespace Microsoft.R.Host.Client {
     public static class GuidList {
         public static readonly Guid InteractiveWindowRSessionGuid = new Guid("77E2BCD9-BEED-47EF-B51E-2B892260ECA7");
+        public static readonly Guid PackageManagerRSessionGuid = new Guid("61C93E8D-D24D-4012-82F4-093086A4FB08");
+        public static readonly Guid IntellisenseRSessionGuid = new Guid("8BEF9C06-39DC-4A64-B7F3-0C68353362C9");
     }
 }
--- a/src/Host/Client/Impl/Host/RemoteBrokerClient.cs
+++ b/src/Host/Client/Impl/Host/RemoteBrokerClient.cs
@@ -77,8 +77,8 @@ namespace Microsoft.R.Host.Client.Host {
                     Log.WriteAsync(LogVerbosity.Minimal, MessageCategory.Warning, Resources.Trace_UntrustedCertificate.FormatInvariant(certificate.Subject)).DoNotWait();
 
                     var message = Resources.CertificateSecurityWarning.FormatInvariant(Uri.Host);
-                    var task = _services.Security.ValidateX509CertificateAsync(certificate, message);
-                    _services.Tasks.Wait(task);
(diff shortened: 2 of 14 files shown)


src/Host/Client/Impl/Definitions/GuidList.cs
```diff
@@ -6,5 +6,7 @@
 namespace Microsoft.R.Host.Client {
     public static class GuidList {
         public static readonly Guid InteractiveWindowRSessionGuid = new Guid("77E2BCD9-BEED-47EF-B51E-2B892260ECA7");
+        public static readonly Guid PackageManagerRSessionGuid = new Guid("61C93E8D-D24D-4012-82F4-093086A4FB08");
```
AlexanderSher: No, it actually should be the opposite - `InteractiveWindowRSessionGuid` should be removed from the `Microsoft.R.Host.Client`. And actually right now nothing prevents us from doing it

src/Host/Client/Impl/Host/RemoteBrokerClient.cs
```diff
@@ -67,6 +68,13 @@ public RemoteBrokerClient(string name, Uri brokerUri, ICoreServices services, IC
                 return false;
             }
 
+            if (_services.MainThread.ThreadId == Thread.CurrentThread.ManagedThreadId) {
```
MikhailArkhipov: Nope. I've seen it about three times but will keep trying.

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -81,7 +81,9 @@ public partial class ConnectionManagerControl : UserControl {
         }
 
         private void Connection_MouseDoubleClick(object sender, MouseButtonEventArgs e) {
-            if ((e.Source as ListBoxItem)?.IsSelected == false) {
+            var connection = GetConnection(e);
+            if (connection != Model.EditedConnection) {
+                Model?.CancelEdit();
```
AlexanderSher: Do we cancel edit when user simply clicks on connection button?

src/Host/Client/Impl/Definitions/GuidList.cs
AlexanderSher: How is it different for tracing if this file is in `Microsoft.R.Host.Client` or `Microsoft.R.Components`?

src/Host/Client/Impl/Host/RemoteBrokerClient.cs
AlexanderSher: ok

src/Host/Client/Impl/Definitions/GuidList.cs
AlexanderSher: I mean - it is good to have all session guids in one place, but I think it should be in `Microsoft.R.Components` instead of `Microsoft.R.Host.Client`.

MikhailArkhipov: OK

AlexanderSher:
👍

karthiknadig:
:shipit:

PR #2760, opened 2016-11-28
Two bug fixes

### LINKED ISSUE #2722: Use host load information retrieval to track connection state

### LINKED ISSUE #2749: Switching sessions should only be switching primary (REPL)
Intellisense session will be re-created as needed. Package manager should auto-close (simpler) when targets switch or should auto-update itself.

AlexanderSher:
- Fix #2749: Switching sessions should only be switching primary (REPL)
- Fix #2722: Use host load information retrieval to track connection state
- Part of the fix for #2070: Save As... does not change file name

### DIFF
--- a/src/Host/Client/Impl/Session/RSessionProvider.cs
+++ b/src/Host/Client/Impl/Session/RSessionProvider.cs
@@ -14,13 +14,13 @@ using Microsoft.Common.Core.Services;
 using Microsoft.Common.Core.Telemetry;
 using Microsoft.Common.Core.Threading;
 using Microsoft.R.Host.Client.Host;
+using Microsoft.R.Host.Protocol;
 using Microsoft.R.Interpreters;
 
 namespace Microsoft.R.Host.Client.Session {
     public class RSessionProvider : IRSessionProvider {
         private readonly ConcurrentDictionary<Guid, RSession> _sessions = new ConcurrentDictionary<Guid, RSession>();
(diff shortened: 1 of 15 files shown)


src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -326,26 +298,26 @@ public class RSessionProvider : IRSessionProvider {
...
-
-            public Task<REvaluationResult> EvaluateAsync(string expression, REvaluationKind kind, CancellationToken cancellationToken = new CancellationToken()) {
-                return _evaluation.EvaluateAsync(expression, kind, cancellationToken);
+        private async Task UpdateHostLoadAsync(CancellationToken ct = default(CancellationToken)) {
+            HostLoad hostLoad;
+            using (await _connectArwl.ReaderLockAsync(ct)) {
+                try {
+                    hostLoad = await Broker.GetHostInformationAsync<HostLoad>(ct);
+                } catch (RHostDisconnectedException) {
+                    hostLoad = null;
```
MikhailArkhipov: You can just init it to null

MikhailArkhipov:
👍

PR #2766, opened 2016-11-29
Tweaks to avoid authenticating load and about requests.

karthiknadig:


src/Host/Broker/Impl/About/AboutInfo.cs
```diff
@@ -5,27 +5,21 @@
...
 using Microsoft.R.Host.Broker.Sessions;
 using Microsoft.R.Host.Protocol;
 
 namespace Microsoft.R.Host.Broker.About {
-    [Authorize(Policy = Policies.RUser)]
-    [Route("/info/about")]
-    public class AboutController : Controller {
-        private readonly InterpreterManager _interpManager;
-        private readonly SessionManager _sessionManager;
+    public class AboutInfo {
```
int19h: It should either be a static class, or a service. I think the latter would be preferable (you can then get components injected via constructor etc).

src/Host/Broker/Impl/About/LoadInfo.cs
```diff
@@ -8,21 +8,16 @@
...
-using Microsoft.AspNetCore.Authorization;
-using Microsoft.AspNetCore.Mvc;
-using Microsoft.R.Host.Broker.Security;
 using Microsoft.R.Host.Protocol;
 
 namespace Microsoft.R.Host.Broker.About {
-    [Authorize(Policy = Policies.RUser)]
-    [Route("/info/load")]
-    public class LoadController : Controller {
+    public class LoadInfo {
```
int19h: Ditto.

int19h:
This goes very low level to circumvent all the ASP.NET Core authentication machinery. Is `[AllowAnonymous]` not sufficient? If so, perhaps we can still do something higher level by controlling the middleware chain better or more precisely.

One thing that I notice is that in `SecurityManager.SignInAsync` (which is called to create the principal and add claims to it), we have access to `BasicSignInContext`, which in turn has `HttpContext`. This should give us the URL of the request, and decide whether we need to run the logon code path and add the corresponding claim, or else just set the ticket to some anon principal.

karthiknadig:
AllowAnonymous check happens after the claims have been acquired. So it goes through the Sign In before it even checks for anonymous. I like the use HttpContext with anon principle idea. Let me give that a try.

src/Host/Broker/Impl/Extensions/BasicSignInContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Odachi.AspNetCore.Authentication.Basic;
+
+namespace Microsoft.R.Host.Broker {
+    public static class BasicSignInContextExtensions {
+        public static bool IsSignInRequired(this BasicSignInContext context) {
+            string path = context.HttpContext.Request.Path.ToString();
+            if (path.Equals("/info/load") || path.Equals("/info/about")) {
```
int19h: Let's make it a `static readonly HashSet` so that it's easy to add new paths as needed.

(Later, might want to look at detecting `[AllowAnonymous]` attributes instead, but for now let's keep it simple.)

MikhailArkhipov:
LGTM. 👍 Let's merge and build for the BB today.

int19h:
👍

Issue #2771, opened 2016-11-29
Crash while typing in REPL connected to remote R


huguesv:
AggregateException

```
>	mscorlib.dll!System.Threading.Tasks.Task.ThrowIfExceptional(bool includeTaskCanceledExceptions)	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.RFunctionCompletion.TryFetchDescription()	Unknown
 	Microsoft.R.Editor.dll!Microsoft.R.Editor.Completion.RFunctionCompletion.Description.get()	Unknown
 	Microsoft.VisualStudio.Platform.VSEditor.dll!Microsoft.VisualStudio.Language.Intellisense.Implementation.DefaultCompletionSetPresenter.ShowToolTip()	Unknown
 	Microsoft.VisualStudio.Platform.VSEditor.dll!Microsoft.VisualStudio.Language.Intellisense.Implementation.DefaultCompletionSetPresenter.OnShowTooltipTimer_Tick(object sender, System.EventArgs e)	Unknown
 	WindowsBase.dll!System.Windows.Threading.DispatcherTimer.FireTick(object unused)	Unknown
 	WindowsBase.dll!System.Windows.Threading.ExceptionWrapper.InternalRealCall(System.Delegate callback, object args, int numArgs)	Unknown
 	WindowsBase.dll!System.Windows.Threading.ExceptionWrapper.TryCatchWhen(object source, System.Delegate callback, object args, int numArgs, System.Delegate catchHandler)	Unknown
 	WindowsBase.dll!System.Windows.Threading.DispatcherOperation.InvokeImpl()	Unknown
 	WindowsBase.dll!System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(object state)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
[...]

Issue #2757, opened 2016-11-27
Changing file content type causes NRE in RSuggestedActionSource.HasSuggestedActionsAsync


AlexanderSher:
- Create new R project 
- Rename `script.r` to file with no extension
- Rename to R file again

Changing file extension causes change of content type and disposes `RSuggestedActionSource`, but `Dispose` doesn't remove `RSuggestedActionSource` from `ServiceProvider`. As a result, instance is reused when file is renamed again.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/e223045a9636ce7a070e28485d5ef629305c6b90

PR #2775, opened 2016-11-30
Bug fixes

### LINKED ISSUE #2708: Connection manager (workspaces) doesn't respect VS theme

MikhailArkhipov:
#2757 Changing file content type causes NRE
#2771 Crash while typing in REPL connected to remote R


AlexanderSher:
👍

PR #2780, opened 2016-11-30
Three bug fixes

### LINKED ISSUE #2773: pkg mgr says disconnected
1.  connected to remote, works fine.  tried installing pkg, got:

image

2. the msg should be in RED and BOLD if in fact we are connected to remote but pkg mgr is connected to local or unconnected.  perhaps not allow pkg mgmt.

3. i dismissed the msg, switched to local, got the error again using a local interp:

image

UPDATE: 

when switching back & forth between local Rs, the msg appears and disappears.  it should probably instead:

"Currently not connected to an R interpreter" 

(because failed implies something else and RHost isnt a known entity for users)

### LINKED ISSUE #2777: Long delays may cause incorrect connection indicator
Better repro in debugger. May be product issue - say, `GetHostInformationAsync` takes a long time

1. Set bp in `PrintBrokerInformationAsync`
2. Connect to remote
3. bp is hit.
4. Wait a few min
5. Continue

image

### LINKED ISSUE #2778: Deadlock on fast switch
- Open VS with no REPL open and no workspace connected. Two workspaces should be available.
- Switch to the first of them
- Switch to the another one right after progress bar disappears

Progress bar stays forever.

Issue happens because right after switch ends REPL tries to start its session. `RInteractiveEvaluator.InitializeAsync` is called on UI thread, takes the lock and switches to the background thread to start RHost. When second switch starts, it blocks UI thread and doesn't allow `InitializeAsync` continuation to return back to UI thread and release lock

AlexanderSher:
- Fix #2778: Deadlock on fast switch
- Fix #2777: Long delays may cause incorrect connection indicator
- Fix #2773: pkg mgr says disconnected

### DIFF
--- a/src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
+++ b/src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Common.Core.Threading {
             ///          ↓     ↓            ↓        ↓
             ///    R−→W−→W−→E−→R  ―►  R−→W−→W−→E−→E−→R
             ///             ↑                     ↑
-            ///   erLock.Tail               erlTail
+            ///       erlTail               erlTail
             /// </remarks>
             public void AddExclusiveReader(ExclusiveReaderLockSource item, out bool isAddedAfterWriterOrExclusiveReader) {
                 lock (this) {
@@ -134,7 +134,7 @@ namespace Microsoft.Common.Core.Threading {
             public LockSource[] Remove(LockSource item) {
                 lock (this) {
                     var next = item.Next;
-                    Interlocked.CompareExchange(ref _head, next, item);
+                    var oldHead = Interlocked.CompareExchange(ref _head, next, item);
 
                     var previous = item.Previous;
                     Interlocked.CompareExchange(ref _tail, previous, item);
@@ -151,7 +151,7 @@ namespace Microsoft.Common.Core.Threading {
                         return new LockSource[0];
                     }
 
-                    if (_head.IsWriter && next == _head) {
+                    if (_head.IsWriter && item == oldHead) {
                         return new[] { _head };
                     }
 
@@ -159,10 +159,14 @@ namespace Microsoft.Common.Core.Threading {
                         return FilterAndCopyToArray(next);
                     }
 
+                    if (erLock != null) {
+                        return FindFirstExclusiveReader(erLock);
+                    }
+
                     return new LockSource[0];
                 }
             }
-            
+
             private void UpdateTail(LockSource item) {
                 var tail = Interlocked.Exchange(ref _tail, item);
                 if (tail == null) {
@@ -194,6 +198,27 @@ namespace Microsoft.Common.Core.Threading {
                 }
 
                 return items;
+            }
+
+            private LockSource[] FindFirstExclusiveReader(ExclusiveReaderLock erLock) {
+                if (erLock.Tail == null) {
+                    return new LockSource[0];
+                }
+
+                var item = _head;
+                while (item != erLock.Tail) {
+                    if (item.IsWriter) {
+                        return new LockSource[0];
+                    }
+
+                    if ((item as ExclusiveReaderLockSource)?.ExclusiveReaderLock == erLock) {
+                        return new[] { item };
+                    }
+
+                    item = item.Next;
+                }
+
+                return new[] { item };
             }
 
             private bool ReaderCanBeReleased(LockSource source) {
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -215,27 +215,33 @@ namespace Microsoft.R.Host.Client.Session {
         }
 
         public async Task EnsureHostStartedAsync(RHostStartupInfo startupInfo, IRSessionCallback callback, int timeout = 3000, CancellationToken cancellationToken = default(CancellationToken)) {
-            using (_disposeToken.Link(ref cancellationToken))
-            using (await _initializationLock.WaitAsync(cancellationToken)) {
-                if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
-                    await StartHostAsyncBackground(startupInfo, callback, timeout, cancellationToken);
+            using (_disposeToken.Link(ref cancellationToken)) {
+                await TaskUtilities.SwitchToBackgroundThread();
+
+                using (await _initializationLock.WaitAsync(cancellationToken)) {
+                    if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
+                        await StartHostAsyncBackground(startupInfo, callback, timeout, cancellationToken);
+                    }
                 }
             }
         }
 
         public async Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callback, int timeout = 3000, CancellationToken cancellationToken = default(CancellationToken)) {
-            using (_disposeToken.Link(ref cancellationToken))
-            using (await _initializationLock.WaitAsync(cancellationToken)) {
-                if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
-                    await StartHostAsyncBackground(startupInfo, callback, timeout, cancellationToken);
-                } else {
-                    throw new InvalidOperationException("Another instance of RHost is running for this RSession. Stop it before starting new one.");
+            using (_disposeToken.Link(ref cancellationToken)) {
+                await TaskUtilities.SwitchToBackgroundThread();
+
+                using (await _initializationLock.WaitAsync(cancellationToken)) {
+                    if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
+                        await StartHostAsyncBackground(startupInfo, callback, timeout, cancellationToken);
+                    } else {
+                        throw new InvalidOperationException("Another instance of RHost is running for this RSession. Stop it before starting new one.");
+                    }
                 }
             }
         }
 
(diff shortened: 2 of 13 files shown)


src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
```diff
@@ -134,7 +134,7 @@ private class Queue {
             public LockSource[] Remove(LockSource item) {
                 lock (this) {
                     var next = item.Next;
-                    Interlocked.CompareExchange(ref _head, next, item);
+                    var oldHead = Interlocked.CompareExchange(ref _head, next, item);
```
MikhailArkhipov: Seems locked already, do we need interlocked?

src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
```diff
@@ -70,7 +70,7 @@ private class Queue {
             ///          ↓     ↓            ↓        ↓
             ///    R−→W−→W−→E−→R  ―►  R−→W−→W−→E−→E−→R
             ///             ↑                     ↑
-            ///   erLock.Tail               erlTail
+            ///       erlTail               erlTail
             /// </remarks>
             public void AddExclusiveReader(ExclusiveReaderLockSource item, out bool isAddedAfterWriterOrExclusiveReader) {
```
MikhailArkhipov: The lock may benefit from detailed documentation. What is exactly the sync pattern? Single writer/multiple readers? What is the example of exclusive reader? Can it simply be a writer that actually doesn't write anything?

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -215,27 +215,33 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
         public async Task EnsureHostStartedAsync(RHostStartupInfo startupInfo, IRSessionCallback callback, int timeout = 3000, CancellationToken cancellationToken = default(CancellationToken)) {
-            using (_disposeToken.Link(ref cancellationToken))
-            using (await _initializationLock.WaitAsync(cancellationToken)) {
-                if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
-                    await StartHostAsyncBackground(startupInfo, callback, timeout, cancellationToken);
+            using (_disposeToken.Link(ref cancellationToken)) {
+                await TaskUtilities.SwitchToBackgroundThread();
+
+                using (await _initializationLock.WaitAsync(cancellationToken)) {
+                    if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
```
MikhailArkhipov: What about 'Faulted' or 'Canceled' state?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -194,33 +197,50 @@ public class RSessionProvider : IRSessionProvider {
                 }
 
                 OnBrokerStateChanged(new HostLoad());
```
MikhailArkhipov: When we are only testing connection but not actually switching - it seems like 'broker changed' should not be firing?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -194,33 +197,50 @@ public class RSessionProvider : IRSessionProvider {
...
+                if (transaction != null) {
+                    transactions.Add(transaction);
+                } else {
+                    sessionsToStop.Add(session);
+                }
+            }
+
             if (transactions.Any()) {
-                await SwitchSessionsAsync(transactions, cancellationToken);
+                await SwitchSessionsAsync(transactions, sessionsToStop, cancellationToken);
```
MikhailArkhipov: So if we are only to switch interactive session - do we still need transactions and multiple sessions support?

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Fault or canceled mean that it has failed to start host last time, so we allow to try again

src/R/Components/Impl/ConnectionManager/Implementation/View/ConnectionManagerControl.xaml.cs
```diff
@@ -94,16 +94,7 @@ public partial class ConnectionManagerControl : UserControl {
...
-            if (model != null) {
-                var connection = GetConnection(e);
-                if (connection != model.EditedConnection) {
-                    model.CancelEdit();
-                } else {
-                    model.Save(connection);
-                }
-                model.Connect(connection);
-            }
+            Model?.Connect(GetConnection(e));
```
MikhailArkhipov: Does cancels editing if it is in progress?

MikhailArkhipov: nvm, see it below

src/R/Components/Impl/PackageManager/Implementation/ViewModel/RPackageManagerViewModel.cs
```diff
@@ -429,14 +429,20 @@ internal class RPackageManagerViewModel : BindableBase, IRPackageManagerViewMode
             }
         }
 
-        private async Task LoadInstalledAndLoadedPackagesAsync() {
+        private async Task LoadInstalledAndLoadedPackagesAsync(bool reload) {
```
MikhailArkhipov: What happens when user reconnects elsewhere while packages are still being updated? It may be simpler just to close the window. BTW, enumerating packages will put load on the connection and stall REPL intellisense, for example.

smortaz:
i dont know all the intricacies, but logically switching repls is like
logging out of X and logging into Y.  hence, stopping everything is
perfectly acceptable, esp if it reduces stateg mgmt complexity.

MikhailArkhipov:
We can't be stopping active REPL session until the switch completes successfully since it would delete all the user data. It is safe to stop other sessions though. However, we could go to the model similar to reloading projects: if user wants to switch, then give warning so they can save rdata then proceed with stopping everything. If connection fails they end up with not connected anywhere. I agree with @smortaz, it is not that different from logging on/off or solution reload.

src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
AlexanderSher: `Interlocked.CompareExchange` is translated into `lock cmpxchg` on multi-core machine, so there is no need to use `volatile` and it is just one operation instead of 
```cs
var oldHead = _head;
if (item == _head) {
    _head = next;
}
```

MikhailArkhipov:
In fact same as VS close solution: all related widows automatically close - so we can just close PM

src/Common/Core/Impl/Threading/AsyncReaderWriterLock.Queue.cs
MikhailArkhipov: I mean the operation looks like concurrent double-linked list. Maybe just make a class?

AlexanderSher: That was my original idea, but it happened that it is not enough - locks around several operations with the list are still required. There is an existing `LinkedList<T>`, but it is slower and creates tons of garbage in memory.

MikhailArkhipov:
Make this simple may also reduce cost of #2733 and #2734. We may choose to push these to S9 and instead just make connection switching as modal and simple as possible (but not simpler - (c) Einstein :-))

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -215,27 +215,33 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
         public async Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callback, int timeout = 3000, CancellationToken cancellationToken = default(CancellationToken)) {
-            using (_disposeToken.Link(ref cancellationToken))
-            using (await _initializationLock.WaitAsync(cancellationToken)) {
-                if (_initializationTcs.Task.Status != TaskStatus.RanToCompletion || !_isHostRunning) {
-                    await StartHostAsyncBackground(startupInfo, callback, timeout, cancellationToken);
-                } else {
-                    throw new InvalidOperationException("Another instance of RHost is running for this RSession. Stop it before starting new one.");
+            using (_disposeToken.Link(ref cancellationToken)) {
+                await TaskUtilities.SwitchToBackgroundThread();
+
```
AlexanderSher: Ok, I'll merge implementation into one method.

MikhailArkhipov: Let's instead consider @smortaz  proposal

AlexanderSher:
Guys, this area isn't a bug prone. I've spent the whole night hunting two bugs that were in completely different places. Changes in RSessionProvider aren't directly related to the bugs, they are to avoid creating RHost instances that won't run the loop (which is more important to the local scenario than to remote).

Changes that you propose won't simplify a lot - we still have to tread two different types of sessions differently, still have to stop everything if connection fails and still have to explicitly wait until non-REPL sessions are stopped. And you still need tests with remote host because bugs happen in some weird places.

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: This code path is used for switching only. Test connection has its own method.

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Save method?

MikhailArkhipov: *Same

MikhailArkhipov:
Let's get this in, then proceed further 👍

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: Yes, that is why I fixed it so `await TaskUtilities.SwitchToBackgroundThread();` is called before lock is acquired.

Issue #2696, opened 2016-11-15
Use TelemetryPiiProperty


huguesv:
We're currently hashing telemetry value ourselves. We should set each event property to TelemetryPiiProperty instance instead of a hashed string. VS takes care of hashing TelemetryPiiProperty, and this will make sure that all data is consistently processed across our products (PTVS already uses it).

MikhailArkhipov:
We need predictable values. I.e. we need to know that some hash means 'utils'.

huguesv:
TelemetryPiiProperty is predictable.

It's a built-in way of reporting PII values, and can make it possible for the server/dashboard to have tools that let you query the data more easily.  Don't know if they actually have that or not, but it is possible.

If you pre-hash it and send it like any regular value, server/dashboard doesn't know it's private info.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/9a664aa23f112177dc3747f5653a650dfcbb6aba

Issue #2628, opened 2016-11-03
UI issues


rnpandya:
Some small but annoying UI issues:
1. I turn off outlining and the R toolbar, and it does not remember that - it comes back very time when I restart VS
2. Ctrl-shift-S in VS usually means Save All Files, but if I'm in an R window it sources that file - very unexpected
3. The R Interactive tab keeps popping to front at random times when I'm debugging a C++ program in the same VS instance

Additional information:

OS Information
    Version:       Microsoft Windows NT 6.3.9600.0

RTVS Information:
    Assembly: Microsoft.VisualStudio.R.Package, Version=0.4.20623.1711, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a

R Install path:
    C:\Program Files\Microsoft\MRO-3.3.1\

MikhailArkhipov:
(1) is by design in VS. The editor does not keep outlining settings.
(2) - you have a choice to use 'data science settings' that match R Studio or use default VS. You can also re-bind keystrokes to your preference in Tools | Options | Keyboard.
(3) - please update to RTVS 0.5 at https://aka.ms/rtvs-current

rnpandya:
(1) I have outlining turned off globally in Settings, but the R windows do not respect that - all my other windows come up without outlining. And hiding the toolbar is persistent for all the other toolbars, but not for the R toolbar.

Thanks,

Ravi

cg31:
The Toolbar is really annoying when it can't remember the status.

MikhailArkhipov:
There is no global setting for outlining, some languages offer the option:

![image](https://cloud.githubusercontent.com/assets/12820357/20771484/728576f8-b6fe-11e6-95ba-eb49d2b80a88.png)

Basic == Visual Basic rather than global setting. So only VB, C#, C++ and XML offer the setting. All other languages only have it per file (Edit | Outlining | Stop Outlining) and that one is not persistent.

MikhailArkhipov:
Will be adding the option and will fix the toolbar visibility

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/9a664aa23f112177dc3747f5653a650dfcbb6aba

Issue #2736, opened 2016-11-22
Connection should not turn green until REPL opens


MikhailArkhipov:
Without REPL opened, open Wokspaces and try connecting to remote. Progress is displayed, then progress dialog closes, connection turns green but REPL doesn't open. After several seconds (~10) REPL finally opens.

the operation should be considered complete when everything is in place and ready.

![image](https://cloud.githubusercontent.com/assets/12820357/20542247/ad05861e-b0b5-11e6-9c09-908d98f5f5f1.png)

AlexanderSher:
We have no way to control when REPL will appear. What happens is that we send command to show REPL and it tries to create its `RSession` from UI. Since we lock `RSessionProvider`, it can't create its session until switch is completed and so can't appear. If we show REPL before we start switching, user will see error message from NullBroker.

Proper solution is to add 4th status - connected but no REPL session. Can we request icons for the statuses?

MikhailArkhipov:
More statuses confuses users. Perhaps null broker should never be showing messages. Then it would be safe to open repl first.

AlexanderSher:
It doesn't show anything by itself. Interactive Window tries to connect to it and shows an error, which is a correct behavior if you're not connected to any other session. Interactive Window has no idea that we're going to try to connect right after it creates the session.

AlexanderSher:
Ok, the main delay happens on the UI thread. Interactive Window shows up several seconds after RHost receives `?>` message, which itself happens after `RInteractiveEvaluator.InitializeAsync` is completed.
So if we want to show green only when interactive window is initialized, we need to track this specific moment.
There is a `LastActiveWindowChanged` event in `IActiveRInteractiveWindowTracker`, but no one uses it. So I remove it from there and add `ActiveWindowChanged` to the `RInteractiveWorkflow` itself.

AlexanderSher:
Btw, `RInteractiveWorkflow` provides access to `ActiveWindow` before it is initalized. I was sure that I've fixed it for the 0.4 release.

PR #2784, opened 2016-11-30
#2696, #2628, consistent output of session info

MikhailArkhipov:
- Make sure session info prints when REPL is ready
#2696 Use TelemetryPiiProperty
#2628 UI issues
 - Add outlining option as in C#/VB/C++
 - Store visibility of R Toolbar so we don't show it forcibly all the time (with mock and test)


src/Package/Impl/Packages/RToolbar.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+
+        public void SaveState() {
+            RToolbarAction((cb) => {
+                _settings.ShowRToolbar = cb.Visible;
+            });
+        }
+
+        private void RToolbarAction(Action<CommandBar> action) {
```
AlexanderSher: Not sure I understand this. Can't we just create `GetToolbar` method?

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
```diff
@@ -35,12 +39,28 @@ public sealed class SessionInformationCommand : IAsyncCommand {
             return CommandResult.Executed;
         }
 
-        private void OnBrokerChanged(object sender, EventArgs e) {
+        private void OnRSessionInitialized(object sender, EventArgs e) {
             if (_interactiveWorkflow.RSession.IsRemote) {
-                PrintBrokerInformationAsync(reportTelemetry: true).DoNotWait();
+                ReplInitComplete().ContinueWith(async (t) => await PrintBrokerInformationAsync(reportTelemetry: true)).DoNotWait();
```
AlexanderSher:
```cs
_interactiveWorkflow.RSession.HostStarted.ContinueWith(
    t => TaskContinuationOptions.PrintBrokerInformationAsync(reportTelemetry: true).DoNotWait(),
    TaskContinuationOptions.OnlyOnRanToCompletion)
```

src/Package/Test/Package/RToolbarTest.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.VisualStudio.R.Package.Packages;
+using NSubstitute;
+using Microsoft.VisualStudio.Shell.Mocks;
+
+namespace Microsoft.VisualStudio.R.Package.Test.Package {
+    [ExcludeFromCodeCoverage]
+    public sealed class RToolbarTest {
+        [Test]
+        public void Visibility() {
+            var settings = Substitute.For<IRToolsSettings>();
```
AlexanderSher: `TestRToolsSettings`?

huguesv:
The telemetry change LGTM.

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
AlexanderSher: Actually, no. We need to either simplify `_interactiveWorkflow.GetOrCreateVisualComponent` (can be done easily) or provide an awaiter, something like `_interactiveWorkflow.ActiveWindowInitialized`
Maybe both. I can create a PR

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
```diff
@@ -197,6 +197,8 @@ internal sealed class RToolsSettingsImplementation : BindableBase, IRToolsSettin
             set { SetProperty(ref _logLevel, value); }
         }
 
+        public bool ShowRToolbar { get; set; } = true;
```
AlexanderSher: Is it different from what we have in VSCT?

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
AlexanderSher: https://github.com/Microsoft/RTVS/pull/2785
And then you can change `PrintBrokerInformationAsync` to:
```cs
        private async Task PrintBrokerInformationAsync(bool reportTelemetry = false) {
            AboutHost aboutHost;
            try {
                aboutHost = await _interactiveWorkflow.RSessions.Broker.GetHostInformationAsync<AboutHost>();
                if (aboutHost == null) {
                    return;
                }
                return;
            }

            await _interactiveWorkflow.Shell.SwitchToMainThreadAsync();
            var window = (await _interactiveWorkflow.GetOrCreateVisualComponentAsync()).InteractiveWindow;

            window.WriteErrorLine(Environment.NewLine + Resources.RServices_Information);
            window.WriteErrorLine("\t" + Resources.Version.FormatInvariant(aboutHost.Version));
            window.WriteErrorLine("\t" + Resources.OperatingSystem.FormatInvariant(aboutHost.OS.VersionString));
            window.WriteErrorLine("\t" + Resources.ProcessorCount.FormatInvariant(aboutHost.ProcessorCount));
            window.WriteErrorLine("\t" + Resources.PhysicalMemory.FormatInvariant(aboutHost.TotalPhysicalMemory, aboutHost.FreePhysicalMemory));
            window.WriteErrorLine("\t" + Resources.VirtualMemory.FormatInvariant(aboutHost.TotalVirtualMemory, aboutHost.FreeVirtualMemory));
            window.WriteErrorLine("\t" + Resources.ConnectedUserCount.FormatInvariant(aboutHost.ConnectedUserCount));
            window.WriteErrorLine(string.Empty);

            if (reportTelemetry) {
                var services = _interactiveWorkflow.Shell.Services;
                foreach (var name in aboutHost.Interpreters) {
                    services.Telemetry.ReportEvent(TelemetryArea.Configuration, "Remote Interpteter", name);
                    services.Telemetry.ReportEvent(TelemetryArea.Configuration, "Remote OS", aboutHost.OS.VersionString);
                    services.Telemetry.ReportEvent(TelemetryArea.Configuration, "Remote CPUs", aboutHost.ProcessorCount);
                    services.Telemetry.ReportEvent(TelemetryArea.Configuration, "Remote RAM", aboutHost.TotalPhysicalMemory);
                }
            }
        }
```
and get rid of event subscription

src/R.sln
```diff
@@ -202,6 +202,18 @@ Global
 		{58D86BE4-FA8B-4F59-91FE-A9B348C70ED2}.Release|x64.Build.0 = Release|Any CPU
 		{58D86BE4-FA8B-4F59-91FE-A9B348C70ED2}.Release|x86.ActiveCfg = Release|Any CPU
 		{58D86BE4-FA8B-4F59-91FE-A9B348C70ED2}.Release|x86.Build.0 = Release|Any CPU
+		{AD8A000B-320D-4B11-B7FF-95C314A40E7B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{AD8A000B-320D-4B11-B7FF-95C314A40E7B}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{AD8A000B-320D-4B11-B7FF-95C314A40E7B}.Debug|x64.ActiveCfg = Debug|Any CPU
```
AlexanderSher: Is it expected?

MikhailArkhipov: VS moves project guids around for some reason.

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
MikhailArkhipov: Also, this is the event when connection should turn green and the progress dialog should go away.

MikhailArkhipov: We can change name to 'Interactive' like in DTHML: Empty -> Connecting -> Connected -> Loading -> Loaded -> Interactive -> Complete.

AlexanderSher: You can move RData into separate class, but you can't use new `Initialized` event for it, cause we have to execute initialization code in a specific order, and events have no determination on that.

In fact, this is fired in the middle of nowhere. On the one hand, there is `Connected` event already, which is raised when RHost loop is ready for input (and if you call any evaluation at that moment, it will be executed after all that initialization code). On the other hand, some complex functions (like rtvs:::install.packages) require RHost to get to the state when it can accept user input, which has nothing to do with `Initialized` event either. And this event has nothing to do with Interactive Window readiness that you need in this particular case (cause you don't really care about session - you need broker and console to print information).

AlexanderSher: > Also, this is the event when connection should turn green and the progress dialog should go away.

No. If we want to turn green only when Interactive Window is visible, we need to track both broker and window. Since window is on UI thread, there can be a big gap between the moment when RHost is ready for user input and Interactive Window has completed its initialization (which is what you actually see in #2736).

src/Package/Impl/Options/R/Tools/RToolsSettingsImplementation.cs
AlexanderSher: ok

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
MikhailArkhipov: We want to turn green when REPL has prompt (i.e. ready for the interaction), not when it is still initializing. Same for the the info printing or asking if user wants to reload rdata. 'Connected' i insufficient, I tried that. In any case we need this kind of event to UI gets reenabled.

MikhailArkhipov: As for init code, sure, it gets in a specific order - but loading RD is always at the end. See `ITextBuffer.HighPriorityChanged` etc.

CPS "eventing" model is hideous and everyone hates it.

AlexanderSher: CPS has a total mess of over engineered dataflow blocks, MEF imports and async events, and I'm not trying to mimic it. And yes, `async/await` itself is a heavily leaking abstraction. 

None of these though makes C# events a silver bullet to control program flow:
- There is no good way to control how many times event must be handled. Great example is WPF 3.5, where you could easily get millions of `PropertyChanged` without an idea if handler should take care of current notification or skip it because the same property will send another `PropertyChanged`. Message pipes with predefined enumeration order are much better in these cases, and dataflow blocks suck because of the way they are implemented, not because pattern itself is bad. 
- There is no good way to get notified by event only once. This is our case. It is good to have a set of events that notify listeners about the state of the object, but in the multi-threaded environment, you sign for notification or execute handler if object is in the required state already, so code usually looks like this:
```cs
if (_obj.IsReady) {
    handle(_obj);
} else {
    _obj.Ready += HandleIsReady;
    if (_obj.IsReady) {
        _obj.Ready -= HandleIsReady;
        handle(_obj);
    }
}

private void HandleIsReady(object sender, EventArgs args) {
    _obj.Ready -= HandleIsReady;
    handle(_obj);
}
```
while with tasks and async/await, you need just 
```cs
await _obj.IsReady;
handle(_obj)
```
- Normal events do not allow raiser to wait for handlers in other threads without blocking the thread, because you can't automatically combine several delegates with non-void return type. You'll have to manually iterate through `MulticastDelegate` list of delegates and raise them one by one, ending up with CPS async event, where people forget to use special extension methods and get very unpleasant bugs. They've been hunting these bugs for years. 
- You'll also have to manually iterate through delegates if you want to swallow specific types of exceptions in failed handler and allow other handlers to proceed.

And so on. Every tool has its own scope and set of pros and cons. I don't see why we need to use event when there is a method that does everything we need by itself. On the other hand, if for #2736 event is better, I'll use an event.

AlexanderSher: > Side note: you can combine non-void returning delegates. But if you invoke the result, you'll only get the last returned value; all others are silently discarded.

I don't remember anything in MSDN that exactly the last one is returned. I think it is considered undetermined. 

As for the event, I think we can raise it when we receive `?>`. What about `RExecutionTracer`? Does it need access to the `RSession` ASAP?

MikhailArkhipov: @int19h - Interesting question. There may be code that is actually part of initialization and needs to know when session is connected so it can complete init process. On the other hand, this is probably a minority and may have special early event. In HTML docs there are multiple usable stages such as Loaded -> Interactive -> Complete.

MikhailArkhipov: As for events I'd rather not invent models on a case by case basis. If we are not happy with basic subscriber pattern we should first consider what exactly do we need in practice and then implement the model. I.e. keep things simple and invent new things when they are really needed. Makes code easier to maintain by new hires for example.

int19h: I think right now all initialization code is in `RSession` itself.

The problem with initialization code living outside (if we do have it, or will ever have it) is that it's not clear how it's supposed to be ordered with respect to each other. Presumably some of that code depends on other initialization code running before. For example, for code that is inside `RSession` right now, it needs to load the `rtvs` package first, before it can start doing things like setting the working directory.

When it's all in one place inside the class itself, and everything outside of that class can only start using it once the init stuff is complete, then all is good - we can arrange the init code inside in the sequence that it needs to be in, and any outside user has the guarantee that they can rely on init. So in that world, there should be no need for a _public_ `RSession.Connected` event.

MikhailArkhipov: OK, so move `Connected` and fire it after initialization then?

src/Package/Test/Package/RToolbarTest.cs
AlexanderSher: ok

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
MikhailArkhipov: Actually not sure about changing `Connected` semantics. This may be destabilizing. If we fire `Connected` when user actually can interact with REPL, we will break code that still needs to complete before allowing user to start work. There is plot manager that perform device assignments, working directory dropdown that is being populated on connect and so on.

int19h: I don't have a strong opinion on this. If there's some code that you suspect could be destabilized, let's keep it as is for now. But I think it's worthwhile to revisit the overall design at some later point, and see if we can simplify it to reduce the likelihood of future bugs around current semantics.

PR #2788, opened 2016-12-01
Fix #2736: Connection should not turn green until REPL opens

### LINKED ISSUE #2736: Connection should not turn green until REPL opens

AlexanderSher:
- Added `RInteractiveWorkflow.ActiveWindowChanged` event that is used by `ConnectionManager`
- Changed `RInteractiveWorkflow.ActiveWindow` to return null until REPL is actually initialized
- Removed unused event from `IActiveRInteractiveWindowTracker`


MikhailArkhipov:
👍

PR #2789, opened 2016-12-01
Explicit broker switch

AlexanderSher:
This is an attempt to reduce ambiguity in REPL/RSessionProvider operation.
Previously, `SwitchBrokerToLastConnection` was scheduled into UI thread to run at some point after `RInteractiveWorkflow` completes initialization. That could happen before or after REPL becomes visible, which itself was scheduled to show up during first broker switch. But because of the locks, it couldn't show up at the time the switch is happening.

With this change, behavior is determined:
- While REPL isn't visible and user haven't chosen workspace explicitly, no connection is established
- When user opens REPL (explicitly or by opening a project or by opening R or RMD file), it first checks if it can start local broker and only after that shows the window. 
- When user switches to the connection with closed REPL, `ConnectionManager` will first try to connect and then will open the REPL. Both operations will happen under progress bar.


MikhailArkhipov:
👍

Issue #2769, opened 2016-11-29
initial experience when i have local R's is bad


smortaz:
i have two local R's.  when i start RTVS for the 1st time i get a RED msg saying not connected.  i should just pick one (or the last locally connected one) and use that.  it also says open workspaces, which is probably an unknown concept at the beginning.

MikhailArkhipov:
I think we can do this

a. When there is no previous connection stored, try use highest version from local R. If no R is present, ask if user wants to install MRO or do they want to connect to remote. If the latter, open Workspaces window.

b. When previous connection was local, re-establish it.
c. If previous was remote, do nothing.

smortaz:
i like.

AlexanderSher:
Should we ask user only once about MRO?

MikhailArkhipov:
I think it would be once automatically (almost). If they don't have R and install R or MRO we won't be asking again b/c we will pick up the available R. Or they connect to remote and then per (c) we won't be asking.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ffb8e6a6e4ab20fc5aee7fab37c33675e1d88bcc

PR #2791, opened 2016-12-01
#2769 Initial experience when i have local R's is bad

MikhailArkhipov:
Where there is no last used connection, try automatically pick local R if available


int19h:
👍

Issue #2781, opened 2016-11-30
OOM leaving VS sitting overnight


MikhailArkhipov:
F5
Connected to team server
Left VS idling for the night
OOM in the morning

![image](https://cloud.githubusercontent.com/assets/12820357/20763473/046f38ee-b6df-11e6-8ae1-ccb6896684f9.png)

MikhailArkhipov:
Growing by 1MB per ~10 seconds.

MikhailArkhipov:
We constantly getting

```
>	Microsoft.R.Wpf.dll!Microsoft.R.Wpf.Controls.Watermark.OnTextBoxHintChanged(System.Windows.DependencyObject obj, System.Windows.DependencyPropertyChangedEventArgs args) Line 35	C#
 	WindowsBase.dll!System.Windows.DependencyObject.OnPropertyChanged(System.Windows.DependencyPropertyChangedEventArgs e)	Unknown
 	PresentationFramework.dll!System.Windows.FrameworkElement.OnPropertyChanged(System.Windows.DependencyPropertyChangedEventArgs e)	Unknown
 	PresentationFramework.dll!System.Windows.Controls.TextBox.OnPropertyChanged(System.Windows.DependencyPropertyChangedEventArgs e)	Unknown
 	WindowsBase.dll!System.Windows.DependencyObject.NotifyPropertyChange(System.Windows.DependencyPropertyChangedEventArgs args)	Unknown
 	WindowsBase.dll!System.Windows.DependencyObject.UpdateEffectiveValue(System.Windows.EntryIndex entryIndex, System.Windows.DependencyProperty dp, System.Windows.PropertyMetadata metadata, System.Windows.EffectiveValueEntry oldEntry, ref System.Windows.EffectiveValueEntry newEntry, bool coerceWithDeferredReference, bool coerceWithCurrentValue, System.Windows.OperationType operationType)	Unknown
 	PresentationFramework.dll!System.Windows.StyleHelper.ApplyTemplatedParentValue(System.Windows.DependencyObject container, MS.Internal.FrameworkObject child, int childIndex, ref MS.Utility.FrugalStructList<System.Windows.ChildRecord> childRecordFromChildIndex, System.Windows.DependencyProperty dp, System.Windows.FrameworkElementFactory templateRoot)	Unknown
[...]

Issue #2770, opened 2016-11-29
VS doesn't terminate properly after using remote R


huguesv:
Open VS
Connect to Karthik's server.
Run a plot: `plot(0:10)`
Close VS

Check Task Manager, you'll see devenv.exe is still running.

MikhailArkhipov:
Specific to plotting. Just doing a bit of calculations or bringing up signature help doesn't repro. VS is waiting for the host to close. Task is in 'WaitingForActivation`

```
 	Microsoft.VisualStudio.R.Package.dll!Microsoft.VisualStudio.R.Package.Shell.VsTaskService.Wait(System.Threading.Tasks.Task task, System.Threading.CancellationToken cancellationToken, int ms) Line 14	C#
>	Microsoft.R.Host.Client.dll!Microsoft.R.Host.Client.Session.RSessionProvider.Dispose() Line 66	C#
 	Microsoft.R.Common.Core.dll!Microsoft.Common.Core.Disposables.DisposableBag.TryDispose() Line 58	C#
 	Microsoft.R.Components.dll!Microsoft.R.Components.InteractiveWorkflow.Implementation.RInteractiveWorkflow.Dispose() Line 180	C#
```

MikhailArkhipov:
```
var request = EvaluationRequest.SendAsync("q()"...) 
await request.Task; 
```
last await never returns

MikhailArkhipov:
Fix for #2749 removed the offending code. Need to verify that it fixes the case.

MikhailArkhipov:
Still reproes

MikhailArkhipov:
`await brokerClient.TerminateSessionAsync(hostName)` never returns

MikhailArkhipov:
When VS is shutting down it cannot transition to the main thread. Since we make call to the broker to delete session, it goes through `GetUserCredentialsAsync` that wants to switch to the UI thread. It is not technically necessary when actually displaying UI.

MikhailArkhipov:
@karthiknadig - is it safe to invoke `GetUserCredentialsAsync` on bg thread when credentials are not invalidated?

AlexanderSher:
Maybe make `SwitchToMainThreadAsync` throw exception when UI thread is shutting down?

MikhailArkhipov:
Well we need to execute that code...

AlexanderSher:
Ok, then its not an option

int19h:
[...]

PR #2793, opened 2016-12-02
#2770, #2781

MikhailArkhipov:
#2781 OOM leaving VS sitting overnight
 - separate broker state change and host load change events
#2770 VS doesn't terminate properly after using remote R 
 - partial: make sure we terminate interactive workflow when main thread is still up

### DIFF
--- a/src/Common/Core/Impl/Microsoft.Common.Core.csproj
+++ b/src/Common/Core/Impl/Microsoft.Common.Core.csproj
@@ -42,7 +42,6 @@
     <Reference Include="System.IO.Compression.FileSystem" />
     <Reference Include="System.Security" />
     <Reference Include="System.Windows.Forms" />
-    <Reference Include="WindowsBase" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\..\..\GlobalAssemblyInfo.cs">
--- a/src/Common/Core/Impl/Microsoft.Common.Core.csproj
+++ b/src/Common/Core/Impl/Microsoft.Common.Core.csproj
@@ -42,6 +42,7 @@
     <Reference Include="System.IO.Compression.FileSystem" />
     <Reference Include="System.Security" />
     <Reference Include="System.Windows.Forms" />
+    <Reference Include="WindowsBase" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\..\..\GlobalAssemblyInfo.cs">
--- a/src/Common/Core/Impl/Shell/CoreShellExtensions.cs
+++ b/src/Common/Core/Impl/Shell/CoreShellExtensions.cs
@@ -6,22 +6,18 @@ using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Windows.Threading;
 using Microsoft.Common.Core.Threading;
 
 namespace Microsoft.Common.Core.Shell {
     public static class CoreShellExtensions {
-        public static MainThreadAwaitable SwitchToMainThreadAsync(this ICoreShell coreShell, CancellationToken cancellationToken = default(CancellationToken)) {
-            var disp = Dispatcher.FromThread(coreShell.MainThread);
-            Debug.Assert(!disp.HasShutdownStarted, "Unable to transition to UI thread: dispatcher has started shutdown.");
-            return ((IMainThread)coreShell).SwitchToAsync(cancellationToken);
-        }
+        public static MainThreadAwaitable SwitchToMainThreadAsync(this ICoreShell coreShell, CancellationToken cancellationToken = default(CancellationToken))
+             => ((IMainThread)coreShell).SwitchToAsync(cancellationToken);
 
         public static async Task ShowErrorMessageAsync(this ICoreShell coreShell, string message, CancellationToken cancellationToken = default(CancellationToken)) {
             await coreShell.SwitchToMainThreadAsync(cancellationToken);
             coreShell.ShowErrorMessage(message);
         }
-        
+
         public static async Task<MessageButtons> ShowMessageAsync(this ICoreShell coreShell, string message, MessageButtons buttons, CancellationToken cancellationToken = default(CancellationToken)) {
             await coreShell.SwitchToMainThreadAsync(cancellationToken);
             return coreShell.ShowMessage(message, buttons);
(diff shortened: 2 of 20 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -325,8 +325,7 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
             // Try graceful shutdown with q() first.
             if (host != null) {
                 try {
-                    host.QuitAsync().SilenceException<Exception>().DoNotWait();
```
int19h: Will this still exit early if `q()` succeeds, as the old code did?

src/Common/Core/Impl/Microsoft.Common.Core.csproj
```diff
@@ -42,6 +42,7 @@
     <Reference Include="System.IO.Compression.FileSystem" />
     <Reference Include="System.Security" />
     <Reference Include="System.Windows.Forms" />
+    <Reference Include="WindowsBase" />
```
AlexanderSher: `Common.Core` should have no references to `WindowsBase`. The whole purpose of `IMainThread` abstraction is to avoid referencing WPF assemblies in utility library.

src/Common/Core/Impl/Shell/CoreShellExtensions.cs
```diff
@@ -6,12 +6,16 @@
...
+using System.Windows.Threading;
 using Microsoft.Common.Core.Threading;
 
 namespace Microsoft.Common.Core.Shell {
     public static class CoreShellExtensions {
-        public static MainThreadAwaitable SwitchToMainThreadAsync(this ICoreShell coreShell, CancellationToken cancellationToken = default(CancellationToken))
-            => ((IMainThread)coreShell).SwitchToAsync(cancellationToken);
+        public static MainThreadAwaitable SwitchToMainThreadAsync(this ICoreShell coreShell, CancellationToken cancellationToken = default(CancellationToken)) {
+            var disp = Dispatcher.FromThread(coreShell.MainThread);
+            Debug.Assert(!disp.HasShutdownStarted, "Unable to transition to UI thread: dispatcher has started shutdown.");
```
AlexanderSher: This should be part of `IMainThread` implementation. And I think we should actually throw exception, not just do `Debug.Assert`

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -325,8 +325,7 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
...
-                    await Task.WhenAny(hostRunTask, Task.Delay(2000)).Unwrap();
+                    host.QuitAsync().SilenceException<Exception>().Wait(10000);
```
AlexanderSher: How does it actually work? Blocks UI thread?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -338,7 +344,7 @@ private static Task CompleteSwitchingBrokerAsync(IRSessionSwitchBrokerTransactio
             using (await _connectArwl.ReaderLockAsync(ct)) {
                 try {
                     var hostLoad = await Broker.GetHostInformationAsync<HostLoad>(ct);
-                    OnBrokerStateChanged(hostLoad);
+                    OnHostLoadChanged(hostLoad);
```
AlexanderSher: That is too much of a split. If connection dies, we won't be notified about it anymore.
I'd rather leave one method, call it something like `OnBrokerStateOrHostLoadChanged`, but raise `BrokerStateChanged` only if `IsConnected` has changed, while `HostLoadChanged` would've been raised all the time.

src/Package/Impl/Shell/VsAppShell.cs
```diff
@@ -70,7 +70,13 @@ public sealed class VsAppShell : IApplicationShell, IMainThread, IIdleTimeServic
             }
         }
 
+        public static void Terminate() {
+            _instance?.Dispose();
+        }
+
         private void Initialize() {
+            AppConstants.Initialize();
```
AlexanderSher: MEF allows export an instance not through its ctor, but through a property of another object. We use it in `RProjectLoadHooks`. In this case you can explicitly instantiate `AppConstants` and don't have to introduce `Initialize` method in the interface.

src/Common/Core/Impl/Microsoft.Common.Core.csproj
MikhailArkhipov: It is for the AF below `Dispatcher.FromThread` I guess we could add some sort of 'has shutdown sharted' to `ICoreShell`

src/Common/Core/Impl/Shell/CoreShellExtensions.cs
MikhailArkhipov: OK

src/Host/Client/Impl/Session/RSession.cs
MikhailArkhipov: Waits for task completion or 10 seconds, whichever comes first. Yes, it blocks. If it doesnt block then VS will proceed with shutdown. Here we are in 'IOleComponent.Terminate` call path

src/Package/Impl/Shell/VsAppShell.cs
MikhailArkhipov: Will MEF import use same instance?

AlexanderSher: AFAIK it should

AlexanderSher: You just need to remove `ImportingConstructor` attribute

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: > Actually, it can simply fire `OnBrokerStateChanged` along with `OnHostLoadChanged`
That is exactly what I meant :)

src/Host/Client/Impl/Session/RSession.cs
MikhailArkhipov: Yes, task completed or timeout, whichever comes first.

src/Host/Client/Impl/Session/RSession.cs
MikhailArkhipov: There seem to be more than one caller of this method. Do we want to block in one case and not the other?

MikhailArkhipov: Actually it is called on bg thread in both cases

AlexanderSher: > Do we want to block in one case and not the other?

Exactly. This code is used every time you want to stop RHost (which includes REPL reset, connection switching, etc.), not just when you dispose RSession. I'd like it to block only in case of shutdown.

PR #2796, opened 2016-12-02
Fix #2671: Interrupt R is disabled but REPL is busy

### LINKED ISSUE #2671: Interrupt R is disabled but REPL is busy
RTVS_2016-11-08.3

I've issued some plot commands to remote, and it was stuck, so I used Interrupt R.  I did that twice.

Then I tried `1+1` which also hanged, but Interrupt R is disabled.

image

Dump in \\pytools\r\Bugs\2671

Things work fine again after clicking Reset.

int19h:
When interrupting R, do not cancel the current interaction request task until R confirms completion.

Disable "Interrupt R" command while it's running to prevent re-entrancy.

Fix interrupt test to wait for prompt before first checking command status.

### DIFF
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -203,14 +203,13 @@ namespace Microsoft.R.Host.Client.Session {
 
         public async Task CancelAllAsync(CancellationToken cancellationToken = default(CancellationToken)) {
             using (_disposeToken.Link(ref cancellationToken)) {
-                var cancelTask = _host.CancelAllAsync(cancellationToken);
+                await _host.CancelAllAsync(cancellationToken);
 
                 var currentRequest = Interlocked.Exchange(ref _currentRequestSource, null);
                 var exception = new OperationCanceledException();
                 currentRequest?.TryCancel(exception);
-                ClearPendingRequests(exception);
 
-                await cancelTask;
+                ClearPendingRequests(exception);
             }
         }
 
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -203,13 +203,13 @@ namespace Microsoft.R.Host.Client.Session {
 
         public async Task CancelAllAsync(CancellationToken cancellationToken = default(CancellationToken)) {
             using (_disposeToken.Link(ref cancellationToken)) {
+                var exception = new OperationCanceledException();
+                ClearPendingRequests(exception);
+
                 await _host.CancelAllAsync(cancellationToken);
 
                 var currentRequest = Interlocked.Exchange(ref _currentRequestSource, null);
-                var exception = new OperationCanceledException();
                 currentRequest?.TryCancel(exception);
-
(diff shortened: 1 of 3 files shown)


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -203,14 +203,13 @@ internal sealed class RSession : IRSession, IRCallbacks {
...
-                var cancelTask = _host.CancelAllAsync(cancellationToken);
+                await _host.CancelAllAsync(cancellationToken);
 
                 var currentRequest = Interlocked.Exchange(ref _currentRequestSource, null);
                 var exception = new OperationCanceledException();
                 currentRequest?.TryCancel(exception);
-                ClearPendingRequests(exception);
 
-                await cancelTask;
+                ClearPendingRequests(exception);
```
AlexanderSher: Do you really need to call ` ClearPendingRequests` after `CancelAllAsync`? I thought it was about `currentRequest?.TryCancel(exception)`

int19h: This is not new code - that line was there already.

src/R/Components/Test/InteractiveWorkflow/RInteractiveWorkflowCommandTest.cs
```diff
@@ -106,6 +106,7 @@ public class RInteractiveWorkflowCommandTest : IDisposable {
             command.Should().BeInvisibleAndDisabled();
 
             using (await UIThreadHelper.Instance.Invoke(() => _workflow.GetOrCreateVisualComponentAsync())) {
+                using (await _workflow.RSession.BeginInteractionAsync()) { }
```
AlexanderSher: I've seen this issue too.
Not part of this change, but maybe we should set `RSession._initializationTcs` not in `IRCallbacks.Connected`, but in `IRCallbacks.ReadConsole`?

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: It was, but previously it was before `await cancelTask;`

src/R/Components/Test/InteractiveWorkflow/RInteractiveWorkflowCommandTest.cs
int19h: These sorts of issues were part of my motivation suggesting that `Connected` (as in, the event that's exposed outside of `RSession`) should perhaps only be fired on the first prompt, and everything before that prompt should be treated as internal initialization. What you suggest would be one way to do just that, if I understand correctly :)

src/Host/Client/Impl/Session/RSession.cs
int19h: Indeed. Will do.

src/R/Components/Test/InteractiveWorkflow/RInteractiveWorkflowCommandTest.cs
int19h: Either way, I'm all for it.

AlexanderSher:
👍

Issue #2797, opened 2016-12-02
First time connection displayed in red


MikhailArkhipov:
Although it is connected

MikhailArkhipov:
`IsConnected = eventArgs.IsConnected && _interactiveWorkflow.ActiveWindow != null;`

REPL is not up yet

AlexanderSher:
Requested by #2736

MikhailArkhipov:
Yeah but it doesn't change b/c (a) there is no `UpdateActiveConnection` in  `ActiveWindowChanged` and (b) ` _sessionProvider.IsConnected` is still false when REPL becomes active and CM is not listening to `Connected`

AlexanderSher:
` ActiveWindowChanged` is raised only after RSession is connected:

```cs
            if (!RSessions.HasBroker) {
                var connectedToBroker = await Connections.TryConnectToPreviouslyUsedAsync();
                if (!connectedToBroker) {
                    Connections.GetOrCreateVisualComponent().Container.Show(focus:false, immediate:false);
                }
            }

            await interactiveWindow.InitializeAsync();

            ActiveWindow = window;
            ActiveWindow.Container.UpdateCommandStatus(true);
            _visualComponentTcs.SetResult(ActiveWindow);
            ActiveWindowChanged?.Invoke(this, new ActiveWindowChangedEventArgs(window));
```

MikhailArkhipov:
`_sessionProvider.IsConnected` is false

AlexanderSher:
That is a regression from #2793 - event is raised but property isn't updated

MikhailArkhipov:
Because host info really should not have been used for the connectivity and broker state tracking in the first place.

Issue #2794, opened 2016-12-02
Add GPU report to the session info


MikhailArkhipov:
http://stackoverflow.com/questions/29667666/how-get-gpu-information-in-c

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/bfde563e3fd5ba6f9a51568cdf9558d43de93867

PR #2798, opened 2016-12-02
#2797, #2794 and fix MEF failures

MikhailArkhipov:
#2797 First time connection displayed in red
- IsConnected changes after interlocked, need to cache original state. 

#2794 Add GPU report to the session info

Revert MEF export as property since exporting as property does not work, MEF imports fail.

### DIFF
--- a/src/Host/Broker/Impl/About/AboutController.cs
+++ b/src/Host/Broker/Impl/About/AboutController.cs
@@ -37,25 +37,39 @@ namespace Microsoft.R.Host.Broker.About {
             a.WorkingSet = Environment.WorkingSet;
             a.ConnectedUserCount = _sessionManager.GetUsers().Count();
 
-            var query = new SelectQuery(@"Select * from Win32_OperatingSystem");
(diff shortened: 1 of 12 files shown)


src/Host/Broker/Impl/About/AboutController.cs
```diff
@@ -37,25 +37,39 @@ public class AboutController : Controller {
...
+                    break;
+                }
+            }
+        }
+
+        private void GetVideoControllerInformation(ref AboutHost a) {
+            using (var search = new ManagementObjectSearcher("Select * from Win32_VideoController")) {
+                foreach (var mo in search.Get()) {
+                    a.VideoCardName = mo["Name"].ToString();
+                    a.VideoRAM = ToLong(mo, "AdapterRAM", 1024 * 1024);
```
AlexanderSher: If there is more than one controller, do we multiply total AdapterRAM?

MikhailArkhipov:
Actually I don't what happens with multiple and MO search. I guess We'd have to return a collection

PR #2800, opened 2016-12-02
Fix MEF failures + video info

MikhailArkhipov:
May need to support multiple video adapter reporting in the future


AlexanderSher:
👍

Issue #2806, opened 2016-12-02
Still occasionally get package load failures


MikhailArkhipov:
```
An exception was encountered while constructing the content of this frame.  This information is also logged in "C:\Users\mikhaila\AppData\Roaming\Microsoft\VisualStudio\14.0\ActivityLog.xml".

Exception details:
System.SystemException: System error.
   at Microsoft.VisualStudio.Shell.Interop.IVsShell5.LoadPackageWithContext(Guid& packageGuid, Int32 reason, Guid& context)
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.GetPackage()
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.ConstructContent(
```

MikhailArkhipov:
Launching second time went fine. Either some async operation or maybe VS was updating MEF in a background?

AlexanderSher:
I have finished with #2801 (`CoreServises` isn't exported throw MEF anymore, and so the `ApplicationConstants`), probably that can help

MikhailArkhipov:
Unable to create `Frame caption: Workspaces` when loading package. I guess something depends on the package being fully initialized.

MikhailArkhipov:
Variant

```
LegacySitePackage failed for package [RPackage]Source: &apos;PresentationFramework&apos; Description: Could not load file or assembly &apos;Microsoft.R.Wpf, PublicKeyToken=b03f5f7f11d50a3a&apos; or one of its dependencies.
```

MikhailArkhipov:
Specifically reproes if REPL is up before VS was closed. If VE or Workspaces was up then we get 'cannot construct tool windows content' if just REPL was up, we get 

```
System.Windows.Markup.XamlParseException: Could not load file or assembly &apos;Microsoft.R.Wpf, PublicKeyToken=b03f5f7f11d50a3a&apos; 
or one of its dependencies. The system cannot find the file specified. 

---&gt; System.IO.FileNotFoundException: Could not load file or assembly &apos;Microsoft.R.Wpf, PublicKeyToken=b03f5f7f11d50a3a&apos; or one of its dependencies. 
The system cannot find the file specified.&#x000D;&#x000A;   
[...]

PR #2816, opened 2016-12-02
Fix #2801: Simplify MEF dependencies in CoreServices

### LINKED ISSUE #2801: Simplify MEF dependencies in CoreServices
We are constantly getting MEF composition bugs because of the increasing complexity of this type dependencies (which originally was designed as lightweight entry point). It makes harder to write tests as well.

AlexanderSher:
- CoreServices is no longer a MEF export, and is available through `ICoreShell.Services` property only
- Common.Core is added to vsix manifest as assembly


AlexanderSher:
Probably should also fix #2806

MikhailArkhipov:
👍

PR #2824, opened 2016-12-04
Fix #2810: Switching workspaces during a long-running operation gets stuck

### LINKED ISSUE #2810: Switching workspaces during a long-running operation gets stuck
While connected to a remote machine, do:
```r
plot(1:1000000)
```
Then, while it's doing that (should take on the order of dozens of seconds), switch from remote to local.

VS gets stuck on the "Switching connection" dialog, even long after the plot should have completed. Clicking "Cancel" in that dialog hangs VS outright (window not being redrawn).

AlexanderSher:
- `_cancelAllCts` now affects all requests in the loop
- `CompleteSwitchingBrokerAsync` waits for cancellation only 10 seconds, after that host will be terminated
- `RHost.CancelAllAsync` can be canceled even when message loop is stuck
- Couple of tests added

### DIFF
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -173,9 +173,9 @@ namespace Microsoft.R.Host.Client {
                     response = "Y";
                     break;
                 default: {
-                    FormattableString error = $"YesNoCancel: callback returned an invalid value: {input}";
-                    Trace.Fail(Invariant(error));
-                    throw new InvalidOperationException(Invariant(error));
+                    var error = Invariant($"YesNoCancel: callback returned an invalid value: {input}");
+                    Trace.Fail(error);
+                    throw new InvalidOperationException(error);
                 }
             }
 
@@ -357,26 +357,25 @@ namespace Microsoft.R.Host.Client {
                 return;
             }
 
-            var cancellationRegistration = cancellationToken.Register(() => tcs.TrySetCanceled(cancellationToken));
-
-            try {
-                // Cancel any pending callbacks
-                _cancelAllCts.Cancel();
-                var cts = CancellationTokenSource.CreateLinkedTokenSource(_cts.Token, cancellationToken);
+            using (cancellationToken.Register(() => tcs.TrySetCanceled(cancellationToken))) {
                 try {
-                    await NotifyAsync("!//", cts.Token);
-                } catch (OperationCanceledException) {
-                    return;
-                } catch (MessageTransportException) {
-                    return;
-                } finally {
-                    cts.Dispose();
-                }
+                    // Cancel any pending callbacks
+                    _cancelAllCts.Cancel();
+                    var cts = CancellationTokenSource.CreateLinkedTokenSource(_cts.Token, cancellationToken);
+                    try {
+                        await NotifyAsync("!//", cts.Token);
+                    } catch (OperationCanceledException) {
+                        return;
+                    } catch (MessageTransportException) {
+                        return;
+                    } finally {
+                        cts.Dispose();
+                    }
 
-                await tcs.Task;
-            } finally {
-                cancellationRegistration.Dispose();
-                Volatile.Write(ref _cancelAllTcs, null);
+                    await tcs.Task;
+                } finally {
+                    Volatile.Write(ref _cancelAllTcs, null);
+                }
             }
         }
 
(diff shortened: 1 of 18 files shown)


src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -422,13 +418,14 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
             }
         }
 
-        private async Task<Message> RunLoop(CancellationToken ct) {
+        private async Task<Message> RunLoop(CancellationToken loopCt) {
             TaskUtilities.AssertIsOnBackgroundThread();
-
+            var cancelAllCtsLink = CancellationTokenSource.CreateLinkedTokenSource(loopCt, _cancelAllCts.Token);
```
MikhailArkhipov: `using(var cancelAllCtsLink = ...` ?

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -567,17 +565,30 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
...
             } finally {
+                cancelAllCtsLink.Dispose();
                 _log.ExitRLoop(--_rLoopDepth);
             }
 
             return null;
         }
 
+        private CancellationToken UpdateCancelAllCtsLink(ref CancellationTokenSource cancelAllCtsLink, CancellationToken loopCt) {
+            cancelAllCtsLink.Dispose();
```
MikhailArkhipov: Looks like it may be disposed twice. OK?

src/Host/Client/Test/Session/RSessionTest.CancelAll.cs
```diff
@@ -54,6 +59,60 @@ public class CancelAll : IAsyncLifetime {
                 _session.IsHostRunning.Should().BeTrue();
                 responceTask.Status.Should().Be(TaskStatus.Canceled);
             }
+
+            [Test]
+            [Category.R.Session]
+            public async Task CancelAll_HandingLoop() {
```
MikhailArkhipov: HangingLoop?

MikhailArkhipov: Or HandlingLoop?

src/Package/Impl/DataInspect/Viewers/ObjectDetailsViewerProvider.cs
```diff
@@ -28,8 +29,8 @@ public sealed class ObjectDetailsViewerProvider : IObjectViewer {
             }
         }
 
-        public async Task ViewFile(string fileName, string tabName, bool deleteFile) {
-            await VsAppShell.Current.SwitchToMainThreadAsync();
+        public async Task ViewFile(string fileName, string tabName, bool deleteFile, CancellationToken cancellationToken) {
+            await VsAppShell.Current.SwitchToMainThreadAsync(cancellationToken);
```
MikhailArkhipov: There are probably many other places that do it in non-cancellable way. Do we care?

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -149,62 +149,55 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
...
+                    response = "N";
+                    break;
+                case MessageButtons.Cancel:
+                    response = "C";
+                    break;
+                case MessageButtons.Yes:
+                    response = "Y";
+                    break;
+                default: {
+                    FormattableString error = $"YesNoCancel: callback returned an invalid value: {input}";
```
AlexanderSher: Ok

src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -380,6 +375,7 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
 
                 await tcs.Task;
             } finally {
+                cancellationRegistration.Dispose();
```
AlexanderSher: Sure

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: Won't compile. Variable in `using` can't be changed.

src/Host/Client/Test/Session/RSessionTest.CancelAll.cs
AlexanderSher: yes

src/Package/Impl/DataInspect/Viewers/ObjectDetailsViewerProvider.cs
AlexanderSher: Yeah, that's probably a good idea to verify all the usages of `SwitchToMainThreadAsync`. I've made this change because I've been checking all the methods that are called from `RHost` message loop for possible UI deadlocks. #2825

MikhailArkhipov:
👍

Issue #2826, opened 2016-12-04
When no R present, show option to connect to remote


MikhailArkhipov:
Rather than just offering MRO install. And open the Workspaces window if user says yes.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/1f31a5903fb46fb91f87d90f6c1a39f72b07a954

Issue #2817, opened 2016-12-02
Different prompt for remote


MikhailArkhipov:
`*>` or something

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/1f31a5903fb46fb91f87d90f6c1a39f72b07a954

Issue #2808, opened 2016-12-02
Clicking on library link in package manager in remote case crashes VS


MikhailArkhipov:
May either do nothing or remove the field when connection is remote.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/1f31a5903fb46fb91f87d90f6c1a39f72b07a954

Issue #2792, opened 2016-12-01
Microsoft.Data.ConnectionUI binaries are unsigned


MikhailArkhipov:
Nuget brings unsigned bits. Need to try and not ship those.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/1f31a5903fb46fb91f87d90f6c1a39f72b07a954

PR #2835, opened 2016-12-05
#2792, #2808, #2817, #2826

MikhailArkhipov:
#2826 When no R present, show option to connect to remote
 - show WS window when user says NO to install MRC

#2808 Clicking on library link in package manager in remote case
 - make plain text when session is remote. Add check for folder existence

#2792 Microsoft.Data.ConnectionUI binaries are unsigned
- remove binaries since VS ships them and install even without SQL tools

#2817 Different prompt for remote 

image

### DIFF
--- a/src/Host/Client/Impl/Session/RSession.cs
+++ b/src/Host/Client/Impl/Session/RSession.cs
@@ -20,8 +20,8 @@ using static System.FormattableString;
 
 namespace Microsoft.R.Host.Client.Session {
     internal sealed class RSession : IRSession, IRCallbacks {
-        private static readonly string LocalPrompt = "> ";
-        private static readonly string RemotePrompt = "*> ";
+        private static readonly string RemotePromptPrefix = "*";
+        private static readonly string DefaultPrompt = "> ";
 
         private static readonly Task<IRSessionEvaluation> CanceledBeginEvaluationTask;
         private static readonly Task<IRSessionInteraction> CanceledBeginInteractionTask;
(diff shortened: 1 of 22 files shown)


int19h:
Unicode has a nifty cloud symbol: ☁- maybe use that instead of *?

MikhailArkhipov:
I tried it looks extremely ugly in the font we have. Doesn't look like this nice cloud at all.

int19h:
I just pasted it into REPL and I see what you mean :) too bad.

int19h:
👍

MikhailArkhipov:
I tried 2601 ☁ CLOUD and 2605 ★ BLACK STAR and 26AB ⚫ MEDIUM BLACK CIRCLE. Technically we can make this customizable.

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -100,6 +102,13 @@ internal sealed class RSession : IRSession, IRCallbacks {
             _afterHostStartedTask = TaskUtilities.CreateCanceled(new RHostDisconnectedException());
         }
 
+        private string GetDefaultPrompt(string requestedPrompt = null) {
```
AlexanderSher: `GetPrompt` probably?

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -216,6 +225,8 @@ internal class ConnectionManager : IConnectionManager {
         }
 
         private Dictionary<string, IConnection> CreateConnectionList() {
+            _shell.AssertIsOnMainThread();
```
AlexanderSher: `CreateConnectionList` is called from `ConnectionManager` constructor which is accessed from `RInteractiveWorkflow` constructor. There is no guarantee that it is called from UI thread, so this assert will eventually fail.

MikhailArkhipov: OK. The code was doing ShowMessage and such so it had to be on UI thread.

PR #2836, opened 2016-12-06
Remove UI thread-dependent code from `ConnectionManager` constructor.

AlexanderSher:
`ConnectionManager` constructor is called directly from `RInteractiveWorkflow` constructor. `IRInteractiveWorkflowProvider.GetOrCreate` is accessed from many MEF importing constructors, so there is no guarantee that `ConnectionManager` instance will be created from UI.

Logic to show `Resources.NoLocalR` and launch R Client setup is moved to the `RInteractiveWorkflow.CreateVisualComponentAsync`. It is guaranteed to be called from UI thread and before REPL becomes visible, so there will be no false error messages.

@MikhailArkhipov , can you please take a look at the change?

### DIFF
--- a/src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
+++ b/src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
@@ -34,10 +34,8 @@ namespace Microsoft.R.Components.ConnectionManager.Implementation {
         private readonly DisposableBag _disposableBag;
         private readonly ConnectionStatusBarViewModel _statusBarViewModel;
         private readonly HostLoadIndicatorViewModel _hostLoadIndicatorViewModel;
-        private readonly ConcurrentDictionary<string, IConnection> _userConnections;
+        private readonly ConcurrentDictionary<string, IConnection> _connections;
         private readonly ISecurityService _securityService;
-
-        private bool _showWindowAtStartup;
 
         public bool IsConnected { get; private set; }
         public IConnection ActiveConnection { get; private set; }
@@ -66,20 +64,13 @@ namespace Microsoft.R.Components.ConnectionManager.Implementation {
 
             _sessionProvider.BrokerStateChanged += BrokerStateChanged;
             _interactiveWorkflow.ActiveWindowChanged += ActiveWindowChanged;
-            _shell.Started += OnApplicationStarted;
 
             // Get initial values
-            var userConnections = CreateConnectionList();
-            _userConnections = new ConcurrentDictionary<string, IConnection>(userConnections);
+            var connections = CreateConnectionList();
+            _connections = new ConcurrentDictionary<string, IConnection>(connections);
 
             UpdateRecentConnections(save: false);
             CompleteInitializationAsync().DoNotWait();
-        }
-
-        private void OnApplicationStarted(object sender, EventArgs e) {
-            if (_showWindowAtStartup) {
-                GetOrCreateVisualComponent().Container.Show(focus: true, immediate: false);
-            }
         }
 
         private async Task CompleteInitializationAsync() {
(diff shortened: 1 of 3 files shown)


src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -174,7 +170,18 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
...
+                    if (!showConnectionsWindow){
+                        var message = Resources.NoLocalR.FormatInvariant(Environment.NewLine + Environment.NewLine, Environment.NewLine);
+                        showConnectionsWindow = Shell.ShowMessage(message, MessageButtons.YesNo) == MessageButtons.Yes;
+                    }
+
+                    if (!showConnectionsWindow) {
+                        var installer = Shell.ExportProvider.GetExportedValue<IMicrosoftRClientInstaller>();
+                        installer.LaunchRClientSetup(Shell);
+                    } else {
+                        Connections.GetOrCreateVisualComponent().Container.Show(focus: false, immediate: false);
```
MikhailArkhipov: No, you don't want to show when there were recent connections. `showConnectionsWindow = Connections.RecentConnections.Any()`. If user closed the window, it should stay closed. Example: single local R.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
```diff
@@ -253,28 +242,9 @@ internal class ConnectionManager : IConnectionManager {
...
-                    } else {
-                        // If VS is still stating delay until complete.
-                        _showWindowAtStartup = true;
-                    }
-                    return connections;
-                }
-
-                // No connections, may be first use or connections were removed.
-                // Add local connections so there is at least something available.
-                foreach (var e in localEngines) {
```
MikhailArkhipov: Where did auto-population with local Rs go?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -93,10 +93,6 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
                 .Add(Connections);
         }
 
-        private void OnApplicationTerminating(object sender, EventArgs e) {
```
MikhailArkhipov: nvm, i see it got moved

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
MikhailArkhipov: I guess it may be OK if user uninstalled R they used to use, but then we should pick up another one automatically.

MikhailArkhipov:
LGTM except I can't find that fragment

```
               // No connections, may be first use or connections were removed.		
               // Add local connections so there is at least something available.		
                foreach (var e in localEngines) {		
                    var c = CreateConnection(e.Name, e.InstallPath, string.Empty, isUserCreated: false);		
                    connections[e.Name] = c;		
                }
```

That auto-populated connection window with local Rs.

src/R/Components/Impl/ConnectionManager/Implementation/ConnectionManager.cs
AlexanderSher: We have this code in the beginning of the method:
```cs
            // Add newly installed engines
            foreach (var e in localEngines) {
                if (!connections.Values.Any(x => x.Path.PathEquals(e.InstallPath))) {
                    connections[e.Name] = CreateConnection(e.Name, e.InstallPath, string.Empty, isUserCreated: false);
                }
            }
```

PR #2841, opened 2016-12-07
Fix connection manager bugs

### LINKED ISSUE #2805: Saving unchanged connection marks it unconnected in Workspaces window
- Add remote connection
- Connect to it
- Click "edit"
- Click "save"

Connection is marked as not connected in Workspaces. Clicking on "connect" doesn't do anything.

Bonus:
- Click "edit" again
- Change the name

New connection is added while connection with the old name is still there.

### LINKED ISSUE #2815: Cannot edit an existing Remote R Workspace entry
I'm trying to click on the gear to edit an existing entry in R Remote Workspace window. Sometimes the edit controls appear but then disappear right away.

### LINKED ISSUE #2823: Cannot edit workspace item after creating it
Repro steps:
1. Click add to create works space item.
2. Type some URL and hit save.
3. Edit the URL for the same item again and hit save

Expected:
The item must have updated URL.

Actual:
The item remains unchanged.

### LINKED ISSUE #2829: After changing port in URL connection is deleted
Add new remote connection
Type URL without port such as `https://machine`
Save connection - port defaults to 443
Edit connection and change port to 5444
Save

Observed: connection is removed.

### LINKED ISSUE #2831: Auto-fill of connection name works oddly
Add new remote connection
Type URL without https:// 
- Connection name auto-fills with machine name
Type `:` after the machine name
- Connection name clears
Type port number
- Connection name is now the same as the port

Expected: connection name always remain as machine name unless non-empty (entered by the user)

### LINKED ISSUE #2837: SwitchToTheSameBroker test hangs cosistently
on `await Task.WhenAll(switch1Task, switch2Task);`

AlexanderSher:
- Fix #2837: SwitchToTheSameBroker test hangs cosistently
- Fix #2831: Auto-fill of connection name works oddly
- Fix #2829: After changing port in URL connection is deleted
- Fix #2823: Cannot edit workspace item after creating it
- Fix #2815: Cannot edit an existing Remote R Workspace entry
- Fix #2805: Saving unchanged connection marks it unconnected in Workspaces window


src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
```diff
@@ -206,18 +206,18 @@ internal sealed class ConnectionManagerViewModel : BindableBase, IConnectionMana
 
         public void Save(IConnectionViewModel connectionViewModel) {
             _shell.AssertIsOnMainThread();
-            if (connectionViewModel == null) {
+            if (connectionViewModel == null || !connectionViewModel.HasChanges) {
                 return;    
             }
 
-            var connection = _connectionManager.AddOrUpdateConnection(
+            _connectionManager.AddOrUpdateConnection(
```
MikhailArkhipov: Frankly in UI list with in-place editing would be simpler than the dictionary ;-) It's like 5 elements max.

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
```diff
@@ -221,19 +258,95 @@ public class UIThreadHelper {
...
+            public override void Send(SendOrPostCallback d, object state) {
+                _threadHelper._application.Dispatcher.Invoke(DispatcherPriority.Send, d, state);
+            }
+
+            public override void Post(SendOrPostCallback d, object state) {
+                var bl = _threadHelper._blockingLoop.Value;
+                if (bl != null) {
+                    bl.Post(() => d(state));
+                } else {
+                    _threadHelper._application.Dispatcher.BeginInvoke(DispatcherPriority.Send, d, state);
```
MikhailArkhipov: I thought in VS using dispatcher is a bad form? VsAppShell is using ThreadHelper.

MikhailArkhipov:
👍

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
AlexanderSher: Yeah, I'll try to do it with the next set of bug fixes.

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
AlexanderSher: We don't load VS threading assemblies in tests cause they will make tests much slower and `ThreadHelper` don't provide enough level of control. This code emulates `JoinableTaskFactory` behavior - it creates a blocking loop that can be accessed only from the code that has specific `AsyncLocal` value. `JoinableTaskFactory` does pretty much the same with much bigger amount of code and with respect to COM messages.

PR #2851, opened 2016-12-07
Fix #2849: When broker dies connection remains green

### LINKED ISSUE #2849: When broker dies connection remains green
Stop broker on remote machine. We get 'host process stopped'. Indicator is green and host load indicator is there (albeit showing last known state). Reset REPL. Now you get 'R Services is not running' but the connection is still green. And there is no 'connect' button obviously.

image

AlexanderSher:


MikhailArkhipov:
👍

Issue #2853, opened 2016-12-07
Error adding database connection


huguesv:
On VS 2015.
```
---------------------------
Microsoft Visual Studio
---------------------------
Could not load file or assembly 'Microsoft.Data.ConnectionUI.Dialog, Version=0.0.0.0, Culture=neutral, PublicKeyToken=f4ca07f51760da93' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
---------------------------
OK   
---------------------------

```

MikhailArkhipov:
Referencing versioned one

Issue #2843, opened 2016-12-07
Failure to set `AfterHostStarted` working directory stops RHost


AlexanderSher:
On my laptop I have (for some unknown reason) two user directories: `alsher` and `alsher.REDMOND`. The first one is not accessible without admin permissions and isn't really used. When I switch to local, RTVS peeks `alsher` as a user directory and since R has no access to it, `setwd` fails:
```
Error in setwd("C:/Users/alsher/Documents") : 
  cannot change working directory

   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions.ThrowOnError(String expression, REvaluationResult res) in D:\GitHub\RTVS\src\Host\Client\Impl\Definitions\IRExpressionEvaluator.cs:line 146
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions.<ExecuteAsync>d__3.MoveNext() in D:\GitHub\RTVS\src\Host\Client\Impl\Definitions\IRExpressionEvaluator.cs:line 117
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Microsoft.R.Host.Client.Session.RSession.<AfterHostStarted>d__112.MoveNext() in D:\GitHub\RTVS\src\Host\Client\Impl\Session\RSession.cs:line 412
R Host process is stopped. Click "Reset" to start a new one.
```

MikhailArkhipov:
I thought we added check for the directory existence. But I guess it also needs to handle access denied.

MikhailArkhipov:
`setwd` should be fixed.

PR #2857, opened 2016-12-08
#2843, #2806

MikhailArkhipov:
#2843 Failure to set `AfterHostStarted` working directory stops RHost
- add ACL check on the directory

#2806 Still occasionally get package load failures
- store provider rather than create workflow when editor instantiates keyboard processor

#2853 Error adding database connection
- reference signed and versioned binaries from lib

Update System.Collections.Immutable to 1.3.0


src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -95,5 +97,33 @@ public static class IOExtensions {
             }
             return Path.DirectorySeparatorChar.ToString();
         }
+
+        public static bool HasWritePermission(this string path) {
```
MikhailArkhipov: OK. I wanted stricter check since writing generally assumes reading (in common sense case anyway)

MikhailArkhipov: OK. I wanted stricter check since writing generally assumes reading (in common sense case anyway)

int19h: I don't think that is the case for Windows ACLs, at least - I just tried, and you can deny read and allow write to the same user at the same time, and it actually enforces it in that manner.

MikhailArkhipov: Yes, it is possible in general case since ACLs may apply to, say, pipes.

MikhailArkhipov: I'll leave writing check method for now in case we need it

Issue #2854, opened 2016-12-07
Session (host) information is not reported with some remote


MikhailArkhipov:
On my VM host load indicator works, but about host does not. Host Information command doesn't work either.

![image](https://cloud.githubusercontent.com/assets/12820357/20989501/b5ea3b0a-bc89-11e6-90a4-e450df59e9be.png)

MikhailArkhipov:
Seems to be specific to VMs.

MikhailArkhipov:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.R.Host.Broker.About.AboutController.GetVideoControllerInformation(AboutHost& a)
   at Microsoft.R.Host.Broker.About.AboutController.Get()
```

AlexanderSher:
Lack of permissions?

MikhailArkhipov:
No, it looks like on VM `ManagementObjectSearcher.Get("AdapterRAM")` returns null.

PR #2860, opened 2016-12-08
#2854 Session (host) information is not reported with VMs

MikhailArkhipov:


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -410,7 +410,11 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
 
                     var wd = startupInfo.WorkingDirectory;
                     if (!IsRemote && !string.IsNullOrEmpty(wd) && wd.HasReadPermissions()) {
-                        await evaluation.SetWorkingDirectoryAsync(wd);
+                        try {
+                            await evaluation.SetWorkingDirectoryAsync(wd);
+                        } catch(REvaluationException) {
+                            await evaluation.SetDefaultWorkingDirectoryAsync();
```
AlexanderSher: Is it really `REvaluationException`?

MikhailArkhipov:
It was 

```
Error in setwd("C:/Users/alsher/Documents") : 
  cannot change working directory
```

I think it translates to REval. I added access check, but hardening it more

src/Host/Client/Impl/Session/RSession.cs
MikhailArkhipov: `cannot change working directory` comes as REval?

MikhailArkhipov:
OK, I need to build this along with earlier merge to verify SQL issue is fixed

src/Host/Client/Impl/Session/RSession.cs
int19h: The only thing that can come from` EvaluateAsync` (which is what it ultimately does) is `REvaluationException` and `OperationCanceledException` - if there's something else, it's a bug in and of itself.

Issue #2812, opened 2016-12-02
users need a way to delete sensitive files on remote machine


smortaz:
either in the docs or a button that basically does a clean up of remote temp files (plots, files, etc.) for security purposes.

PR #2862, opened 2016-12-08
User Profile cleanup for remote machines.

karthiknadig:
Adds a feature to support #2812


src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -59,6 +60,41 @@ ILogger<Process> hostOutputLogger
             }
         }
 
+        public void CloseAndBlockSessionsCreationForUser(IIdentity user) {
+            lock (_sessions) {
```
int19h: I'd rather use a single lock for both collections, to reduce the potential for wrong order deadlock bugs. It would lock for longer than it has to, but session creation is not an operation that happens frequently, so this shouldn't be a perf concern.

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -59,6 +60,41 @@ ILogger<Process> hostOutputLogger
...
+                lock (_blockedUsers) {
+                    var userSessions = GetOrCreateSessionList(user);
+                    var sessions = userSessions.ToArray();
+                    foreach (var session in sessions) {
+                        try {
+                            userSessions.Remove(session);
+                            Task.Run(() => session.KillHost()).DoNotWait();
+                        } catch (Exception) { }
```
int19h: Is this meant to catch exceptions from `KillHost`? If so, it won't, because the latter is in a separate task that's not awaited. You need to put the handler inside the lambda, or use `SilenceException()` on the task.

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -59,6 +60,41 @@ ILogger<Process> hostOutputLogger
...
+
+                        session.State = SessionState.Terminated;
+                    }
+
+                    if (!_blockedUsers.Contains(user.Name)) {
```
int19h: You don't need to check, it's always safe to add to a hashset (`Add` will return `false` if element was already there).

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -59,6 +60,41 @@ ILogger<Process> hostOutputLogger
...
+                        _blockedUsers.Add(user.Name);
+                    }
+                }
+            }
+        }
+
+        public void UnblockSessionCreationForUser(IIdentity user) {
+            lock (_blockedUsers) {
+                if (_blockedUsers.Contains(user.Name)) {
```
int19h: Ditto.

src/Host/Broker/Impl/Sessions/SessionsController.cs
```diff
@@ -35,6 +35,10 @@ public class SessionsController : Controller {
                 return Task.FromResult<IActionResult>(new ApiErrorResult(BrokerApiError.NoRInterpreters));
             }
 
+            if (_sessionManager.IsSessionCreationBlockedForUser(User.Identity)) {
```
int19h: This checking should really be in `SessionManager`.

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -59,6 +60,41 @@ ILogger<Process> hostOutputLogger
...
+                    _blockedUsers.Remove(user.Name);
+                }
+            }
+        }
+
+        public bool IsSessionCreationBlockedForUser(IIdentity user) {
```
int19h: There's no way to use this API safely - by the time it returns, it has released the lock, and the return value might already be invalid (if session creation was blocked in the meantime).

I don't think this needs to be exposed at all, actually. Just lock and make the check directly in `CreateSession`, and throw if it was denied. But the crucial point is that when session is created, it should take the lock, check, and if the check passed, then hold the lock until it finishes session creation (so that any code trying to come and remove profile would have to wait until that completes).

This will also be easier to do if `_sessions` and `_blockedUsers` share a single lock, since `CreateSession` already acquires the former for most of its body.

src/Host/Broker/Impl/UserProfile/DeleteProfileController.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Broker.UserProfile {
+    [Authorize(Policy = Policies.RUser)]
+    [Route("/deleteuser")]
```
int19h: Per REST, the route itself should just be `/user` or something similar (maybe `/profile`? we might have `/users` later as an admin endpoint for user management) - the action is represented as HTTP method. We might add an HTTP GET endpoint here later for things like user querying info about his profile.

src/Host/Broker/Impl/UserProfile/DeleteProfileController.cs
```diff
@@ -0,0 +1,62 @@
...
+    public class DeleteProfileController : Controller {
```
int19h: And so this should be `ProfileController` (or `UserContoller` - either way, matching the route name).

src/Host/Broker/Impl/Sessions/SessionManager.cs
```diff
@@ -59,6 +60,41 @@ ILogger<Process> hostOutputLogger
             }
         }
 
+        public void CloseAndBlockSessionsCreationForUser(IIdentity user) {
```
int19h: It might be better to make this return an `IDisposable`, such that disposing it would unblock, to enforce the pairing. And hide `Unblock` as a public method completely.

Also, perhaps make it optional to kill exising sessions? Something like `BlockSessionCreationForUser(..., bool terminateExisting)`?

src/Host/Broker/Impl/UserProfile/DeleteProfileController.cs
```diff
@@ -0,0 +1,62 @@
...
+                result = await _userProfileManager.DeleteProfileAsync(RUserProfileServiceRequest.Create(username.ToString(), domain.ToString(), password), cts.Token);
+            } finally {
+                _sessionManager.UnblockSessionCreationForUser(User.Identity);
+            }
+
+            if(result.Error == 0) {
+                return Ok();
+            }
+
+            return BadRequest(result.Error.ToString());
```
int19h: We should probably propagate the full error info to the client (via `ApiErrorResult` and friends, like `SessionsController` does), so that it can be checked there, and rich error reporting done if something goes wrong.  As it is, if I understand correctly, this will just display the Win32 GetLastError code, as a standalone number, without any further explanations?

src/Host/Broker/Impl/UserProfile/UserProfileManager.cs
```diff
@@ -0,0 +1,52 @@
...
+        public Task<RUserProfileServiceResponse> DeleteProfileAsync(RUserProfileServiceRequest request, CancellationToken ct) {
+            return ProfileWorkerAsync(NamedPipeServerStreamFactory.DeletorName, Resources.Error_ProfileDeletionFailedIO, request, ct);
+        }
+
+        private async Task<RUserProfileServiceResponse> ProfileWorkerAsync(string name, string log, RUserProfileServiceRequest request, CancellationToken ct) {
+            using (NamedPipeClientStream client = new NamedPipeClientStream(name)) {
+                try {
+                    await client.ConnectAsync(ct);
+
+                    string jsonReq = JsonConvert.SerializeObject(request);
```
int19h: You can serialize directly to the stream if you wrap it into a `TextWriter`, via `JsonSerializer.Serialize` (and ditto for deserialization below with `TextWriter`, although you may need to add another helper to `Json` there). Then you don't need to deal with buffers, and it'll also properly handle the case when response is larger than 1K.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -95,6 +95,19 @@ internal abstract class BrokerClient : IBrokerClient {
             }
         }
 
+        public async Task<bool> DeleteRemoteUserProfileAsync(CancellationToken cancellationToken) {
```
int19h: It's not remote if it's called on a local broker :) I think this should just be `DeleteProfileAsync`.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -95,6 +95,19 @@ internal abstract class BrokerClient : IBrokerClient {
...
+            try {
+                if (HttpClient != null) {
+                    var response = await HttpClient.DeleteAsync("deleteuser", cancellationToken);
```
int19h: There should be a service class for it, similar to `SessionsWebService`. This endpoint requires authentication, so you want the code that is there to deal with it to be involved. It will also support `BrokerApiError` correctly, and throw the corresponding exception, which can then be caught here (like in `ConnectAsync` below).

src/Host/Protocol/Impl/RUserProfile/RUserProfileServiceResponse.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.R.Host.Protocol {
+    public class RUserProfileServiceResponse : IUserProfileServiceResult {
+        [JsonConstructor]
+        private RUserProfileServiceResponse() { }
+
+        public uint Error { get; set; }
+        public bool ProfileExists { get; set; }
+        public string ProfilePath { get; set; }
+
+        public static RUserProfileServiceResponse Blank => new RUserProfileServiceResponse() { Error = 13, ProfileExists = false, ProfilePath = string.Empty };
```
int19h: `readonly`

src/Host/Protocol/Impl/RUserProfile/RUserProfileServiceResponse.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public static RUserProfileServiceResponse Create(uint error, bool profileExists, string profilePath) {
```
int19h: This looks like it could be a constructor?

src/Host/Protocol/Impl/RUserProfile/RUserProfileServices.cs
```diff
@@ -69,7 +75,7 @@ public class RUserProfileCreator {
             }
         }
 
-        private static async Task CreateProfileHandleUserCredentails(IUserProfileServices userProfileService, Stream stream,  CancellationToken ct, ILogger logger = null) {
+        private static async Task HandleUserCredentailsAsync(Stream stream, Func<IUserCredentials, ILogger, IUserProfileServiceResult> operation, CancellationToken ct, ILogger logger = null) {
```
int19h: (spelling) Credentials

src/Host/Protocol/Impl/RUserProfile/RUserProfileServices.cs
```diff
@@ -12,16 +12,22 @@
 using Microsoft.Common.Core.Json;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
+using System;
 
 namespace Microsoft.R.Host.Protocol {
-    public class RUserProfileCreator {
+    public class RUserProfileServices {
```
int19h: This class, as well as the one below, should not be in the `Protocol` assembly. `Protocol` is only for interfaces and DTOs that are shared between the broker, the client, and the user profile service, and the associated infrastructure like JSON converters.  These two belong to `R.Host.UserProfile`.

src/Host/Protocol/Impl/RUserProfile/RUserProfileServicesImpl.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Security.Principal;
+using System.Text;
+using System.Runtime.InteropServices;
+using Microsoft.Common.Core.OS;
+using Microsoft.Common.Core.Security;
+using Microsoft.Extensions.Logging;
+using static Microsoft.R.Host.Protocol.NativeMethods;
+
+namespace Microsoft.R.Host.Protocol {
+    class RUserProfileServicesImpl : IUserProfileServices {
```
int19h: I think it would be better to call this class `RUserProfileServices` (since it implements the corresponding interface), and rename the class above. For the latter, so far as I can see, its main purpose is to accept and parse a named pipe request, and delegate to this class for actual execution. So the other class should be named something like `RUserProfileServer`.

int19h: Or actually that is also wrong, since the request loop is not there. It would be something like `RUserProfileRequestHandler`.

src/Host/Protocol/Impl/RUserProfile/RUserProfileServices.cs
int19h: The name seems to be misleading in general - I don't see any code in the body that deals with credentials? It just deserializes request, invokes the operation, and serializes response.

src/Host/Broker/Impl/UserProfile/UserProfileManager.cs
karthiknadig: This might not be possible. I tried using JsonSerializer. I seems to disconnect the client after writing.

---
In reply to: [91633320](https://github.com/Microsoft/RTVS/pull/2862#discussion_r91633320) [](ancestors = 91633320)

int19h: It calls `Dispose` on the underlying `StreamWriter`, which will in turn call the same on the underlying `Stream`. But `StreamWriter` has a constructor that lets you tell it to leave the stream alone - try that:

https://msdn.microsoft.com/EN-US/library/gg712853(v=VS.110,d=hv.2).aspx

int19h:
👍

src/Host/Broker/Impl/UserProfile/UserProfileManager.cs
karthiknadig: That works but still no dice. Stream writer is able to keep the connection alive with the `leaveOpen` flag set. But the reader during de-serialization does not wait like ReadAsync does. It instead always reads null (it does not wait for the server to flush before it reads). I will revisit this during the broker as service refactor.

PR #2865, opened 2016-12-09
Roaming settings

MikhailArkhipov:
+ some reference fixes.


src/Package/Impl/Options/R/RToolsOptionsPage.cs
```diff
@@ -232,7 +235,10 @@ class SettingsHolder {
                 _dict[name] = value;
             }
 
-            public void Apply() => _settings.SetProperties(_dict);
+            public void Apply() {
+                _settings.SetProperties(_dict);
+                ((IRPersistentSettings)_settings).SaveSettingsAsync().DoNotWait();
```
MikhailArkhipov: I guess... Point was to limit widely exposed functionality, sort of like IWpfTextView/ITextView. Maybe do it the other way around: derive IRPersistentSettings from IRToolsSettings and export it.

MikhailArkhipov: Cleaned it up

AlexanderSher:
👍

PR #2868, opened 2016-12-09
Add cancellations to data transfer operations

MikhailArkhipov:
There is more in plot manager etc - next time.

### DIFF
--- a/src/Common/Core/Impl/Extensions/StreamExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/StreamExtensions.cs
@@ -3,25 +3,26 @@
 
 using System;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Common.Core {
     public static class StreamExtensions {
-        public static async Task CopyAndFlushAsync(this Stream source, Stream destination, IProgress<long> progress = null) {
-                await source.CopyToAsync(destination, progress);
+        public static async Task CopyAndFlushAsync(this Stream source, Stream destination, CancellationToken cancellationToken, IProgress<long> progress = null) {
+                await source.CopyToAsync(destination, cancellationToken, progress);
                 await destination.FlushAsync();
         }
 
-        public static async Task CopyToAsync(this Stream source, Stream destination, IProgress<long> progress) {
+        public static async Task CopyToAsync(this Stream source, Stream destination, CancellationToken cancellationToken, IProgress<long> progress) {
             byte[] buffer = new byte[81920];
             int bytesRead = 0;
             long bytesTotal = 0;
-            while ((bytesRead = await source.ReadAsync(buffer, 0, buffer.Length)) > 0) {
-                await destination.WriteAsync(buffer, 0, bytesRead);
+            while ((bytesRead = await source.ReadAsync(buffer, 0, buffer.Length, cancellationToken)) > 0) {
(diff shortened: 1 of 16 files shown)


src/Common/Core/Impl/Extensions/StreamExtensions.cs
```diff
@@ -3,25 +3,26 @@
...
 using System;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Common.Core {
     public static class StreamExtensions {
-        public static async Task CopyAndFlushAsync(this Stream source, Stream destination, IProgress<long> progress = null) {
-                await source.CopyToAsync(destination, progress);
+        public static async Task CopyAndFlushAsync(this Stream source, Stream destination, CancellationToken cancellationToken, IProgress<long> progress = null) {
```
AlexanderSher: Usually in public methods it is a last parameter and has a default value `default(CancellationToken)`. Do we want to follow this pattern?

src/Common/Core/Impl/Extensions/StreamExtensions.cs
```diff
@@ -3,25 +3,26 @@
...
+                await source.CopyToAsync(destination, cancellationToken, progress);
                 await destination.FlushAsync();
```
AlexanderSher: You can pass `cancellationToken` here as well

src/Common/Core/Impl/Extensions/StreamExtensions.cs
MikhailArkhipov: I know, but there are two more defaults and I actually wanted to make ct required.

src/Common/Core/Impl/Extensions/StreamExtensions.cs
MikhailArkhipov: OK

src/Common/Core/Impl/Extensions/StreamExtensions.cs
AlexanderSher: Ok. What about `CopyToAsync`? It doesn't have defaults.

MikhailArkhipov: Sure, makes sense

AlexanderSher: Actually, I'd rather remove default `null` value for `progress` here. By default, we want to provide feedback about the long operation, and user should explicitly opt-out from that feedback.

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -73,7 +73,7 @@ public class RemoteUriWebService : IRemoteUriWebService {
                         SetResponseHeaders(response, context.Response, localBaseUrl, remoteBaseUrl);
                         using (Stream respStream = response.GetResponseStream())
                         using (Stream outStream = context.Response.OutputStream) {
-                            await respStream.CopyAndFlushAsync(outStream);
+                            await respStream.CopyAndFlushAsync(outStream, ct);
```
AlexanderSher: Async call inside finally block? @int19h , @karthiknadig , is it really required?

MikhailArkhipov: Sure thing

AlexanderSher:
👍

PR #2874, opened 2016-12-09
Fixes credentials dialog isn't modal

### LINKED ISSUE #2871: Credentials dialog isn't modal
When it appears, user can switch back to VS without closing the dialog and interact with UI

karthiknadig:
Fixes #2871

### DIFF
--- a/src/Package/Impl/Shell/VsAppShell.cs
+++ b/src/Package/Impl/Shell/VsAppShell.cs
@@ -84,13 +84,12 @@ namespace Microsoft.VisualStudio.R.Package.Shell {
             _compositionService = componentModel.DefaultCompositionService;
             _exportProvider = componentModel.DefaultExportProvider;
 
-            _appConstants.Initialize();
-
             CheckVsStarted();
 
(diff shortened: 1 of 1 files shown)


src/Package/Impl/Shell/VsAppShell.cs
```diff
@@ -88,6 +86,7 @@ public sealed class VsAppShell : IApplicationShell, IMainThread, IIdleTimeServic
         }
 
         private void OnApplicationStarted(object sender, EventArgs e) {
+            _appConstants.Initialize();
```
MikhailArkhipov: This only comes when VS was still starting. You need to add this to

```
           if (value is bool) {
                if ((bool)value) {
                    Started?.Invoke(this, EventArgs.Empty);
```

karthiknadig: I did not see `CheckVsStarted` till i rebased. will update.

karthiknadig:
@MikhailArkhipov does this look good. I will merge if there are no more comments.

MikhailArkhipov:
👍

Issue #2879, opened 2016-12-13
Select() brings signature help on lm.ridge


MikhailArkhipov:
Not sure about the case since select() is not in the intellisense list

![image](https://cloud.githubusercontent.com/assets/12820357/21151136/3ff7baee-c116-11e6-8a60-4ac5ed4767c8.png)

MikhailArkhipov:
```
\name{lm.ridge}\alias{lm.ridge}\alias{plot.ridgelm}\alias{print.ridgelm}\alias{select}\alias{select.ridgelm}\keyword{models}\description{
Fit a linear model by ridge regression.
```
The signature is correct. However, we should probably be showing actual typed alias name rather than the primary.

int19h:
Isn't this the wrong `select`? I thought this was about the one from `dplyr`?

int19h:
Which, speaking of - what happens when there are several packages installed that offer the same function name? Execution-wise, you get the one that you've attached last (e.g. via `library`). What about Intellisense?

MikhailArkhipov:
Currently completion search picks first in the list of packages, unrelated to the attachment list.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/c9fd701c9712d2b8e7f674578c78b15ad398d3c7

PR #2884, opened 2016-12-14
#2878, #2879, test fixes

### LINKED ISSUE #2878: Snippets are broken in REPL

MikhailArkhipov:
#2878 Snippets are broken in REPL
#2879 Select() brings signature help on lm.ridge

Test fixes
Switch to direct references to CPS since there is no nuget that matches U3 exactly anymore


src/ProjectSystem/Test/Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.Test.csproj
```diff
@@ -33,6 +33,13 @@
     <IntermediateOutputPath>$(BaseIntermediateOutputPath)\$(Configuration)\$(AssemblyName)\</IntermediateOutputPath>
     <OutputPath>$(BaseOutputPath)\$(Configuration)\</OutputPath>
   </PropertyGroup>
+  <ItemGroup Condition="'$(VisualStudioVersion)'=='14.0'" >
```
int19h: Why not reference these directly from VS install folder? That way we won't have to keep checking the new ones in. And if a new VS version breaks things again, you'll find out during build.

src/R/Editor/Test/QuickInfo/QuickInfoSourceTest.cs
```diff
@@ -71,7 +71,7 @@ public class FunctionIndexTest : FunctionIndexBasedTest {
 
             applicableSpan.Should().NotBeNull();
             quickInfoContent.Should().ContainSingle()
-                .Which.ToString().Should().StartWith("as.Date(x, ...)");
+                .Which.ToString().Should().StartWith("as.Date.character(x, ...)");
```
int19h: Looks like the comment at the beginning of the test also needs updating.

src/R/Support/Test/Functions/FunctionInfoTest.cs
```diff
@@ -62,13 +64,25 @@ public class FunctionInfoTest : IAsyncLifetime {
...
             signature.Should().Be("eval(expr, envir = parent.frame(), enclos = if(is.list(envir) || is.pairlist(envir)) parent.frame() else baseenv(), n)");
             locusPoints.Should().Equal(5, 11, 35, 116, 117);
         }
+
+        [Test]
+        public async Task AliasesTest() {
+            var functionInfo = await PackageIndexUtility.GetFunctionInfoAsync(_functionIndex, "select");
+
+            functionInfo.Should().NotBeNull();
+            functionInfo.Name.Should().Be("lm.ridge");
```
int19h: We need to check with the user who reported it that this is what they actually expect. I suspect that a lot more people use `dplyr` than the package where this comes from - we might have to special-case that for now (and come up with some long term plan to fix it in general).

src/ProjectSystem/Test/Microsoft.VisualStudio.ProjectSystem.FileSystemMirroring.Test.csproj
MikhailArkhipov: Primarily to make sure we reference proper version since installed VS may or may not be appropriate (i.e. we may have Update N but target Update N-x for customers.

int19h:
👍

MikhailArkhipov:
I may have relatively simple solution to the duplicate names, will PR separately

Issue #2887, opened 2016-12-14
Handle non-unique names in completion as R does


MikhailArkhipov:
For `select` we need to be able to fetch correct description depending if `MASS` or `dplyr` was loaded last.

Hong-Revo:
This would apply to any situation where the same object is exported by different packages. dplyr just happens to be a notorious offender.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/52cf566ee7726979fa6a15cfd414cab922b25a0a

PR #2890, opened 2016-12-15
#2887 Handle non-unique names in completion as R does

MikhailArkhipov:
- store list of packages per function rather than single entry
- if package name is unknown and function has single associated package use it
- if function name is associated with multiple packages fetch package name from the session


src/R/Support/Impl/Help/IntelliSenseRSession.cs
```diff
@@ -36,10 +40,32 @@ public sealed class IntelliSenseRSession : IIntellisenseRSession {
...
+                session = _workflow.RSession;
+            } else if (_coreShell.IsUnitTestEnvironment) {
+                session = Session;
+            }
+
+            if (session != null && session.IsHostRunning) {
+                try {
+                    packageName = await session.EvaluateAsync<string>(
+                        Invariant(
+                            $"if(exists({functionName.ToRStringLiteral()})) getPackageName(environment({functionName})) else NULL"
```
int19h: `environment(f)` will not necessarily give a correct result, because it is the environment that was active when the closure was created, which is not necessarily the same as the package from which the function is exported. It should work most of the time, but things that will fail are stuff like:
```r
f <- function() {
  return(function() { ... })
}

g <- f()
```
In this case, `environment(g)` will return the local environment inside `f` when it was called, rather than whatever namespace the binding `g` is in.

What you actually want here is the `find` function, that takes an identifier, and gives you the environments on the search path that contains a binding for that identifier. Like so:
```r
find("cat", mode = "function")[1]
```
(You need `[1]` because it will return a vector of multiple elements if several things on the search path contain that identifier. You also no longer need `exists`, since you'll just get an empty vector from `find` in that case.)
 
Also, if the environment is not actually a package, `getPackageName` will return ".GlobalEnv". Because that is not a package name, it can't then be used with `::` - which, so far as I can see, the rest of the code will try to do.  So that needs a check also.

MikhailArkhipov: Thanks. Yes, I know my variant doesn't work when user overrides well known name i.e. `lm.fit <- function() { }` but it seemed to be a bit exotic. However, I'll try `find`.

int19h:
👍

PR #2893, opened 2016-12-15
Fix  #2828: Renaming project in RTVS breaks it.

### LINKED ISSUE #2828: Renaming project in RTVS breaks it.
Renaming project in RTVS breaks it, because [projectName].rxproj content is not updated. 

**Steps to reproduce:**
1. Create new project called "test"
2. In VS rename project to "test-2"
3. Save changes and reopen project.
4. Project files do not show in the Solution Explorer and its not possible to add new items into the project.

**Fix:**
1. Open test-2.rxproj in text editor.
2. As you can see, this line does not reflect the new name:
`<Import Project="test.InMemory.Targets" Condition="Exists('test.InMemory.Targets')" />`
3. Update the project name to new one:
`<Import Project="test-2.InMemory.Targets" Condition="Exists('test-2.InMemory.Targets')" />`

AlexanderSher:


AlexanderSher:
Not sure if we should rename *.rproj file as well

MikhailArkhipov:
👍

MikhailArkhipov:
Probably yes. I think C#/C++ do that?

AlexanderSher:
I don't know if there is anything like that in C#/C++. We don't really use *.rproj except for creating *.rxproj in a first place. On the other hand, if user still wants to open project in RStudio, old path in recent projects won't work after renaming. Project folders aren't renamed as well. So I'll leave it as it is for now.

PR #2896, opened 2016-12-15
Addresses the issue with help getting stuck on remote.

karthiknadig:
Added more logging and info text to diagnose this better.

### DIFF
--- a/src/Host/Client/Impl/BrokerServices/WebServer.cs
+++ b/src/Host/Client/Impl/BrokerServices/WebServer.cs
@@ -2,26 +2,38 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Common.Core;
+using Microsoft.Common.Core.Logging;
+using Microsoft.Common.Core.Services;
 using static System.FormattableString;
 
 namespace Microsoft.R.Host.Client.BrokerServices {
     public class WebServer {
-        private static Dictionary<int, WebServer> Servers { get; } = new Dictionary<int, WebServer>();
+        private static ConcurrentDictionary<int, WebServer> Servers { get; } = new ConcurrentDictionary<int, WebServer>();
 
         private readonly IRemoteUriWebService _remoteUriService;
+        private readonly ICoreServices _services;
+        private readonly IConsole _console;
+        private readonly string _name;
+
         private HttpListener _listener;
+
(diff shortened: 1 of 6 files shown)


src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -51,62 +65,86 @@ public class WebServer {
...
+            WebServer server;
+            if(!Servers.TryGetValue(remoteUri.Port, out server)) {
+                server = new WebServer(remoteUri.Host, remoteUri.Port, baseAddress, name, services, console);
                 await server.InitializeAsync(ct);
-                Servers.Add(remoteUri.Port, server);
+                Servers.TryAdd(remoteUri.Port, server);
             }
 
-            server.DoWorkAsync(CancellationToken.None).DoNotWait();
+            server.DoWorkAsync().DoNotWait();
```
MikhailArkhipov: Should be passing incoming `ct` I think

MikhailArkhipov:
👍

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -30,14 +42,16 @@ public class WebServer {
         }
 
         public Task InitializeAsync(CancellationToken ct) {
-            return Task.Run(() => {
+            return Task.Run(async () => {
```
int19h: `await TaskUtilities.SwitchToBackgroundThread()`.

int19h:
👍

src/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -51,62 +65,86 @@ public class WebServer {
...
                         throw new OperationCanceledException();
                     }
                     break;
                 }
+
+                try {
+                    await Log.WriteLineAsync(LogVerbosity.Minimal, MessageCategory.General, Resources.Info_RemoteWebServerStarted.FormatInvariant(_name, LocalHost, LocalPort));
+                    _console.WriteLine(Resources.Info_RemoteWebServerStarted.FormatInvariant(_name, LocalHost, LocalPort));
+                    _console.WriteLine(Resources.Info_RemoteWebServerDetails.FormatInvariant(Environment.MachineName, LocalHost, LocalPort,_name, RemoteHost, RemotePort));
+                } catch {
```
AlexanderSher: What kind of failure do you expect here?

karthiknadig: Closing VS while listener just as listener is about to finish finding an available port. My concern was that it might lead to using a invalid console.

PR #2900, opened 2016-12-16
Fix connection manager bugs

### LINKED ISSUE #2807: No way to switch between connections with identical path
Case 1:
- Create 2 identical connections with just different name
- Connect to first one
- Click "Connect" on the second one

Nothing happens. While not trying to reconnect to the same connection is correct, user should be either notified about that or active connection should change do the new one.

Case 2:
- Create 2 identical connections with different name and different command line arguments
- Connect to first one
- Click "Connect" on the second one

Nothing happens. These connections are different, so we should switch and update command line parameters for the newly started sessions

### LINKED ISSUE #2845: When connection is deleted it is not really disconnected
Deleted connection to the server (while connected). Operation on REPL says disconnected, but title of the window still lists the URL and attemts to reset REPL actually tries to reconnect.

image

### LINKED ISSUE #2898: Package manager does not update after uninstall
I installed 'shiny' in REPL. Opened package manager, searched for 'shiny' hit uninstall. After the package was removed, package manager did not update. I verified that the package was removed.

AlexanderSher:
- Fix #2807: No way to switch between connections with identical path
- Fix #2845: When connection is deleted it is not really disconnected
- Fix #2898: Package manager does not update after uninstall


src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -46,17 +49,20 @@ internal abstract class BrokerClient : IBrokerClient {
...
+        protected BrokerClient(string name, Uri brokerUri, string rCommandLineArguments, string interpreterId, ICredentialsDecorator credentials, IActionLog log, IConsole console) {
             Name = name;
             Uri = brokerUri;
             Log = log;
 
+            RCommandLineArguments = rCommandLineArguments?.Trim() ?? string.Empty;
             _interpreterId = interpreterId;
             _credentials = credentials;
             _console = console;
+            _parametersHash = CalculateAdditionalParametersHash();
```
AlexanderSher: `_parametersHash` is required to distinguish different connections from the same user to the same machine for different parameters (command line arguments, interpreter, etc.).

MikhailArkhipov: Theoretically it is possible to override GetHashString on the connection and compare those.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -109,20 +115,21 @@ internal abstract class BrokerClient : IBrokerClient {
             DisposableBag.ThrowIfDisposed();
             await TaskUtilities.SwitchToBackgroundThread();
 
+            var sessionName = connectionInfo.Name + _parametersHash;
```
MikhailArkhipov: Make small `GetUniqueName`?

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -136,14 +143,42 @@ internal abstract class BrokerClient : IBrokerClient {
         protected virtual Task<Exception> HandleHttpRequestExceptionAsync(HttpRequestException exception)
             => Task.FromResult<Exception>(new RHostDisconnectedException(Resources.Error_HostNotResponding.FormatInvariant(Name, exception.Message), exception));
 
+        private string CalculateAdditionalParametersHash() {
+            if (string.IsNullOrEmpty(RCommandLineArguments)) {
+                return string.Empty;
+            }
+
+            var bytes = Encoding.Unicode.GetBytes(RCommandLineArguments);
```
MikhailArkhipov: `RCommandLineArguments.GetSHA512Hash()`

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -136,14 +143,42 @@ internal abstract class BrokerClient : IBrokerClient {
...
+            }
+
+            var hashCharsLength = (int)(hashBytes.Length * 4.0d / 3.0d);
+            if (hashCharsLength % 4 != 0) {
+                hashCharsLength += 4 - hashCharsLength % 4;
+            }
+
+            var hashChars = new char[hashCharsLength];
+
+            Convert.ToBase64CharArray(hashBytes, 0, hashBytes.Length, hashChars, 0);
```
MikhailArkhipov: This might be another useful extension next to `GetSHA512Hash`

src/Host/Client/Impl/Host/IBrokerClient.cs
```diff
@@ -11,6 +11,7 @@ public interface IBrokerClient : IDisposable {
         string Name { get; }
```
MikhailArkhipov: Starts looking a lot like `ConnectionInfo`

src/Host/Client/Impl/Host/BrokerClient.cs
MikhailArkhipov: The other thing we could just add Guid to `ConnectionInfo` and use it (and persist in settings) and distinguish connections by guids.

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -239,10 +242,10 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
             TaskUtilities.AssertIsOnBackgroundThread();
 
             _callback = callback;
-            _startupInfo = startupInfo;
+            _startupInfo = startupInfo ?? new RHostStartupInfo();
             RHost host;
             try {
-                var connectionInfo = new BrokerConnectionInfo(startupInfo.Name, this, startupInfo.RHostCommandLineArguments, timeout);
+                var connectionInfo = new BrokerConnectionInfo(Name, this, startupInfo.UseRHostCommandLineArguments, timeout);
```
MikhailArkhipov: So why empty/null command line args are insufficient and we need a separate flag?

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -84,15 +84,15 @@ public class RSessionProvider : IRSessionProvider {
             Broker.Dispose();
         }
 
-        private RSession CreateRSession(Guid guid) {
-            var session = new RSession(Interlocked.Increment(ref _sessionCounter), Broker, _connectArwl.CreateExclusiveReaderLock(), () => DisposeSession(guid));
+        private RSession CreateRSession(string sessionId) {
```
MikhailArkhipov: If we add Guid to connection info, we can as well use it as session id

src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -145,11 +145,25 @@ public class RSessionProvider : IRSessionProvider {
...
+                }
+            }
+        }
+
+        public async Task<bool> TrySwitchBrokerAsync(string name, string path = null, string rCommandLineArguments = null, CancellationToken cancellationToken = default(CancellationToken)) {
             using (_disposeToken.Link(ref cancellationToken)) {
                 await TaskUtilities.SwitchToBackgroundThread();
 
-                var brokerClient = CreateBrokerClient(name, path, cancellationToken);
+                var brokerClient = CreateBrokerClient(name, path, rCommandLineArguments, cancellationToken);
```
MikhailArkhipov: Just pass IConnectionInfo?

MikhailArkhipov:
So basically question is: like COM objects, connections can have unique guids. This way property value won't matter. Then, session ID == connection info guid. And we can pass around connection info around.

src/Host/Client/Impl/Host/BrokerClient.cs
AlexanderSher: Sure.

src/Host/Client/Impl/Session/RSession.cs
AlexanderSher: command line args is a connection level setting. Individual sessions may or may not want to use arguments. REPL session uses args, Package Manager session doesn't need them. I'm not sure about intellisense session: command line args may contain a script that loads a package. Does it matter?

src/Host/Client/Impl/Host/BrokerClient.cs
AlexanderSher: It is already a combination of session id and `_parametersHash`. As you mentioned, `_parametersHash` is basically a connection id.

The problem with guid as a connection id is that it is random. To find matching connection, you have to either fetch data about all existing connections and find the right one, or ask broker to do it by providing all the parameters. In both cases, it requires changes in broker code and may have potential versioning issues (if we add or remove parameter). 
With hash instead of guid, it is guaranteed that for the same parameters you'll get the same id, and broker doesn't have to know anything about those parameters if they are VS or RHost specific.

src/Host/Client/Impl/Host/BrokerClient.cs
AlexanderSher: Ok

src/Host/Client/Impl/Session/RSessionProvider.cs
AlexanderSher: No, sessions are orthogonal to connections.

MikhailArkhipov: I'd say 45 degrees but OK. ;-)

MikhailArkhipov:
👍

src/Host/Client/Impl/Host/IBrokerClient.cs
AlexanderSher: Well, looks like an idea to merge these two interfaces was a bit reckless.
- IConnectionInfo has a `Path` property, which is a string. It is good for both UI and settings storage. We actually don't want to replace it with URL because we want to preserve user input after he has saved it.
- IBrokerClient has a `Uri` property, which is URI. It is good for web service operation. Also, on the level of `Host.Client` assembly we don't care about user input format. 

Equality logic is also different:
- IConnectionInfo instances are considered equal if they have identical names.
- IBrokerClient instances are considered equal if they have equal Id's.

So I'd rather not touch it for now, cause it is not that simple and also we introduce new bugs.

MikhailArkhipov: Let's leave it alone then. It is probably possible to have some sort of comparable indentifications but it may be too much of a change at this point.

PR #2911, opened 2016-12-19
#2361 and #2908

### LINKED ISSUE #2361: Workspaces window doesn't allow multiple connections that differ only by URL fragment
Try adding two remote connections like so:
- http://localhost#foo
- http://localhost#bar

The second one ends up overwriting the first. This shouldn't be happening, since fragment is a meaningful part of the URL (it designates which version of R to use, if broker has more than one configured).

### LINKED ISSUE #2908: Any # added to the server URL is immediately removed when you save
In workspaces window, Add new remote server
Try to add a # in the URL, and click save.
Result: The # and anything after is removed

The use case for this is to specify the interpreter guid on the remote machine when multiple interpreters are available. While we don't plan on documenting this, being able to specify it lets us test it and may be able to help customers on case-by-case basis.

AlexanderSher:
- Fix #2908: Any # added to the server URL is immediately removed when you save
- Fix #2361: Workspaces window doesn't allow multiple connections that differ only by URL fragment
- Fix `RInstallationTest`
- More tests for `ConnectionViewModel` and `ConnectionManagerViewModel`


MikhailArkhipov:
👍

PR #2916, opened 2016-12-20
Fix #2732: No error is displayed for failed connection if REPL was not up

### LINKED ISSUE #2732: No error is displayed for failed connection if REPL was not up
Without opening REPL try conneting to remote that is offline

- Long progress
- The dialog closes, REPL opens but there is no 'machine offline' message.

image

Try connecting again

- Now you get the error message with the reason

AlexanderSher:
Issues was caused by two bugs:
- REPL wasn't scheduled for showing up when first errors come in.
- No errors were sent to REPL if there was no RSession instances created.


MikhailArkhipov:
👍

Issue #2621, opened 2016-11-02
Dev15: interactive tests are broken


MikhailArkhipov:
Due to changes in test runner working directory code cannot find editor assemblies for the test MEF catalog

AlexanderSher:
Do we have code that relies on test runner working directory? I thought we specify assembly paths explicitly.

MikhailArkhipov:
Well, it was never recommended to figure out VS installation path so we used to load via `CommonExtensions\Editor\...` relying on the fact that runner directory is in `Common7\IDE`. This no longer works. To add insult to injury, there is no registry key for VS15 so we'd have to resort to Win32.GetModuleFileName(NULL).

MikhailArkhipov:
Need proper nuget modules for IDE references

Issue #983, opened 2016-02-24
REPL buttons appear on C# interactive window


MikhailArkhipov:
We need to assign different IDs for menu and toolbar commands and handle toolbar commands in the REPL controller and top level menu commands separately (although calling same code on invoke).

Continuation of #813

MikhailArkhipov:
Upgrading per user feedback

```
When debugging C# code (not using the RTVS tools at all), the R Interactive window pops up every time I hit F11 to step into some code.  The R Interactive window hides the Watch window.  This is pretty much a deal-breaker, and I’ll probably be uninstalling RTVS and going back to RStudio until it’s resolved.
```

MikhailArkhipov:
Dev15 only

MikhailArkhipov:
Waiting for IW references updates to RC

MikhailArkhipov:
Considering PR in IW, then integration to Dev15 then getting new build - this won't fit into this week.

MikhailArkhipov:
https://github.com/dotnet/interactive-window/pull/23

MikhailArkhipov:
IW part merged as https://github.com/dotnet/interactive-window/commit/257e2d274ee09c26a405631632ae6e3075c796f9

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/51de46d315fc0e0cc89fd3e795bbb409f39b4db1

PR #2919, opened 2016-12-20
#983 REPL buttons appear on C# interactive window 

MikhailArkhipov:
Integrate new IW toolbar functionality. Running requires D15Prerel 12/15+
Also partial fixes for #2621 Dev15: interactive tests are broken (actual issue is that public nugets for VS IDE are old)


MikhailArkhipov:
Actually need to finish VSCT conditionals

src/UnitTests/Core/Impl/Paths.cs
```diff
@@ -3,22 +3,43 @@
...
     public static class Paths {
-        private static Lazy<string> VsRootLazy { get; } = Lazy.Create(() => (string)Registry.GetValue(Invariant($"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\VisualStudio\\{Toolset.Version}"), "InstallDir", string.Empty));
-        private static Lazy<string> VsPrivateAssembliesLazy { get; } = Lazy.Create(() => Path.Combine(VsRootLazy.Value, @"PrivateAssemblies\"));
-        private static Lazy<string> VsCommonExtensionsLazy { get; } = Lazy.Create(() => Path.Combine(VsRootLazy.Value, @"CommonExtensions\"));
+        private static string _vsRoot;
+        public static string VsRoot {
+            get {
+                if (_vsRoot == null) {
+#if VS14
+                    _vsRoot = (string)Registry.GetValue(Invariant($"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\VisualStudio\\{Toolset.Version}"), "InstallDir", string.Empty);
```
AlexanderSher: We should either use `Lazy` or wrap it with `lock`, cause this field is accessed in tests from multiple threads.
I'd prefer `Lazy`.

PR #2923, opened 2016-12-21
4 bug fixes + tests

### LINKED ISSUE #2811: Give warning that session will be terminated when user tries connect to another R
Otherwise users may think they can go local while remote is still doing calculations.

### LINKED ISSUE #2918: Creating connection with same name and URL bu different arguments de-sync indicators
Connect to remote
Try creating connection with same name and URL but different arguments
Save

Remote disconnects, but status bar is still green

image

### LINKED ISSUE #2920: Display warning dialog with "don't show me this again" checkbox when switching Workspaces.
We were going to do this as part of the larger work around simplifying the semantics of switching behavior, but we didn't implement all of that feature. This was lost somewhere along the way.

From: Hong Ooi's mail:

"There should be a message/warning somewhere that switching workspaces means your old workspace will be closed. Maybe a separate dialog box with the option “don’t show this again” so it won’t get too disruptive."

### LINKED ISSUE #2921: Not loading last used local workspace on startup
From a mail with Hong Ooi:

1.	Start Visual Studio, where MRC 3.3.2 is selected automatically
2.	Manually switch to MRS 3.3.2 (and presumably typed some commands into the REPL or run some code)
3.	Restart Visual Studio, and MRC 3.3.2 is selected again after restarting

AlexanderSher:
- Fix #2920: Display warning dialog with "don't show me this again" сheckbox when switching Workspaces.
- Fix #2921: Not loading last used local workspace on startup
- Fix #2918: Creating connection with same name and URL bu different arguments de-sync indicators
- Fix #2811: Give warning that session will be terminated when user tries connect to another R

### DIFF
--- a/src/Common/Core/Impl/Enums/YesNo.cs
+++ b/src/Common/Core/Impl/Enums/YesNo.cs
@@ -1,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core.Enums {
+    public enum YesNo {
+        Yes,
+        No
+    }
+}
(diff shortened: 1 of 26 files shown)


MikhailArkhipov:
👍

PR #2928, opened 2016-12-23
Three bug fixes

### LINKED ISSUE #2790: Simplifying the connection switching experience
There are three things that we need to do to simplify the connection switching experience:

1. Remote the test connection button from the Workspaces window
1. Extend the duration of the modal dialog box during connection establishment to not return until all secondary sessions (e.g., IntelliSense) are established
1. Remove the transactional nature of switching connections. We first terminate the existing session (with prompt to user to save state) before establishing a new connection. Yes, this introduces a new state "no connection established" into the process in the case where a new connection cannot be established. This applies to both local-local, local-remote and remote-local switches.

This issue references #2643 as well as #2736 which will be closed when this issue is complete.

### LINKED ISSUE #2917: Remote HTTP server messages need to be fixed
image

Machine address is incorrect (shows localhost instead of the actual IP)
'HTTP server allows you to connect' is probably not quite right. Maybe `HTTP server transfers data from remote machine...'

### LINKED ISSUE #2927: Provide better error message when machine is unreachable
```
Connecting to R Workspace failed.
Reason: Machine 'hostname' did not respond to a ping.
The machine may be offline or the network has been disconnected.
```
is a bit misleading. We’ve first failed to reach the web service, then have tried to ping the host, and ping also has failed.

AlexanderSher:
- Fix #2927: Provide better error message when machine is unreachable
- Fix #2917: Remote HTTP server messages need to be fixed
- Fix #2790: Simplifying the connection switching experience (test connection button hidden)

### DIFF
--- a/src/Host/Client/Impl/BrokerServices/WebServer.cs
+++ b/src/Host/Client/Impl/BrokerServices/WebServer.cs
@@ -16,6 +16,7 @@ namespace Microsoft.R.Host.Client.BrokerServices {
         private static ConcurrentDictionary<int, WebServer> Servers { get; } = new ConcurrentDictionary<int, WebServer>();
 
         private readonly IRemoteUriWebService _remoteUriService;
+        private readonly string _baseAddress;
         private readonly ICoreServices _services;
         private readonly IConsole _console;
         private readonly string _name;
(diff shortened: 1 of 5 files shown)


MikhailArkhipov:
👍

PR #2930, opened 2016-12-23
#2929 Simplify selection of remote Rs 

MikhailArkhipov:
- Change guilds to ints
- Output interpreter names and IDs
- Provide guidance message to the user

### DIFF
--- a/src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
+++ b/src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
@@ -35,7 +35,7 @@ namespace Microsoft.R.Host.Broker.Interpreters {
 
             var sb = new StringBuilder(Invariant($"{Interpreters.Count} interpreters configured:"));
             foreach (var interp in Interpreters) {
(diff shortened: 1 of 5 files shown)


MikhailArkhipov:
For review only so far. Not merging just yet.

MikhailArkhipov:
@int19h - please access the risk.

image

image

image

src/Host/Broker/Impl/Interpreters/InterpreterManager.cs
```diff
@@ -19,6 +19,7 @@ public class InterpreterManager {
         private readonly ROptions _options;
         private readonly ILogger _logger;
         private IFileSystem _fs;
+        private int _intepretedId = 1;
```
int19h: `_nextInterpreterId`?

But given that `Initialize` is only supposed to be called once, and overwriters `Interpreters` entirely, it could be a local variable inside that method.

int19h:
I consider this very low risk, and safe to take for the closest available milestone.

PR #2935, opened 2016-12-28
User-friendly URL correction

MikhailArkhipov:


src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -248,42 +251,47 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
...
+            }
 
-                    if (hasPort) {
-                        components |= UriComponents.StrongPort;
-                    }
+            if (uri != null && !string.IsNullOrEmpty(uri.Host)) {
+                var hasPort = uri.Port >= 0 && (!uri.IsDefaultPort || path.IndexOfOrdinal(":443") > 0);
+                var port = hasPort ? uri.Port : DefaultPort;
+                var hasPathOrQuery = !string.IsNullOrEmpty(uri.PathAndQuery) && uri.PathAndQuery != "/";
+                var mainPart = Invariant($"{Uri.UriSchemeHttps}{Uri.SchemeDelimiter}{uri.Host.ToLower()}:{port}");
```
int19h: `UriBuilder` (so that it can do all the proper escaping etc).

int19h: Also, can you make it so that it preserves the input scheme in Debug builds? With this auto-correct, it will be impossible to connect to non-SSL servers, and having to set up SSL makes reproing remote stuff really complicated (which is why we allow no SSL in the broker for Debug).

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -248,42 +251,47 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
...
+                try {
+                    Uri.TryCreate(path, UriKind.Absolute, out uri);
+                } catch (InvalidOperationException) { } catch (ArgumentException) { } catch (UriFormatException) { }
+            }
 
-                    if (hasPort) {
-                        components |= UriComponents.StrongPort;
-                    }
+            if (uri != null && !string.IsNullOrEmpty(uri.Host)) {
+                var hasPort = uri.Port >= 0 && (!uri.IsDefaultPort || path.IndexOfOrdinal(":443") > 0);
```
int19h: Why check for anything other than `IsDefaultPort`? If they used `:port`, it should be false, and if they omitted it, it should be true, so all other checks are redundant.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
MikhailArkhipov: OK

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
MikhailArkhipov: Actually if port is 443 `IsDefaultPort` is true. I hit that in tests.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
MikhailArkhipov: However, tests will fail in debug then :-(

MikhailArkhipov: I see why @AlexanderSher didnt use `UriBuilder` - it add `/` at the end which then needs to be trimmed. So I have to leave manual construction in place.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
int19h: Indeed, you're right, `new Uri("http://localhost:443").IsDefaultPort` is true.

However, you can check `uri.Port == 443` for that case, instead of a substring check.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
int19h: We can `#ifdef` those tests out for Debug then? 

Regarding `UriBuilder`, what's wrong with adding `/` - why does it need to be trimmed? The URLs `http://foo` and `http://foo/` should be identical in all respects. In fact, the latter is a canonical form, and as soon as you put that string into an `Uri` instance, it will also add `/` - e.g. try `new Uri("http://foo").ToString()`.

MikhailArkhipov: So we don't get `host/#123` when we append fragment

MikhailArkhipov: wrt test, we can do that but that will reduce test coverage since they are rarely ran in release mode, most folks run in debug only.

int19h:
👍

PR #2941, opened 2016-12-29
Workspaces UI icons

### LINKED ISSUE #2940: Incorrect connection status display after terminating R
Connect to Local
Session -> Terminate R
Connect to remote

- The status is yellow. should be green.

Reset REPL - now it is green.

MikhailArkhipov:
Control with overlays
Hide URL for now to reduce clutter
Fix #2940 Incorrect connection status display after terminating R 
- sync issue since broker changed and session connection events fire from different threads


MikhailArkhipov:
image

Issue #2948, opened 2017-01-03
NRE is markdown region highlighter


MikhailArkhipov:
From user logs
```
System.NullReferenceException: Object reference not set to an instance of an object.
at Microsoft.Markdown.Editor.Classification.CodeBackgroundTextAdornment.ShouldHighlightEntireLine(IWpfTextViewLine line)
at Microsoft.Markdown.Editor.Classification.CodeBackgroundTextAdornment.CreateVisuals(IWpfTextViewLine line)
at Microsoft.Markdown.Editor.Classification.CodeBackgroundTextAdornment.ProcessLine(ITextViewLine line)
at Microsoft.Markdown.Editor.Classification.CodeBackgroundTextAdornment.OnLayoutChanged(Object sender, TextViewLayoutChangedEventArgs e)
at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.RaiseEvent[TArgs](Object sender, EventHandler`1 eventHandlers, TArgs args)
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/184e299a926b811d100edb3984dfe4835a880101

Issue #2931, opened 2016-12-23
Check remote version and output message if client is different


MikhailArkhipov:
Build number is irrelevant but we need to pop up Yes/No if server is 0.N and client is 0.X.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/184e299a926b811d100edb3984dfe4835a880101

Issue #2850, opened 2016-12-07
Need link to instructions page for the remote setup


MikhailArkhipov:
Currently menu item in R Tools | Remote takes you to microsoft.com

MikhailArkhipov:
Putting `http://aka.ms/rtvs-remote` in code

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/184e299a926b811d100edb3984dfe4835a880101

PR #2949, opened 2017-01-03
#2850 #2931 #2948

MikhailArkhipov:
#2948 NRE is markdown region highlighter 
#2931 Check remote version and output message if client is different 
#2850 Need link to instructions page for the remote setup


int19h:
👍

PR #2950, opened 2017-01-03
Fixes the issue where sent to remote did not apply the filter.

karthiknadig:
Applies file filter while sending files using "send to remote" command. Exception to this rule is when the user selects file(s) to be sent, then that/those file(s) will be sent always.

### DIFF
--- a/src/Common/Core/Impl/Extensions/MatcherExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/MatcherExtensions.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.FileSystemGlobbing;
+
+namespace Microsoft.Common.Core {
+    public static class MatcherExtensions {
+        public static IEnumerable<string> GetMatchedFiles(this Matcher matcher,  IEnumerable<string> dirPaths) {
+            List<string> files = new List<string>();
+            foreach(string dir in dirPaths) {
+                files.AddRange(matcher.GetResultsInFullPath(dir));
+            }
+
+            return files.Distinct();
+        }
+    }
+}
--- a/src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
+++ b/src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
@@ -58,19 +58,14 @@ namespace Microsoft.VisualStudio.R.Package.ProjectSystem.Commands {
             string projectDir = Path.GetDirectoryName(_configuredProject.UnconfiguredProject.FullPath);
 
             string fileFilterString = await properties.GetFileFilterAsync();
-            Matcher matcher = new Matcher(StringComparison.InvariantCultureIgnoreCase);
+            Matcher matcher = new Matcher(StringComparison.OrdinalIgnoreCase);
             matcher.AddIncludePatterns(fileFilterString.Split(new string[] { ";" }, StringSplitOptions.RemoveEmptyEntries));
 
             List<string> filteredFiles = new List<string>();
             filteredFiles.AddRange(matcher.GetMatchedFiles(nodes.GetAllFolderPaths(_configuredProject.UnconfiguredProject)));
 
-            foreach (IProjectTree node in nodes) {
-                if (node.IsFile()) {
-                    // Add any file that user specifically selected. 
-                    // This can contain a file ignored by the filter.
-                    filteredFiles.Add(node.FilePath);
-                } 
-            }
+            // Add any file that user specifically selected. This can contain a file ignored by the filter.
+            filteredFiles.AddRange(nodes.Where(n => n.IsFile()).Select(n => n.FilePath));
 
             string projectName = properties.GetProjectName();
             string remotePath = await properties.GetRemoteProjectPathAsync();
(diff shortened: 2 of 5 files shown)


src/Common/Core/Impl/Extensions/MatcherExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.FileSystemGlobbing;
+
+namespace Microsoft.Common.Core {
+    public static class MatcherExtensions {
+        public static IEnumerable<string> GetMatchedFiles(this Matcher matcher,  IEnumerable<string> dirPaths) {
+            List<string> files = new List<string>();
+            foreach(string dir in dirPaths) {
+                files.AddRange(matcher.GetResultsInFullPath(dir));
```
MikhailArkhipov: files.AddRange(dirPaths.Select(p => matcher.GetResultsInFullPath(p))) ?

src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
```diff
@@ -52,10 +56,28 @@ internal sealed class SendToRemoteCommand : SendFileCommandBase, IAsyncCommandGr
 
             var properties = _configuredProject.Services.ExportProvider.GetExportedValue<ProjectProperties>();
             string projectDir = Path.GetDirectoryName(_configuredProject.UnconfiguredProject.FullPath);
+
+            string fileFilterString = await properties.GetFileFilterAsync();
+            Matcher matcher = new Matcher(StringComparison.InvariantCultureIgnoreCase);
```
MikhailArkhipov: Might be current culture? Not sure about file name comparo

src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
```diff
@@ -52,10 +56,28 @@ internal sealed class SendToRemoteCommand : SendFileCommandBase, IAsyncCommandGr
...
+            matcher.AddIncludePatterns(fileFilterString.Split(new string[] { ";" }, StringSplitOptions.RemoveEmptyEntries));
+
+            List<string> filteredFiles = new List<string>();
+            filteredFiles.AddRange(matcher.GetMatchedFiles(nodes.GetAllFolderPaths(_configuredProject.UnconfiguredProject)));
+
+            foreach (IProjectTree node in nodes) {
+                if (node.IsFile()) {
+                    // Add any file that user specifically selected. 
+                    // This can contain a file ignored by the filter.
+                    filteredFiles.Add(node.FilePath);
```
MikhailArkhipov: filteredFiles.AddRange(nodes.Where(n => n.IsFile()).Select(n => n.FilePath))

src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
```diff
@@ -52,10 +56,28 @@ internal sealed class SendToRemoteCommand : SendFileCommandBase, IAsyncCommandGr
...
+                } 
+            }
+
             string projectName = properties.GetProjectName();
             string remotePath = await properties.GetRemoteProjectPathAsync();
 
-            await SendToRemoteAsync(nodes.GetAllFilePaths(), projectDir, projectName, remotePath, CancellationToken.None);
+            if(filteredFiles.Count > 0) {
+                await SendToRemoteAsync(filteredFiles.Distinct(), projectDir, projectName, remotePath, CancellationToken.None);
```
MikhailArkhipov: So there is no progress and operation is not cancellable. What if it is a huge project?

src/ProjectSystem/Impl/Extensions/ProjectTreeExtensions.cs
```diff
@@ -85,5 +85,25 @@ public static class ProjectTreeExtensions {
             }
             return string.Empty;
         }
+
+        public static IEnumerable<string> GetAllFolderPaths(this IEnumerable<IProjectTree> nodes, UnconfiguredProject unconfiguredProject) {
+            List<string> paths = new List<string>();
```
MikhailArkhipov: You may want try-catch IOExceptions here

MikhailArkhipov: And UnautorizedAccessException

src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
int19h: For file names it's normally `OrdinalIgnoreCase`.

src/Common/Core/Impl/Extensions/MatcherExtensions.cs
int19h: Might as well just do `Select(...).Distinct().ToArray()` and return that directly.

karthiknadig: `GetResultsInFullPath` returns `IEnumerable<string>`. This will require flattening at some point, before returning.

src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
karthiknadig: `SendToRemoteAsync` shows status and progress via statusbar.

src/ProjectSystem/Impl/Extensions/ProjectTreeExtensions.cs
karthiknadig: Other than `Directory.Exisits` I don't actually access any directories here. `Directory.Exisits`, returns `false` in case of any error while trying to determine if the directory exists. So, I don't think this is necessary.

src/Package/Impl/ProjectSystem/Commands/SendToRemoteCommand.cs
karthiknadig: #2961 to track this.

---
In reply to: [94520413](https://github.com/Microsoft/RTVS/pull/2950#discussion_r94520413) [](ancestors = 94520413)

MikhailArkhipov:
👍

Issue #2954, opened 2017-01-03
Can't connect to custom local interpreter


huguesv:
In Workspace window, click Add.
Click on the browse button next to the path text box, and pick the same folder R Client\RSERVER folder as in the picture below:
![image](https://cloud.githubusercontent.com/assets/1696845/21625279/c41a3a18-d1bf-11e6-9c60-83bf8898a4e0.png)
Save and click connect.  Progress bar appears and it never connects.
If you cancel VS hangs. Dump here: \\pytools\r\Bugs\2954

MikhailArkhipov:
Reproes for me as well.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/16f3e664e51f2f88e8ad7220eed50bcff3358709

PR #2958, opened 2017-01-04
#2954 Can't connect to custom local interpreter

MikhailArkhipov:
- Pipe must be opened in async mode for the WaitForConnectionAsync to be cancellable. Otherwise token is ignored.
- We need to cancel on either process exit or user clicks cancel
- Actual issue is that broker derives log name from the connection name. Hence we need to make sure the name is reasonably clean.
- Also fixed the issue when typing bad connection names crashed VS

### DIFF
--- a/src/Mocks/VisualStudio/project.15.0.json
+++ b/src/Mocks/VisualStudio/project.15.0.json
@@ -5,7 +5,7 @@
     "Microsoft.VisualStudio.Editor": "15.0.25824-RC",
     "Microsoft.VisualStudio.ImageCatalog": "15.0.25824-RC",
     "Microsoft.VisualStudio.ProjectSystem": "15.0.594-pre",
-    "Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime": "14.3.25407",
+    "Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime": "15.0.25726-Preview5",
     "Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime": "15.0.25726-Preview5",
     "Microsoft.VisualStudio.SDK.EmbedInteropTypes": "14.1.4",
     "Microsoft.VisualStudio.Text.Data": "15.0.25901-RC",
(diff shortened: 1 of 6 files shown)


src/Mocks/VisualStudio/project.15.0.json
```diff
@@ -5,7 +5,7 @@
     "Microsoft.VisualStudio.Editor": "15.0.25824-RC",
     "Microsoft.VisualStudio.ImageCatalog": "15.0.25824-RC",
     "Microsoft.VisualStudio.ProjectSystem": "15.0.594-pre",
-    "Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime": "14.3.25407",
```
int19h: Probably wrong.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
```diff
@@ -192,7 +194,7 @@ internal sealed class ConnectionViewModel : BindableBase, IConnectionViewModel {
 
             Uri uri;
             var isPathValid = Uri.TryCreate(Path, UriKind.Absolute, out uri);
-            if (string.IsNullOrEmpty(Name)) {
+            if (!IsValidConnectionName(Name)) {
```
int19h: We should also validate this in broker side, and reject invalid names in some sensible way (i.e. as API error).

src/Mocks/VisualStudio/project.15.0.json
MikhailArkhipov: Weird. I have `    "Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime": "15.0.25726-Preview5"` locally...

MikhailArkhipov: That's 15.0 and what do you know, `15.0.25726-Preview5` is the latest. go figure.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionViewModel.cs
MikhailArkhipov: It fails to start the process, throws in n`ew FileLoggerProvider(_startupOptions.Name)` so there is no connection to report the API error.

int19h:
👍

PR #2960, opened 2017-01-04
Public R Host API to .NET

MikhailArkhipov:
Product code changes are minimal. Includes nuget packaging necessary assemblies (broker, host and friends). Local only.


src/Host/API/Impl/Converter.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.Common.Core;
+using static System.FormattableString;
+
+namespace Microsoft.R.Host.Client {
+    public static class Converter {
+        public static List<T> ToListOf<T>(this IEnumerable<object> e) {
+            return new List<T>(e.Select(x => (T)Convert.ChangeType(x, typeof(T))));
```
AlexanderSher: `=> e.Select(x => (T)Convert.ChangeType(x, typeof(T))).ToList();`

src/Host/API/Impl/Definitions/IRHostSession.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// To return collections use <see cref="GetListAsync"/> and <see cref="GetDataFrameAsync"/>
+        /// </typeparam>
+        /// <param name="expression">Expression or block of R code to execute</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>The variable or expression value</returns>
+        /// <exception cref="ArgumentException" />
+        /// <exception cref="REvaluationException" />
+        /// <exception cref="OperationCanceledException" />
+        /// <exception cref="RHostDisconnectedException" />
+        Task<T> EvaluateAsync<T>(string expression, CancellationToken cancellationToken = default(CancellationToken));
```
AlexanderSher: @int19h , do we want `EvaluateAsync` in a separate interface?

src/Host/API/Impl/Definitions/IRHostSession.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// Invokes R function. 
+        /// </summary>
+        /// <param name="function">Function name</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <param name="args">Function arguments</param>
+        /// <exception cref="ArgumentException" />
+        /// <exception cref="REvaluationException" />
+        /// <exception cref="OperationCanceledException" />
+        /// <exception cref="RHostDisconnectedException" />
+        Task InvokeAsync(string function, CancellationToken cancellationToken = default(CancellationToken), params object[] args);
```
AlexanderSher: What's the difference between this method and a previous one?

src/Host/API/Impl/Definitions/IRHostSession.cs
MikhailArkhipov: For simplicity I don't want user to jump through multiple interfaces.

src/Host/API/Impl/Definitions/IRHostSession.cs
MikhailArkhipov: `function` is a name (i.e no arguments). Yes, one can do `EvalAsync("c(1:10)") but how to pass arguments from C#? `InvokeAsync` constructs R call from function name and arguments passed as .NET types. I.e. InvokeAsync("x", new List<int>() {1,2,3})` turns into `EvalAsync("x(c(1, 2, 3))")`

src/Host/API/Impl/Definitions/IRHostSession.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// Retrieves list of specific type from R variable
+        /// </summary>
+        /// <param name="expression">Expression (variable name) to fetch as list</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>List of values of the provided type</returns>
+        /// <exception cref="ArgumentException" />
+        /// <exception cref="REvaluationException" />
+        /// <exception cref="OperationCanceledException" />
+        /// <exception cref="RHostDisconnectedException" />
+        Task<List<T>> GetListAsync<T>(string expression, CancellationToken cancellationToken = default(CancellationToken));
```
AlexanderSher: Is there a difference between calling this method and `EvaluateAsync<List<T>>`?

int19h:
I suggest doing some high-level design discussions about what the API should look like overall, before committing to an implementation. Given that this would be a stable API that we will commit to maintain for a long time, it's not something that should be rushed.

This PR closely resembles our internal API right now, but I think that for the high-level layer, we can do a lot simpler, especially if we take proper advantage of `dynamic`.

Most clients shouldn't care about things like callbacks at all. The single most common scenario is evaluating an R expression and fetching back data in format that is readily usable by R. It would be great if we could automatically wire up the defaults (auto-detecting which R to use etc), such that something like this would work:
```cs
var r = new R();
int scalar = r("1 + 2"); // returns "dynamic", C# takes care of casting
DataTable tbl1 = r["read.csv"]("foo.csv"); // method calls translate to R function calls
DataTable tbl2 = r.head(iris, n = 5); // for function names that are valid C# identifiers, .name can be used instead of ["name"]; also note that named arguments etc all work as intended
```
some more advanced stuff:
```
dynamic iris = r("iris");
// Because we didn't convert to DataTable but left it as dynamic, tbl still corresponds to original R object - so we can do further R calls on it
int nrows = r.nrows(iris);
string[] species = iris["Species"];
// Format-style splicing of R values into expressions - takes care of literal escaping etc
DataTable setosa = r("{0}[{0}['Species'] == {1}]", iris, "Setosa");
// Same thing with named arguments
DataTable setosa = r("{iris}[{iris}['Species'] == {species}]", iris: iris, species: "Setosa");
```
I have some more ideas on how this all could work in more detail.

It might also be beneficial to enlist Dino, since his experience with IronPython, and dealing with the impedance mismatch between .NET and Python, would be relevant in this context. He could also tell us what is and isn't possible with `dynamic` (but so far as I know, everything I suggested above is possible).

MikhailArkhipov:
Certainly. However, there is balance of time/effort. Also, I definitely don't want to make something like R.NET (i.e. write R code in C#). The intent is to be clear about cost of calls and actually discourage code like above (i.e. Office API dynamic style).

src/Host/API/Impl/Definitions/IRHostSessionCallback.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Implemented by the application that uses Microsoft.R.Host.Client.API
+    /// Provides facilities to respond to the R engine requests, if any.
+    /// Stub class <see cref="RHostSessionCallback"/> provides basic implementation.
+    /// </summary>
+    public interface IRHostSessionCallback {
+        /// <summary>
+        /// Displays error message in the application-specific UI
+        /// </summary>
+        Task ShowErrorMessage(string message, CancellationToken cancellationToken = default(CancellationToken));
```
AlexanderSher: `ShowErrorMessageAsync`

int19h:
I do not suggest that we do it R.NET style, where every step yields a handle to an R object. But this is orthogonal to `dynamic`. We can still do the latter, offering a much simpler API without the need to juggle several interfaces and use strings heavily, at least for simple operations - generic expression evaluation, simple function calls, and accessing variables inside environments. I'm not suggesting that we do a full-fledged mapping that allows you to do things like `a+b` etc in C#, with R semantics carrying through.

Since R is inherently a dynamic language, I think `dynamic` is very appropriate here.

int19h:
Also, for a long-term supported API that a lot of people are going to use, and changing which later is going to be a breaking change, I think we'd do well to err on the side of caution (i.e. slow, deliberative design process) here.

src/Host/API/Impl/Definitions/IRHostSession.cs
MikhailArkhipov: It explicitly forces `as.list` and converts returned `JArray` to the specified type.

src/Host/API/Impl/Definitions/IRHostSessionCallback.cs
MikhailArkhipov: OK

Issue #2952, opened 2017-01-03
Output name of the R interpreter when it is not found


MikhailArkhipov:
I mistyped URL and couldn't figure out why the remote connection fails

![image](https://cloud.githubusercontent.com/assets/12820357/21624370/9e1792ec-d1bb-11e6-83f4-4a9fbdb2186d.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5146a2a262e879e23b940bf6dce0aad0056cfb52

Issue #2937, opened 2016-12-28
After changing connection URL message is misleading


MikhailArkhipov:
Yes, we disconnect correctly, but we should say so since Reset doesn't actually work here.

![image](https://cloud.githubusercontent.com/assets/12820357/21530685/5a66a20e-ccf7-11e6-9c98-6aba1c3e4896.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/5146a2a262e879e23b940bf6dce0aad0056cfb52

PR #2962, opened 2017-01-04
#2937 #2952 + hardened URL checks

MikhailArkhipov:
#2952 Output name of the R interpreter when it is not found
#2937 After changing connection URL message is misleading
Harden URL checks
Disable connect button when URL is invalid
Handle Enter better
Add bad URL check in HTTP client it case it still leaks there

### DIFF
--- a/src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
+++ b/src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Reflection;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Common.Core;
@@ -22,7 +23,7 @@ namespace Microsoft.R.Components.InteractiveWorkflow.Commands {
             _interactiveWorkflow = interactiveWorkflow;
             _console = console;
 
-            _interactiveWorkflow.RSession.Interactive += OnRSessionInteractive;
+            _interactiveWorkflow.RSession.Connected += OnRSessionConnected;
             _interactiveWorkflow.RSession.Disposed += OnRSessionDisposed;
         }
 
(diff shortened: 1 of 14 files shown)


src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
```diff
@@ -43,9 +44,15 @@ public sealed class SessionInformationCommand : IAsyncCommand {
...
+        private void OnRSessionConnected(object sender, EventArgs e) {
             if (_interactiveWorkflow.RSession.IsRemote) {
-                ReplInitComplete().ContinueWith(async (t) => await PrintBrokerInformationAsync(reportTelemetry: true)).DoNotWait();
+                // Host signals 'connected' after R loop was set up (and R banner is printed).
+                // We want to complete printing synchorously here so output does not come
+                // after user started typing in the REPL. But don't wait forever.
+                Task.Run(async () => {
+                    await ReplInitComplete();
+                    await PrintBrokerInformationAsync(reportTelemetry: true);
+                }).Wait(10000);
```
int19h: Will this actually block the REPL prompt? And if it does, doesn't that rely on the order of `Connected` events being fired?

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -32,35 +33,49 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
...
             _console = console;
             _evaluatorRequest = new CountdownDisposable();
-        }
 
-        public void Dispose() {
-            if (CurrentWindow != null) {
-                CurrentWindow.TextView.VisualElement.SizeChanged -= VisualElement_SizeChanged;
-            }
+            _disposableBag
+                .Add(() => Session.Output -= SessionOnOutput)
```
int19h: Might as well put all the event deregistration in the single `{}` block that's already there below.

MikhailArkhipov: OK. Alex tends to add them separately so I kept the style

src/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
MikhailArkhipov: I looked at R and our host code and the banner is printed in `setupRLoop` which we call (and then offer MRO), then host calls 'connected'. So by the 'connected' time the banner should be in the REPL queue. However, releasing control here will allow other interactions to happen - such as initlialization process and possible even user input.

int19h:
👍

Issue #2974, opened 2017-01-08
Can not input line break between " and other command.


Marchen:
Between the " and 'x' in the following code, hitting Return key doesn't insert line break between them and insert line break next line.
```R
test <- function() {
	for (i in 1:10) {
		"
		Test string.
		"x <- 1
	}
}


```

MikhailArkhipov:
So the caret does not move

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/c0da9ce1d309332d3a115d8d5a74b0ee3e0cdf75

Marchen:
Thanks!

PR #2989, opened 2017-01-10
#2974 Can not input line break between " and other command.

MikhailArkhipov:
The issue was twofold:

1). Caret positioning after formatting was incorrect. For example, in 

`x |<- 1`

typing ENTER produced

```
x
<|- 1
```

because caret position tracker only tracked changed range and offset from `x` was incorrect since autoformat removes trailing space. The fix is to for the tracking range to include next line do next token after the CRLF will be seen.

2). In

```
s <- 
    "
     string
    "|x <- 1
```

upon ENTER formatter attempted to format `    "` not seeing that it is inside the string.


AlexanderSher:
👍

Issue #2990, opened 2017-01-10
Credentials dialog should have meaningful caption


MikhailArkhipov:
`devenv.exe` is kinda odd :-)

![image](https://cloud.githubusercontent.com/assets/12820357/21827144/30b1a6f0-d73f-11e6-94f0-b4bdafdd2c62.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/c30317e2343da479acc89a665c72a7abe8fb236a

PR #2992, opened 2017-01-10
Enable non-generic login prompt.

karthiknadig:


MikhailArkhipov:
👍

MikhailArkhipov:
Can you also fix #2990 same PR

karthiknadig:
Will do.

karthiknadig:
For now it shows all the available options for the logon prompt. Pin pass will not work, user will need to select the username password option and enter it there. I have yet to figure out a clean way to unpack the pin-pass data. There is no indication that the packed data i receive from the prompt is username-pass or pin-pass.

MikhailArkhipov:
👍

Issue #2968, opened 2017-01-06
Missing R Host produces error that is not informative


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/21728679/6941f81c-d3fc-11e6-80da-4c86c1454c79.png)

For local we should not say 'on machine ABC' and rather say something like 'Installation of R is missing or corrupted at `path`' and check for the R Host existence (which we used to to in 0.5 and earlier)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/541aabfcf429f850d4a153c9e75c6e63e8cf4d5b

PR #2993, opened 2017-01-10
#2968 Missing R Host produces error that is not informative

MikhailArkhipov:


src/Host/Client/Test/Session/RSessionTest.cs
```diff
@@ -183,7 +183,7 @@ public partial class RSessionTest : IDisposable {
             var session = new RSession(0, _testMethod.Name, brokerClient, new AsyncReaderWriterLock().CreateExclusiveReaderLock(), () => { });
             Func<Task> start = () => session.StartHostAsync(new RHostStartupInfo(), null, 10000);
 
-            start.ShouldThrow<RHostBrokerBinaryMissingException>();
+            start.ShouldThrow<ComponentBinaryMissingException>();
```
AlexanderSher: Should it be `RHostBinaryMissingException` in this case?

MikhailArkhipov: Yes

src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -283,9 +283,11 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
             // From R docs:  Valid values are 10...10000 with default normally 80.
             _terminalWidth = Math.Max(10, Math.Min(10000, width));
 
-            Session.OptionsSetWidthAsync(_terminalWidth)
-                .SilenceException<RException>()
-                .DoNotWait();
+            if (Session.IsHostRunning) {
+                Session.OptionsSetWidthAsync(_terminalWidth)
```
AlexanderSher: `Session.IsHostRunning` may change right after you've checked it. We should add `SilenceException<RHostDisconnectedException>()` to the `OptionsSetWidthAsync` call instead.

MikhailArkhipov: That think is extension to `IRExpressionEvaluator` so it doesn't have that information. I guess I can just leave this alone

AlexanderSher:
👍

PR #2997, opened 2017-01-11
Show username-password prompt by default

karthiknadig:
Instead of the device-pin that currently does not work.


MikhailArkhipov:
👍

PR #2998, opened 2017-01-12
Fix #2956: Switching connection using status bar doesn't warn that session will be terminated

### LINKED ISSUE #2956: Switching connection using status bar doesn't warn that session will be terminated
Like you get if you switch using the workspace window.

AlexanderSher:


MikhailArkhipov:
👍

Issue #2613, opened 2016-11-01
VS shows error message when typing Tab in comments


Marchen:
Typing Tab key after in comments  produce following error message.

>Index and length must refer to a location within the string.
>Parameter name: length

![comment-tab](https://cloud.githubusercontent.com/assets/3863577/19876619/12da91ce-a01b-11e6-9580-cd0a6599c2cd.PNG)

Repro:
1. Open VS.
2. Make new R script.
3. Type #.
4. Hit Tab.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/698b187013402962ac7ff35ccdf9ae4d321a1869

Marchen:
Thanks!

Marchen:
This error message is still shown in my environment (RTVS 0.6 with VS 2015 Update 3).
Please reopen the issue.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/bb09d540badb5b43d02b6966880cb1315abf35f2

Issue #2569, opened 2016-10-25
Settings page, edit db connection string button missing


huguesv:
We used to have a button on the right side of the text box when it's a database connection string, and that brings up the database connection dialog.

![image](https://cloud.githubusercontent.com/assets/1696845/19703620/57f4ddbc-9ab9-11e6-94da-4b1b4f004771.png)

Same issue on Win 7 and Win 10.

huguesv:
Weird, it's working on my laptop:
![image](https://cloud.githubusercontent.com/assets/1696845/19703818/1b6f411a-9aba-11e6-9a80-56d232d1a6c9.png)

huguesv:
Ran under debugger, no exceptions being thrown on Win7 VM when opening project properties/settings.

huguesv:
I installed RTVS on Win 10 and I'm seeing the same issue.  I checked on another development machine, and it works there.  So it's a bug with running on installed version versus development machine in experimental hive.

PR #3002, opened 2017-01-13
#2613

MikhailArkhipov:
#2613 VS shows error message when typing Tab in comments
Allow apostrophe in the connection name


MikhailArkhipov:
Added fix for

#2569 Settings page, edit db connection string button missing

which most probably will also fix #3003. The problem is that .NET property grid has issues loading type editors from a different assembly. The fix is to move UI editor to the package.

AlexanderSher:
👍

PR #3005, opened 2017-01-14
Broker as a service

karthiknadig:
Runs broker as a service in the remote case:
* uses **--start.as service** switch to run broker as a service, otherwise runs as normal exe
* Monitor process is no longer needed.
* No need to create RBrokerUser
* Installs Broker as a Network Service
* Removed the need to transfer username and password to user profile service. User Profile service requires the user SID only.
* Removed the Password Claim from Broker. Password is NOT needed after creating primary user token.
* User Profile service named pipe is ACL-ed to only accept connections from process with Network Service SID.


src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,102 @@
...
+                if (CreateEnvironmentBlock(out env, token, inherit)) {
+                    IntPtr ptr = env;
+                    while (true) {
+                        string envVar = Marshal.PtrToStringUni(ptr);
+                        byte[] data = Encoding.Unicode.GetBytes(envVar);
+                        if (data.Length <= 2) {
+                            // detected double null
+                            break;
+                        } else {
+                            int idx = envVar.IndexOf('=');
```
MikhailArkhipov: String.Split(env, '=') ?

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,102 @@
...
+                DestroyEnvironmentBlock(env);
+            }
+            return eb;
+        }
+
+        public byte[] ToByteArray() {
+            using (MemoryStream ms = new MemoryStream()) {
+                byte[] nulls = { 0, 0 };
+                foreach (var p in this) {
+                    string envData = $"{p.Key}={p.Value}";
```
MikhailArkhipov: Invariant

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            Marshal.Copy(data, 0, ptr, data.Length);
+
+            length = data.Length;
+            return new Win32NativeEnvironmentBlock(ptr);
+        }
+
+        public new string this[string key] {
+            get {
+                string mapped = GetMappedKey(key);
```
MikhailArkhipov: this.TryGetValue ?

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,102 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+using static Microsoft.Common.Core.NativeMethods;
+
+namespace Microsoft.Common.Core.OS {
+    public class Win32EnvironmentBlock : Dictionary<string, string> {
```
MikhailArkhipov: Does this need to be thread-safe? I'd then consider implementing IDictionary instead and having a lock over inner dictionary

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -0,0 +1,102 @@
...
+                if (!string.IsNullOrWhiteSpace(mapped)) {
+                    base[mapped] = value;
+                } else {
+                    Add(key, value);
+                }
+            }
+        }
+
+        private string GetMappedKey(string key) {
+            string low = key.ToLowerInvariant();
```
MikhailArkhipov: I think TryGetValue does this more efficiently

src/Common/Core/Impl/OS/Win32NativeEnvironmentBlock.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+using static Microsoft.Common.Core.NativeMethods;
+
+namespace Microsoft.Common.Core.OS {
+    public class Win32NativeEnvironmentBlock : IDisposable {
```
MikhailArkhipov: sealed? or there are derived classes?

src/Common/Core/Impl/OS/Win32Process.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            PROCESS_INFORMATION pi;
+            if (!CreateProcessAsUser(
+                winIdentity.Token, applicationName, commandLine, ref processAttr, ref threadAttr, true,
+                (uint)(CREATE_PROCESS_FLAGS.CREATE_UNICODE_ENVIRONMENT | CREATE_PROCESS_FLAGS.CREATE_NO_WINDOW),
+                environment.NativeEnvironmentBlock,
+                workingDirectory,
+                ref si,
+                out pi)) {
+                throw new Win32Exception(Marshal.GetLastWin32Error());
```
MikhailArkhipov: You may want to free `processAttr` and `threadAttr` if `CreateProcessAsUser` fails

src/Host/Broker/Impl/RBrokerServiceInstaller.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Configuration.Install;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.R.Host.Broker {
+    [RunInstaller(true)]
+    public partial class RBrokerServiceInstaller : System.Configuration.Install.Installer {
```
MikhailArkhipov: If it needs icon, use RFile.ico

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
karthiknadig: This mapping was needed to maintain the original casing of the environment block. TryGetValue does not seem to have a way to set the string comparer. If we don't use this mapping we can create, for example, "PATH" and "Path" as environment variables, while we add the interpreter path. Depending on which evnvariable windows picks-up, LoadLibrary can file while loading RHost since it cannot find the R.dll

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
karthiknadig: See comment below on GetMappedKey. I had to do this to preserve original case of the key. Also because Dictionary does not support  Case-insensitive comparison.

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
karthiknadig: cannot split it by '=', '=' is a valid file name character. I need to split by the first '=' only.
image

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
int19h: This should definitely implement `IDictionary` rather than deriving from it, since a bunch of stuff is shadowed with `new` rather than properly overridden. So if you upcast to `Dictionary`, you lose all that special semantics.

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
int19h: It does actually, you need to pass the corresponding `IEqualityComparer` to the constructor of `Dictionary`.

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
AlexanderSher: `new Dictionary(StringComparer.OrdinalIgnoreCase)`

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -87,96 +87,113 @@ public class Session {
...
-            if (useridentity != null && WindowsIdentity.GetCurrent().User != useridentity.User && password != null) {
-                uint error = NativeMethods.CredUIParseUserName(User.Name, username, username.Capacity, domain, domain.Capacity);
+            var usernameBldr = new StringBuilder(NativeMethods.CREDUI_MAX_USERNAME_LENGTH + 1);
+            var domainBldr = new StringBuilder(NativeMethods.CREDUI_MAX_DOMAIN_LENGTH + 1);
+
+            var shortHome = new StringBuilder(NativeMethods.MAX_PATH);
+            NativeMethods.GetShortPathName(Interpreter.Info.Path, shortHome, shortHome.Capacity);
+
+            Stream stdout, stdin;
+            if (useridentity != null && WindowsIdentity.GetCurrent().User != useridentity.User) {
```
int19h: It might be worth reusing the logic in this block for the no-identity case as well. If I understand it correctly, the only thing we'd need is to handle username/password conditionally, and use `CreateProcess` in lieu of `CreateProcessAsUser`. I think it would be better, because we'd then have only one codepath handling environment variable and pipe redirections; as it is, it may be that subtle differences or even bugs in `Process` might cause things to deviate.

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
int19h: `Split` has an overload where you can limit the number of components. If you set it to 2, it'll split on the first one only.

MikhailArkhipov: Oh I see, you are concerned with the `=` in the value.

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
AlexanderSher: Why do you need to derive from `Dictionary` in first place?

karthiknadig: I need a way to manipulate the environment block and tailor it to each user before creating a process. I decided to wrap the environment key value pair with a dictionary, this allowed me to manipulate the environment easily and convert back and fourth between native and managed environment block.

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -0,0 +1,155 @@
...
+        private static ILogger _logger;
+        private static readonly StartupOptions _startupOptions = new StartupOptions();
+        private static readonly SecurityOptions _securityOptions = new SecurityOptions();
+        private static readonly CancellationTokenSource _cts = new CancellationTokenSource();
+
+        internal static IConfigurationRoot Configuration { get; set; }
+
+        internal static CancellationToken CancellationToken => _cts.Token;
+
+        internal static void CommonStartupInit(IConfigurationRoot configuration) {
```
AlexanderSher: Why everything is static?

karthiknadig: This was stuff in Program.Main, I moved it out of there so it could be accessed from service startup.

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
AlexanderSher: Ok, but for that you don't need to implement `IDictionary` - you need a pair of `get`/`set` methods or an indexer. The fact that you use a dictionary to store key/value pairs (or maybe `ConcurrentDictionary`) isn't important for the code that uses `Win32EnvironmentBlock` and can be encapsulated.

karthiknadig: Correct.

src/Common/Core/Impl/OS/Win32NativeEnvironmentBlock.cs
```diff
@@ -0,0 +1,38 @@
...
+        public IntPtr NativeEnvironmentBlock { get; }
+
+        public Win32NativeEnvironmentBlock(IntPtr env) {
+            NativeEnvironmentBlock = env;
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing) {
```
AlexanderSher: http://blog.stephencleary.com/2009/08/third-rule-of-implementing-idisposable.html

karthiknadig: I used the Implement Disposable pattern that C# generates. What is the change you are suggesting? the pointer that is used here is obtained from AllocHGlobal when i convert the environment key-value pair back to native.

src/Host/Broker/Impl/Startup/CommonStartup.cs
AlexanderSher: Ok, how hard is it to make `CommonStartup` an instance type?

src/Common/Core/Impl/OS/Win32NativeEnvironmentBlock.cs
AlexanderSher: Can you use `SafeBuffer` instead?
https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.safebuffer.aspx

src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -40,63 +61,24 @@ public class Win32EnvironmentBlock : Dictionary<string, string> {
...
 
-        public byte[] ToByteArray() {
-            using (MemoryStream ms = new MemoryStream()) {
-                byte[] nulls = { 0, 0 };
-                foreach (var p in this) {
-                    string envData = $"{p.Key}={p.Value}";
-                    byte[] data = Encoding.Unicode.GetBytes(envData);
-                    ms.Write(data, 0, data.Length);
+        private byte[] ToByteArray() {
+            lock (_environment) {
```
int19h: This doesn't do anything useful now, since nothing else uses `lock` on that object. I'm not sure if it's even safe to do for `ConcurrentDictionary`.

If the only thing you care about is for enumeration to not throw, then you don't need locking here at all - `ConcurrentDictionary` guarantees that enumeration will succeed even if elements are added or removed concurrently. However, it doesn't guarantee anything about which elements might or might not show up. It sounds like in this case we don't actually care? So you can just ditch this.

If you actually need snapshot semantics here, the best way to achieve it is to call `ToArray` on the dictionary. It has its own implementation that will make a snapshot under internal lock.

karthiknadig: Will update. I needed a snapshot here, so ToArray is the solution.

src/Host/Broker/Impl/Startup/CommonStartup.cs
karthiknadig: I can take a work-item to convert it to an instance.

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -104,108 +109,57 @@ public class Session {
...
-                stdout = _process.StandardOutput.BaseStream;
+            _sessionLogger.LogInformation(Resources.Info_StartingRHost, Id, User.Name, rhostExePath, arguments);
+            using (Win32NativeEnvironmentBlock nativeEnv = eb.GetNativeEnvironmentBlock()) {
+                int pid = 0;
+                if (loggedOnUser) {
+                    pid = Win32Process.StartProcessAsUser(useridentity, rhostExePath, arguments, Path.GetDirectoryName(rhostExePath), nativeEnv, out stdin, out stdout, out stderror);
+                }else {
+                    pid = Win32Process.StartProcessAsUser(null, rhostExePath, arguments, Path.GetDirectoryName(rhostExePath), nativeEnv, out stdin, out stdout, out stderror);
+                }
+                _process = Process.GetProcessById(pid);
```
int19h: There is a race condition here, since the process might crash before we try to look it up. 

We can handle `ArgumentException`, but there's still a potential race if a new process starts concurrently, and reuses the same PID. So it seems that the only safe way to handle this is to not use `Process` at all.

It looks like at this point, the only things that we're actually using it for is checking if the process is still running, and registering a handler for when it exists. BTW, the latter might also be a race now, since we register the handler after starting the process, not before; I'm not sure what happens if process terminates (e.g. crashes) before we register the handler, but I wouldn't be surprised if the handler doesn't get called.

So, I think we should just bite the bullet and move completely to Win32 here for dealing with processes. Stash away the process handle, not just PID; and use `WaitOne` and `ThreadPool.RegisterWaitForSingleObject` to detect process termination via the handle.

karthiknadig: Agreed Will update.

src/Common/Core/Impl/OS/Win32Process.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            si.dwFlags = STARTF_USESTDHANDLES;
+            si.hStdInput = stdinRead;
+            si.hStdOutput = stdoutWrite;
+            si.hStdError = stderrorWrite;
+
+            SECURITY_ATTRIBUTES processAttr = CreateSecurityAttributes(winIdentity == null ? WellKnownSidType.AuthenticatedUserSid : WellKnownSidType.NetworkServiceSid);
+            SECURITY_ATTRIBUTES threadAttr = CreateSecurityAttributes(winIdentity == null ? WellKnownSidType.AuthenticatedUserSid : WellKnownSidType.NetworkServiceSid);
+
+            PROCESS_INFORMATION pi;
```
int19h: MSDN says:

> Handles in PROCESS_INFORMATION must be closed with CloseHandle when they are no longer needed.

> If the function succeeds, be sure to call the CloseHandle function to close the hProcess and hThread handles when you are finished with them. Otherwise, when the child process exits, the system cannot clean up the process structures for the child process because the parent process still has open handles to the child process. 

Given the other comment regarding the use of `Process`, I think this class just needs to become non-static, and manage the lifetime of these handles, just like `Process` does.

src/Common/Core/Impl/OS/Win32NativeEnvironmentBlock.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Common.Core.OS {
+    public sealed class Win32NativeEnvironmentBlock : SafeBuffer {
```
AlexanderSher: I meant that you may want to use `SafeBuffer` to read/write values directly. If you just need a `SafeHGlobalHandle`, `SafeHandleZeroOrMinusOneIsInvalid` as a base class is enough:

```cs
public sealed class SafeHGlobalHandle : SafeHandleZeroOrMinusOneIsInvalid {
    public int Length { get; }

    private SafeHGlobalHandle(int length) : base(true) {
        Length = length;
    }

    protected override bool ReleaseHandle() {
        Marshal.FreeHGlobal(handle);
        return true;
    }

    public static SafeHGlobalHandle Create(byte[] bytes) {
        var handle = Create(bytes.Length);
        Marshal.Copy(bytes, 0, result.DangerousGetHandle(), bytes.Length);
        return handle;
    }

    private static SafeHGlobalHandle Create(int length) {
        var handle = new SafeHGlobalHandle(length);
 
        RuntimeHelpers.PrepareConstrainedRegions();
        try { 
        } finally {
            var ptr = Marshal.AllocHGlobal(length);
            handle.SetHandle(ptr);
        }
        return handle;
    }
}
```

karthiknadig: updated.

src/Common/Core/Impl/OS/Win32Process.cs
```diff
@@ -8,11 +8,54 @@
...
     public class Win32Process {
-        public static int StartProcessAsUser(WindowsIdentity winIdentity, string applicationName, string commandLine, string workingDirectory, Win32NativeEnvironmentBlock environment, out Stream stdin, out Stream stdout, out Stream stderror) {
+        private Win32Process(PROCESS_INFORMATION pi) {
+            _hasExited = false;
+            ProcessId = pi.dwProcessId;
+            MainThreadId = pi.dwThreadId;
+            _processHandle = new SafeProcessHandle(pi.hProcess, true);
+            _threadHandle = new SafeThreadHandle(pi.hThread);
+            ProcessWaitHandle wait = new ProcessWaitHandle(_processHandle);
+            ThreadPool.RegisterWaitForSingleObject(wait, (o, t) => {
```
int19h: It's still not clear to me who does `CloseHandle(wait)`. Looking at the docs for `RegisterWaitForSingleObject`, it doesn't look like it does that?

int19h: Also, it looks like they recommend calling `RegisteredWaitHandle.Unregister` in the callback.

int19h:
👍

Issue #2976, opened 2017-01-08
Tooltip displayed for dplyr functions when dplyr not loaded


Hong-Revo:
Here's an odd glitch I ran into this week. The help for dplyr functions is displayed even when the package isn't loaded.

![image](https://cloud.githubusercontent.com/assets/11328890/21746675/e0090200-d5a1-11e6-9471-f474bc01797d.png)

I use dplyr a lot of the time, but not always. While it's nice to have the help available, I think it would be less surprising if it was only done when the package was loaded.

MikhailArkhipov:
I don't think we ever checked against loaded packages. 0.5 and before we collected function names from all installed packages by parsing HTML, 0.6 switched to fetching names from R so it would take care or multiple library locations. So function is located through packages and then, having package name we fetch data for the signature. This can be checked against loaded packages technically.

Hong-Revo:
Actually it seems a bit inconsistent. `foreach`, `select` and `gam` don't show tooltips, and `filter` shows the tooltip for the function in the stats package (not dplyr). In the case of `gam` I have both the mgcv and gam packages installed, which both export a `gam` function which might be confusing RTVS. But it seems to cope ok with `filter` which is also exported by multiple packages.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/245ae71f86247a73e526c37b16d081d3be306bbe

PR #3007, opened 2017-01-17
#2976 Tooltip displayed for dplyr functions when dplyr not loaded

MikhailArkhipov:
1. Move loaded package tracking to the intellisense host
2. Verify against loaded packages when retrieving function info
3. Add test


MikhailArkhipov:
Will reopen later

PR #3010, opened 2017-01-17
#2976 Tooltip displayed for dplyr functions when dplyr not loaded

MikhailArkhipov:
- Move loaded packages provider to intellisense host
- Verify function info against loaded packages 
- Add test
- Fix help test instability
- Fix huge R logo


src/R/Support/Impl/Help/IntelliSenseRSession.cs
```diff
@@ -37,27 +46,32 @@ public sealed class IntelliSenseRSession : IIntellisenseRSession {
         public IRSession Session { get; private set; }
 
         public void Dispose() {
+            if (_workflow?.RSession != null) {
```
AlexanderSher: `RSession` can't be null

MikhailArkhipov: It can be in tests that never started interactive

MikhailArkhipov: Or can it

AlexanderSher: `RSession` is set in ctor, so even if session wasn't started, it still shouldn't be null.

src/R/Support/Impl/Help/IntelliSenseRSession.cs
```diff
@@ -79,11 +96,54 @@ public sealed class IntelliSenseRSession : IIntellisenseRSession {
...
+
+        private void OnInteractiveSessionMutated(object sender, EventArgs e)
+             => UpdateListOfLoadedPackagesAsync().DoNotWait();
+
+        private async Task UpdateListOfLoadedPackagesAsync() {
+            try {
+                await StartSessionAsync();
+                var session = GetLoadedPackagesInspectionSession();
+                if (session != null) {
+                    var response = await session.EvaluateAsync<string>("paste0(.packages(), collapse = ' ')", REvaluationKind.Normal);
```
int19h: You should be able to just do `.packages()` and retrieve the result as JSON array of strings, without having to join it and then split it back.

int19h: (not sure if they can have spaces in their names, but if they do, it'd also deal with that)

MikhailArkhipov: OK. O don't think they ever do since otherwise unclear how would `library(name)` works as it does without quotes. I guess theoretically it might be and then one would require quotes in `library` but it doesn't seem like normal practice

MikhailArkhipov: According to https://journal.r-project.org/archive/2012-2/RJournal_2012-2_Baaaath.pdf it doesn't seem happening

int19h: Remember that you can always quote identifiers with backticks - ``library(`foo bar`)`` etc. I don't think that's the standard naming convention, but users can ignore naming conventions, so it's still potential input that can be handled.

Anyway, I think it's just simpler to treat it as an array all around, since we have the code to support marshaling it in JSON serializer - you can just do `EvaluateAsync<string[]>()` and get the values back.

MikhailArkhipov: Sure

int19h:
👍

PR #3014, opened 2017-01-18
Fixes to user profile tests.

### LINKED ISSUE #3012: UserProfileServicePipeTest.CreateProfileTest fails with UnauthorizedAccessException

karthiknadig:
Fixes #3012. The named pipe server in user-profile service only accepts connections from a network service. Since the tests run as user the connection is not authorized. Added a factory interface to create a server pipe for testing that does not require client to be a network service.


MikhailArkhipov:
👍

PR #3031, opened 2017-01-20
Fix #3030: Revert default log path to the temp folder

### LINKED ISSUE #3030: Revert default log path to the temp folder

AlexanderSher:


MikhailArkhipov:
👍 if local R works again :-)

Issue #3036, opened 2017-01-20
Remove 'Check for updates' from VS 2017


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/6fbd45c85fbd9a94903268b8bebb8b1d5f56e7e4

Issue #3034, opened 2017-01-20
Microsoft R client detection does not work


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/22166288/3509a378-df15-11e6-86dc-9cdcf5f606fd.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/6fbd45c85fbd9a94903268b8bebb8b1d5f56e7e4

Issue #3025, opened 2017-01-20
Saying NO when prompted to install R client should mean NO


huguesv:
Brand new VM, I installed RTVS, created my first project.
I get the following prompt:
![image](https://cloud.githubusercontent.com/assets/1696845/22159817/8e952458-def7-11e6-997c-520b31101be0.png)

I am 100% certain that I clicked No.

Next I see this:
![image](https://cloud.githubusercontent.com/assets/1696845/22159851/b592e39c-def7-11e6-927b-b8d0be940b49.png)

I thought this happened a week ago, but I figured I must have clicked Yes accidentally.  Now I know it wasn't my fault ;)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/6fbd45c85fbd9a94903268b8bebb8b1d5f56e7e4

Issue #3023, opened 2017-01-20
Certificate warning dialog always appears on primary monitor


MikhailArkhipov:
New Win32 implementation is not parented to VS so it always appears on the primary monitor. Need to better specify hidden form position.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/6fbd45c85fbd9a94903268b8bebb8b1d5f56e7e4

PR #3039, opened 2017-01-20
Menu, version and message box positioning fixes

### LINKED ISSUE #3022: Incorrect wording (or version comparison) with remote machine
image

Should say 0.6 is LOWER than 1.0

MikhailArkhipov:
#3022 Incorrect wording (or version comparison) with remote machine 
- use Version to compare rather than compare by fields

#3025 Saying NO when prompted to install R client should mean NO 
- fix incorrect comparison

#3034 Microsoft R client detection does not work 
- removed 'Switch to MRC' command since we support multiple Rs and there is MRO, MRO from SQL, etc

#3023 Certificate warning dialog always appears on primary monitor 
 - position message box properly and above the progress dialog

#3036 Remove 'Check for updates' from VS 2017 
 - removed check for updates and survey/news. In VS 2017+ we should be getting VS notifications and should join start page feed for news.


src/Common/Core/Impl/OS/Win32MessageBox.cs
```diff
@@ -72,8 +73,48 @@ public enum Result : uint {
...
+                return (Result)NativeMethods.MessageBox(form.Handle, message, Resources.MessageBoxTitle, (uint)flags);
             }
         }
+
+        private static Task MoveMessageBoxAsync(string title, int centerX, int centerY) {
+            return Task.Run(() => {
+                IntPtr msgBoxHandle = IntPtr.Zero;
+                var startTime = DateTime.Now;
+
+                while (msgBoxHandle == IntPtr.Zero && (DateTime.Now - startTime).TotalMilliseconds < 2000) {
```
AlexanderSher: What is this for?

MikhailArkhipov: It is trying to locate message box handle. This task starts BEFORE the message box is shown so it can'd find the handle right away.

PR #3042, opened 2017-01-21
Installer now adds firewall exception and auto starts seclogon

### LINKED ISSUE #3028: Remote installer should add firewall exception

karthiknadig:
Fixes #3028 #3029


int19h:
👍

PR #3043, opened 2017-01-22
3041 REPL may lose intellisense after reset

MikhailArkhipov:
Make sure we find proper R buffer if multiples are attached


int19h:
👍

PR #3048, opened 2017-01-23
Prevent hang after using locator()

### LINKED ISSUE #2939: LocatorCommandNoClick test fails with NRE
```
locatorModeTask = EventTaskSources.IRPlotDevice.LocatorModeChanged.Create(_workflow.Plots.ActiveDevice);
```

`_workflow.Plots.ActiveDevice` is null

```
 	Microsoft.R.Components.Test.dll!Microsoft.R.Components.Test.EventTaskSources.IRPlotDevice..cctor.AnonymousMethod__5_2(Microsoft.R.Components.Plots.IRPlotDevice o, System.EventHandler<Microsoft.R.Components.Plots.RPlotDeviceEventArgs> e) Line 48	C#
 	Microsoft.R.Common.Core.dll!Microsoft.Common.Core.Tasks.EventTaskSource<Microsoft.R.Components.Plots.IRPlotDevice, System.EventHandler<Microsoft.R.Components.Plots.RPlotDeviceEventArgs>, Microsoft.R.Components.Plots.RPlotDeviceEventArgs>.Create(Microsoft.R.Components.Plots.IRPlotDevice instance, System.Action<Microsoft.R.Components.Plots.RPlotDeviceEventArgs> callback, System.Threading.CancellationToken cancellationToken) Line 42	C#
 	Microsoft.R.Common.Core.dll!Microsoft.Common.Core.Tasks.EventTaskSource<Microsoft.R.Components.Plots.IRPlotDevice, System.EventHandler<Microsoft.R.Components.Plots.RPlotDeviceEventArgs>, Microsoft.R.Components.Plots.RPlotDeviceEventArgs>.Create(Microsoft.R.Components.Plots.IRPlotDevice instance, System.Threading.CancellationToken cancellationToken) Line 33	C#
>	Microsoft.R.Components.Test.dll!Microsoft.R.Components.Test.Plots.RPlotIntegrationTest.LocatorCommandNoClick() Line 817	C#
 	[Resuming Async Method]	
 	[External Code]	
 	Microsoft.UnitTests.Core.dll!Microsoft.UnitTests.Core.Threading.UIThreadHelper.RunMainThread.AnonymousMethod__31_0() Line 223	C#
 	Microsoft.UnitTests.Core.dll!Microsoft.UnitTests.Core.Threading.UIThreadHelper.CallSafe.AnonymousMethod__0() Line 254	C#
 	Microsoft.UnitTests.Core.dll!Microsoft.UnitTests.Core.Threading.UIThreadHelper.CallSafe<object>(System.Func<object> func) Line 261	C#
 	Microsoft.UnitTests.Core.dll!Microsoft.UnitTests.Core.Threading.UIThreadHelper.CallSafe(System.Action action) Line 253	C#
 	Microsoft.UnitTests.Core.dll!Microsoft.UnitTests.Core.Threading.UIThreadHelper.RunMainThread(object obj) Line 223	C#
```

### LINKED ISSUE #3011: Assert and sometimes hang when quitting VS after using locator function
```
plot(euro)
locator()
```

Test 1 - close vs

result: assert then hang

Test 2 - click end locator then close vs

result: assert then everything is fine

```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
EndLocatorMode at C:\Projects\huguesv\RTVS\src\R\Components\Impl\Plots\Implementation\ViewModel\RPlotDeviceViewModel.cs:154 was incorrectly called from a background thread.



   at Microsoft.Common.Core.Shell.CoreShellExtensions.AssertIsOnMainThread(ICoreShell coreShell, String memberName, String sourceFilePath, Int32 sourceLineNumber) in C:\Projects\huguesv\RTVS\src\Common\Core\Impl\Shell\CoreShellExtensions.cs:line 29

   at Microsoft.R.Components.Plots.Implementation.ViewModel.RPlotDeviceViewModel.EndLocatorMode() in C:\Projects\huguesv\RTVS\src\R\Components\Impl\Plots\Implementation\ViewModel\RPlotDeviceViewModel.cs:line 154

   at System.Threading.CancellationToken.ActionToActionObjShunt(Object obj)

   at System.Threading.CancellationCallbackInfo.ExecutionContextCallback(Object obj)

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.ExecutionContext.Run(ExecutionContext execu......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------
```

huguesv:
Fix #3011 Assert and sometimes hang when quitting VS after using locator function
Fix #2939 LocatorCommandNoClick test fails with NRE

Change plot manager so `StartLocatorModeAsync` doesn't need to switch to the main thread, same for the code that ends locator mode.

This means that the `_devices` list is now being accessed from multiple threads, before it used to be guaranteed from main thread only. I'm locking around the use of the list, but I don't know if I need to go further and protect the device object themselves. I'm worried about introducing new deadlocks so close to release.

The view model state updates itself by listening to the plot manager plot device object LocatorModeChanged event, by posting to the main thread.

Scenarios that now work that used to be causing issues:
- Call `locator` and then quit VS
- Call `locator`, and then reset REPL
- Call `locator`, ..., click End Locator, Reset REPL or quit VS

For the test issue, R `dev.cur()` was returning null device, and I didn't find out why. I've tested similar scenario in the product, and that never seem to happen. Anyway the test will pass now.

### DIFF
--- a/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceCommand.cs
+++ b/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceCommand.cs
@@ -19,7 +19,7 @@ namespace Microsoft.R.Components.Plots.Implementation.Commands {
 
         protected bool IsInLocatorMode {
             get {
-                return VisualComponent.LocatorMode;
+                return VisualComponent.Device.LocatorMode;
             }
         }
 
--- a/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceCommand.cs
+++ b/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceCommand.cs
@@ -19,7 +19,7 @@ namespace Microsoft.R.Components.Plots.Implementation.Commands {
 
         protected bool IsInLocatorMode {
             get {
-                return VisualComponent.Device.LocatorMode;
+                return VisualComponent.Device?.LocatorMode ?? false;
             }
         }
 
--- a/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
+++ b/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.Common.Core;
 using Microsoft.R.Components.Controller;
 using Microsoft.R.Components.InteractiveWorkflow;
 using Microsoft.R.Host.Client;
@@ -23,7 +24,7 @@ namespace Microsoft.R.Components.Plots.Implementation.Commands {
         }
 
         public Task<CommandResult> InvokeAsync() {
-            InteractiveWorkflow.Plots.EndLocatorModeAsync(VisualComponent.Device, LocatorResult.CreateNotClicked());
+            InteractiveWorkflow.Plots.EndLocatorModeAsync(VisualComponent.Device, LocatorResult.CreateNotClicked()).DoNotWait();
 
             return Task.FromResult(CommandResult.Executed);
         }
--- a/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
+++ b/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
@@ -4,6 +4,7 @@
 using System.Threading.Tasks;
 using Microsoft.R.Components.Controller;
 using Microsoft.R.Components.InteractiveWorkflow;
+using Microsoft.R.Host.Client;
 
 namespace Microsoft.R.Components.Plots.Implementation.Commands {
     internal sealed class PlotDeviceEndLocatorCommand : PlotDeviceCommand, IAsyncCommand {
@@ -22,7 +23,8 @@ namespace Microsoft.R.Components.Plots.Implementation.Commands {
         }
 
         public Task<CommandResult> InvokeAsync() {
-            VisualComponent.EndLocatorMode();
+            InteractiveWorkflow.Plots.EndLocatorModeAsync(VisualComponent.Device, LocatorResult.CreateNotClicked());
+
             return Task.FromResult(CommandResult.Executed);
         }
     }
--- a/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
+++ b/src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
@@ -24,7 +24,7 @@ namespace Microsoft.R.Components.Plots.Implementation.Commands {
         }
 
         public Task<CommandResult> InvokeAsync() {
-            InteractiveWorkflow.Plots.EndLocatorModeAsync(VisualComponent.Device, LocatorResult.CreateNotClicked()).DoNotWait();
+            InteractiveWorkflow.Plots.EndLocatorMode(VisualComponent.Device, LocatorResult.CreateNotClicked());
 
             return Task.FromResult(CommandResult.Executed);
         }
--- a/src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
+++ b/src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
@@ -32,7 +32,7 @@ namespace Microsoft.R.Components.Plots.Implementation {
         private readonly ICoreShell _shell;
 
         private TaskCompletionSource<LocatorResult> _locatorTcs;
-        private CancellationTokenRegistration? _locatorCancelTokenRegistration;
+        private CancellationTokenRegistration _locatorCancelTokenRegistration;
 
         public event EventHandler<RPlotDeviceEventArgs> ActiveDeviceChanged;
         public event EventHandler<RPlotDeviceEventArgs> DeviceAdded;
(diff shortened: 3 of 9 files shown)


src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceCommand.cs
```diff
@@ -19,7 +19,7 @@ internal abstract class PlotDeviceCommand : InteractiveWorkflowAsyncCommand {
 
         protected bool IsInLocatorMode {
             get {
-                return VisualComponent.LocatorMode;
+                return VisualComponent.Device.LocatorMode;
```
MikhailArkhipov: I think `Device` can be null? Can it?

huguesv: Thanks. Yes it can. I didn't notice because VS command handling swallows that exception.

src/R/Components/Impl/Plots/Implementation/Commands/PlotDeviceEndLocatorCommand.cs
```diff
@@ -22,7 +23,8 @@ public PlotDeviceEndLocatorCommand(IRInteractiveWorkflow interactiveWorkflow, IR
         }
 
         public Task<CommandResult> InvokeAsync() {
-            VisualComponent.EndLocatorMode();
+            InteractiveWorkflow.Plots.EndLocatorModeAsync(VisualComponent.Device, LocatorResult.CreateNotClicked());
```
AlexanderSher: If it is async, you should either await it or call `DoNotWait`

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -181,16 +189,43 @@ internal class RPlotManager : IRPlotManager {
...
+
+            return Task.CompletedTask;
+        }
+
+        private void CancelLocatorMode(IRPlotDevice device) {
+            device.LocatorMode = false;
+
+            var tcs = _locatorTcs;
+            _locatorTcs = null;
+            tcs?.SetCanceled();
```
AlexanderSher: It's better to call `TrySetCanceled`

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -181,16 +189,43 @@ internal class RPlotManager : IRPlotManager {
...
+
+            return _locatorTcs.Task;
+        }
+
+        public Task EndLocatorModeAsync(IRPlotDevice device, LocatorResult result) {
+            device.LocatorMode = false;
+
+            var tcs = _locatorTcs;
+            _locatorTcs = null;
+            tcs?.SetResult(result);
```
AlexanderSher: `TrySetResult`

huguesv: ok will change both

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -181,16 +189,43 @@ internal class RPlotManager : IRPlotManager {
...
+
+            _shell.DispatchOnUIThread(() => {
+                var visualComponent = GetVisualComponentForDevice(deviceId);
+                visualComponent?.Container.Show(focus: false, immediate: true);
+            });
+
+            return _locatorTcs.Task;
+        }
+
+        public Task EndLocatorModeAsync(IRPlotDevice device, LocatorResult result) {
```
AlexanderSher: Actually, why `EndLocatorModeAsync` is async? It returns only `Task.CompletedTask`

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
```diff
@@ -24,12 +24,16 @@ internal class RPlotManager : IRPlotManager {
...
         private readonly DisposableBag _disposableBag;
+        private readonly object _devicesLock = new object();
         private readonly List<IRPlotDevice> _devices = new List<IRPlotDevice>();
         private readonly Dictionary<int, IRPlotDeviceVisualComponent> _visualComponents = new Dictionary<int, IRPlotDeviceVisualComponent>();
         private readonly Dictionary<Guid, IRPlotDeviceVisualComponent> _assignedVisualComponents = new Dictionary<Guid, IRPlotDeviceVisualComponent>();
         private readonly List<IRPlotDeviceVisualComponent> _unassignedVisualComponents = new List<IRPlotDeviceVisualComponent>();
         private readonly ICoreShell _shell;
 
+        private TaskCompletionSource<LocatorResult> _locatorTcs;
+        private CancellationTokenRegistration? _locatorCancelTokenRegistration;
```
AlexanderSher: I don't think you really need to make it nullable. It isn't checked for null, and `Dispose` can be called on `CancellationTokenRegistration` multiple times.

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: Just for symmetry, the rest of the API is pretty much all Async.

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
huguesv: Tested it, looks like it's fine as a struct, will change.

src/R/Components/Impl/Plots/Implementation/RPlotManager.cs
AlexanderSher: Async/await pattern with default task awaiters is quite expensive. Any task awaiter creates garbage in memory, copies all values for `AsyncLocal`, captures context, etc. It also may create additional problems for debugging if stack trace is broken. One of the best practices for async/await is to use it only if it is really required.

huguesv: Alright, changing it.

AlexanderSher:
👍

PR #3049, opened 2017-01-24
Fix #3037: Fix hack for debugger breakpoints focus in VS 2017

### LINKED ISSUE #3037: Fix hack for debugger breakpoints focus in VS 2017
Since it has proper mechanisms

int19h:
Register REPL tool window via IVsDebugger6.RegisterFocusPreservingWindow in VS 2017, and disable the workaround.

Fix VS 2015 focus workaround not working when there is an active project.


src/Package/Impl/project.15.0.json
```diff
@@ -14,6 +14,7 @@
     "Microsoft.VisualStudio.Shell.Interop.10.0": "10.0.30319",
     "Microsoft.VisualStudio.Shell.Interop.11.0": "11.0.61030",
     "Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime": "15.0.25726-Preview5",
+    "Microsoft.VisualStudio.Shell.Interop.15.0.DesignTime": "15.0.25726-Preview5",
```
AlexanderSher: Do we still need `Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime`?

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
```diff
@@ -52,6 +53,20 @@ internal class VsRInteractiveWindowComponentContainerFactory : IInteractiveWindo
             var componentContainer = new VisualComponentToolWindowAdapter<IInteractiveWindowVisualComponent>(toolWindow);
             var component = new RInteractiveWindowVisualComponent(vsWindow.InteractiveWindow, componentContainer, sessionProvider, _shell);
             componentContainer.Component = component;
+
+#if !VS14
+            var frame = toolWindow.Frame as IVsWindowFrame;
```
AlexanderSher: Can we extract it in a method?

src/Package/Impl/project.15.0.json
int19h: Yeah, these assemblies don't include old types in newer versions. So whatever we were using from 14.0 (which I think is a different `IVsDebuggerX` interface), we're still using it, and need it.

src/Package/Impl/Repl/VsRInteractiveWindowComponentContainerFactory.cs
int19h: Sure.

AlexanderSher:
👍

PR #3053, opened 2017-01-24
Fixes a bug with returning error codes while creating process

karthiknadig:


src/Common/Core/Impl/OS/Win32Process.cs
```diff
@@ -63,6 +63,18 @@ public class Win32Process {
             }
         }
 
+        private void SetExitState() {
+            lock (this) {
```
AlexanderSher: Never use `lock (this)` - your lock can be taken by code outside of class scope and result in a dead-lock.

src/Common/Core/Impl/OS/Win32Process.cs
```diff
@@ -117,43 +129,48 @@ public class Win32Process {
...
-            try {
-                if(winIdentity == null) {
-                    if(!CreateProcess(applicationName, commandLine, ref processAttr, ref threadAttr, true,
-                        (uint)(CREATE_PROCESS_FLAGS.CREATE_UNICODE_ENVIRONMENT | CREATE_PROCESS_FLAGS.CREATE_NO_WINDOW),
-                        environment.NativeEnvironmentBlock,
-                        workingDirectory,
-                        ref si,
-                        out pi)) {
-                        throw new Win32Exception(Marshal.GetLastWin32Error());
+            lock (_createProcessLock) {
```
AlexanderSher: What's the problem if two processes are started concurrently?

karthiknadig: I was using the lock to keep the Error mode in sync. Error mode can impact how the system handles critical errors for the entire process. In this particular case SetThreadErrorMode does not work. Hence i had to use a process wide setting.

karthiknadig: SetErrorMode is needed to prevent system from showing a message box if a critical error occurs.

MikhailArkhipov:
👍

Issue #3052, opened 2017-01-24
REPL becomes unusable after inputting malformed text


Hong-Revo:
At the prompt, type:

    f("a")")

Press Enter. REPL displays continuation prompt, but pressing Enter and Esc do nothing, and the inputted text can't be edited.

![image](https://cloud.githubusercontent.com/assets/11328890/22246310/587bc7c0-e288-11e6-9268-e93750e235a7.png)

Hong-Revo:
Actually, on testing this some more, typing `"` (ie, closing the string) fixes this. I think pressing Esc should immediately break back to the prompt, though; that's what rgui.exe does.

MikhailArkhipov:
`ESC`: #1066. REPL appears alive (enter works). If I do `Up Arrow` it bring up original expression from the history and it can be edited (or executed which brings back `>`)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/6dfa10c44d8fd01bd13400c9188d3bc631b9099e

PR #3062, opened 2017-01-25
#3052 REPL becomes unusable after inputting malformed text

MikhailArkhipov:


src/Package/Impl/Repl/Commands/ReplCommandController.cs
```diff
@@ -151,6 +152,7 @@ public ReplCommandController(ITextView textView, ITextBuffer textBuffer)
         }
 
         private void HandleCancel(RCompletionController controller) {
+            Workflow.Operations.CancelAsync().DoNotWait();
```
AlexanderSher: Do you need to wait for cancellation?

MikhailArkhipov: I don't think it is really necessary. We post `interrupt` as well and don't wait.

AlexanderSher:
👍

Issue #3051, opened 2017-01-24
Can't open pdf when using internal help browser


Hong-Revo:
This might actually be an interaction with Adobe Acrobat Reader.

R packages often have vignettes which are pdf files. Trying to open these from the internal help fails. An example is the Matrix package:

    library(Matrix)
    ?Matrix

In the internal help, navigate to the main Matrix help page, then go to the vignettes link (2nd from top). Click on any of the pdf links. The help pane goes blank, and when I try to go back (click on the Back button), VS spins its wheels for a few seconds. Acrobat Reader also starts up in the background, and then disappears.

MikhailArkhipov:
By design. internal browser is IE and has limitations. For complex content external browser is the only option.

Hong-Revo:
Would it be possible to display an alert at least, if this happens? Or is it unrelated to RTVS?

MikhailArkhipov:
I have simple solution

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/1e205ba4476f9703a1e00bc5c2c17907255b7e50

PR #3064, opened 2017-01-25
#3051 Can't open pdf when using internal help browser

MikhailArkhipov:
Open PDF externally

### DIFF
--- a/src/Common/Core/Impl/Extensions/IOExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/IOExtensions.cs
@@ -100,8 +100,10 @@ namespace Microsoft.Common.Core {
 
         public static bool HasReadPermissions(this string path) {
             try {
-                Directory.GetFiles(path);
-                return true;
+                if (Directory.Exists(path)) {
+                    Directory.GetFiles(path);
+                    return true;
+                }
             } catch(IOException) { } catch(UnauthorizedAccessException) { }
             return false;
         }
--- a/src/Package/Impl/Help/HelpVisualComponent.cs
+++ b/src/Package/Impl/Help/HelpVisualComponent.cs
@@ -266,6 +266,9 @@ namespace Microsoft.VisualStudio.R.Package.Help {
 
         private static bool IsHelpUrl(string url) {
             Uri uri = new Uri(url);
+            if(uri.AbsoluteUri.EndsWith(".pdf")) {
+                return false;
+            }
             // dynamicHelp.R (startDynamicHelp function):
             // # Choose 10 random port numbers between 10000 and 32000
             // ports <- 10000 + 22000*((stats::runif(10) + unclass(Sys.time())/300) %% 1)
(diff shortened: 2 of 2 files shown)


src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -100,8 +100,10 @@ public static class IOExtensions {
 
         public static bool HasReadPermissions(this string path) {
             try {
-                Directory.GetFiles(path);
-                return true;
+                if (Directory.Exists(path)) {
```
AlexanderSher: `Directory.GetFiles` should've throw `IOException` and then you return false. Didn't it work?

src/Package/Impl/Help/HelpVisualComponent.cs
```diff
@@ -266,6 +266,9 @@ internal sealed class HelpVisualComponent : IHelpVisualComponent {
 
         private static bool IsHelpUrl(string url) {
             Uri uri = new Uri(url);
+            if(uri.AbsoluteUri.EndsWithIgnoreCase(".pdf")) {
```
AlexanderSher: Do you care if it is a local uri?

src/Common/Core/Impl/Extensions/IOExtensions.cs
MikhailArkhipov: It is annoying in debugger

MikhailArkhipov: Really the issue is that we store folder as a singleton but really should be storing per connection

src/Package/Impl/Help/HelpVisualComponent.cs
MikhailArkhipov: No, still should open PDF in external. Should I also be checking AbsolutePath? Ie `IsFile && AbsolutePath.Ends...` ?

MikhailArkhipov: But in help case it is never local AFAIK

AlexanderSher: > No, still should open PDF in external.
I'm not sure if such check is reliable for external url, which may have no file name in it at all.

src/Common/Core/Impl/Extensions/IOExtensions.cs
AlexanderSher: Do we have a bug for that?

src/Package/Impl/Help/HelpVisualComponent.cs
MikhailArkhipov: That would difficult to solve since we don't know content then. R Vignettes seem to work normally (i.e. they do have pdf extension)

src/Common/Core/Impl/Extensions/IOExtensions.cs
MikhailArkhipov: Nope. This throws b/c it stored my Azure's `c:\users\user2\Documents`

AlexanderSher:
👍

PR #3073, opened 2017-01-26
Moving RHost Firewall exception to the RHost Remote installer

### LINKED ISSUE #3071: Firewall exception to rhost.exe is not working
We had to edit the path to the exe in order for it to work.

What this means is that my outgoing connections (like package repository) were not working.

karthiknadig:
Fixes #3071


int19h:
👍

PR #3080, opened 2017-01-28
 Fix #3068: F5 produces "Object reference not set to instance of object" error if REPL is not visible

### LINKED ISSUE #3068: F5 produces "Object reference not set to instance of object" error if REPL is not visible
subj - repros with any newly created project from R Project template.

int19h:


.gitignore
```diff
@@ -76,7 +76,7 @@ src/Mocks/Editor/project.json
 src/Debugger/Impl/project.json
 src/R/Support/Impl/project.json
 src/R/Sql/Impl/project.json
-
+src/R/Wpf/Impl/project.json
```
AlexanderSher: Do we have project-dependent json in R.Wpf?

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
```diff
@@ -36,6 +38,7 @@ namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
     internal class RDebugLaunchProvider : DebugLaunchProviderBase {
         private readonly ProjectProperties _properties;
         private readonly IRInteractiveWorkflow _interactiveWorkflow;
+        private readonly IApplicationShell _shell;
```
AlexanderSher: Do you use this field?

.gitignore
int19h: Yep. Apparently it was added some time ago, but project.json wasn't removed accordingly, so it kept popping up in the changelist and annoying me :)

src/Package/Impl/ProjectSystem/RDebugLaunchProvider.cs
int19h: My bad, something that I forgot to remove from an older iteration.

AlexanderSher:
👍

Issue #3075, opened 2017-01-27
View() doesn't support 2D vectors that aren't matrices


ebs238:
RTVS silently produces no output View(x)  even though x can be coerced to a data.frame
this is related to Issue #3000 where View is identified as a workaround

source of Issue:
```
library(quantmod)
View(getSymbols("SPY", auto.assign = FALSE))
```

`tail(as.data.frame(getSymbols("SPY", auto.assign = FALSE)))`

produces:

```
           SPY.Open SPY.High SPY.Low SPY.Close SPY.Volume SPY.AdjFactor
2017-01-18   226.54   226.80  225.90    226.75   51146400             1
2017-01-19   226.84   227.00  225.41    225.91   65357500             1
2017-01-20   226.70   227.31  225.97    226.74  116431700             1
2017-01-23   226.74   226.81  225.27    226.15   72331900             1
2017-01-24   226.40   228.08  226.27    227.60   89144800             1
2017-01-25   228.70   229.57  228.51    229.57   80932500             1
```

Also, this works as expected:

`View(as.data.frame(getSymbols("SPY", auto.assign = FALSE)))`

int19h:
`View` does not coerce to dataframe by design - it is a lossy operation on some types, and it creates a copy, which can be very expensive on large datasets. It will try to work directly on the original variable, using `[ ]` and `[[ ]]`. 

If you do want to specifically show something that isn't a dataframe as if it were one, you need to manually do the cast, i.e. `View(as.data.frame(getSymbols("SPY", auto.assign = FALSE)))`.

However, in this case, the returned value is a 2D vector of doubles, so `View` should display it in a grid just fine.

int19h:
Hm, looks like we're specifically excluding vectors with dimensions for some reason - this is in `VectorView`:
```cs
        public override bool CanView(IRValueInfo evaluation) {
            if (evaluation != null && !evaluation.Classes.Any(t => _excludedClasses.Contains(t))) {
                return evaluation.Dim == null && evaluation.Length > 1;
            }
            return false;
        }
```
And then in `GridViewer`, we whitelist by classname, including `array` and `matrix`. But in this case we have a 2D vector that is *not* a matrix, so it doesn't apply either.

ebs238:
but the docs on View say: 

> an R object which can be **coerced** to a data frame with non-zero numbers of rows and columns

also, MSR Client works as expected for:

`View(getSymbols("SPY", auto.assign = FALSE))`

[...]

PR #3086, opened 2017-01-31
Port fix for #3075 to Release/1_0

int19h:


src/Package/Test/DataInspect/ViewersTest.cs
```diff
@@ -18,19 +18,23 @@
...
+        private const REvaluationResultProperties AllFields = unchecked((REvaluationResultProperties)~0);
+
+        private readonly TestMethodFixture _testMethod;
         private readonly IRSessionProvider _sessionProvider;
         private readonly IObjectDetailsViewerAggregator _aggregator;
         private readonly IRInteractiveWorkflow _workflow;
 
-        public ViewersTest() {
+        public ViewersTest(TestMethodFixture testMethod) {
+            _testMethod = testMethod;
```
AlexanderSher: Do you need `TestMethodFixture` here?

int19h: I think it's not actually needed, this was from an earlier iteration when I was trying to start the session differently. But it's a port of the other commit that just went in, so I'd rather not touch anything here unless it's specifically required to merge into this branch.

AlexanderSher:
👍 
Do we need a second reviewer?

int19h:
@MikhailArkhipov ?

MikhailArkhipov:
👍

PR #3087, opened 2017-02-01
Handles issues with blocked REPL due to UI in remote

### LINKED ISSUE #3069: Packages loaded on remote machines which popup UI will cause RTVS to hang
An example of this is the `rattle` package.

If you do this:

*> library(rattle)

You will get the screenshot below, where the REPL is hung. The reason why is that during load of this package the first time, it tries to popup a dialog box to ask the user whether they want to install GTK+ on their local machine.

image

karthiknadig:
Fixes #3069

### DIFF
--- a/src/Host/Broker/Impl/Sessions/Session.cs
+++ b/src/Host/Broker/Impl/Sessions/Session.cs
@@ -83,11 +83,12 @@ namespace Microsoft.R.Host.Broker.Sessions {
 
             var useridentity = User as WindowsIdentity;
             // In remote broker User Identity type is always WindowsIdentity
-            string suppressUI = (useridentity == null) ? string.Empty : "--suppress-ui ";
+            string suppressUI = (useridentity == null) ? string.Empty : "--rhost-suppress-ui ";
+            string isRepl = Id.StartsWithIgnoreCase("REPL") ? string.Empty : "--rhost-interactive ";
             string brokerPath = Path.GetDirectoryName(typeof(Program).Assembly.GetAssemblyPath());
             string rhostExePath = Path.Combine(brokerPath, RHostExe);
             string logFolderParam = string.IsNullOrEmpty(logFolder) ? string.Empty : Invariant($"--rhost-log-dir \"{logFolder}\"");
-            string commandLine = Invariant($"\"{rhostExePath}\" {suppressUI}--rhost-name \"{Id}\" {logFolderParam} --rhost-log-verbosity {(int)verbosity} {CommandLineArguments}");
+            string commandLine = Invariant($"\"{rhostExePath}\" {suppressUI}{isRepl}--rhost-name \"{Id}\" {logFolderParam} --rhost-log-verbosity {(int)verbosity} {CommandLineArguments}");
             var usernameBldr = new StringBuilder(NativeMethods.CREDUI_MAX_USERNAME_LENGTH + 1);
             var domainBldr = new StringBuilder(NativeMethods.CREDUI_MAX_DOMAIN_LENGTH + 1);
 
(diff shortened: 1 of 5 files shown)


src/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveEvaluator.cs
```diff
@@ -99,7 +99,13 @@ public sealed class RInteractiveEvaluator : IInteractiveEvaluator {
 
         public async Task<ExecutionResult> ResetAsync(bool initialize = true) {
             try {
+                bool isResetting = true;
                 if (Session.IsHostRunning) {
+                    if(MessageButtons.Yes == await _coreShell.ShowMessageAsync(Resources.Warning_SaveOnReset, MessageButtons.YesNo)) {
+                        await Session.ExecuteAsync("rtvs:::save_state()", REvaluationKind.NoResult);
```
AlexanderSher: What should happen in case of `RHostDisconnectedException` here?

int19h: You don't need `NoResult`, it's implied by the use of `ExecuteAsync` rather than `EvaluateAsync`.

And yes, this should catch exceptions and log to REPL. Something along the lines of "failed to save state, resetting anyway".

src/R/Components/Impl/Resources.resx
```diff
@@ -630,4 +630,7 @@ Please upgrade remote R Services installation to match.</value>
   <data name="BrokerDisconnected" xml:space="preserve">
     <value>Connection to the workspace has been terminated.</value>
   </data>
+  <data name="Warning_SaveOnReset" xml:space="preserve">
+    <value>Resetting R Host will loose the current state. Would you like to save state?</value>
```
MikhailArkhipov: lose. loose = is to make something less tight :-). Maybe 'erase all variables from the workspace' instead.

int19h: It's not just variables, it's also loaded packages etc. I think "... will clear the workspace" should be clear enough.

int19h:
👍

MikhailArkhipov:
Please port to `Release/1_0`

MikhailArkhipov:
👍

src/Host/Broker/Impl/Sessions/Session.cs
```diff
@@ -83,11 +83,12 @@ public class Session {
 
             var useridentity = User as WindowsIdentity;
             // In remote broker User Identity type is always WindowsIdentity
-            string suppressUI = (useridentity == null) ? string.Empty : "--suppress-ui ";
+            string suppressUI = (useridentity == null) ? string.Empty : "--rhost-suppress-ui ";
+            string isRepl = Id.StartsWithIgnoreCase("REPL") ? string.Empty : "--rhost-interactive ";
```
int19h: @AlexanderSher Is this the best way to flow this info through? I think we might want a new boolean flag here as well, instead of checking names.

int19h: Actually, come to think of it, when the session is created, the creator can specify the arbitrary command line parameters. So let's do that all on the broker client side instead. This shouldn't care about anything except for log verbosity, and that only because it comes from global broker settings.

MikhailArkhipov: As long as this is not too expensive. We want this change in for the BB tomorrow.

int19h: It can be the exact same string check if need be, just on the other side of the wire (in `BrokerClient` presumably, or really anywhere the command line arguments for the host flow through before they get to the client). @AlexanderSher might have a more specific preference.

karthiknadig: Created a task to track this. I will be merging the current changes.

PR #3089, opened 2017-02-01
Fix #3046: PackageIndex.BuildRemainingPackagesFunctionListAsync very slow in tests

### LINKED ISSUE #3046: PackageIndex.BuildRemainingPackagesFunctionListAsync very slow in tests
`PackageIndex.BuildRemainingPackagesFunctionListAsync` makes two calls to RHost for every package. This can be combined into single call.

AlexanderSher:


src/Host/Client/Impl/rtvs/R/packages.R
```diff
@@ -1,11 +1,40 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See LICENSE in the project root for license information.
 
-packages.installed <- function() {
+packages.installed.functions <- function() {
+	unname(lapply(packages.installed.named(), function(p) {
```
MikhailArkhipov: Do you want to `tryCatch` this?

MikhailArkhipov:
👍

src/Host/Client/Impl/rtvs/R/packages.R
AlexanderSher: AFAIK `unname` doesn't produce any errors, and `lapply` can produce error only if underlying function fails.

PR #3100, opened 2017-02-02
Fix  #3091: WebSocketMessageTransport.ReceiveAsync null return value isn't handled

### LINKED ISSUE #3091: WebSocketMessageTransport.ReceiveAsync null return value isn't handled
It is possible that new socket become closed before first request to it was made. `RHost.RunWorker` should handle possible null value and exit without throwing `InvalidDataException`.

AlexanderSher:
RHost can fail at any time after start, including the moment when RTVS requests the handshake.

### DIFF
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -81,6 +81,8 @@ namespace Microsoft.R.Host.Client {
 
             if (message != null) {
                 Log.Response(message.ToString(), _rLoopDepth);
+            } else {
+                Log.Response(_transport.CloseStatusDescription, _rLoopDepth);
             }
 
             return message;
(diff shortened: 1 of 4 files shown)


src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -81,6 +81,8 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
 
             if (message != null) {
                 Log.Response(message.ToString(), _rLoopDepth);
+            } else {
+                Log.Response(_transport.CloseStatusDescription, _rLoopDepth);
```
MikhailArkhipov: Does it handle null?

MikhailArkhipov: Oh, nvm it is at least empty

MikhailArkhipov:
👍

src/Host/Client/Impl/Host/RHost.cs
AlexanderSher: Yes, we return `null` when transport is closed

PR #3116, opened 2017-02-03
Fix #3101: Microsoft.R.Common.Wpf is not versioned

### LINKED ISSUE #3101: Microsoft.R.Common.Wpf is not versioned
Signed, but version is 0.0.0.0

AlexanderSher:


MikhailArkhipov:
👍

PR #3118, opened 2017-02-03
Fix #3076: Getting the class of an object after a failed View() hangs the REPL

### LINKED ISSUE #3076: Getting the class of an object after a failed View() hangs the REPL
Following on from my comment in https://github.com/Microsoft/RTVS/issues/3075

    tib <- tibble::as_tibble(mtcars)
    View(tib)
    class(tib)

Can't stop execution or reset the R session.

int19h:
Set task completion sources for request handlers asynchronously, so that the continuation for the task does not hijack the message handling thread and block it.

Log errors in VisualGridScroller instead of asserting.

### DIFF
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -413,7 +413,7 @@ namespace Microsoft.R.Host.Client {
                             throw ProtocolError($"Mismatched response - message name does not match request '{request.MessageName}':", message);
                         }
 
-                        request.Handle(this, message);
+                        Task.Run(() => request.Handle(this, message)).DoNotWait();
                         continue;
                     }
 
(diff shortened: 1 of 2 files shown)


MikhailArkhipov:
👍

int19h:
@AlexanderSher, can you also take a look. This is to address this problem:

http://blog.stephencleary.com/2012/12/dont-block-in-asynchronous-code.html

except in our case, this is happening because we call `SetException` due to evaluation failing, which results in the exception handler in the grid view executing on the thread, and then proceeding to the next loop iteration where it blocks on a concurrent collection (of scroll events), thus effectively blocking the `RunLoop` thread entirely.

AlexanderSher:
No, awaited task can inline continuation and that is a normal behavior. Lock around awaits is not. `BlockingCollection<ScrollCommand>` should be replaced with `BufferBlock<ScrollCommand>`, and `_scrollCommands.GetConsumingEnumerable` with `await _scrollCommands.ReceiveAsync`

int19h:
Roger that, I'll start a new PR for that change.

int19h:
https://github.com/Microsoft/RTVS/pull/3128

PR #3128, opened 2017-02-06
Fix #3076: Getting the class of an object after a failed View() hangs the REPL

### LINKED ISSUE #3076: Getting the class of an object after a failed View() hangs the REPL

int19h:
Use BufferBlock instead of BlockingCollection in VisualGridScroller, to avoid blocking waits across awaits.


AlexanderSher:
👍

PR #3137, opened 2017-02-07
Disable auto restart for services after first restart.

### LINKED ISSUE #3136: Do not endlessly try to restart the R Host Broker if TLS cert cannot be found
When you create a VM on Azure, you need a fully qualified internet-reachable domain name which is *always* different than the machine name (unlike NETBIOS name resolution).

The error logs say this:

```
[2017-02-07 10:12:11Z] <Microsoft.R.Host.Broker.Startup.Program> (C):
TLS certificate not found: CN=<server name>
```

The service crashes with this log entry and then immediately tries to restart. It is stuck in an infinite loop creating the service and crashing.

What we need to add the following to this file: `C:\Program Files\R Remote Service for Visual Studio\1.0\Microsoft.R.Host.Broker.Config.json`

```
{
  "server.urls": "https://0.0.0.0:5444",
  "security": {
    "X509CertificateName": "CN=<server name>.westus.cloudapp.azure.com"
  }
}
```

At a minimum, I think we need to not endlessly try to restart the service when we encounter this error.

karthiknadig:


MikhailArkhipov:
I am not convinced we should disable restarts. If service *crashed* we DO want to restart. If admin made a mistake in configuration then maybe we shouldn't be restarting.

karthiknadig:
Closing this. We have discussed a different way to solve this problem.

PR #3139, opened 2017-02-07
Fix #3120

### LINKED ISSUE #3120: Message to reload autosave after reset is wrong
When you click Reset on REPL, it will prompt you to save.
Click Yes.
You will see a new session start, and then this other prompt appears:
```
---------------------------
Microsoft Visual Studio
---------------------------
Previous R session terminated unexpectedly, and its global workspace has been saved to image "~/.Autosave.RData". Would you like to reload it?
---------------------------
Yes   No   
---------------------------
```

It didn't really terminate **unexpectedly**.

Also, that first prompt when you reset, we probably want that an option to control whether we prompt or not.

MikhailArkhipov:
Add prompt suppression option. We already handle reset and do not show reload prompt


int19h:
👍

AlexanderSher:
👍

PR #3149, opened 2017-02-08
Fix #3140: Move message text from R code to C#

### LINKED ISSUE #3140: Move message text from R code to C#
So we can localize them in a regular way. Perhaps pass some form of ID like `#125` similarly to what VS SDK does? Or more readable `#message_id` similar to `Resources.message_id`

int19h:
Add an R helper function to get localizable strings from the IDE, and use it for all user-visible string literals in R code.

This depends on https://github.com/Microsoft/R-Host/pull/109

### DIFF
--- a/src/Host/API/Impl/RSessionCallback.cs
+++ b/src/Host/API/Impl/RSessionCallback.cs
@@ -51,5 +51,7 @@ namespace Microsoft.R.Host.Client {
         public Task ViewFile(string fileName, string tabName, bool deleteFile, CancellationToken cancellationToken = default(CancellationToken)) => Task.CompletedTask;
         public Task ViewLibraryAsync(CancellationToken cancellationToken = default(CancellationToken)) => Task.CompletedTask;
         public Task ViewObjectAsync(string expression, string title, CancellationToken cancellationToken = default(CancellationToken)) => Task.CompletedTask;
+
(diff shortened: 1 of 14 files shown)


src/Host/API/Impl/RSessionCallback.cs
```diff
@@ -51,5 +51,7 @@ public Task<MessageButtons> ShowMessageAsync(string message, MessageButtons butt
         public Task ViewFile(string fileName, string tabName, bool deleteFile, CancellationToken cancellationToken = default(CancellationToken)) => Task.CompletedTask;
         public Task ViewLibraryAsync(CancellationToken cancellationToken = default(CancellationToken)) => Task.CompletedTask;
         public Task ViewObjectAsync(string expression, string title, CancellationToken cancellationToken = default(CancellationToken)) => Task.CompletedTask;
+
+        public string GetLocalizedString(string id) => null;
```
AlexanderSher: Can we provide all those strings at startup in a batch?

int19h: We can, but then it won't reflect culture changes until the session is restarted.

MikhailArkhipov:
👍

karthiknadig:
👍

PR #3159, opened 2017-02-09
Fix #2726: Workspaces window is not shown by default in Data Science profile

### LINKED ISSUE #2726: Workspaces window is not shown by default in Data Science profile
This leads to a rather confusing initial experience, because the REPL is there, but it tells you to connect to something in Workspaces - and there's no Workspaces. You have to know it's under R Tools | Windows.

int19h:


int19h:
This also goes into 1.0, right?

MikhailArkhipov:
Yes

PR #3162, opened 2017-02-09
Handle endless service restart.

### LINKED ISSUE #3136: Do not endlessly try to restart the R Host Broker if TLS cert cannot be found

karthiknadig:
Fixes #3136 

Broker service logs error to Application log:
image

image

### DIFF
--- a/src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
+++ b/src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
@@ -7,8 +7,8 @@ using System.Globalization;
 using Microsoft.Common.Core.Disposables;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.R.Host.UserProfile {
-    internal sealed class ServiceLogger : ILogger, IDisposable {
+namespace Microsoft.Common.Core.Logging {
+    public sealed class ServiceLogger : ILogger, IDisposable {
         private readonly string _category;
 
         public ServiceLogger(string category) {
--- a/src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
+++ b/src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
@@ -10,9 +10,13 @@ using Microsoft.Extensions.Logging;
 namespace Microsoft.Common.Core.Logging {
     public sealed class ServiceLogger : ILogger, IDisposable {
         private readonly string _category;
+        private readonly LogLevel _logLevel;
+        private readonly string _eventLogSource;
 
-        public ServiceLogger(string category) {
+        public ServiceLogger(string category, LogLevel minLogLevel, string eventLogSource) {
             _category = category;
+            _logLevel = minLogLevel;
+            _eventLogSource = eventLogSource;
         }
 
         public void Dispose() {
@@ -26,12 +30,15 @@ namespace Microsoft.Common.Core.Logging {
             return true;
         }
 
-
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter) {
+            if(logLevel < _logLevel) {
+                return;
+            }
+
             string message = formatter(state, exception);
             using (EventLog eventLog = new EventLog("Application")) {
                 EventLogEntryType logType = GetLogType(logLevel);
-                eventLog.Source = "Application";
+                eventLog.Source = _eventLogSource;
                 string logMessage;
                 if (exception != null) {
                     logMessage = string.Format(CultureInfo.CurrentCulture, "[{0:u}] <{1}> ({2}):{3}{4}{5}Exception: {6}", DateTime.Now, _category, logLevel.ToString()[0], Environment.NewLine, message, Environment.NewLine, exception);
--- a/src/Host/Broker/Impl/Resources.resx
+++ b/src/Host/Broker/Impl/Resources.resx
@@ -250,12 +250,12 @@
     <value>Creates and manages R Host (Microsoft.R.Host.exe) sessions.</value>
   </data>
   <data name="Text_ServiceDisplayName" xml:space="preserve">
-    <value>R Host Broker Service</value>
+    <value>R Remote Service</value>
   </data>
   <data name="Error_ExitRHost" xml:space="preserve">
-    <value>RHost Exited with Win32 error : {0}</value>
+    <value>R Host Exited with Win32 error : {0}</value>
   </data>
   <data name="Text_ServiceName" xml:space="preserve">
-    <value>RHostBrokerService</value>
+    <value>RRemoteService</value>
   </data>
 </root>
--- a/src/Host/Broker/Impl/Resources.resx
+++ b/src/Host/Broker/Impl/Resources.resx
@@ -255,4 +255,7 @@
   <data name="Error_ExitRHost" xml:space="preserve">
     <value>RHost Exited with Win32 error : {0}</value>
   </data>
+  <data name="Text_ServiceName" xml:space="preserve">
+    <value>RHostBrokerService</value>
+  </data>
 </root>
--- a/src/Host/Broker/Impl/Resources.resx
+++ b/src/Host/Broker/Impl/Resources.resx
@@ -258,4 +258,7 @@
   <data name="Text_ServiceName" xml:space="preserve">
     <value>RRemoteService</value>
   </data>
+  <data name="Error_ConfigFailed" xml:space="preserve">
+    <value>Command Line or Config File error: {0}</value>
+  </data>
 </root>
--- a/src/Host/Broker/Impl/Startup/BrokerService.cs
+++ b/src/Host/Broker/Impl/Startup/BrokerService.cs
@@ -4,6 +4,7 @@
 using System.ServiceProcess;
 using System.Threading.Tasks;
 using Microsoft.Common.Core;
+using Microsoft.Extensions.Logging;
 using Microsoft.R.Host.Broker.Startup;
 
 namespace Microsoft.R.Host.Broker {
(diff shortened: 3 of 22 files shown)


karthiknadig:
Changes made here are as follows:
1. If it is a Windows Service related failure windows will handle restart.
2. if the failure is due to certificate or configuration. We will log error and stop service. Service will not be restarted automatically.
3. Broker will log warning, errors and critical failures to the Windows> Application log. 
4. Broker will log all Traces to Broker log file in the Network Services TEMP folder.

src/SetupRemote/SetupRemote.wixproj
```diff
@@ -43,6 +43,10 @@
     <SuppressSpecificWarnings>1076</SuppressSpecificWarnings>
   </PropertyGroup>
   <ItemGroup>
+    <WixExtension Include="WixNetFxExtension">
+      <HintPath>$(WixExtDir)\WixNetFxExtension.dll</HintPath>
```
AlexanderSher: What is this for?

karthiknadig: it is needed to get the path to EventLogMessages.dll in .net framework install directory. It is a required field to set while creating event log source.

src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
```diff
@@ -26,12 +30,15 @@ internal sealed class ServiceLogger : ILogger, IDisposable {
             return true;
         }
 
-
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter) {
+            if(logLevel < _logLevel) {
```
AlexanderSher: This should affect `IsEnabled` as well

src/SetupRemote/Product.wxs
```diff
@@ -74,6 +78,7 @@
         </ServiceInstall>
         <ServiceControl Id="RemoteRBrokerServiceUninstall" Name="RHostBrokerService" Remove="uninstall" Stop="uninstall"/>
         <fire:FirewallException Id="RBrokerSvcFirewallException" Name="Microsoft.R.Host.Broker" Program="[INSTALLFOLDER]Microsoft.R.Host.Broker.exe" Scope="any" Profile="all"/>
+        <util:EventSource Log="Application" Name="RHostBrokerService" EventMessageFile="[NETFRAMEWORK40CLIENTINSTALLROOTDIR]EventLogMessages.dll"/>
```
AlexanderSher: Do we check anywhere that `NETFRAMEWORK40FULLINSTALLROOTDIR64` exists?

karthiknadig: .net 4.6 is required before installing. The check for this is in the bundle project.

src/Host/Broker/Impl/Startup/BrokerService.cs
```diff
@@ -13,11 +14,10 @@ partial class BrokerService : ServiceBase {
...
 
         protected override void OnStart(string[] args) {
-            Task.Run(() => CommonStartup.CreateAndRunWebHostForService()).DoNotWait();
-        }
-
-        protected override void OnStop() {
-            CommonStartup.Exit();
+            Task.Run(() => {
+                CommonStartup.CreateAndRunWebHostForService();
+                Stop();
```
AlexanderSher: Does it affect a scenario when user explicitly stops the service?

src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
```diff
@@ -7,12 +7,16 @@
 using Microsoft.Common.Core.Disposables;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.R.Host.UserProfile {
-    internal sealed class ServiceLogger : ILogger, IDisposable {
+namespace Microsoft.Common.Core.Logging {
+    public sealed class ServiceLogger : ILogger, IDisposable {
```
AlexanderSher: Probably we should call it `EventLogLogger`, cause it is in `Common.Core` now and can be reused in code that isn't related to service.

src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
```diff
@@ -7,12 +7,16 @@
...
         private readonly string _category;
+        private readonly LogLevel _logLevel;
+        private readonly string _eventLogSource;
 
-        public ServiceLogger(string category) {
+        public ServiceLogger(string category, LogLevel minLogLevel, string eventLogSource) {
```
MikhailArkhipov: I think service logger is never 'None' and generally always 'Minimal'. It is not controlled by user settings

src/Host/Broker/Impl/Resources.resx
```diff
@@ -255,4 +255,7 @@
   <data name="Error_ExitRHost" xml:space="preserve">
     <value>RHost Exited with Win32 error : {0}</value>
   </data>
+  <data name="Text_ServiceName" xml:space="preserve">
+    <value>RHostBrokerService</value>
```
MikhailArkhipov: Should be something more descriptive. Human friendly name is 'R Remote Service'

MikhailArkhipov: We don't use terms 'broker' or 'host' in user-facing messages since they are specific to our implementation and user has no idea what do they mean

src/Host/Broker/Impl/Startup/BrokerService.cs
karthiknadig: No. This should not affect that case.

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -4,21 +4,21 @@
...
 using Microsoft.Extensions.Logging;
 using Microsoft.R.Host.Broker.Logging;
 using Microsoft.R.Host.Broker.Security;
 using Microsoft.R.Host.Protocol;
 using Newtonsoft.Json;
-using System.Diagnostics;
-using System.ServiceProcess;
 
 namespace Microsoft.R.Host.Broker.Startup {
     internal class CommonStartup {
```
MikhailArkhipov: IMO too many if(IsService). This is worth separate class, maybe some abstract base and then App and Service

src/Host/Broker/Impl/Resources.resx
karthiknadig: The name here is the Service Name, e.g, seclogon. User friendly service name is Service Display Name. If you are suggesting that we use user friendly name for the event log, I can make that change. If we want to rename the display name for Broker Service, then i can do that. Should it be 'R Service' ?

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -142,16 +156,37 @@ internal class CommonStartup {
...
-                // but if it didn't work, just terminate it.
-                await Task.Delay(10000);
-                _logger.LogCritical(Resources.Critical_TimeOutShutdown);
-                Environment.Exit((int)BrokerExitCodes.Timeout);
+                if (IsService) {
+                    ServiceExit();
+                } else {
+                    // Give cooperative cancellation 10 seconds to shut the process down gracefully,
+                    // but if it didn't work, just terminate it.
+                    await Task.Delay(10000);
```
MikhailArkhipov: This seem just to wait for 10 sec and then log critical. Is it expected to check for successful termination somewhere?

karthiknadig: @int19h do you know why we wait here?

src/Host/Broker/Impl/Startup/CommonStartup.cs
karthiknadig: I think `CommonStartup` has to be refactored. For now, I see that `IsService` is used in Start and Exit. So, i would suggest such a change for the future.

src/Common/Core/Impl/Logging/Implementation/ServiceLogger.cs
karthiknadig: I added this setting so that RUserProfileService can log at `Trace` level and Broker can log at `Warning` level. Since RUserProfileService only logs via the event logs, it should log everything via the event logs. For broker, all the log messages go into the log file, and anything at warning or greater level goes into event logs.

src/Host/Broker/Impl/Startup/CommonStartup.cs
```diff
@@ -37,34 +38,65 @@ internal class CommonStartup {
...
+                    if (configFile != null) {
+                        var configBuilder = new ConfigurationBuilder().AddJsonFile(configFile, optional: false);
+                        configuration = configBuilder.Build();
+                    }
+
+                    Configuration = configuration;
+                    Configuration.GetSection("startup").Bind(_startupOptions);
+                    Configuration.GetSection("security").Bind(_securityOptions);
+                    Configuration.GetSection("logging").Bind(_loggingOptions);
+                } catch (Exception ex) when (!ex.IsCriticalException()) {
```
int19h: I think this might also throw if command line arguments are messed up, so you might want to change the error message to be more generic to reflect that.

src/Host/Broker/Impl/Startup/CommonStartup.cs
int19h: (sorry, can't see the change in context anymore)

In the original code, it was done in `Task.Run` after triggering the cancellation token. The idea is that this token should be propagated everywhere to serve as an exit trigger for the whole app. In this case, the wait wouldn't complete because the main thread would exit first.

It looks like we have lost a bunch of places that were doing it somewhere along the line though - I only see it ever being used in one place.

Not sure if we still need it at all.

MikhailArkhipov:
Let's finish this soon so we can merge for RC2

Issue #3163, opened 2017-02-09
Dismissing signature help with ESC adds line break in REPL


MikhailArkhipov:
Since it also propagates to REPL as 'Cancel'. See also #1066 and #3052. Most probably side effect of https://github.com/Microsoft/RTVS/commit/6dfa10c44d8fd01bd13400c9188d3bc631b9099e

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/a5c501548e392c0352a850cd2ddcd00f3f4968dd

jflam:
I still see this issue if I hit ESC on an empty line in the REPL - a blank line is inserted.

MikhailArkhipov:
By design. Same as in RGui and RStudio. The bug was NOT to do it when user dismissed intellisense list or signature help.

Issue #3161, opened 2017-02-09
Completion shows signature from wrong function


MikhailArkhipov:
```
library(raster)
```

type `graphics::plot` - plot signature is still from raster, not from graphics

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/a5c501548e392c0352a850cd2ddcd00f3f4968dd

Issue #3095, opened 2017-02-02
Function arguments in tooltips messed up


Hong-Revo:
... when you have multiple functions listed in a single help file.

![image](https://cloud.githubusercontent.com/assets/11328890/22545270/200dbee2-e98c-11e6-8c82-35aa601c91c4.png)

This is because `?parLapply` (in the parallel package) shares the same help page as a bunch of related functions:

```
clusterCall(cl = NULL, fun, ...)
clusterApply(cl = NULL, x, fun, ...)
clusterApplyLB(cl = NULL, x, fun, ...)
clusterEvalQ(cl = NULL, expr)
clusterExport(cl = NULL, varlist, envir = .GlobalEnv)
clusterMap(cl = NULL, fun, ..., MoreArgs = NULL, RECYCLE = TRUE,
           SIMPLIFY = FALSE, USE.NAMES = TRUE,
           .scheduling = c("static", "dynamic"))
clusterSplit(cl = NULL, seq)

parLapply(cl = NULL, X, fun, ...)
parSapply(cl = NULL, X, FUN, ..., simplify = TRUE,
          USE.NAMES = TRUE)
parApply(cl = NULL, X, MARGIN, FUN, ...)
parRapply(cl = NULL, x, FUN, ...)
parCapply(cl = NULL, x, FUN, ...)

parLapplyLB(cl = NULL, X, fun, ...)
parSapplyLB(cl = NULL, X, FUN, ..., simplify = TRUE,
            USE.NAMES = TRUE)
```

Hong-Revo:
I just noticed from the screenshot: should also tidy up handling of URLs in tooltips.

MikhailArkhipov:
We don't currently show overrides list like C# does. Also, we use standard tooltip (i.e. we provide a string and that's it). It is possible to use custom WPF control with clickable links, eventually we'll provide that - see #1400

Hong-Revo:
You'll have to explain it to me like I'm a total C# and VS noob (which I am ;) ). What's an override list?

MikhailArkhipov:
So this appear to be effect of #1060 which has been there since at least 0.3. Apparently some documentation provides complete list of arguments in one of the signatures and incomplete in others. Also, most detailed list propagates to aliases

`parLapplyLB` ans `parSapply` are both listed as alias or `clusterApply`.

If this is incorrect, it is easy to revert #1060 and completion with then just show exactly what it is RD.

MikhailArkhipov:
[...]

PR #3170, opened 2017-02-10
#3095 #3161 #3163

MikhailArkhipov:
#3095 Function arguments in tooltips messed up 
- this was techically by design since 0.3 but apparently wrong. I thought that among function aliases signature would be the same but it is not true. Alias means more 'function from the same family'. We now match RS.

#3161 Completion shows signature from wrong function
- make sure we extract package name and use it when fetching function information. This used to work when we had functions extracted from all packages but stopped when we fixed behaviour so only function from the last loaded package is shown. Test TBA, need to figure out stability since tests with package load/unload are not very stable.

#3163 Dismissing signature help with ESC adds line break in REPL
- check that there is no completion or signature sessions up


src/R/Editor/Impl/Signatures/ParametersInfo.cs
```diff
@@ -35,19 +35,25 @@ public sealed class ParameterInfo {
         public string FunctionName { get; private set; }
 
         /// <summary>
+        /// Function name
```
AlexanderSher: Wrong comment?

src/Package/Impl/Repl/Commands/ReplCommandController.cs
```diff
@@ -152,10 +152,12 @@ public ReplCommandController(ITextView textView, ITextBuffer textBuffer)
             }
         }
 
-        private void HandleCancel(RCompletionController controller) {
-            Workflow.Operations.CancelAsync().DoNotWait();
-            // Post interrupt command which knows if it can interrupt R or not
-            VsAppShell.Current.PostCommand(RGuidList.RCmdSetGuid, RPackageCommandId.icmdInterruptR);
+        private void HandleCancel(RCompletionController controller, ITextView textView) {
```
AlexanderSher: `TextView` is a field of the same class. Did you want to make this method static?

src/R/Editor/Impl/Signatures/ParametersInfo.cs
MikhailArkhipov: Yes

src/Package/Impl/Repl/Commands/ReplCommandController.cs
MikhailArkhipov: I'll use the field instead

src/Package/Impl/Repl/Commands/ReplCommandController.cs
```diff
@@ -152,10 +152,12 @@ public ReplCommandController(ITextView textView, ITextBuffer textBuffer)
...
+            if (!controller.HasActiveCompletionSession && !controller.HasActiveSignatureSession(textView)) {
```
AlexanderSher: I don't understand this. `InterruptRCommand` command calls `_session.CancelAllAsync`, which is also called from `Workflow.Operations.CancelAsync()`. Why do we call it twice?

MikhailArkhipov: InterruptR may or may not work. Example is when typing incomplete expression user wants out and hits ESC. It should return user to `>`. However, InterruptR doesnt work for that. Now, if there is a long running process then InterruptR works (and clears pending inputs). Also, InterruptR is disabled when in debugger break mode but ESC should still work.

AlexanderSher:
👍

PR #3179, opened 2017-02-11
Fixes the bug with credential blob storage.

karthiknadig:


src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -17,21 +17,26 @@ internal class RemoteCredentialsDecorator : ICredentialsDecorator {
...
 
         public RemoteCredentialsDecorator(string credentialAuthority, string workspaceName, ISecurityService securityService, IMainThread mainThread) {
             _securityService = securityService;
             _mainThread = mainThread;
             _authority = credentialAuthority;
             _lock = new AsyncReaderWriterLock();
-            _credentialsAreValid = true;
+            _credentialsLock = new object();
             _workspaceName = workspaceName;
+            _credentials = null;
```
AlexanderSher: `_credentials` is null by default

src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -41,24 +46,25 @@ internal class RemoteCredentialsDecorator : ICredentialsDecorator {
...
 
-            try {
-                _credentials = await _securityService.GetUserCredentialsAsync(_authority, _workspaceName, cancellationToken);
-            } catch (Exception) {
-                token.Dispose();
-                throw;
+            lock (_credentialsLock) {
+                try {
+                    _credentials = _credentials ?? _securityService.GetUserCredentials(_authority, _workspaceName, cancellationToken);
+                    SecurityUtilities.WriteCredentials(_authority, _credentials);
```
AlexanderSher: Do you need to call `WriteCredentials` under lock?

src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -41,24 +46,25 @@ internal class RemoteCredentialsDecorator : ICredentialsDecorator {
...
             return Disposable.Create(() => {
                 token.Dispose();
             });
         }
 
         public void InvalidateCredentials() {
-            Volatile.Write(ref _credentialsAreValid, false);
-            SecurityUtilities.DeleteCredentials(_authority);
+            lock (_credentialsLock) {
+                SecurityUtilities.DeleteCredentials(_authority);
```
AlexanderSher: Do you need to set `_credentials` to `null`?

src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -17,21 +17,26 @@ internal class RemoteCredentialsDecorator : ICredentialsDecorator {
...
-        private bool _credentialsAreValid;
         private readonly string _workspaceName;
 
         public RemoteCredentialsDecorator(string credentialAuthority, string workspaceName, ISecurityService securityService, IMainThread mainThread) {
             _securityService = securityService;
             _mainThread = mainThread;
             _authority = credentialAuthority;
             _lock = new AsyncReaderWriterLock();
-            _credentialsAreValid = true;
+            _credentialsLock = new object();
```
AlexanderSher: There is already lock in this class - `_lock`. Since you need both sync and async locking, you can use `SemaphoreSlim` instead of `AsyncReaderWriterLock`

src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
karthiknadig: If we set _credentials to null upon invalidation then yes this has to be written under the lock.

src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -38,31 +32,22 @@ internal class RemoteCredentialsDecorator : ICredentialsDecorator {
             }
         }
 
-        public async Task<IDisposable> LockCredentialsAsync(CancellationToken cancellationToken = default(CancellationToken)) {
+        public async Task LockCredentialsAsync(CancellationToken cancellationToken = default(CancellationToken)) {
```
AlexanderSher: This method doesn't lock anything anymore (and so the comment below isn't relevant). The purpose of the lock and returned `IDisposable` was to prevent concurrent change of credentials. I don't know if it is still required, but there should be a different name for the method.

int19h: Yeah, this looks more like `GetCredentialsFromUser` or something like that now.

AlexanderSher: Do you think we still need that lock functionality?

karthiknadig: The locking mechanism is needed for invalidation.

karthiknadig:
Ping.

karthiknadig:
Merging this since we have already discussed this for 1_0.

PR #3183, opened 2017-02-13
Surgical fix for the password bug.

karthiknadig:
This is a surgical fix for 1_0

### DIFF
--- a/src/Common/Core/Impl/Security/SecurityService.cs
+++ b/src/Common/Core/Impl/Security/SecurityService.cs
@@ -23,12 +23,12 @@ namespace Microsoft.Common.Core.Security {
             _coreShell = coreShell;
         }
 
-        public Task<Credentials> GetUserCredentialsAsync(string authority, string workspaceName, CancellationToken cancellationToken = default(CancellationToken)) {
+        public Credentials GetUserCredentials(string authority, string workspaceName, CancellationToken cancellationToken = default(CancellationToken)) {
             _coreShell.AssertIsOnMainThread();
 
             var credentials = SecurityUtilities.ReadCredentials(authority);
             if (credentials != null) {
-                return Task.FromResult(credentials);
+                return credentials;
             }
 
             var credui = new CREDUI_INFO {
@@ -66,7 +66,7 @@ namespace Microsoft.Common.Core.Security {
                     throw new Win32Exception(Marshal.GetLastWin32Error());
                 }
 
-                return Task.FromResult(Credentials.CreateCredentails(userNameBuilder.ToString(), SecurityUtilities.SecureStringFromNativeBuffer(passwordStorage), save));
+                return Credentials.CreateCredentails(userNameBuilder.ToString(), SecurityUtilities.SecureStringFromNativeBuffer(passwordStorage), save);
             } finally {
                 if (inCredBuffer != IntPtr.Zero) {
                     Marshal.FreeCoTaskMem(inCredBuffer);
--- a/src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
+++ b/src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
@@ -16,12 +16,11 @@ namespace Microsoft.R.Host.Client.Host {
     internal class RemoteCredentialsDecorator : ICredentialsDecorator {
         private readonly ISecurityService _securityService;
         private readonly IMainThread _mainThread;
-        private Credentials _credentials;
+        private volatile Credentials _credentials;
         private readonly string _authority;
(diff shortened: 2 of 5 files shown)


AlexanderSher:
So this is pretty much PR #3179 with the same double-locking.
Can you give the repro steps for the bug and explain why this lock fixes it?

MikhailArkhipov:
Some users cant't log on to remote and getting 401. This was identified as a race condition. Hence the fix.

MikhailArkhipov:
👍 @AlexanderSher - need 2 reviews

AlexanderSher:
Ok, there can't be race between two calls of `LockCredentialsAsync` (this method is protected by lock already) or `LockCredentialsAsync` and `InvalidateCredentials` (the second one is called inside the lock that is acquired by the first one). So the race was between what and what?

src/Common/Core/Impl/Security/SecurityService.cs
```diff
@@ -23,12 +23,12 @@ public class SecurityService : ISecurityService {
             _coreShell = coreShell;
         }
 
-        public Task<Credentials> GetUserCredentialsAsync(string authority, string workspaceName, CancellationToken cancellationToken = default(CancellationToken)) {
+        public Credentials GetUserCredentials(string authority, string workspaceName, CancellationToken cancellationToken = default(CancellationToken)) {
             _coreShell.AssertIsOnMainThread();
 
             var credentials = SecurityUtilities.ReadCredentials(authority);
```
MikhailArkhipov: Actually, should this be under the lock?

MikhailArkhipov: nvm

karthiknadig:
We can further minimize the fix to just adding the null.

MikhailArkhipov:
`GetCredential` and `LockCredential`

karthiknadig:
@AlexanderSher Do you recommend any changes?

MikhailArkhipov:
Or @int19h can help with 2nd review. Need to kick 1.0 RC1 build.

AlexanderSher:
> `GetCredential` and `LockCredential`

I can't find any code that calls `GetCredential`. Is it used at all?

src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -21,6 +21,7 @@ internal class RemoteCredentialsDecorator : ICredentialsDecorator {
         private readonly AsyncReaderWriterLock _lock;
         private bool _credentialsAreValid;
         private readonly string _workspaceName;
+        private object _credentialsLock;
```
int19h: Why does this need a new lock? There's already `_lock` above, and `LockCredentialsAsync` already takes that lock. So it would seem that `GetCredential` and `InvalidateCredentials` should just do the same.

MikhailArkhipov: Actually yes, I missed this in the diff-only. Doesn't `_credentialsLock` remove need for the token?

AlexanderSher: `InvalidateCredentials` is already called inside the lock, and `GetCredential` belongs to the framework interface, so I'm not sure if it is safe to put the lock inside it.

karthiknadig: I did a bunch of tests. Taking the change just to the Write Credentials seems to be enough. I was not able to hit this issue after making just that. However, it requires deleting the credential entry (either by removing the work-space and re-adding it or deleting it from credential manager).

AlexanderSher: What I think we should do is make _credentials volatile, implement GetCredential like this:
```cs
public NetworkCredential GetCredential(Uri uri, string authType) {
    var credentials = _credentials;
    return credentials != null ? new NetworkCredential(credentials.UserName, credentials.Password) : new NetworkCredential();
}
```

and don't add any new locks.

karthiknadig: ok

MikhailArkhipov: FYI this is port of the tested fix from master. So we need to be careful here and not diverge. I'd suggest going same path as exists in master, then mod master per suggestions for the BB on RC2.

karthiknadig:
@AlexanderSher I have made the changes you suggested. If there are no more comments, I will check this in.

AlexanderSher:
👍

Issue #3196, opened 2017-02-14
Logging options combo box empty


huguesv:
![image](https://cloud.githubusercontent.com/assets/1696845/22952608/8021700a-f2c2-11e6-98f4-23eb9ee14091.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/79e4ef08a4dfc226d3f293f8ebd813eca2756d76

Issue #3195, opened 2017-02-14
REvaluationException crashed VS


huguesv:
I was connected to local R, I had just detached the REPL from debugger not long before it crashed.
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.R.Host.Client.REvaluationException
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions.ThrowOnError(System.String, Microsoft.R.Host.Client.REvaluationResult)
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions+<EvaluateAsync>d__1`1[[System.__Canon, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.DataInspection.RSessionExtensions+<DescribeChildrenAsync>d__4.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Editor.Data.RSessionDataObject+<GetChildrenAsyncInternal>d__21.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.VisualStudio.R.Package.DataInspect.WorkspaceVariableProvider+<UpdateList>d__11.MoveNext()
   at Microsoft.Common.Core.TaskExtensions+<>c.<DoNotWaitThreadContinuation>b__6_0(System.Object)
   at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
[...]

Issue #3193, opened 2017-02-14
Add database connection, cancel, and it still adds it to your settings.R


huguesv:
Create new R project
RTools, Data, Add Database Connection
Cancel

Result: Settings.R is created, with a database connection in it.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/79e4ef08a4dfc226d3f293f8ebd813eca2756d76

PR #3201, opened 2017-02-16
Bug fixes

### LINKED ISSUE #3186: Shortcuts don't show up in menus RTools > Windows
image

Ignore High-contrast.

MikhailArkhipov:
#3186 Shortcuts don't show up in menus RTools > Windows
#3196 Logging options combo box empty (with test)
#3195 REvaluationException crashed VS
#3193 Add database connection, cancel, and it still adds it to your settings

### DIFF
--- a/src/Common/Core/Impl/Logging/Implementation/LoggingPermissions.cs
+++ b/src/Common/Core/Impl/Logging/Implementation/LoggingPermissions.cs
@@ -89,7 +89,7 @@ namespace Microsoft.Common.Core.Logging {
 
         private int? GetValueFromRegistry(string name, int minValue, int maxValue) {
             if(_appConstants.LocalMachineHive == null) {
-                return minValue;
+                return maxValue;
             }
             using (var hlkm = _registry.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32)) {
                 try {
--- a/src/Package/Impl/Options/Attributes/EnumTypeConverter.cs
+++ b/src/Package/Impl/Options/Attributes/EnumTypeConverter.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
-using System.Resources;
 
 namespace Microsoft.VisualStudio.R.Package.Options.Attributes {
     internal class EnumTypeConverter<T> : TypeConverter {
@@ -63,7 +62,7 @@ namespace Microsoft.VisualStudio.R.Package.Options.Attributes {
         }
 
         public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType) {
-            if (value.GetType() == destinationType) {
+            if (value == null || value.GetType() == destinationType) {
                 return value;
             }
 
--- a/src/Package/Impl/Options/Attributes/EnumTypeConverter.cs
+++ b/src/Package/Impl/Options/Attributes/EnumTypeConverter.cs
@@ -62,7 +62,7 @@ namespace Microsoft.VisualStudio.R.Package.Options.Attributes {
         }
 
         public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType) {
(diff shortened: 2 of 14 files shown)


MikhailArkhipov:
Ping so I can kick build for the BB

src/Common/Core/Impl/Logging/Implementation/LoggingPermissions.cs
```diff
@@ -89,7 +89,7 @@ public sealed class LoggingPermissions : ILoggingPermissions {
 
         private int? GetValueFromRegistry(string name, int minValue, int maxValue) {
             if(_appConstants.LocalMachineHive == null) {
-                return minValue;
+                return maxValue;
```
AlexanderSher: Does it mean we will have `maxValue` by default?

src/Package/Impl/Options/Attributes/EnumTypeConverter.cs
```diff
@@ -63,7 +62,7 @@ protected EnumTypeConverter(params string[] localizedDisplayNames)
         }
 
         public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType) {
-            if (value.GetType() == destinationType) {
+            if (value == null || value.GetType() == destinationType) {
```
AlexanderSher: `value?.GetType() == destinationType`

src/R/Editor/Impl/Data/RSessionDataObject.cs
```diff
@@ -108,22 +108,24 @@ public class RSessionDataObject : IRSessionDataObject {
...
-                    AttributeCountProperty |
-                    DimProperty |
-                    FlagsProperty |
-                    (RToolsSettings.Current.EvaluateActiveBindings ? ComputedValueProperty : 0);
-                var children = await valueEvaluation.DescribeChildrenAsync(properties, RValueRepresentations.Str(MaxReprLength), MaxChildrenCount);
-                return EvaluateChildren(children);
-            }
+            try {
+                if (valueEvaluation.HasChildren) {
+                    await TaskUtilities.SwitchToBackgroundThread();
```
AlexanderSher: Let's move `await TaskUtilities.SwitchToBackgroundThread();` before try/catch block, otherwise `catch` will switch to UI thread.

src/Common/Core/Impl/Logging/Implementation/LoggingPermissions.cs
MikhailArkhipov: This is test only case (no registry)

src/Package/Impl/Options/Attributes/EnumTypeConverter.cs
MikhailArkhipov: OK

src/Package/Impl/Options/R/RToolsOptionsPage.cs
```diff
@@ -198,19 +198,15 @@ public class RToolsOptionsPage : DialogPage {
...
             get {
-#if DEBUG
-                return _holder.GetValue<LogVerbosity>(LogVerbosity.Traffic);
-#else
-                return _holder.GetValue<LogVerbosity>(LogVerbosity.Normal);
-#endif
+                var permissions = VsAppShell.Current.Services.LoggingPermissions;
+                var value = _holder.GetValue<LogVerbosity>(permissions.MaxVerbosity);
+                value = MathExtensions.Min<LogVerbosity>(value, permissions.MaxVerbosity);
+                _holder.SetValue(value);
```
AlexanderSher: Doesn't really understand this piece. Shouldn't we restrict max value in setter?

MikhailArkhipov: This happens if value stored earlier is read by VS and set here. Earlier value may be higher than the most recently set limit in registry.

MikhailArkhipov: Or we can move this to load settings?

karthiknadig:
👍

Issue #3194, opened 2017-02-14
When package is installed intellisense doesn't always pick it up


MikhailArkhipov:
Sometimes when I freshly install dplyr it doesn't show up in intellisense. Restart VS and it comes up.

MikhailArkhipov:
Need to figure out unit test

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/b0b2a48a889c3c7e263913f0e9cdc1170ab1774f

PR #3210, opened 2017-02-20
#3194 When package is installed intellisense doesn't always pick it up

MikhailArkhipov:


src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -201,38 +200,55 @@ public sealed class PackageIndex : IPackageIndex {
             }
         }
 
+        private async Task UpdateInstalledPackagesAsync() {
+            var token = await _buildIndexLock.WaitAsync();
```
AlexanderSher: - `token` isn't released, so there can be a deadlock
- I think there should be `ResetAsync` instead of `WaitAsync`

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -201,38 +200,55 @@ public sealed class PackageIndex : IPackageIndex {
...
+                    info = await AddPackagesToIndexAsync(packagesNotInIndex);
+                } catch (RHostDisconnectedException) { }
+            }
+            return info;
+        }
 
-                        await info.LoadFunctionsIndexAsync();
-                        _functionIndex.RegisterPackageFunctions(info);
+        private async Task<IEnumerable<IPackageInfo>> AddPackagesToIndexAsync(IEnumerable<RPackage> packages) {
+            var list = new List<IPackageInfo>();
```
AlexanderSher: If `List` here is important, I'd rather make this method return `Task<IList<IPackageInfo>>`

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -201,38 +200,55 @@ public sealed class PackageIndex : IPackageIndex {
             }
         }
 
+        private async Task UpdateInstalledPackagesAsync() {
+            var token = await _buildIndexLock.WaitAsync();
+            try {
+                var installed = await GetInstalledPackagesAsync();
+                var installedNames = installed.Select(p => p.Package).Append("rtvs");
+
+                var currentNames = _packages.Keys.ToArray();
```
AlexanderSher: Why do you need `ToArray()`?
`_packages.Keys.Except(installedNames)` should work

MikhailArkhipov: Since it is concurrent bag, I wanted to capture its current, rather than changing, state.

src/R/Support/Impl/Help/Packages/PackageIndex.cs
MikhailArkhipov: Doesn't matter much. Would you rather prefer `yield return'? I just don't want to change logic too much at this point.

src/R/Support/Impl/Help/Packages/PackageIndex.cs
MikhailArkhipov: ResetAsync makes code think that the index is not ready. Basically it is not set when index is empty and set otherwise. Technically states are 'empty', 'ready' and 'needs update'. Perhaps I should add `finally` then

src/R/Support/Impl/Help/Packages/PackageIndex.cs
AlexanderSher: No, `yield return` over the async call may change behavior significantly.

src/R/Support/Impl/Help/Packages/PackageIndex.cs
AlexanderSher: What's the difference between `empty` and `need update`?

MikhailArkhipov: Before code was doing `_buildIndexLock.EnqueueReset();` and then trying to rebuild the entire index. Now it only handles change in the set of packages, leaving the rest of the index intact.

src/R/Support/Impl/Help/Packages/PackageIndex.cs
MikhailArkhipov: I should probably do `installedNames.ToArray/List` as well then

src/R/Support/Impl/Help/Packages/PackageIndex.cs
AlexanderSher: Yes, it tried to rebuild entire index but not because of the lock, but because `ScheduleIdleTimeRebuild` calls `RebuildIndexAsync`, which isn't called anymore. If you don't call `ResetAsync`, there can be concurrent calls to the `UpdateInstalledPackagesAsync` which may impact performance.
So I believe it should be 
```cs
var token = await _buildIndexLock.ResetAsync();
if (!token.IsSet) {
    try {
...
        token.Set();
    } finally {
        token.Reset();
    }
}
```

src/R/Support/Impl/Help/Packages/PackageIndex.cs
AlexanderSher: `installed.ToList()` I think.

src/R/Support/Impl/Help/Packages/PackageIndex.cs
MikhailArkhipov: OK, makes sense

MikhailArkhipov:
Ping. Anything else?

AlexanderSher:
👍

PR #3221, opened 2017-02-22
Fix #3205: VE crash for objects with length exceeding 32-bit

### LINKED ISSUE #3205: VE crash for objects with length exceeding 32-bit
Open Variable Explorer, and do:
```r
m <- matrix(0L, .Machine$integer.max / 9, 10)
```
Note that you'll need enough RAM to actually allocate this matrix, and it will take a while.

When VE refreshes to show the new variable, it crashes with `OverflowException` "Value was either too large or too small for an Int32." when extracting evaluation result data from JSON:

```
Microsoft.R.Host.Client.dll!Microsoft.R.DataInspection.RValueInfo.RValueInfo(Microsoft.R.Host.Client.IRExpressionEvaluator evaluator, string environmentExpression, string expression, string name, Newtonsoft.Json.Linq.JObject json) Line 101	C#
Microsoft.R.Host.Client.dll!Microsoft.R.DataInspection.REvaluationResultInfo.Parse(Microsoft.R.Host.Client.IRExpressionEvaluator evaluator, string environmentExpression, string name, Newtonsoft.Json.Linq.JObject json) Line 43	C#
Microsoft.R.Host.Client.dll!Microsoft.R.DataInspection.RSessionExtensions.DescribeChildrenAsync(Microsoft.R.Host.Client.IRExpressionEvaluator session, string environmentExpression, string expression, Microsoft.R.DataInspection.REvaluationResultProperties properties, string repr, int? maxCount, System.Threading.CancellationToken cancellationToken) Line 168	C#
```

The problem is that we assume that length is 32-bit there:
```cs
Length = json.Value<int?>(FieldNames.Length);
```
which is obviously not the case for really large vectors of matrices.

Note that this can also potentially apply to `AttributeCount`, `SlotCount` and `NameCount`, since these are all vectors. It might also apply to `Dim`.

int19h:
Use 64-bit integers for number of elements / rows / columns, and the corresponding indices.

Guard against objects with extremely large extents in one or more dimensions.

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/eval.R
+++ b/src/Host/Client/Impl/rtvs/R/eval.R
@@ -39,6 +39,12 @@ describe_object <- function(obj, res, fields, repr = NULL) {
 
     if (field('classes')) {
         classes <- NA_if_error(class(obj));
+
+        # Since the entire class vector is serialized, restrict exceedingly large numbers of classes.
+        if (length(classes) > 1000) {
+            classes <- classes[1:1000];
+        }
+
         res$classes <- lapply(classes, function(cls) {
             if (!is.character(cls) || length(cls) != 1 || is.na(cls)) NA else cls
         });
@@ -62,7 +68,8 @@ describe_object <- function(obj, res, fields, repr = NULL) {
 
     if (field('dim')) {
         dim <- NA_if_error(dim(obj));
-        if (is.integer(dim) && !anyNA(dim)) {
+        # Since the entire dimension vector is serialized, restrict exceedingly large numbers of dimensions.
+        if (is.integer(dim) && !anyNA(dim) && length(dim) < 1000) {
             res$dim <- as.list(dim);
         }
     }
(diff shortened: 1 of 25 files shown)


karthiknadig:
👍

src/Host/Client/Impl/rtvs/R/eval.R
```diff
@@ -39,6 +39,12 @@ describe_object <- function(obj, res, fields, repr = NULL) {
 
     if (field('classes')) {
         classes <- NA_if_error(class(obj));
+
+        # Since the entire class vector is serialized, restrict exceedingly large numbers of classes.
+        if (length(classes) > 1000) {
+            classes <- classes[1:1000];
```
AlexanderSher: Shouldn't we do this on the RHost side?

AlexanderSher:
Would be great to have couple of tests

MikhailArkhipov:
Should probably add some tests for large sizes/overflows. 👍

src/Host/Client/Impl/rtvs/R/eval.R
int19h: This is the R-Host side, since that's where R code runs.

int19h:
Tests are a bit tricky, because it requires actually creating that humongous matrix that would otherwise crash us, which is slow. I'll see if I can simulate that with a custom type with a redefined indexer.

MikhailArkhipov:
That's OK then for now. Get the fix in and file new sprint 12 task.

int19h:
https://github.com/Microsoft/RTVS/issues/3225

PR #3228, opened 2017-02-23
Fix #3209: After running tests R.Host.exe processes do not terminate

### LINKED ISSUE #3209: After running tests R.Host.exe processes do not terminate
There is no client so they should be auto-terminating I think... Should they?

`is_terminating` is true, `response_state == RESPONSE_EXPECTED`

but

```
            for (;;) {
                message msg;
                for (;;) { 
```

never terminates.

int19h:
Do not request localized strings from client in autosave-on-shutdown code.

Dispose broker when it is removed from the session provider.

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/util.R
+++ b/src/Host/Client/Impl/rtvs/R/util.R
@@ -299,9 +299,12 @@ query_reload_autosave <- function() {
 }
 
 save_state <- function() {
-    message(sprintf(locstr('rtvs_AutosavingWorkspace'), autosave_filename));
+    # This function runs when client is already disconnected, so locstr() cannot be used, since it requires
+    # a connected client to provide translated strings. However, since messages are not user-visible and are
+    # here for logging purposes only, they don't need to be localized in the first place.
+    message(sprintf('Autosaving workspace to image "%s" ...', autosave_filename), appendLF = FALSE);
     save.image(autosave_filename);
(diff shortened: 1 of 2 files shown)


int19h:
Also see https://github.com/Microsoft/R-Host/pull/112

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -299,9 +299,12 @@ query_reload_autosave <- function() {
 }
 
 save_state <- function() {
-    message(sprintf(locstr('rtvs_AutosavingWorkspace'), autosave_filename));
+    # This function runs when client is already disconnected, so locstr() cannot be used, since it requires
+    # a connected client to provide translated strings. However, since messages are not user-visible and are
+    # here for logging purposes only, they don't need to be localized in the first place.
+    message(sprintf('Autosaving workspace to image "%s" ...', autosave_filename), appendLF = FALSE);
```
MikhailArkhipov: localize?

MikhailArkhipov: nvm, I see

MikhailArkhipov:
👍

AlexanderSher:
👍

src/Host/Client/Impl/rtvs/R/util.R
int19h: See the comment :)

Issue #3217, opened 2017-02-21
Sometime package manager appears empty


MikhailArkhipov:
Switching between tabs populates it

![image](https://cloud.githubusercontent.com/assets/12820357/23182401/4474a692-f82d-11e6-807a-6a158e9e2bc5.png)

AlexanderSher:
Any specific steps or conditions?

MikhailArkhipov:
nope, just connected to local and opened it. Seen this periodically but not too often, looks like async issue. Trying to debug.

MikhailArkhipov:
So the repro is

- make sure all windows except Workspaces is closed
- Start VS

You will get BP hit in `GetInstalledPackagesAsync`. But since there is no connection, I guess it ends with nothing.

- Open PM

Two BPs hit: `GetInstalledPackagesAsync` and then `GetAvailablePackagesAsync`. 

The window remains empty.

MikhailArkhipov:
Interestingly enough, if I 

1. Open PM (with local R active)
2. Close VS
3. Start VS

PM window appears with 'loading' and then auto-closes. B/c there is no R connected?

MikhailArkhipov:
There are 2 concurrent executions of `GetPackagesAsync` on the same session since `GetInstalledPackagesAsync` and `GetAvailablePackagesAsync` operate concurrently.

AlexanderSher:
That is fine, we can get into this state by switching tabs fast enough

MikhailArkhipov:
But is it OK to concurrently start and initialize sessions?

```
                await _session.EnsureHostStartedAsync(new RHostStartupInfo(), null, cancellationToken: cancellationToken);
                await _session.SetVsCranSelectionAsync(CranMirrorList.UrlFromName(_settings.CranMirror), cancellationToken);
                await _session.SetCodePageAsync(_settings.RCodePage, cancellationToken);
```

AlexanderSher:
Session is initialized after `EnsureHostStartedAsync` already.
`SetVsCranSelectionAsync` and `SetCodePageAsync` are called to update code page and CRAN mirror from REPL session.

MikhailArkhipov:
`GetInstalledPackagesAsync` is not called at all when this reproes

AlexanderSher:
`LoadInstalledAndLoadedPackagesAsync` fails with the following error:
```
Error: R evaluation failed:

rtvs:::packages.installed()

Error in loadNamespace(name) : 
[...]

Issue #3226, opened 2017-02-22
Need better message and handling of R connectivity in package manager


MikhailArkhipov:
1. Open PM when R is not connected:

![image](https://cloud.githubusercontent.com/assets/12820357/23227090/bb5d92a4-f8ec-11e6-8474-189567cdbd29.png)

- We should give better message since users don't know what 'R Host process' is. Like `R is not running` or `Workspace disconnected` or something.

Now try connecting to local R

You may see (sometimes it goes away but sometimes sticks) 'R Evaluation failed' message.

MikhailArkhipov:
Just the message change https://github.com/Microsoft/RTVS/commit/a068c51d930f91f4c1b5d277c46d8faa2478803f

PR #3230, opened 2017-02-23
Theming and high contrast fixes

### LINKED ISSUE #3189: High contrast in package manager has invisible text
Using `High Contrast #1` theme on Windows 10.

The mouse hover text is white on white.

image

MikhailArkhipov:
#3189 High contrast in package manager has invisible text 
#3226 Need better message and handling of R connectivity in package manager

- Fixes highlight on hover in connections
- Fixes image background in dark/light theme switching (use standard VS icons)

### DIFF
--- a/src/Package/Impl/DataInspect/VariableView.xaml.cs
+++ b/src/Package/Impl/DataInspect/VariableView.xaml.cs
@@ -49,19 +49,16 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
         public VariableView(IRToolsSettings settings, ICoreShell shell) {
             _settings = settings;
             _shell = shell;
+            _shell.UIThemeChanged += OnUIThemeChanged;
 
             InitializeComponent();
+            SetImageBackground();
 
             _aggregator = _shell.ExportProvider.GetExportedValue<IObjectDetailsViewerAggregator>();
-
             SetRootNode(VariableViewModel.Ellipsis);
 
             SortDirection = ListSortDirection.Ascending;
             RootTreeGrid.Sorting += RootTreeGrid_Sorting;
-
-            var theme = shell.ExportProvider.GetExportedValue<IThemeUtilities>();
-            theme.SetImageBackgroundColor(RootTreeGrid, Brushes.ToolWindowBackgroundColorKey);
-            theme.SetThemeScrollBars(RootTreeGrid);
 
             var workflow = VsAppShell.Current.ExportProvider.GetExportedValue<IRInteractiveWorkflowProvider>().GetOrCreate();
             _session = workflow.RSession;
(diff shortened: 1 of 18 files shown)


src/Package/Impl/DataInspect/VariableView.xaml.cs
```diff
@@ -71,6 +68,16 @@ public partial class VariableView : UserControl, ICommandTarget, IDisposable {
             _environmentProvider.RefreshEnvironmentsAsync().DoNotWait();
         }
 
+        private void OnUIThemeChanged(object sender, EventArgs e) {
+            SetImageBackground();
+        }
+
+        private void SetImageBackground() {
```
MikhailArkhipov: This fragment is duplicated across 3 controls. We could define some `ThemedUserControl` and derive from it, but that would be a much more extensive change.

AlexanderSher: We can use an `AttachedProperty` for that

src/Package/Impl/Shell/VsAppShell.cs
```diff
@@ -448,18 +453,32 @@ public sealed class VsAppShell : IApplicationShell, IMainThread, IIdleTimeServic
...
+                if (_vsShellBroadcastEventsCookie != 0) {
+                    _vsShell.UnadviseBroadcastMessages(_vsShellBroadcastEventsCookie);
+                    _vsShellBroadcastEventsCookie = 0;
+                }
             }
         }
 
+        #region IVsBroadcastMessageEvents
+        public int OnBroadcastMessage(uint msg, IntPtr wParam, IntPtr lParam) {
+            if (msg == 0x15 /* WM_SYSCOLORCHANGE */) {
```
AlexanderSher: Let's define it as a constant

src/Package/Impl/DataInspect/VariableView.xaml.cs
MikhailArkhipov: I mean, package event handling in the `ThemedUserControl`

src/Package/Impl/Shell/VsAppShell.cs
MikhailArkhipov: Sure

src/Package/Impl/DataInspect/VariableView.xaml.cs
AlexanderSher: Yes, but you can encapsulate it inside `AttachedProperty` and set it in ctor of the control like:
```cs
UITheme.SetShell(this, _shell);
```

This is better than adding a base class because we may need more than one base class for it

src/R/Components/Impl/Resources.resx
```diff
@@ -304,7 +304,7 @@
     <value>R host process is not available.</value>
   </data>
   <data name="PackageManager_TransportError" xml:space="preserve">
-    <value>Connection to R host process failed.</value>
+    <value>Workspace is disconnected or R session is not running.</value>
```
AlexanderSher: To fix #3217, I've split this message into three (individual for every tab), so you may skip this change.
Up to you.

src/Package/Impl/DataInspect/VariableView.xaml.cs
AlexanderSher: You can assign a refactoring task on me, for now I think it's good to go.

src/R/Components/Impl/PackageManager/Implementation/View/PackageList.xaml
```diff
@@ -143,7 +145,7 @@
                                         </TextBlock.Text>
                                     </TextBlock>
                                 </Button.ToolTip>
-                                <Rectangle Fill="{StaticResource IconUpdateAvailableIndicator}" Width="16" Height="16" />
+                                <imaging:CrispImage Width="16" Height="16" Moniker="{x:Static imagecatalog:KnownMonikers.Upload}" />
```
AlexanderSher: Are they similar to the old icons?

MikhailArkhipov: image

image

AlexanderSher:
👍

PR #3236, opened 2017-02-24
Fix #3217: Sometime package manager appears empty

### LINKED ISSUE #3217: Sometime package manager appears empty

AlexanderSher:
The reason why we get "cyclic namespace dependency" error isn't clear, especially after CRAN mirror was successfully updated, but error message now remain visible until list installed packages are loaded or explicitly dismissed.


MikhailArkhipov:
👍

Issue #3232, opened 2017-02-23
R Tools Default Working Directory


tlong296:
By default R Tools for Visual Studio sets the working directory to whatever ~ references. On a lot of machines whatever ~ references is not what is needed as the working directory in order to get RevoScaleR and MicrosoftML to work appropriately.

The work around for this is to not-so-obviously run the command setwd("c:\\users\\username"). This will solve any strange error you receive such as "file.exists(defaultDataDir) is not TRUE". It took me 3 weeks to figure out this error. The fact that the working directory has to be set to c:\users\username by default is still curious to me since R Client and its libraries do not live anywhere in that directory.

The underlying issue in R Tools is that you cannot change the default working directory. I find this really odd and almost unbelievable, like I've overlooked how to do this, but as I continue to look I cannot find a place to change the default working directory. 

This should be added to the tool. I prefer using R Studio right now for the sole reason that I don't have to run setwd(......) at the beginning of every session (Since it let me change the default working directory).

MikhailArkhipov:
@AlexanderSher, @int19h  - do you think it is safe not to call `setwd(~)` ?

int19h:
Working directory is a UX concept - we won't break any other code by not doing this.

But also, there was a reason why we decided to do it in the first place ... and presumably whatever scenario was behind it, it's still valid?

If I remember correctly, if we don't do `setwd` at all, it ends up being something really useless - some global directory where you don't have write access, and so cannot create files (I think it was R installation folder?)

It sounds like the fundamental issue here isn't that we always do `setwd`, but rather that we need to let users specify the target directory in options, rather than always assuming `~`.

int19h:
At the same time, I'm curious why `~` did *not* resolve to the same thing as `c:\users\username`. That's where it's supposed to be...

MikhailArkhipov:
I can make it optional with default to `off`. Thoughts?

int19h:
Unless it's going to end up in the desired directory, I don't see how it solves the problem - @tlong296 will still have to manually do `setwd` to get there every time the session starts.

MikhailArkhipov:
Yes, it would only control "don't change whatever is set in R by default"

Hong-Revo:
I think I understand @tlong296 's issue. Most R code will translate `~` to your Documents folder, but RevoScaleR and MicrosoftML have some problems with it on Windows (since `~` is a Unix/Linux concept).

[...]

PR #3239, opened 2017-02-24
#3232 R Tools Default Working Directory 

MikhailArkhipov:
image

### DIFF
--- a/src/Host/Client/Impl/Session/RSessionProvider.cs
+++ b/src/Host/Client/Impl/Session/RSessionProvider.cs
@@ -24,7 +24,6 @@ namespace Microsoft.R.Host.Client.Session {
         private readonly BrokerClientProxy _brokerProxy;
         private readonly ICoreServices _services;
         private readonly IConsole _console;
-        private readonly bool _workingDirectory;
 
         private volatile bool _isConnected;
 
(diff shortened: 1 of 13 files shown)


src/Host/Client/Impl/Session/RSessionProvider.cs
```diff
@@ -24,6 +24,7 @@ public sealed class RSessionProvider : IRSessionProvider {
         private readonly BrokerClientProxy _brokerProxy;
         private readonly ICoreServices _services;
         private readonly IConsole _console;
+        private readonly bool _workingDirectory;
```
int19h: Unused?

int19h:
I might be confused, but I can't see the code that actually uses the new `WorkingDirectory` option?

MikhailArkhipov:
The field was already in settings we just never exposed it. It is used in `AfterHostStarted`

src/Host/Client/Impl/Session/RSessionProvider.cs
MikhailArkhipov: Yep

int19h:
👍

smortaz:
can i specify the initial directory to be where the project is created?

ie: [$Project_dir                       ]

(or whatever it is called)

MikhailArkhipov:
No, no macros.

PR #3240, opened 2017-02-25
Remove forced background paint

MikhailArkhipov:
And simplify main thread paint a bit

### DIFF
--- a/src/Package/Impl/DataInspect/VisualGrid/VisualGridScroller.cs
+++ b/src/Package/Impl/DataInspect/VisualGrid/VisualGridScroller.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks.Dataflow;
 using System.Windows;
 using Microsoft.Common.Core;
 using Microsoft.Common.Core.Logging;
+using Microsoft.Common.Core.Shell;
 using Microsoft.VisualStudio.PlatformUI;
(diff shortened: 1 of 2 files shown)


src/Package/Impl/DataInspect/VisualGrid/VisualGridScroller.cs
```diff
@@ -197,22 +195,19 @@ internal sealed class VisualGridScroller {
                     throw new InvalidOperationException("Couldn't acquire enough data");
                 }
 
-                // actual drawing runs in UI thread
-                await Task.Factory.StartNew(
-                    () => DrawVisuals(newViewport, data, updateType, visualViewport, suppressNotification),
-                    token,
-                    TaskCreationOptions.None,
-                    _ui);
+                await DrawVisualsAsync(newViewport, data, updateType, visualViewport, suppressNotification);
```
AlexanderSher: Need to pass `token` to the `SwitchToMainThreadAsync`

MikhailArkhipov: Sure

MikhailArkhipov:
This change in attempt to fix grid being blank while data is actually there.

AlexanderSher:
👍

Issue #3158, opened 2017-02-09
.Net Standard 2.0 Dependencies


AlexanderSher:
This issue contains a list of types that have to be revised when .Net Standard 2.0 is published. NETSTANDARD1_6 conditional symbol is used for `#if` directive. 
- `Microsoft.Common.Core.Threading.BackgroundThreadAwaiter` - `ThreadPool.UnsafeQueueUserWorkItem` doesn't exist
- `Microsoft.Common.Core.Security.Credentials` - `NetworkCredential` doesn't have a ctor that accepts secure string
- `Microsoft.Common.Core.ExceptionExtensions` - `StackOverflowException`, `ThreadAbortException` and `AccessViolationException` don't exist 
- `Microsoft.Common.Core.Logging.FileLogWriter` - `AppDomain.CurrentDomain` is excluded from .Net Standard build
- `Microsoft.Common.Core.IO.IUserProfileNamedPipeFactory` - `NamedPipeServerStream.MaxAllowedServerInstances` doesn't exist
- `Microsoft.Common.Core.TaskUtilities` - `Thread.IsThreadPoolThread` doesn't exist

AlexanderSher:
- `Microsoft.Languages.Core.Text.CharacterStream` - `ExcludeFromCodeCoverage` is missing.
- `Microsoft.Languages.Core.Text.TextChangeEventArgs` - `ExcludeFromCodeCoverage` is missing.
- `Microsoft.Languages.Core.Text.TextRange` - `ExcludeFromCodeCoverage` is missing.

MikhailArkhipov:
Yeah, I was going to remove these attributes. In fact, I have a branch where it is sort of done.

AlexanderSher:
So I can just remove them as well?

MikhailArkhipov:
Yes

MikhailArkhipov:
Folding into #4086

PR #3246, opened 2017-02-28
Extract VS-dependent and platform-dependent code from Common.Core for out-of-proc

AlexanderSher:


src/Common/Core/Impl/Microsoft.Common.Core.DNC.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
MikhailArkhipov: How does it get built in MB, signed and versioned?

src/Common/Core/Impl/Security/Credentials.cs
```diff
@@ -15,77 +12,18 @@ public sealed class Credentials : ICredentials {
...
-            var creds = new Credentials() { UserName = userName, Password = password, Source = source };
+        public static Credentials Create(string userName, SecureString password) {
+            var creds = new Credentials { UserName = userName, Password = password };
             creds.Password.MakeReadOnly();
             return creds;
         }
 
         public NetworkCredential GetCredential(Uri uri, string authType) {
+#if NETSTANDARD1_6
+            return new NetworkCredential(UserName, Password.ToUnsecureString());
```
MikhailArkhipov: eeek

src/Languages/Editor/Impl/Microsoft.Languages.Editor.csproj
```diff
@@ -156,6 +156,10 @@
       <Project>{8d408909-459f-4853-a36c-745118f99869}</Project>
       <Name>Microsoft.Common.Core</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\Windows\Core\Impl\Microsoft.Windows.Core.csproj">
```
MikhailArkhipov: Perhaps follow suffix scheme and name Microsoft.Common.Core.Windows?

src/Package/Impl/Browsers/WebBrowserServices.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.VisualStudio.R.Package.Shell;
 using Microsoft.VisualStudio.R.Packages.R;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.Windows.Core.OS;
```
MikhailArkhipov: I kept same namespace in Windows and common i.e. There is no 'using Microsoft.R.Editor.Windows`.

MikhailArkhipov: Up to you though since these may be really OS specific

src/Package/Impl/RClient/MicrosoftRClientInstaller.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.R.Interpreters;
 using Microsoft.VisualStudio.R.Package.Telemetry;
 using Microsoft.VisualStudio.R.Package.Utilities;
+using Microsoft.Windows.Core.Network;
```
MikhailArkhipov: Also, Microsoft.Windows sounds like this is OS assembly

src/Common/Core/Impl/Security/Credentials.cs
AlexanderSher: See https://github.com/Microsoft/RTVS/issues/3158

src/Languages/Editor/Impl/Microsoft.Languages.Editor.csproj
AlexanderSher: Maybe. Btw, what should be the name for the assembly?

MikhailArkhipov: I changed editor stuff to

`Microsoft.R.Editor` - common code
`Microsoft.R.Editor.Windows` - OS specific

namespace is the same in both and is `Microsoft.R.Editor`

So - `Microsoft.R.Common.Core` & `Microsoft.R.Common.Core.Windows` ?

MikhailArkhipov: And namespace is `Microsoft.Common.Core` like before?

src/Common/Core/Impl/Security/Credentials.cs
MikhailArkhipov: Yeah I know. Still eeeew. :-)

src/Languages/Editor/Impl/Microsoft.Languages.Editor.csproj
AlexanderSher: Ok, though I'm not sure about namespace. Extracted code is indeed platform-specific.

MikhailArkhipov: I kept NS the same since OS-specific part is effectively an extension to the common code so, say, in the package we won't have to have `using ABC` and `using ABC.Windows` all over the place.

AlexanderSher: Ok, I'll change default namespace.

src/Windows/Core/Impl/Security/SecurityService.cs
```diff
@@ -5,30 +5,81 @@
...
+        public void DeleteCredentials(string authority) {
+            if(!CredDelete(authority, CRED_TYPE.GENERIC, 0)) {
+                int err = Marshal.GetLastWin32Error();
+                if(err != ERROR_NOT_FOUND) {
+                    throw new Win32Exception(err);
+                }
+            }
+        }
+
+        public bool DeleteUserCredentials(string authority) {
```
karthiknadig: Can this one be removed? It does not handle the win32 error.

MikhailArkhipov:
If this is g2g we should merge so I can start changing services and languages on top.

AlexanderSher:
Since there will be 2 separate .sln files for VS14 and VS15, I need to make changes to the build definition. Can we disable building for VS15 for now?

MikhailArkhipov:
Actually I'd expect master to be 2017 going forward. Release/1_0 remains 2015. So we can disable VS14 instead.

Issue #3248, opened 2017-03-01
Some autoformatting annoyances


Hong-Revo:
These are with Automatic Formatting = On in the Editor Options.

When creating a dplyr chain, the editor keeps unindenting lines that I indent because they're part of the chain. Desired formatting:
```
df <- df %>%
    verb1(...) %>%
    verb2(...)
```
Actual formatting:
```
df <- df %>%
verb1(...) %>%
verb2(...)
```

Also, my preferred style is not to have spaces between argument names and values inside a function call (or a function header). The editor inserts spaces when I press Enter. Desired formatting:
```
func(x=1, y=2, z=something, ...)
```
Actual formatting:
```
func(x = 1, y = 2, z = something, ...)
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/bd30fdccb5ad88a13db543cf9be4d5150a8152c2

MikhailArkhipov:
Please try https://aka.ms/rtvs-1.0rc3

Issue #3235, opened 2017-02-24
Still Can't view the table


SenZHANG-GitHub:
I know it's a repeated issue (same as #3150, #2996). But this issue still exists after I try https://aka.ms/rtvs-1.0rc1

![image](https://cloud.githubusercontent.com/assets/25522404/23295132/f427011a-faa9-11e6-9c7d-db921bb18458.png)

The version of the VS and R tools I'm using is:
![image](https://cloud.githubusercontent.com/assets/25522404/23289407/4eac97ba-fa84-11e6-80aa-434d80e51b02.png)

I can still view the table using the excel button. But it would be much better that I can view it directly in VS. 

Thanks in advance!

int19h:
Could you run this code:
```r
rtvs:::grid_data(sigma, 1:2, 1:2)
```
and post the output here?

iRoss2007:
Exactly the same problem I and my friends encountered. Viewing data within VS has never been a success since the first public version of RTVS. Tricks as in #3150, #2996 did nothing to it. 
@int19h Could you please tell me where to run the grid_date code? I run it in R interactive and it prompted "Error in as.vector(x, mode) : cannot coerce type 'closure' to vector of type 'any' ". 
Thanks.

SenZHANG-GitHub:
@int19h 

Thanks for your reply! The output is shown as below

![image](https://cloud.githubusercontent.com/assets/25522404/23292366/457c2f14-fa9a-11e6-8f5c-98d0bc3c7497.png)

Seems OK though...

SenZHANG-GitHub:
@iRoss2007 I just run  rtvs:::grid_data(sigma, 1:2, 1:2) in R intearctive (Microsoft R Client (3.2.2.803)). And it works

SenZHANG-GitHub:
@iRoss2007

The following link showed a similar error,

https://stat.ethz.ch/pipermail/r-help/2010-December/261939.html

I guess maybe this happens when you assign the row and col parameters in an improper way  to the function grid_data()

Could you plz show the data.frame you wanted to investigate and the command you used?

Thanks

int19h:
Yep, the output is correct, and seems to indicate that the R backend is working as it should and retrieving the data correctly. This narrows down the area where the bug is occurring to the presentation layer. I wonder if the real issue here is that the grid doesn't refresh properly after it gets the data.

What happens if you change the value of `sigma` after the grid tab is visible? Try doing something like this:
```
[...]

PR #3252, opened 2017-03-02
3235

MikhailArkhipov:
#3235 Can't view the table
- Removed typeface calculation and replaced with IDE fonts and fallback

CHS OS:
image

#3248 Some autoformatting annoyances
- Added the requested options
- Range formatter should not use 'exact' version directly, only after it figures out correct range.

Hardened exception handling in the package and function index

### DIFF
--- a/src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
+++ b/src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
@@ -53,8 +53,7 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
 
         #region Font
 
-        public static readonly DependencyProperty FontFamilyProperty =
-                TextElement.FontFamilyProperty.AddOwner(typeof(VisualGrid));
+        public static readonly DependencyProperty FontFamilyProperty = TextElement.FontFamilyProperty.AddOwner(typeof(VisualGrid));
 
         [Localizability(LocalizationCategory.Font)]
         public FontFamily FontFamily {
@@ -65,33 +64,40 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
             }
         }
 
-        public static readonly DependencyProperty FontSizeProperty =
-                TextElement.FontSizeProperty.AddOwner(
-                        typeof(VisualGrid));
+        public static readonly DependencyProperty FontSizeProperty = TextElement.FontSizeProperty.AddOwner(typeof(VisualGrid));
 
         [TypeConverter(typeof(FontSizeConverter))]
         public double FontSize {
             get { return (double)GetValue(FontSizeProperty); }
-            set { SetValue(FontSizeProperty, value); }
+            set {
+                _typeFace = null;
+                SetValue(FontSizeProperty, value);
+            }
+        }
+
+        public static readonly DependencyProperty FontWeightProperty = TextElement.FontWeightProperty.AddOwner(typeof(VisualGrid));
+
+        [TypeConverter(typeof(FontWeightConverter))]
+        public FontWeight FontWeight {
+            get { return (FontWeight)GetValue(FontWeightProperty); }
+            set {
+                _typeFace = null;
+                SetValue(FontWeightProperty, value);
+            }
         }
 
         private Typeface _typeFace;
         private Typeface Typeface {
             get {
                 if (_typeFace == null) {
-                    _typeFace = ChooseTypeface();
+                    if (FontFamily != null && FontSize > 0) {
+                        try {
+                            _typeFace = new Typeface(FontFamily, FontStyles.Normal, FontWeight, FontStretches.Normal);
+                        } catch(ArgumentException) { }
+                    }
+                    _typeFace = _typeFace ?? new Typeface(new FontFamily("Segoe UI"), FontStyles.Normal, FontWeights.Normal, FontStretches.Normal);
                 }
                 return _typeFace;
-            }
-        }
-
-        private Typeface ChooseTypeface() {
-            if (ScrollDirection == ScrollDirection.Vertical
-                || ScrollDirection == ScrollDirection.Horizontal) {
-                // TODO: fall back
-                return FontFamily.GetTypefaces().First(tf => tf.Style == FontStyles.Normal && tf.Weight == FontWeights.DemiBold && tf.Stretch == FontStretches.Normal);
-            } else {
-                return FontFamily.GetTypefaces().First(tf => tf.Style == FontStyles.Normal && tf.Weight == FontWeights.Normal && tf.Stretch == FontStretches.Normal);
             }
         }
 
--- a/src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
+++ b/src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
@@ -53,38 +53,34 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect {
 
         #region Font
 
-        public static readonly DependencyProperty FontFamilyProperty = TextElement.FontFamilyProperty.AddOwner(typeof(VisualGrid));
-
+        public static readonly DependencyProperty FontFamilyProperty = TextElement.FontFamilyProperty.AddOwner(typeof(VisualGrid),
+                                                        new FrameworkPropertyMetadata(new FontFamily("Segoe UI"), FrameworkPropertyMetadataOptions.Inherits,
+                                                        new PropertyChangedCallback(OnTypefaceParametersChanged)));
         [Localizability(LocalizationCategory.Font)]
         public FontFamily FontFamily {
             get { return (FontFamily)GetValue(FontFamilyProperty); }
-            set {
-                _typeFace = null;
-                SetValue(FontFamilyProperty, value);
-            }
+            set { SetValue(FontFamilyProperty, value); }
         }
 
-        public static readonly DependencyProperty FontSizeProperty = TextElement.FontSizeProperty.AddOwner(typeof(VisualGrid));
-
+        public static readonly DependencyProperty FontSizeProperty = TextElement.FontSizeProperty.AddOwner(typeof(VisualGrid),
+                                                       new FrameworkPropertyMetadata(12.0, FrameworkPropertyMetadataOptions.Inherits,
+                                                       new PropertyChangedCallback(OnTypefaceParametersChanged)));
         [TypeConverter(typeof(FontSizeConverter))]
         public double FontSize {
             get { return (double)GetValue(FontSizeProperty); }
-            set {
-                _typeFace = null;
-                SetValue(FontSizeProperty, value);
-            }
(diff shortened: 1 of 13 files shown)


MikhailArkhipov:
@AlexanderSher, @int19h  - need two reviews

int19h:
👍

src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
```diff
@@ -65,36 +64,43 @@ internal sealed class VisualGrid : FrameworkElement {
...
-                TextElement.FontSizeProperty.AddOwner(
-                        typeof(VisualGrid));
+        public static readonly DependencyProperty FontSizeProperty = TextElement.FontSizeProperty.AddOwner(typeof(VisualGrid));
 
         [TypeConverter(typeof(FontSizeConverter))]
         public double FontSize {
             get { return (double)GetValue(FontSizeProperty); }
-            set { SetValue(FontSizeProperty, value); }
+            set {
+                _typeFace = null;
```
AlexanderSher: DependencyProperty can be changed by WPF using only its metadata without direct access to the property. `DependencyProperty.AddOwner` has overload that allows to specify `FrameworkPropertyMetadata` instance and add custom `PropertyChangedCallback`

src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
```diff
@@ -65,36 +64,43 @@ internal sealed class VisualGrid : FrameworkElement {
...
+                SetValue(FontWeightProperty, value);
+            }
         }
 
         private Typeface _typeFace;
         private Typeface Typeface {
             get {
                 if (_typeFace == null) {
-                    _typeFace = ChooseTypeface();
+                    if (FontFamily != null && FontSize > 0) {
```
AlexanderSher: Should `Typeface` be a `DependencyProperty` as well?

MikhailArkhipov: Normally it is not exposed AFACT. Controls have FontSize/Weight/Family/Stretch, never seen Typeface though. I guess this is b/s 4 former ones are defining it and not a single property.

src/Package/Impl/DataInspect/VisualGrid/VisualGrid.cs
MikhailArkhipov: OK

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -232,7 +234,7 @@ public sealed class PackageIndex : IPackageIndex {
                     var installedPackages = await GetInstalledPackagesAsync();
                     var packagesNotInIndex = installedPackages.Where(p => packageNames.Contains(p.Package));
                     info = await AddPackagesToIndexAsync(packagesNotInIndex);
-                } catch (RHostDisconnectedException) { }
+                } catch (RException) { }
```
AlexanderSher: `RHostDisconnectedException` isn't derived from `RException`. Also, `GetFunctionNamesAsync` already handles `REvaluationException`

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -80,7 +80,7 @@ internal sealed class PackageInfo : NamedItemInfo, IPackageInfo {
                 try {
                     var result = await _host.Session.PackagesFunctionsNamesAsync(Name, REvaluationKind.BaseEnv);
                     functions = result.Children<JValue>().Select(v => (string)v.Value);
-                } catch (TaskCanceledException) { } catch (REvaluationException) { }
+                } catch (RException) { }
```
AlexanderSher: `} catch (RHostDisconnectedException) { } catch (REvaluationException) {`

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -260,7 +262,7 @@ public sealed class PackageIndex : IPackageIndex {
             try {
                 await _host.StartSessionAsync();
                 return _host.LoadedPackageNames;
-            } catch (OperationCanceledException) { }
+            } catch (RException) { }
```
AlexanderSher: I think `StartSessionAsync` can't throw `RException`, but it can throw `RHostDisconnectedException`

MikhailArkhipov: You are right. But there is also a number of places when we call things on the session and yet not handle disconnected. Let me look there as well.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: In most cases in code I see use of `RException` though

MikhailArkhipov: `catch (RException) { } catch (OperationCanceledException)` should cover most cases I think

AlexanderSher: We should be careful with it. `catch (RException) { } catch (OperationCanceledException) { }` will swallow all exceptions that we expect to get from `RSession`. 

I think we should rather make it a separate PR.

MikhailArkhipov: If code already swallows `RException` it should be handling `OperationCanceledException`. In any case, the code is inconsistent - some handle both and some just `RException`. I think it is the old code that we've missed after adding remote. We don't want to crash VS on canceled, right?

MikhailArkhipov: OK, I pushed related files, will open PR on the rest

AlexanderSher: No, but there can be side effects:
- `OperationCanceledException` may be required by tasks. For example:
```cs
var cts = new System.Threading.CancellationTokenSource();
var t = Task.Run(async () => {
    while (true) {
        await Task.Delay(1000, cts.Token);
    }
});
```
Task `t` will end when `cts.Cancel()` is called. If we wrap `await Task.Delay(1000, cts.Token);`:
```cs
var cts = new System.Threading.CancellationTokenSource();
var t = Task.Run(async () => {
    while (true) {
        try {
            await Task.Delay(1000, cts.Token);
    }
});
```
task will be running forever.

- `RHostDisconnectedException` is captured on the ViewModel layer to show error messages. Swallowing it on the lower levels will break this code.

src/R/Support/Impl/Help/Packages/PackageInfo.cs
```diff
@@ -77,10 +77,8 @@ internal sealed class PackageInfo : NamedItemInfo, IPackageInfo {
         private async Task<IEnumerable<string>> GetFunctionNamesAsync() {
             var functions = TryRestoreFromCache();
             if (functions == null || !functions.Any()) {
-                try {
-                    var result = await _host.Session.PackagesFunctionsNamesAsync(Name, REvaluationKind.BaseEnv);
-                    functions = result.Children<JValue>().Select(v => (string)v.Value);
-                } catch (TaskCanceledException) { } catch (REvaluationException) { }
+                var result = await _host.Session.PackagesFunctionsNamesAsync(Name, REvaluationKind.BaseEnv);
+                functions = result.Children<JValue>().Select(v => (string)v.Value);
```
AlexanderSher: Previous implementation allowed to proceed with next package in case of error. With this change, `LoadRemainingPackagesFunctions` will stop on the first error. Is it intentional?

MikhailArkhipov: Probably should skip package on `RException` but skip all on `OperationCanceledException`

AlexanderSher: Agreed

src/R/Support/Impl/Help/Packages/PackageInfo.cs
MikhailArkhipov: Yes, I just mean that take `LoadWorkspaceCommand` - it handles both

```
        private async Task LoadWorkspace(IRSession session, string file) {
            try {
                await session.LoadWorkspaceAsync(file);
            } catch (RException ex) {
                var message = string.Format(CultureInfo.CurrentCulture, Resources.LoadWorkspaceFailedMessageFormat, file, ex.Message);
                VsAppShell.Current.ShowErrorMessage(message);
            }
        }
```

but in `WorkingDirectoryCommand` `SetDirectory` does not, only handles `RException`

MikhailArkhipov: `DeleteAllVariablesCommand` shouldn't be crashing if host disconnects either

AlexanderSher:
👍 
@int19h , do you want to take a look on the PR after all the changed?

src/R/Support/Impl/Help/Packages/PackageInfo.cs
AlexanderSher: And it won't. One of the reasons why `RHostDisconnectedException` is derived from `OperationCanceledException` is because `DoNotWait` automatically swallows task cancellation, and so it will swallow `RHostDisconnectedException`. So until we don't expect `rm(list = ls(all = TRUE))` to fail with `REvaluationException`, `DeleteAllVariablesCommand` is fine.

AlexanderSher: Same thing with `WorkingDirectoryCommand` - `.SilenceException<RException>()` is enough.

MikhailArkhipov: OK. but there is more that only `catch(RException)`

`AD7Property.SetValueAsStringWithError` - it is COM.
`RExecutionTracer`
`GridDataSource`
`REnvironmentProvider`

and several more

AlexanderSher: `AD7Property.SetValueAsStringWithError` - executed inside `ThreadHelper.JoinableTaskFactory`, which handles task cancellation
`GridDataSource` doesn't catch `RHostDisconnectedException` because calling code relies on it being thrown so that awaiter continuation doesn't execute.
`REnvironmentProvider` catches both

Generally speaking, I'd rather not change anything about handling `RHostDisconnectedException` because:
- for async code, `OperationCanceledException` is a normal thing and doesn't lead to a crash. 
- `RHostDisconnectedException` may be handled up in the stack separately from the `RException` or intentionally not handled, so we can easily break existing functionality
- I don't remember seeing VS crashed because of `RHostDisconnectedException`

MikhailArkhipov: OK ;-) But they would be hard to repro, it needs to disconnect right when code is executing. Which is usually not a very long time.

MikhailArkhipov:
I am going to close this one and merge into Release/1_0 as RC2 is now shipped.

AlexanderSher:
Ok

Issue #3264, opened 2017-03-06
R Interactive steals focus when debugging Python.


Marchen:
After updating RTVS to 1.0.30302.1431, R interactive steals focus from Python debug Interactive .
Maybe related with #2229.

![animation](https://cloud.githubusercontent.com/assets/3863577/23598828/65e607dc-0280-11e7-911a-a6c420af1789.gif)

Repro:
1) Start python debugging with a breakpoint.
2) At the break point, activate Python Debug Interactive. 
3) Push F10 to step over.
4) R Interactive is activated.

Python Tools for Visual Studio   2.2.50113.00
R Tools for Visual Studio   1.0.30302.1431

int19h:
The code we have to restore focus in VS 2015 probably doesn't check whether it's R being debugged, or something else. It should do so, by checking engine GUIDs (probably via DTE.Debugger).

This shouldn't be a problem in VS 2017, since we rely on VS itself to handle focus properly there.

MikhailArkhipov:
Yes, this is resolved in 2017. Unfortunately, interactive window component won't be updated for 2015.

MikhailArkhipov:
Code check if current active project is R project or that there is no project. We probably should also check for

1. If R is actually running so `R Tools | Session | Terminate R` would stop RTVS code.
2. For the debugger type

Marchen:
Additional information, the solution I tested this issue had both Python and R projects and Python project was start up project. I hope this will help you.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/16149bd35ee00d90b6c942b59447030db52da752

Issue #3262, opened 2017-03-04
Typing ::: sometimes gets messed up


Hong-Revo:
When I want to get an unexported function from a package via `:::`, the cursor jumps to the end of the line after I type `::`. Eg `stats:::AIC`, becomes `stats::AIC:`.

This seems to happen only if:

1. I type the function name first, then move to the start of the line and type the package name followed by `:::`
2. The completion list contains exactly one member, eg it happens with `AIC`[home]`stats:::` but not with `lm`[home]`stats:::`.

I can avoid this behaviour by pressing Esc to dismiss the completion list when it pops up, but I'd rather not have to.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/16149bd35ee00d90b6c942b59447030db52da752

Issue #3261, opened 2017-03-04
Inconsistent formatting of spaces around =


Hong-Revo:
Following on from #3260... the way RC3 treats spacing for `=` seems flakey.

If I type the following code in the editor:
```
data <- data %>%
    mutate(a=f(a),
           b=f(b),
           c=f(c))
```
The result is (note spaces around `=`):
```
data <- data %>%
    mutate(a=f(a),
           b = f(b),
           c = f(c))
```
However, if I continue with a `%>%`[enter]`group_by`, the spaces magically disappear:
```
data <- data %>%
    mutate(a=f(a),
           b=f(b),
           c=f(c)) %>%
    group_by(d) %>%
```

MikhailArkhipov:
(3) is expected as complete statement is formatted. (2) appears to be a bug.

Hong-Revo:
Yes, this would also seem to be the root cause of #3260: I've noticed that code throughout a file will have spaces appear and disappear as I type.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/16149bd35ee00d90b6c942b59447030db52da752

Issue #3260, opened 2017-03-04
Making a pipeline marks other code sections as touched


Hong-Revo:
Thanks for the quick turnaround on #3248 guys! I'm testing out RC3 now. I just ran into this:

![image](https://cloud.githubusercontent.com/assets/11328890/23575115/2f7d9cf0-00dc-11e7-8a77-a1d05d7ac8ae.png)

When I type the code at the bottom, it marks several other (untouched) sections of the file as edited.

MikhailArkhipov:
This is might happen in the current implementation. In order to format properly formatter is actually looking at a bigger range. At some point we should look into extracting smaller, original range from the formatted scope when applying change back to the document.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/16149bd35ee00d90b6c942b59447030db52da752

PR #3266, opened 2017-03-06
3260, 3261, 3262, 3264

MikhailArkhipov:
#3264 R Interactive steals focus when debugging Python.
- add explicit check for R debugger session

#3260 Making a pipeline marks other code sections as touched
- start of expression locator returned relative position rather than absolute

#3262 Typing ::: sometimes gets messed up
- in case of typing `stats:::` in front of the existing member last `:`completed to the member and then `:` appeared after the member name. ':` should only complete right after the package name and not after another `:`.

#3261 Inconsistent formatting of spaces around = 
- brace handler is not active for partial formatting so we have to always use option for spaces around `=`, not just in arguments. However, using `=` instead of `<-` is highly discouraged and if someone does want to use it, default is to put spaces around `=` so it will work.

### DIFF
--- a/src/R/Core/Test/Formatting/FormatOperatorsTest.cs
+++ b/src/R/Core/Test/Formatting/FormatOperatorsTest.cs
@@ -9,9 +9,9 @@ using Xunit;
 
 namespace Microsoft.R.Core.Test.Formatting {
     [ExcludeFromCodeCoverage]
+    [Category.R.Formatting]
     public class FormatOperatorsTest {
         [CompositeTest]
-        [Category.R.Formatting]
         [InlineData("-1", "-1")]
         [InlineData("- 1", "-1")]
         [InlineData("x--1", "x - -1")]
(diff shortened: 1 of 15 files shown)


src/R/Editor/Impl/Completion/RCompletionController.cs
```diff
@@ -163,7 +163,16 @@ public sealed class RCompletionController : CompletionController {
                     case ']':
                     case '}':
                     case ';':
-                        return completionSet.SelectionStatus.IsUnique;
+                        return unique;
+
+                    case ':':
+                        // : completes only if not preceded by another :
+                        // so we can avoid false positive when user is typing stats:::
+                        var caretPosition = TextView.Caret.Position.BufferPosition;
```
AlexanderSher: Do you want to add any tests for this?

MikhailArkhipov: I was going to do it in master, but I can as well add it here.

AlexanderSher:
👍

int19h:
👍

src/R/Core/Test/Formatting/FormatOperatorsTest.cs
```diff
@@ -29,5 +29,18 @@ public class FormatOperatorsTest {
             actual.Should().Be(expected);
         }
 
+        [CompositeTest]
+        [InlineData("x=1", true, "x = 1")]
+        [InlineData("x=1", false, "x=1")]
+        [InlineData("x <- function(a=1,b=2)", true, "x <- function(a = 1, b = 2)")]
+        [InlineData("x <- function(a=1,b=2)", false, "x <- function(a=1, b=2)")]
```
AlexanderSher: Do you want to add negative case like `x <- 'x=1'`?

PR #3270, opened 2017-03-07
Fix FormatException occurring while reporting REvaluationException for code that contains {}.

int19h:


AlexanderSher:
What will happen if there is `{{}}`?
Can you add a stack trace for the original bug?

int19h:
Nothing will happen. The problem is that we've been using the text directly as a format string, so any `{}` in it are interpreted as control characters. Using `{0}` and passing the actual string as a first argument will always reliably do the right thing here. This is the same problem as `printf(s)` vs `printf("%s", s)`.

AlexanderSher:
> The problem is that we've been using the text directly as a format string

Then shouldn't we be able to use `Write` instead of `WriteFormat`?

int19h:
Ah, indeed. I assumed that `WriteFormat` was used here in this manner, because it's the only member on that interface. But I see now that it has `Write` as well.

Not sure why format was used in the first place. This method has an overload with `params`, but this particular version doesn't have that, and so it should never get a format string. For added puzzlement, the `params` overload invokes `Format` and then delegates to this one.

int19h:
Fixed to use `Write`.

AlexanderSher:
👍

MikhailArkhipov:
👍

MikhailArkhipov:
Probably should go Release/1_0 as well

PR #3284, opened 2017-03-10
Modal dialogs for exporting plots.

anu-ka:
This is my first check-in in github. I have modified the export image and export pdf commands for plots. The dialog boxes are similar to rstudio. See the screenshots for the dialog boxes.

Exporting Image:
image

Exporting PDF:
image

I looked at the existing files and copied the patterns in it.


src/Common/Core/Impl/Microsoft.Common.Core.csproj
```diff
@@ -88,7 +88,13 @@
     <Compile Include="Services\ICoreServices.cs" />
     <Compile Include="Services\CoreServices.cs" />
     <Compile Include="Shell\CoreShellExtensions.cs" />
+    <Compile Include="Shell\ExportImageParameters.cs" />
```
MikhailArkhipov: Core assembly is generally for non-UI components and commonly used code. It does have some, but we are in process of cleaning this up post 1.0. Please move everything that is this dialog-related to Package/Plots.

src/Common/Core/Impl/Shell/ExportPdfParameters.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.Common.Core {
+    public class ExportPdfParameters {
+        public ExportPdfParameters() {
+            PaperName = "special";
```
MikhailArkhipov: Must be in resources as it is localizable

src/Common/Core/Impl/Shell/ExportPdfParameters.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public ExportPdfParameters(PDFExportOptions options, OrientationEnum orientation, PdfPrintOptionsEnum pdfPrint, bool viewPlot) {
+
+            if (orientation == OrientationEnum.Portrait) {
+                if (options.PaperName == Resources.Combobox_A4) {
+                    PaperName = "a4";
+                } else if (options.PaperName == Resources.Combobox_UsLetter) {
+                    PaperName = "us";
+                } else if (options.PaperName == Resources.Combobox_UsLegal) {
+                    PaperName = "legal";
```
MikhailArkhipov: Must be in resources as it is localizable

src/Common/Core/Impl/Shell/PDFExportOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+                new PDFExportOptions(Resources.Combobox_A6, 4.13, 5.83),
+                new PDFExportOptions(Resources.Combobox_4x6inches, 4.0, 6.0),
+                new PDFExportOptions(Resources.Combobox_5x7inches, 5.0, 7.0),
+                new PDFExportOptions(Resources.Combobox_6x8inches, 6.0, 8.0),
+                new PDFExportOptions(Resources.Combobox_Custom, width, height)
+            };
+            return pdfPapers;
+        }
+
+        public string PaperName { get; set; }
```
MikhailArkhipov: Just get as it is set in ctor. Typical patter is to only 'get'

src/Common/Core/Impl/Shell/PDFExportOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.Common.Core {
+    public class PDFExportOptions {
+        public PDFExportOptions(string paperName, double paperWidth, double paperHeight) {
+            PaperName = paperName;
+            PaperWidth = paperWidth;
+            PaperHeight = paperHeight;
+        }
+
+        public static List<PDFExportOptions> GetPdfPapers(double pixelWidth, double pixelHeight) {
```
MikhailArkhipov: IEnumerable or IReadOnlyCollection. Please do no return objects that can be modified unless it is necessary

src/Common/Core/Impl/Shell/PDFExportOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public string PaperName { get; set; }
+        public double PaperWidth { get; set; }
+        public double PaperHeight { get; set; }
+
+        public override string ToString() => PaperName;
+
+        private static double PixelsToInches(double pixels) {
+            return Math.Round(pixels / 96.0, 2);
```
MikhailArkhipov: Resolution is not always 96. What about high-resolution displays like 120, 132 and 144 dpi?

src/Common/Core/Test/Fakes/Shell/TestFileDialog.cs
```diff
@@ -10,11 +11,18 @@ public class TestFileDialog : IFileDialog {
...
+        public ExportImageParameters ExportImageParameters { get; set; }
+        public ExportPdfParameters ExportPdfParameters { get; set; }
 
         public string ShowOpenFileDialog(string filter, string initialPath = null, string title = null) => OpenFilePath;
 
         public string ShowBrowseDirectoryDialog(string initialPath = null, string title = null) => BrowseDirectoryPath;
 
         public string ShowSaveFileDialog(string filter, string initialPath = null, string title = null) => SaveFilePath;
+
+        public ExportImageParameters ShowExportImageDialog(ExportArguments imageArguements, string filter, string initialPath = null, string title = null) => ExportImageParameters;
```
MikhailArkhipov: These methods are not very common. Open File dialog is needed in many places but this one is very specific. Please avoid adding the to the ICoreShell

src/Package/Impl/ExportDialog/EnumValueConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) {
+            if(targetType == typeof(T) && value is string) {
+                return ConvertStringToEnum(value as string);
+            }
+            return default(T);
+        }
+
+        private T ConvertStringToEnum(string key) {
```
MikhailArkhipov: Are these user-visible strings? They need to be localizable

src/Package/Impl/ExportDialog/EnumValueConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+            IDictionary<string, T> stringToEnumValues = GetStringToEnum();
+            if(stringToEnumValues.ContainsKey(key)) {
```
MikhailArkhipov: TryGetValue and you don't need to check for existence

src/Package/Impl/ExportDialog/EnumValueConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+                return stringToEnumValues[key];
+            }
+            return default(T);
+        }
+
+        private string ConvertEnumToString(T key) {
+            IDictionary<T, string> enumToStringValues = GetEnumToString();
+            if (enumToStringValues.ContainsKey(key)) {
```
MikhailArkhipov: As above

src/Package/Impl/ExportDialog/ExportImageDialog.xaml
```diff
@@ -0,0 +1,52 @@
...
+            <RowDefinition />
+            <RowDefinition Height="auto"/>
+            <RowDefinition Height="auto"/>
+        </Grid.RowDefinitions>
+        <Grid Grid.Row="0">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition/>
+                <ColumnDefinition/>
+                <ColumnDefinition Width="auto"/>
+            </Grid.ColumnDefinitions>
```
MikhailArkhipov: Please make sure all UI is

1. Tested with Dark and Light VS UI themes.
2. Tested with high DPI/Large fonts.
3. Tested with OS High Contrast themes.
4. Fully accessible from keyboard.
5. ESC works as Cancel and Enter as default action.
6. Narrator correctly identifies controls and values to a visually-impaired person.

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+        private const int MAX = 100000;
+        private int _prevValidWidth;
+        private int _prevValidHeight;
+        private bool _maintainAspectRatio;
+        private int _originalWidth;
+        private int _originalHeight;
+
+        public ExportImageDialog(ExportArguments imageArguments) {
+            InitializeComponent();
+            this.HasMaximizeButton = false;
```
MikhailArkhipov: false is default

src/Package/Impl/ExportDialog/ExportImageDialog.xaml
```diff
@@ -0,0 +1,52 @@
...
+        <Separator Grid.Row="2"/>
+        <Grid Grid.Row="3">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition/>
+                <ColumnDefinition/>
+            </Grid.ColumnDefinitions>
+            <CheckBox Name="ViewPlotCheckBox" Content="{x:Static pkg:Resources.CheckBox_ViewPlot}" HorizontalAlignment="Left" Margin="10,10" KeyboardNavigation.TabIndex="3" IsChecked="{Binding ViewPlotAfterSaving,UpdateSourceTrigger=PropertyChanged,Mode=TwoWay}" VerticalAlignment="Center"/> 
+            <StackPanel Grid.Column="1" Orientation="Horizontal" HorizontalAlignment="Right">
+                <Button x:Name="SaveButton" Content="{x:Static components:Resources.Save}" HorizontalAlignment="Right"  VerticalAlignment="Top" Width="75"  Margin="10,10" Click="SaveButton_Click" KeyboardNavigation.TabIndex="4"/>
+                <Button x:Name="CancelButton" Content="{x:Static components:Resources.Cancel}" HorizontalAlignment="Right"  VerticalAlignment="Top" Width="75"  Margin="10,10" Click="CancelButton_Click" KeyboardNavigation.TabIndex="5"/>
```
MikhailArkhipov: `IsCancel = true`

src/Package/Impl/ExportDialog/ExportImageDialog.xaml
```diff
@@ -0,0 +1,52 @@
...
+        </Grid>
+        <Separator Grid.Row="2"/>
+        <Grid Grid.Row="3">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition/>
+                <ColumnDefinition/>
+            </Grid.ColumnDefinitions>
+            <CheckBox Name="ViewPlotCheckBox" Content="{x:Static pkg:Resources.CheckBox_ViewPlot}" HorizontalAlignment="Left" Margin="10,10" KeyboardNavigation.TabIndex="3" IsChecked="{Binding ViewPlotAfterSaving,UpdateSourceTrigger=PropertyChanged,Mode=TwoWay}" VerticalAlignment="Center"/> 
+            <StackPanel Grid.Column="1" Orientation="Horizontal" HorizontalAlignment="Right">
+                <Button x:Name="SaveButton" Content="{x:Static components:Resources.Save}" HorizontalAlignment="Right"  VerticalAlignment="Top" Width="75"  Margin="10,10" Click="SaveButton_Click" KeyboardNavigation.TabIndex="4"/>
```
MikhailArkhipov: `IsDefault = true`

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        private void CancelButton_Click(object sender, RoutedEventArgs e) {
+            DialogResult = false;
+            Close();
+        }
+
+        private void HeightTextbox_LostFocus(object sender, RoutedEventArgs e) {
+            TextBox textBox = (TextBox)sender;
+            int res = ValidateValues(textBox.Text);
```
MikhailArkhipov: Should 'Save' simply be disabled when values are not valid?

src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
```diff
@@ -1442,6 +1452,14 @@
       <Generator>MSBuild:Compile</Generator>
       <SubType>Designer</SubType>
     </Page>
+    <Page Include="ExportDialog\ExportImageDialog.xaml">
```
MikhailArkhipov: Please move it to Plots/ExportDialog as it is plotting specific

src/Package/Impl/Resources.resx
```diff
@@ -1129,4 +1129,52 @@ Click OK to open MiKTeX download page in the default browser. R Remote Services
...
     <value>Determines if formatter should place spaces around equals sign</value>
   </data>
+  <data name="ExportImageDialog_Title" xml:space="preserve">
+    <value>Export Image</value>
+  </data>
+  <data name="label_Height" xml:space="preserve">
+    <value>Height :</value>
+  </data>
+  <data name="label_Width" xml:space="preserve">
+    <value>Width :</value>
```
MikhailArkhipov: Some have space before `:` and some don't. Intentional?

src/Common/Core/Impl/Shell/ExportPdfParameters.cs
huguesv: AFAIK PaperName is what is passed to R, it's not localizable.

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -183,13 +183,13 @@ export_to_image <- function(device_id, plot_id, device, width, height, resolutio
 }
 
 # Helper to export current plot to pdf
-export_to_pdf <- function(device_id, plot_id, width, height) {
+export_to_pdf <- function(device_id, plot_id, pdf_device, width, height, paper) {
```
huguesv: This is tested in IdeGraphicsDeviceTest (see ExportToPdfAsync), so you'll need to make some update there as well.

src/Common/Core/Impl/Shell/ExportPdfParameters.cs
MikhailArkhipov: OK. Maybe then call it 'RInternalPaperName' or something

src/Common/Core/Test/Fakes/Shell/TestFileDialog.cs
huguesv: Do you have a recommended place for it? It has to be in a place where it's easily faked for export image/pdf ui tests, which shouldn't bring up the actual dialog.

huguesv: Speaking of which, please make sure that RPlotIntegrationTest's ExportAsPdf and ExportAsImage still pass.

src/Package/Impl/ExportDialog/ExportPDFDialog.xaml.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        public List<PDFExportOptions> PDFPapers => _pdfPapers;
+        public List<OrientationEnum> Orientations => _orientations;
+        public List<PdfPrintOptionsEnum> PrintOptions => _printOptions;
+
+        public PDFExportOptions SelectedPDFPaperType {
+            get { return _pdfPaperType; }
+            set {
+                _pdfPaperType = value;
+                OnPropertyChanged("SelectedPDFPaperType");
```
huguesv: You can use `nameof(SelectedPDFPaperType)` for these. It's better because refactoring rename operations will automatically update it.  Same for others below.

src/Common/Core/Impl/Enums/OrientationEnum.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core {
+    public enum OrientationEnum {
```
AlexanderSher: While this enum can be related to printing in general, it is still too specific to be referenced everywhere. For instance, editor assemblies don't need it, so I'd prefer to have it in `Microsoft.R.Components`

AlexanderSher: And you don't need `Enum` in the name of the type.

src/Common/Core/Impl/Enums/PrintOptionsEnum.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core {
+    public enum PdfPrintOptionsEnum {
```
AlexanderSher: Unlike `OrientationEnum`, `PdfPrintOptionsEnum` is specific to the R function. So it has to be in the same assembly and namespace with the rest of the "Export to PDF" functionality. Though I'm not sure if it is required at all (see below).

src/Common/Core/Impl/Shell/IFileDialog.cs
```diff
@@ -29,5 +29,25 @@ public interface IFileDialog {
...
+
+        /// <summary>
+        /// Show the export image dialog box.
+        /// </summary>
+        /// <param name="imageArguments"></param>
+        /// <param name="filter"></param>
+        /// <param name="initialPath"></param>
+        /// <param name="title"></param>
+        /// <returns></returns>
+        ExportImageParameters ShowExportImageDialog(ExportArguments imageArguments, string filter, string initialPath = null, string title = null);
```
AlexanderSher: `ShowExportImageDialog` and `ShowExportPdfDialog` aren't file-system related dialogs, so they don't belong to this interface. Again, they are specific to their features.

src/Common/Core/Impl/Shell/PDFExportOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+                new PDFExportOptions(Resources.Combobox_Custom, width, height)
+            };
+            return pdfPapers;
+        }
+
+        public string PaperName { get; set; }
+        public double PaperWidth { get; set; }
+        public double PaperHeight { get; set; }
+
+        public override string ToString() => PaperName;
```
AlexanderSher: Please don't override `ToString`.

src/Common/Core/Impl/Shell/PDFExportOptions.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Common.Core {
+    public class PDFExportOptions {
```
AlexanderSher: Similar to `PdfPrintOptionsEnum`, should be in the same namespace and assembly with either dialog or command.

src/Common/Core/Test/Fakes/Shell/TestFileDialog.cs
AlexanderSher: > Do you have a recommended place for it?

Similar to any other UI abstraction that we have. Dialog here is just another type of `IVisualComponent`

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -191,8 +192,8 @@ public static class RSessionEvaluationCommands {
             return evaluation.EvaluateAsync<ulong>(script, REvaluationKind.Normal);
         }
 
-        public static Task<ulong> ExportToPdfAsync(this IRExpressionEvaluator evaluation, Guid deviceId, Guid plotId, string outputFilePath, double widthInInches, double heightInInches) {
-            string script = Invariant($"rtvs:::export_to_pdf({deviceId.ToString().ToRStringLiteral()}, {plotId.ToString().ToRStringLiteral()}, {widthInInches}, {heightInInches})");
+        public static Task<ulong> ExportToPdfAsync(this IRExpressionEvaluator evaluation, Guid deviceId, Guid plotId, ExportPdfParameters pdfParams) {
```
AlexanderSher: Since it is just a helper extension, I'd rather pass all parameters as they are, without intermediate `ExportPdfParameters`. Also, you need to pass `CancellationToken` for it.

src/Package/Impl/ExportDialog/EnumValueConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace Microsoft.VisualStudio.R.Package.ExportDialog {
+    public abstract class EnumValueConverter<T> : IValueConverter {
```
AlexanderSher: WPF ComboBox can be operated in two modes:
- If list of items is fixed and they are just string key/value pairs, they can be specified directly in XAML:
```xml
<ComboBox>
    <ComboBoxItem Content="components:Resources.PageProperties_Portrait" Tag="portrait" IsSelected="True" />
    <ComboBoxItem Content="components:Resources.PageProperties_Landspace" Tag="landspace" />
</ComboBox>
```
- If list of items isn't fixed or requires a specific ViewModel for its item, you will use `SelectedValuePath` and `DisplayMemberPath` properties.

So this converter isn't required. It also has an implementation overkill with two separate dictionaries, so I'd rather not add it to the repo even for the future.

src/Package/Impl/ExportDialog/ExportImageDialog.xaml
```diff
@@ -0,0 +1,52 @@
...
+            <RowDefinition Height="auto"/>
+        </Grid.RowDefinitions>
+        <Grid Grid.Row="0">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition/>
+                <ColumnDefinition/>
+                <ColumnDefinition Width="auto"/>
+            </Grid.ColumnDefinitions>
+            <StackPanel  Orientation="Horizontal" HorizontalAlignment="Left">
+                <TextBlock  FontWeight="Bold" Text="{x:Static pkg:Resources.label_Width}" VerticalAlignment="Center" Margin="10,10"/>
```
AlexanderSher: Too many panels for a simple dialog box. One top level grid with required amount of rows and columns is enough.

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+        private int _originalWidth;
+        private int _originalHeight;
+
+        public ExportImageDialog(ExportArguments imageArguments) {
+            InitializeComponent();
+            this.HasMaximizeButton = false;
+            this.HasMinimizeButton = false;
+            UserWidth = imageArguments.PixelWidth;
+            UserHeight = imageArguments.PixelHeight;
+            this.DataContext = this;
```
AlexanderSher: Don't use `this` unless there is an ambiguity

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Windows;
+using System.Windows.Controls;
+using System.ComponentModel;
+using Microsoft.Common.Core;
+using Microsoft.VisualStudio.R.Package.Wpf;
+
+namespace Microsoft.VisualStudio.R.Package.ExportDialog {
+    public partial class ExportImageDialog : PlatformDialogWindow, INotifyPropertyChanged {
```
AlexanderSher: This is usually a bad idea to mix `INotifyPropertyChanged` and `DependencyObject` in one class.

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+        public ExportImageDialog(ExportArguments imageArguments) {
+            InitializeComponent();
+            this.HasMaximizeButton = false;
+            this.HasMinimizeButton = false;
+            UserWidth = imageArguments.PixelWidth;
+            UserHeight = imageArguments.PixelHeight;
+            this.DataContext = this;
+        }
+
+        public int UserWidth {
```
AlexanderSher: This is clearly a ViewModel properties which should be moved into a separate object (see any existing `*ViewModel` class in `Microsoft.R.Components` for the reference).

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        private int GetSizeForHeight() {
+            int height = UserWidth * _originalHeight / _originalWidth;
+            int h = ValidateValues(height.ToString());
+            return height = h;
+        }
+
+        private int ValidateValues(string result) {
+            long l = 0;
+            bool isValid = long.TryParse(result, out l);
```
AlexanderSher: It isn't clear what are `MAX`, `l`, `0` and `-1`. Please give names to all magical constants.

AlexanderSher: `ValidateValues` is called over the string that was created from int and parses that string at the very beginning. What's the point?

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+                    int w = GetSizeForWidth();
+                    UserWidth = w;
+                }
+            }
+        }
+
+        private int GetSizeForWidth() {
+            int width = UserHeight * _originalWidth / _originalHeight;
+            int w = ValidateValues(width.ToString());
+            return width = w;
```
AlexanderSher: Last line isn't required, `return ValidateValues(width)` is enough

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
```diff
@@ -0,0 +1,160 @@
...
+                OnPropertyChanged("ViewPlotAfterSaving");
+            }
+        }
+
+        public bool MaintainAspectRatio {
+            get { return _maintainAspectRatio; }
+            set {
+                _maintainAspectRatio = value;
+                if (_maintainAspectRatio) {
+                    _originalWidth = UserWidth;
```
AlexanderSher: Use properties instead of fields, otherwise there will be no `PropertyChanged` event raised.

src/Package/Impl/ExportDialog/ExportImageDialog.xaml.cs
AlexanderSher: `ConnectionViewModel` can be a good example. Please derive your ViewModel from `BindableBase`.

src/Package/Impl/ExportDialog/ExportImageDialog.xaml
anu-ka: I have tested 3 through 6. High contrast screen shot here:
image
image

For High DPI, I don't think there is any additional change needed.
http://stackoverflow.com/questions/2236173/screen-resolution-problem-in-wpf

anu-ka:
I am still working on a few changes. I will send out an update PR, once I merge with main.

anu-ka:
I think I have addressed most of the comments. Please let me know if I need to do some more changes.

anu-ka:
Sorry, I closed this accidentally.

MikhailArkhipov:
We are accumulating massive changes in master, we'll get to this in a week or so when we are done with refactoring. thanks for your patience.

MikhailArkhipov:
Let's resolve conflicts and take

MikhailArkhipov:
👍

PR #3286, opened 2017-03-10
Setup block for unsupported versions of the IDE 

MikhailArkhipov:
Removed custom action since bundle now blocks the setup. Bundle PR incoming.


MikhailArkhipov:
https://github.com/Microsoft/RTVS-pr/pull/21 for the bundle

Issue #3283, opened 2017-03-10
Uploading with Shiny to remote workspace fails


Hong-Revo:
Most of the examples from `shiny::runExample` work fine when run from a remote workspace, except for `"09_upload"`.
```
library(shiny)
runExample("09_upload")

Listening on http://127.0.0.1:4828
HTTP server for remote GPU DSVM starting ...
HTTP server for remote GPU DSVM listening on http://127.0.0.1:60694. 
HTTP server for remote transfers data from remote R machine.
Server at DESKTOP-TBHQGUH http://127.0.0.1:60694 connects to GPU DSVM https://xxx-xxx.cloudapp.azure.com:5444/.
HTTP server for remote stopped with following exception:
You must provide a request body if you set ContentLength>0 or SendChunked==true.  Do this by calling [Begin]GetRequestStream before [Begin]GetResponse.
HTTP server for remote GPU DSVM stopped.
HTTP server for remote stopped with following exception:
An internal WebSocket error occurred. Please see the innerException, if present, for more details. 
```

The error occurs once you select a file to upload.

MikhailArkhipov:
@karthiknadig - what's the cause?

karthiknadig:
There seems to be a issue with the way we handle files when submitted via form input. I figured out the client side problem. Server side seems have a bug as well. I have not investigated server side yet.

Client side issue was, we were looking for a seek-able network stream to get the length of the input data. When a file is being uploaded, the stream is read-only and not seek-able. we ended up skipping the upload. That leads to the error that you see above. However, fixing that on client side is not enough, since we need to pass it all the way to RHost, there is one more change that is needed on the broker side.

MikhailArkhipov:
OK, let's see the PR, perhaps we should take this in 1.0. Shiny demos probably need to work.

PR #3295, opened 2017-03-14
Fixes issue with shiny and uploading files.

karthiknadig:
#3283

### DIFF
--- a/src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
+++ b/src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
@@ -55,7 +55,7 @@ namespace Microsoft.R.Host.Client.BrokerServices {
             var remoteUri = GetRemoteUri(context.Request.Url, remoteBaseUrl);
             request.Headers.Add(CustomHttpHeaders.RTVSRequestedURL, remoteUri.ToString());
 
-            if (context.Request.InputStream.CanSeek && context.Request.InputStream.Length > 0) {
+            if (context.Request.ContentLength64 > 0) {
                 using (Stream reqStream = await request.GetRequestStreamAsync()) {
                     await context.Request.InputStream.CopyAndFlushAsync(reqStream, null, ct);
                 }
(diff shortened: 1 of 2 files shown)


MikhailArkhipov:
👍 . @int19h  - need 2nd.

MikhailArkhipov:
Or @AlexanderSher please

src/Host/Client/Impl/BrokerServices/RemoteUriWebService.cs
```diff
@@ -55,7 +55,7 @@ public class RemoteUriWebService : IRemoteUriWebService {
             var remoteUri = GetRemoteUri(context.Request.Url, remoteBaseUrl);
             request.Headers.Add(CustomHttpHeaders.RTVSRequestedURL, remoteUri.ToString());
 
-            if (context.Request.InputStream.CanSeek && context.Request.InputStream.Length > 0) {
+            if (context.Request.ContentLength64 > 0) {
```
AlexanderSher: Why we don't need to check for `CanSeek` anymore?

karthiknadig: Seek was needed to query the length from a stream for POST messages. Using the stream directly to determine the length requires additional conditions. For, example, when a file is uploaded, the stream is *not* seek-able but read-only. For post messages, the stream may be seek-able and length is only available when it is seek-able. `ContentLength64` stores the message body length for all request objects, and it is simpler. The issue here was caused by read-only case I mentioned above.

AlexanderSher: So `ContentLength64 > 0` only if we can use the stream, correct?

karthiknadig: `ContentLength64 > 0` only if the stream has content. Also, we need to do the copy only if the stream has content. The stream is still usable in the websocket context.

karthiknadig: Opening the stream when there is not supposed to be content will lead to error. So, the request stream has to be accessed only if there is content or in websocket context.

AlexanderSher:
👍

Issue #3294, opened 2017-03-13
File completion resolves ./ incorrectly


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/23879249/23d6b4e8-0809-11e7-813a-13073b7e2e9a.png)

`getwd()` reports current directory set to project but it looks like completion resolves it to RTVS location instead.

MikhailArkhipov:
We do resolve `~/` properly but not `./` or similar relative

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/4e5ef69ffa422dcb50a702da29450be2fbf629ae

Issue #3293, opened 2017-03-13
Multiple nesting in a$b$c does not bring completion


MikhailArkhipov:
User reported
```
d <- list(A = c(1:10), B = c(1:10))
e <- list(A=d, B=d)
```

`e$` brings A and B while `e$A$` is empty. 

This is because we check `exist('e$A') which is wrong.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/4e5ef69ffa422dcb50a702da29450be2fbf629ae

Issue #3292, opened 2017-03-13
Brace matcher exception


MikhailArkhipov:
User-reported

```
System.ArgumentException: The point does not belong to a source snapshot of the projection snapshot
at Microsoft.VisualStudio.Text.Projection.Implementation.ElisionSnapshot.MapFromSourceSnapshot(SnapshotPoint point, PositionAffinity affinity)
at Microsoft.VisualStudio.Text.Projection.Implementation.BufferGraph.MapUpToBufferGuts(SnapshotPoint currentPoint, PositionAffinity affinity, Predicate`1 match)
at Microsoft.VisualStudio.Text.Projection.Implementation.BufferGraph.CheckedMapUpToBuffer(SnapshotPoint point, PointTrackingMode trackingMode, Predicate`1 match, PositionAffinity affinity)
at Microsoft.VisualStudio.Text.Projection.Implementation.BufferGraph.MapUpToBuffer(SnapshotPoint point, PointTrackingMode trackingMode, PositionAffinity affinity, ITextBuffer targetBuffer)
at Microsoft.VisualStudio.Text.Implementation.MappingPoint.GetPoint(ITextBuffer targetBuffer, PositionAffinity affinity)
at Microsoft.VisualStudio.Text.BraceCompletion.Implementation.BraceCompletionManager.IsCaretOnBuffer(ITextBuffer buffer)
at Microsoft.VisualStudio.Text.BraceCompletion.Implementation.BraceCompletionManager.&lt;&gt;c__DisplayClass18_0.&lt;PreTypeChar&gt;b__0()
at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.CallExtensionPoint(Object errorSource, Action call)</description>


System.InvalidOperationException: Master edit already in progress
at Microsoft.VisualStudio.Text.Implementation.BufferGroup.PerformMasterEdit(ITextBuffer buffer, ISubordinateTextEdit xedit, EditOptions options, Object editTag)
at Microsoft.VisualStudio.Text.Projection.Implementation.ProjectionBuffer.ProjectionEdit.PerformApply()
at Microsoft.VisualStudio.Text.Implementation.BaseBuffer.TextBufferEdit.Apply()
at Microsoft.VisualStudio.Text.Operations.Implementation.EditorOperations.InsertText(String text, Boolean final, String undoText, Boolean isOverwriteModeEnabled)
[...]

PR #3296, opened 2017-03-14
3292, 3293, 3294

MikhailArkhipov:
#3292 Brace matcher exception
 - off by one error in markdown tokenizer. Fixed, added test.

#3293 Multiple nesting in a$b$c does not bring completion
- incorrect check for `exists`. Fixed, added test.

#3294 File completion resolves ./ incorrectly
- added support for `./` + test.

### DIFF
--- a/src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
+++ b/src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
@@ -528,18 +528,30 @@ namespace Microsoft.R.Editor.Application.Test.Completion {
                 await ExecuteRCode(_editorHost.HostScript.Session,
 @"
 d <- list(A = c(1:10), B = c(1:10))
-e <- list(A=d, B=d)
+e <- list(D=d, E=d)
 ");
                 PrimeIntellisenseProviders(script);
                 script.DoIdle(500);
 
-                script.Type("e$A$");
+                script.Type("e$");
 
                 var session = script.GetCompletionSession();
                 session.Should().NotBeNull();
                 script.DoIdle(1000);
 
                 var comps = session.SelectedCompletionSet.Completions;
+                comps.Should().HaveCount(2).And
+                    .Contain(x => x.DisplayText == "D").And
+                    .Contain(x => x.DisplayText == "E");
+                script.DoIdle(100);
+
+                script.Type("D$");
+
+                session = script.GetCompletionSession();
+                session.Should().NotBeNull();
+                script.DoIdle(1000);
+
+                comps = session.SelectedCompletionSet.Completions;
                 comps.Should().HaveCount(2).And
                     .Contain(x => x.DisplayText == "A").And
                     .Contain(x => x.DisplayText == "B");
--- a/src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
+++ b/src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
@@ -15,6 +15,7 @@ using Microsoft.R.Editor.Settings;
 using Microsoft.R.Editor.Snippets;
 using Microsoft.R.Editor.Test.Utility;
 using Microsoft.R.Host.Client;
+using Microsoft.R.Host.Client.Session;
 using Microsoft.R.Support.Settings;
 using Microsoft.UnitTests.Core.Mef;
 using Microsoft.UnitTests.Core.Threading;
@@ -188,6 +189,41 @@ namespace Microsoft.R.Editor.Application.Test.Completion {
 
                 var list = session.SelectedCompletionSet.Completions.ToList();
                 var item = list.FirstOrDefault(x => x.DisplayText == "_rtvs_test_");
+                item.Should().NotBeNull();
+
+                if (Directory.Exists(testFolder)) {
+                    Directory.Delete(testFolder);
+                }
+            }
+        }
+
+        [Test]
+        public async Task R_CompletionFilesWorkingDirectory() {
+            using (var script = await _editorHost.StartScript(ExportProvider, RContentTypeDefinition.ContentType, Workflow.RSessions)) {
+                var myDocs = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
+                var folder = "_rtvs_test2_";
+
+                var testFolder = Path.Combine(myDocs, folder);
+                if (!Directory.Exists(testFolder)) {
+                    Directory.CreateDirectory(testFolder);
+                }
+
+                await _editorHost.HostScript.Session.ExecuteAsync(Invariant($"setwd('{myDocs.ToRPath()}')"));
+                var wd = await _editorHost.HostScript.Session.GetWorkingDirectoryAsync();
+                wd.Should().Be(myDocs);
+
+                script.DoIdle(100);
+                script.Type("x <- \"./");
+                script.DoIdle(1000);
+                script.Type("{TAB}");
+                script.DoIdle(500);
+
+                var session = script.GetCompletionSession();
+                session.Should().NotBeNull();
+                script.DoIdle(200);
+
+                var list = session.SelectedCompletionSet.Completions.ToList();
+                var item = list.FirstOrDefault(x => x.DisplayText == folder);
                 item.Should().NotBeNull();
 
                 if (Directory.Exists(testFolder)) {
--- a/src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
+++ b/src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
@@ -487,6 +487,31 @@ namespace Microsoft.R.Editor.Application.Test.Completion {
         }
 
         [Test]
+        public async Task R_VariablesIndexerMemberCompletion03() {
+            using (var script = await _editorHost.StartScript(ExportProvider, RContentTypeDefinition.ContentType, Workflow.RSessions)) {
+                await ExecuteRCode(_editorHost.HostScript.Session,
+@"
+d <- list(A = c(1:10), B = c(1:10))
+e <- list(A=d, B=d)
+");
+                PrimeIntellisenseProviders(script);
+                script.DoIdle(500);
+
+                script.Type("e$A$");
+
+                var session = script.GetCompletionSession();
+                session.Should().NotBeNull();
+                script.DoIdle(1000);
+
+                var comps = session.SelectedCompletionSet.Completions;
+                comps.Should().HaveCount(2).And
+                    .Contain(x => x.DisplayText == "A").And
+                    .Contain(x => x.DisplayText == "B");
+                script.DoIdle(100);
+            }
+        }
+
+        [Test]
         public async Task R_VariablesIndexerMemberCompletion02() {
             using (var script = await _editorHost.StartScript(ExportProvider, RContentTypeDefinition.ContentType, Workflow.RSessions)) {
                 await ExecuteRCode(_editorHost.HostScript.Session, "d1 <- mtcars\r\n");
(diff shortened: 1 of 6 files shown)


src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
```diff
@@ -197,6 +198,41 @@ public class IntellisenseTest : FunctionIndexBasedTest {
         }
 
         [Test]
+        public async Task R_CompletionFilesWorkingDirectory() {
+            using (var script = await _editorHost.StartScript(ExportProvider, RContentTypeDefinition.ContentType, Workflow.RSessions)) {
+                var myDocs = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
+                var folder = "_rtvs_test2_";
+
+                var testFolder = Path.Combine(myDocs, folder);
```
AlexanderSher: Can you use test artifacts folder instead?

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -41,35 +49,42 @@ internal sealed class FilesCompletionProvider : IRCompletionListProvider {
...
+                // If directory is set, then the async task did complete
+                if (!string.IsNullOrEmpty(_rootDirectory)) {
+                    if (_mode == Mode.WorkingDirectory) {
+                        directory = Path.Combine(_rootDirectory, _enteredDirectory);
+                    } else if (_mode == Mode.UserDirectory) {
+                        var subDirectory = _enteredDirectory.Length > 1 ? _enteredDirectory.Substring(2) : _enteredDirectory;
+                        directory = Path.Combine(_rootDirectory, subDirectory);
+                    }
                 }
+            } catch (ArgumentException) { }
```
AlexanderSher: What throws `ArgumentException`?

MikhailArkhipov: `Path.Combine`.

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
MikhailArkhipov: Where is it? I basically derived from existing test above.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -85,7 +85,7 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
                 }
 
                 // May be a package object line mtcars$
-                variableName = TrimToTrailingSelector(variableName);
+                variableName = TrimToFirstSelector(variableName);
```
int19h: For the future, it would be nice to refactor this code so that it doesn't try to distinguish between globals and members at all. Globals are really just members of the global environment object, and so you can start with `.GlobalEnv` as root expression, and then use `DescribeChildrenAsync` to get them. Then the entire code snippet below will be uniform in that regard, with no `exists()` check needed at all.

int19h: Actually, come to think of it, doesn't it do the wrong thing now? If I understand correctly, something like `a$b$` will now be trimmed to the first `$`, so `variableName` becomes `a`. And then you call `DescribeChildrenAsync` on that below. So completion list would that for members of `a`, not for members of `a$b`?

It sounds like it should instead not check `exists()` at all, and instead feed the entire expression up until trailing selector to `DescribeChildrenAsync`, and check if the latter returns an error. If it does, then make member list is empty.

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
```diff
@@ -487,6 +487,31 @@ public class IntellisenseTest : FunctionIndexBasedTest {
         }
 
         [Test]
+        public async Task R_VariablesIndexerMemberCompletion03() {
+            using (var script = await _editorHost.StartScript(ExportProvider, RContentTypeDefinition.ContentType, Workflow.RSessions)) {
+                await ExecuteRCode(_editorHost.HostScript.Session,
+@"
+d <- list(A = c(1:10), B = c(1:10))
```
int19h: Test needs to use different member names for `d` and `e` to ensure that the right completion list is returned for `e$A$`.

MikhailArkhipov: Sure. this is just exact code from the user.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: The problem is that `exists()` was added to avoid that error message in REPL when variable doesn't exist. Can't find the original bug... But let me try some varieties.

And yes, we have task for the cleanup

MikhailArkhipov: #2019

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
MikhailArkhipov: Updated and test extended

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
AlexanderSher: I meant using `TestFilesFixture`. In this case, you don't have to erase the directory at the end of the test.

MikhailArkhipov: I'll open a task for master. I'd rather limit changes in 1.0

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
AlexanderSher: Do you expect invalid characters in `_rootDirectory` or `subDirectory`?

MikhailArkhipov: It comes from the user so possible. Also, we used to catch here so I don't want to change this too much.

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -22,14 +22,22 @@ namespace Microsoft.R.Editor.Completion.Providers {
...
 
-        private Task<string> _userDirectoryFetchingTask;
-        private string _directory;
-        private string _cachedUserDirectory;
-        private bool _forceR; // for tests
+        private readonly string _enteredDirectory;
+        private readonly bool _forceR; // for tests
+
+        private readonly Task<string> _task;
+        private readonly Mode _mode = Mode.Other;
```
AlexanderSher: I don't see how this field is used

MikhailArkhipov: It is set in ctor and then used down in `GetEntries`. I'd like to split cases going forward and derive to separate classes from base rather than use conditional by mode. But that would be much larger change.

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -41,35 +49,42 @@ internal sealed class FilesCompletionProvider : IRCompletionListProvider {
             _glyphService = glyphService;
             _forceR = forceR;
 
-            _directory = ExtractDirectory(directoryCandidate);
-
-            if (_directory.Length == 0 || _directory.StartsWithOrdinal("~\\")) {
-                _directory = _directory.Length > 1 ? _directory.Substring(2) : _directory;
-                _userDirectoryFetchingTask = _workflow.RSession.GetRUserDirectoryAsync();
+            _enteredDirectory = ExtractDirectory(directoryCandidate);
+            if (_enteredDirectory.Length == 0 || _enteredDirectory.StartsWithOrdinal(".")) {
```
AlexanderSher: I'd rather extract this whole part into separate method:
```cs
private async Task GetRootDirectory() {
    if (_enteredDirectory.Length == 0 || _enteredDirectory.StartsWithOrdinal(".")) {
        _rootDirectory = await _workflow.RSession.GetWorkingDirectoryAsync();
    } else if (_enteredDirectory.StartsWithOrdinal("~\\")) {
        _rootDirectory = await _workflow.RSession.GetRUserDirectoryAsync()
    }
}
```
```cs
_task = GetRootDirectory();
```

And we don't need to call `DoNotWait` on it, unless you want to fail fast in `FilesCompletionProvider` constructor on something.

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
AlexanderSher: Ok

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
AlexanderSher: Ok, got it

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
MikhailArkhipov: OK

src/R/Editor/Impl/Completion/Providers/FilesCompletionProvider.cs
```diff
@@ -50,14 +50,18 @@ enum Mode {
             _forceR = forceR;
 
             _enteredDirectory = ExtractDirectory(directoryCandidate);
-            if (_enteredDirectory.Length == 0 || _enteredDirectory.StartsWithOrdinal(".")) {
+            _task = GetRootDirectoryAsync(_enteredDirectory);
+        }
+
+        private Task<string> GetRootDirectoryAsync(string userProvidedDirectory) {
```
AlexanderSher: Why you don't want to make this method async? You don't have to use `Task.Run` in that case.

MikhailArkhipov: There is a problem calling `t = asyncMethod()` in ctor. I spent some time yesterday on it. Basically task never runs and `.Wait` in `GetEntries` never completes (status == WaitingForActivation). As an experiment I removed Wait and just used `t.Result` which hangs. There is something with sync context. Perhaps this is b/c `ctor` and `GetEntries` are called on the same UI thread sequentially and task that is configured to return data to the main thread never completes?

AlexanderSher: Ok, I'll take a look: https://github.com/Microsoft/RTVS/issues/3300

AlexanderSher:
👍

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
int19h: `exists()` won't handle nested members anyway, so we'll have to come up with something else.

So the problem is that `rtvs:::describe_children` doesn't do try/catch around `eval` for the expression that it's given. Technically that's by design, so that it would bubble the error up to the caller - but it also prints it as a side effect, apparently...

One thing that you can do with existing code is do `TryEvaluateAndDescribeAsync` on the expression first (to minimize overhead, pass `REvaluationResultProperties.None`), and check if result is `IRErrorInfo`. If it is, then getting children would fail. If not, then you do `DescribeChildrenAsync` - that one shouldn't throw if the expression itself is valid, it does handle errors when trying to get children. 

For the future, I think it would be good to redesign this API in such a way that `EvaluateAndDescribeAsync` and `DescribeChildrenAsync` are merged - basically, `Children` should be a property in `REvaluationResultProperties`, and if you specify it, then the returned `IRValueInfo` has a filled `Children` collection. Then you could do both of these things in a single operation.

int19h:
👍

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: OK, done. Going to merge so we can build

PR #3302, opened 2017-03-14
Replace MEF by service container (services refactoring, part I)

MikhailArkhipov:
Provide unified access to all services (VS and MEF) via same entry point. Some calls to `GetExportedValue<T>` were replaced by calls to services, but not all. Intent is to encourage the respective code owners to review said code and see what needs to be done. Eventually we should remove ExportProvder from the shell.

This is first round. Next steps are to unify with core services (and rename `GlobalServices` to `Services`), refactor `IApplicationConstants` which has things like window handle, replace editor's static ServiceManager by the new one, move non-UI parts of `ICoreShell` back to portable, and more.


src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            lock (_lock) {
+                object service;
+                if (_services.TryGetValue(type, out service)) {
+                    return service as T;
+                }
+
+                // Try walk through and cast. Perhaps someone is asking for IFoo
+                // that is implemented on class Bar but Bar was added as Bar, not as IFoo
+                foreach (var s in _services.Values) {
```
MikhailArkhipov: Prob should be Linq

src/Common/Core/Impl/Services/IServiceContainer.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Common.Core.Services {
+    public interface IServiceContainer {
+        /// <summary>
+        /// Provides access to global application services
+        /// </summary>
+        T GetService<T>(Type type = null) where T : class;
```
AlexanderSher: How do you use this parameter?

src/Common/Core/Impl/Services/IServiceManager.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <param name="service">Service instance</param>
+        IServiceContainer AddService<T>(T service) where T : class;
+
+        /// <summary>
+        /// Adds on-demand created service
+        /// </summary>
+        /// <param name="type">Service type</param>
+        /// <param name="factory">Optional service factory. If not provided, reflection with default constructor will be used.</param>
+        /// <param name="factoryParameters">Optional arguments to the factory call.</param>
+        IServiceContainer AddService(Type type, IServiceFactory factory = null, params object[] factoryParameters);
```
AlexanderSher: If `IServiceContainer` is `IDisposable`, you don't need `RemoveService`

src/Common/Core/Impl/Services/PropertyDictionary.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Microsoft.Common.Core.Diagnostics;
+
+namespace Microsoft.Common.Core.Services {
+    /// <summary>
+    /// Allows property owners to control the lifetimes of the properties in the collection. 
+    /// </summary>
+    /// <remarks>This collection is synchronized in order to allow access by multiple threads.</remarks>
+    public sealed class PropertyDictionary {
```
AlexanderSher: Class looks like a limited version of a `ConcurrentDictionary`, which also supports factories (`GetOrAdd` method overload)

src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <typeparam name="T">Service type</typeparam>
+        /// <param name="service">Service instance</param>
+        public virtual IServiceContainer AddService<T>(T service) where T : class {
+            var type = typeof(T);
+            lock (_lock) {
+                Check.ArgumentNull(nameof(service), service);
+                Check.InvalidOperation(() => _services.ContainsKey(type));
+
+                _services[type] = service;
+                ServiceAdded?.Invoke(this, new ServiceContainerEventArgs(type));
```
AlexanderSher: It is usually a bad idea to raise event under lock - subscriber can block the whole class.

src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        /// <summary>
+        /// Add service to the service manager container
+        /// </summary>
+        /// <typeparam name="T">Service type</typeparam>
+        /// <param name="service">Service instance</param>
+        public virtual IServiceContainer AddService<T>(T service) where T : class {
+            var type = typeof(T);
+            lock (_lock) {
+                Check.ArgumentNull(nameof(service), service);
```
AlexanderSher: >ArgumentNull [](start = 22, length = 12)

You can check it outside lock

src/Common/Core/Impl/Services/IServiceContainer.cs
MikhailArkhipov: such as `GetService<IVsUIShell>(typeof(SVSUIShell))`

src/Common/Core/Impl/Services/IServiceManager.cs
MikhailArkhipov: I do in the editor. Services attached to buffer and view may come and go

src/Common/Core/Impl/Services/PropertyDictionary.cs
MikhailArkhipov: I started with concurrent but since there are two and we need to move service from deferred to real we'd need lock anyway so it would be atomic. So I ended up just using the lock and regular dictionary.

src/Common/Core/Impl/Services/IServiceManager.cs
AlexanderSher: Actually, do you even need to return a `IServiceContainer`?

---
In reply to: [106049535](https://github.com/Microsoft/RTVS/pull/3302#discussion_r106049535) [](ancestors = 106049535)

src/Common/Core/Impl/Services/ServiceManager.cs
MikhailArkhipov: Yes, good catch

src/Common/Core/Impl/Services/ServiceManager.cs
MikhailArkhipov: Agreed

src/Common/Core/Impl/Services/IServiceManager.cs
MikhailArkhipov: I.e. I will be using this instead of static service manager. So `ServiceManager.GetService<T>(buffer)` will be replaced by `buffer.Services.GetService<T>()`. ditto for the view.

src/Common/Core/Impl/Services/PropertyDictionary.cs
AlexanderSher: Can you point me where it is used?

---
In reply to: [106052699](https://github.com/Microsoft/RTVS/pull/3302#discussion_r106052699) [](ancestors = 106052699)

src/Common/Core/Impl/Shell/IIdleTimeService.cs
```diff
@@ -1,11 +1,14 @@
...
-using System.Threading;
-
-namespace Microsoft.Languages.Editor.Host {
+namespace Microsoft.Common.Core.Shell {
     public interface IIdleTimeService {
-        void DoIdle();
+        /// <summary>
+        /// Fires when host application enters idle state.
+        /// </summary>
+        event EventHandler<EventArgs> Idle;
```
AlexanderSher: Do we want this in cross-plat or in windows-specific part?

src/Common/Core/Impl/Services/IServiceManager.cs
MikhailArkhipov: I am returning if just for convenience like in `DisposableBag` so code can do `s.AddService().AddService()...`

src/Common/Core/Impl/Shell/IIdleTimeService.cs
MikhailArkhipov: This will be in x-plat. Activities related to parsing, task list, squiggles need idle for UI thread activities.

src/Host/API/Test/Microsoft.R.Host.Client.API.Test.csproj
```diff
@@ -48,10 +48,6 @@
       <Project>{5EF2AD64-D6FE-446B-B350-8C7F0DF0834D}</Project>
       <Name>Microsoft.UnitTests.Core</Name>
     </ProjectReference>
-    <ProjectReference Include="..\..\API\Impl\Microsoft.R.Host.Client.API.csproj">
-      <Project>{1a02711f-3b13-48be-8d45-073386354174}</Project>
-      <Name>Microsoft.R.Host.Client.API</Name>
```
AlexanderSher: ?

MikhailArkhipov: There are issues building new nuget, I will fix this in a separate PR. For now we don't need the new API package anyway.

src/Common/Core/Impl/Services/IServiceManager.cs
AlexanderSher: Ok, I'm not there yet :)

AlexanderSher: In this case. it should return `IServiceManager` rather than `IServiceContainer`.

---
In reply to: [106053919](https://github.com/Microsoft/RTVS/pull/3302#discussion_r106053919) [](ancestors = 106053919)

src/Common/Core/Impl/Services/IServiceContainer.cs
AlexanderSher: Got it

---
In reply to: [106052444](https://github.com/Microsoft/RTVS/pull/3302#discussion_r106052444) [](ancestors = 106052444)

src/Host/API/Test/Microsoft.R.Host.Client.API.Test.csproj
AlexanderSher: ok

---
In reply to: [106054311](https://github.com/Microsoft/RTVS/pull/3302#discussion_r106054311) [](ancestors = 106054311)

src/Common/Core/Impl/Services/PropertyDictionary.cs
MikhailArkhipov: Not used anywhere just yet. It will be used in the editor space to implement `IPropertyOwner` that allows attaching things to buffer and view. Current code uses VS implementation but with portable code we'll need our own.

src/Package/Impl/ProjectSystem/OpenRDataVsUiHierarchyWindowCommandGroupHandler.cs
```diff
@@ -21,7 +21,7 @@ public OpenRDataVsUiHierarchyWindowCommandGroupHandler(UnconfiguredProject uncon
             await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
AlexanderSher: While you here, can you please replace this call with our implementation?

src/Package/Impl/Shell/IApplicationShell.cs
```diff
@@ -10,15 +9,5 @@ namespace Microsoft.VisualStudio.R.Package.Shell {
...
-        /// Retrieves global service from the host application.
-        /// This method is not thread safe and should not be called 
-        /// from async methods.
-        /// </summary>
-        /// <typeparam name="T">Service interface type such as IVsUIShell</typeparam>
-        /// <param name="type">Service type if different from T, such as typeof(SVSUiShell)</param>
-        /// <returns>Service instance of null if not found.</returns>
-        T GetGlobalService<T>(Type type = null) where T : class;
-    }
+    public interface IApplicationShell : IEditorShell { }
```
AlexanderSher: Do we need this interface at all?

src/Package/Test/Shell/TestVsAppShell.cs
```diff
@@ -30,17 +30,19 @@ namespace Microsoft.VisualStudio.R.Package.Test.Shell {
...
     [ExcludeFromCodeCoverage]
     sealed class TestVsAppShell : TestShellBase, IApplicationShell {
-        private IServiceProvider _sp;
+        private readonly VsTestServiceManager _serviceManager;
         private static TestVsAppShell _instance;
-        private static object _shellLock = new object();
+        private static readonly object _shellLock = new object();
 
-        private TestVsAppShell() {
+        private TestVsAppShell(): base(VsTestCompositionCatalog.Current.ExportProvider) {
```
AlexanderSher: Why do you need an `ExportProvider` in the base class?

src/Package/Impl/Shell/IApplicationShell.cs
MikhailArkhipov: Probably not unless we'll need something VS specific. But this will be the next round. ICoreShell (basic things), IUiShell (+UI related methods), and _maybe_ editor part. The latter only translates command targets so I am thinking about turning it into a service.

src/Windows/Core/Impl/Shell/ICoreShell.cs
```diff
@@ -12,7 +11,12 @@ namespace Microsoft.Common.Core.Shell {
     /// composition container, export provider, global VS IDE
     /// services and so on.
     /// </summary>
-    public interface ICoreShell : ICompositionCatalog {
+    public interface ICoreShell : IIdleTimeService, ICompositionCatalog {
+        /// <summary>
+        /// Application-global services access
+        /// </summary>
+        IServiceContainer GlobalServices { get; }
```
AlexanderSher: It feels like `ICoreServices` can be derived from `IServiceContainer`. `CoreServices` is also a single object per application.

src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -0,0 +1,149 @@
...
+
+namespace Microsoft.Common.Core.Services {
+    public class ServiceManager : IServiceManager {
+        class ServiceFactory {
+            public IServiceFactory Factory { get; set; }
+            public object[] Parameters { get; set; }
+        }
+
+        private readonly Dictionary<Type, object> _services = new Dictionary<Type, object>();
+        private readonly Dictionary<Type, ServiceFactory> _deferredServices = new Dictionary<Type, ServiceFactory>();
```
AlexanderSher: I originally thought that factories from `_deferredServices` create new object for each call, but it looks like I got it wrong and they are also one per type. In this case, if instead of `IServiceFactory` instance and an array of parameters you provide a simple delegate `Func<T> factory`, you can store it in the dictionary as `new Lazy<T>(factory), and then access it`s value in `GetService`. Another benefit of it is that dictionary don't have to be locked while factory is executed for the first time - only callers that need service of this particular type will wait.

src/Common/Core/Impl/Services/PropertyDictionary.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// </summary>
+        /// <typeparam name="T">The type of the property.</typeparam>
+        /// <param name="key">The key of the property to get or create.</param>
+        /// <param name="creator">The delegate used to create the property (if needed).</param>
+        /// <returns>The property that was requested.</returns>
+        public T GetOrCreateSingletonProperty<T>(object key, Func<T> creator) where T : class {
+            Check.ArgumentNull(nameof(creator), creator);
+
+            lock (_lock) {
+                if (_properties.Value.ContainsKey(key)) {
```
karthiknadig: TryGetValue ?

src/Windows/Core/Impl/Shell/ICoreShell.cs
MikhailArkhipov: I am thinking about making *single* ICoreShell with `Services` and extensions to get `Log`, `Process`, etc via `GetService<T>`. Similarly, `IEditorShell` that is mapping command targets turns into 'controller service`. `IAppConstants` split into Name/Locale which go to ICoreShell and the rest goes to `IPlatformService`. Then we don't need `IApplicationShell`.

src/Common/Core/Impl/Services/ServiceManager.cs
MikhailArkhipov: I though that different ctors can have different number of arguments. Seemed more difficult with Func

src/Common/Core/Impl/Services/PropertyDictionary.cs
MikhailArkhipov: Yep

src/Common/Core/Impl/Services/ServiceManager.cs
AlexanderSher: I'm not sure I get it. Even with current signature of `AddService(Type type, IServiceFactory factory = null, params object[] parameters)` you can do it like `new Func<object>(() => factory(parameters))`

---
In reply to: [106226285](https://github.com/Microsoft/RTVS/pull/3302#discussion_r106226285) [](ancestors = 106226285)

MikhailArkhipov: OK

src/Common/Core/Impl/Services/PropertyDictionary.cs
```diff
@@ -0,0 +1,197 @@
...
+            lock (_lock) {
+                if (!_properties.IsValueCreated) {
+                    throw new KeyNotFoundException(nameof(key));
+                }
+
+                object item = _properties.Value[key];
+
+                // If item is null, it could mean the dictionary has the key but it is just null.
+                // Check Contains() to figure out whether it is a missing key or a null value.
+                if ((item == null) && !_properties.Value.ContainsKey(key)) {
```
karthiknadig: I think, this could also be done with TryGetValue.

karthiknadig: ditto... where similar pattern is used.

src/Windows/Core/Impl/Shell/ICoreShell.cs
AlexanderSher: Yeah, that's also an option. I just don't want to have 2 separate properties for it.

MikhailArkhipov: Yes, current `Services` is temporary. I just want bite-size PRs. Next time there will be only one.

src/Common/Core/Impl/Services/PropertyDictionary.cs
MikhailArkhipov: Yeah, this was lifted from old IDE code, I'll fix it.

src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -0,0 +1,140 @@
...
+            type = type ?? typeof(T);
+
+            lock (_lock) {
+                object service;
+                if (_services.TryGetValue(type, out service)) {
+                    return service as T;
+                }
+
+                // Try walk through and cast. Perhaps someone is asking for IFoo
+                // that is implemented on class Bar but Bar was added as Bar, not as IFoo
```
AlexanderSher: > Perhaps someone is asking for IFoo that is implemented on class Bar but Bar was added as Bar, not as IFoo

So the code that adds a service can't actually specify the type for registration. Not sure that I like it. First, based on the order in which dictionary will enumerate its values, this code can return different instances if there is more than one object that implements requested interface. Second, it doesn't match the logic in the `AddService` method that doesn't check for implemented interfaces.

MikhailArkhipov: It does not specify explicit type. Say, Foo implements `IFoo`, `IBar` or is derived from `Bar` that implements `IBar`. So it calls `Services.AddService(this)`. Requestor just wants `IBar`.

MikhailArkhipov: This is basically copy from static `ServiceManager`.

AlexanderSher: Ok, but what should happen if two services `Service1 : IFoo` and `Service2 : IFoo` both add themselves? Right now there will be no error inside `AddService` and it isn't clear what `GetService` will return in this case.

MikhailArkhipov: Actually also by design. This allows things similar to `ImportMany`. `GetService` will find the first available. However, one can do `Services.Where(s => s is IFoo)` to get multiple. It is not possible to register multiple by `IFoo` but possible by the implementing type. 

Currently used to import multiple controllers and command factories.

AlexanderSher: Maybe we should allow `AddService` to specify type (or types) explicitly?

MikhailArkhipov: We could. Something like ` AddService<T>(T service, Type type = null)` ?

AlexanderSher: Yes. Or `AddService<TRegister>(TService service) where TService : TRegister`.

AlexanderSher: > This allows things similar to `ImportMany`

That is another thing to think about. We don't have a `GetServices` method yet.

MikhailArkhipov: There is `IEnumerable<Type> Services { get; }` though

AlexanderSher: Yes, but it returns list of types, not services.

MikhailArkhipov: Yes. The intent is that enumeration does not instantiate services while Get does. We can add GetServices<T> that will actually instantiate. Otherwise one gets types and then calls `GetService<T>(type)` to get live instances.

I'll add tests when we settle on general approach

AlexanderSher: I'd rather go with `GetServices` approach.

AlexanderSher:
👍

Issue #3305, opened 2017-03-15
File completion on remote workspace duplicates directory entries


jflam:
Here's a screenshot - note that directory entries are duped.

![image](https://cloud.githubusercontent.com/assets/8877/23970630/b2884010-0987-11e7-87c9-61446c2e673d.png)

MikhailArkhipov:
Apparently `list.files(path = '.', include.dirs = FALSE)` returns directory names anyway

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/170e8d3586a1fa42f7a81f9144948921adaa5675

PR #3308, opened 2017-03-15
#3305 File completion on remote workspace duplicates directory entries

MikhailArkhipov:
Also fixes 14.0 references to VS SDK. Not sure how 15.0 references leaked into them.

### DIFF
--- a/src/Package/Impl/Imaging/ImagesProvider.cs
+++ b/src/Package/Impl/Imaging/ImagesProvider.cs
@@ -21,8 +21,14 @@ namespace Microsoft.VisualStudio.R.Package.Imaging {
 
     [Export(typeof(IImagesProvider))]
     internal sealed class ImagesProvider : IImagesProvider {
-        private Dictionary<string, ImageMoniker> _monikerCache = new Dictionary<string, ImageMoniker>();
-        private Lazy<Dictionary<string, string>> _fileExtensionCache = Lazy.Create(() => CreateExtensionCache());
+        private readonly Dictionary<string, ImageMoniker> _monikerCache = new Dictionary<string, ImageMoniker>();
+        private readonly Lazy<Dictionary<string, string>> _fileExtensionCache = Lazy.Create(() => CreateExtensionCache());
+
+        private static IVsImageService2 _imageService;
+
+        public ImagesProvider() {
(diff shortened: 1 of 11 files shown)


src/Package/Impl/Imaging/ImagesProvider.cs
```diff
@@ -21,8 +21,14 @@ namespace Microsoft.VisualStudio.R.Package.Imaging {
...
     internal sealed class ImagesProvider : IImagesProvider {
-        private Dictionary<string, ImageMoniker> _monikerCache = new Dictionary<string, ImageMoniker>();
-        private Lazy<Dictionary<string, string>> _fileExtensionCache = Lazy.Create(() => CreateExtensionCache());
+        private readonly Dictionary<string, ImageMoniker> _monikerCache = new Dictionary<string, ImageMoniker>();
+        private readonly Lazy<Dictionary<string, string>> _fileExtensionCache = Lazy.Create(() => CreateExtensionCache());
+
+        private static IVsImageService2 _imageService;
+
+        public ImagesProvider() {
+            _imageService = _imageService ?? VsAppShell.Current.GetGlobalService<IVsImageService2>(typeof(SVsImageService));
```
MikhailArkhipov: This must be main thread. I can try `DispatchOnUIThread` if MEF might call this async.

AlexanderSher: Why have you moved it from the `GetIconForImageMoniker`?

MikhailArkhipov: Because request for icons in remote file completion case comes from a background thread

MikhailArkhipov: Technically we may want to switch to `ConcurrentDictionary` here although probability of concurrent requests is very low.

AlexanderSher: `ConcurrentDictionary` has overhead only on write operations, so if we don't make thousands of writes here, I think we can switch.

MikhailArkhipov: Yep, done.

AlexanderSher:
👍

MikhailArkhipov:
Ouch, need another - @int19h  ?

Issue #3306, opened 2017-03-15
Reinstalling multiple packages that contain DLL fails


int19h:
```r
install.packages(c('shiny', 'ggvis', 'dplyr', 'RSQLite'))
```
Retry this several times. On some machines it's a reliable repro, with the second install (when packages already exist, and it's installing on top) giving the error. On my VM it doesn't reproduce reliably.

The error you get is "cannot remove prior installation of package 'dplyr'" etc. The errors only happen for packages that contain DLLs. Inspecting the file system shows that the packages are removed entirely, except for the DLL file.

It does not reproduce in RGui on machines on which it reliably reproduces in RTVS.

It doesn't have to be that particular set of packages, but the important part seems to be that it must be multiple packages in a single `install.packages` call, and that more then one of them includes a DLL. In this example, `dplyr` and `RSQLite` are the ones that include DLLs.

int19h:
My theory is that this has something to do with the checks that we have for DLLs being in use to safeguard against inappropriate removal. Perhaps that check happens asynchronously as the package is actually being removed by R, locking the DLL file briefly just as R tries to delete it.

int19h:
Screenshot from @smortaz repro of this:

![image](https://cloud.githubusercontent.com/assets/1175142/23971445/38ad6a96-098b-11e7-97a8-a373ff984f6b.png)

smortaz:
I see this reliably on my machine

int19h:
I was trying to repro on a VM, which does have much slower I/O among other things. So perhaps package removal is just that much slower there, which is why I cannot reliably hit it. You were doing it directly on your laptop, right?

smortaz:
Yep. Sometimes when I install them individually it doesn't happen. But all
together, repros almost all the time

AlexanderSher:
`install.packages` override looks like this:
```r
install.packages <- function(...) {
    utils::install.packages(...)
    invisible(rtvs:::send_notification('!PackagesInstalled'))
}
```
I wasn't able to repro it so far. @int19h , what dll safeguard do you mean?

int19h:
We expose an R function from the host called `get_file_lock_state` that does that check (the actual implementation is in `rstrtmgr.cpp`). It's then wrapped in `rtvs:::package_lock_state`. I thought that was used by package manager somewhere.

[...]

PR #3310, opened 2017-03-16
#3306 Reinstalling multiple packages that contain DLL fails 

MikhailArkhipov:
Stop intellisense host before package installation


src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -212,6 +224,18 @@ public sealed class PackageIndex : IPackageIndex {
                     var currentNames = _packages.Keys.ToArray();
                     var removedNames = currentNames.Except(installedNames);
                     _packages.RemoveWhere((kvp) => removedNames.Contains(kvp.Key));
+                } catch (RException) { } catch (OperationCanceledException) { } finally {
```
MikhailArkhipov: Probably excessive

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -69,10 +69,20 @@ public sealed class PackageIndex : IPackageIndex {
             }
         }
 
-        private void OnPackagesChanged(object sender, EventArgs e) {
+        private void OnBeforePackagesInstalled(object sender, EventArgs e) {
+            try {
+                _host.StopSessionAsync().Wait(3000);
```
AlexanderSher: It is interesting to check what will happen if `StopSessionAsync` throw exception

src/R/Support/Impl/Help/IntelliSenseRSession.cs
```diff
@@ -101,6 +101,21 @@ public sealed class IntelliSenseRSession : IIntellisenseRSession {
             }
         }
 
+        public async Task StopSessionAsync() {
+            var token = await _lock.ResetAsync();
+            try {
+                if (!_sessionProvider.HasBroker) {
```
AlexanderSher: Why do you care if `_sessionProvider.HasBroker == true`? I don't think its even possible to have running session without a broker.

MikhailArkhipov: Not sure. Been a while. I can remove.

MikhailArkhipov:
Added async event handler and made `BeforePackagesInstalledAsync` awaitable

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -38,7 +37,7 @@ internal sealed class RSession : IRSession, IRCallbacks {
         public event EventHandler<EventArgs> Disconnected;
         public event EventHandler<EventArgs> Disposed;
         public event EventHandler<EventArgs> DirectoryChanged;
-        public event EventHandler<EventArgs> BeforePackagesInstalled;
+        public event AsyncEventHandler<EventArgs> BeforePackagesInstalledAsync;
```
AlexanderSher: Async events have been a bug prone in CPS for years and I don't want to introduce them in release branch. We have `IRSessionCallback` where you can add another method.

AlexanderSher: Sorry for not replying earlier.

MikhailArkhipov: OK. What's wrong with them? Currently there is only one listener.  Async call is on `IRCallbacks.BeforePackagesInstalledAsync`. Do we want then to add a method to `IRSessionCallback/RSessionCallback` and call the index directly?

AlexanderSher: > OK. What's wrong with them?

- Async events don't support cancellation tokens. `CancellationToken` can be passed through as a property of `EventArgs`, but it isn't clear what to do if one of the handlers have thrown `OperationCanceledException` by itself.
- Sync event stops invocation when any of the handlers throw exception. Async handlers may or may not follow this rule. With more than one exception being produced during one invocation, we have to throw `AggregateException` which requires lots of code to handle. 
- `OperationCanceledException` vs regular exception. If event was canceled, should we still throw exception from one of the handlers or it should be swallowed?

And so on. Async event requires some good design before it can be widely used in application. Not to mention that people tend to forget to call `InvokeAsync` and use `Invoke` instead, and if there is only one handler, there will be no bug until the next handler is added.

> Do we want then to add a method to `IRSessionCallback/RSessionCallback` and call the index directly?

Yes, similar to plots.

MikhailArkhipov: OK, scary enough :-)

MikhailArkhipov: This is complicated. `IPackageIndex` is in Support that depends on Components but callback is in `Components` and knows nothing about `IPackageIndex`. I can try and move interfaces but change will be huge

MikhailArkhipov: Option 1 is to leave async event alone. There is only one listener anyway. We are sorting out dependencies in master and my plan is to get rid of Support assembly anyway.

AlexanderSher: I believe you need to move 4 interfaces: `IPackageIndex`, `IPackageInfo`, `INamedInfo` and `NamedItemType`. If you're going to get rid of the `Support` assembly, you'll have to do it anyway.

AlexanderSher: We don't have to change their namespaces or add any additional references.

MikhailArkhipov: Yeah but they won't be moving to components... Components is Windows assembly while index is x-plat. I can craft something line `IPackageInstallationListener` or something.

AlexanderSher: Components isn't a Windows assembly - it will be split into Windows and xplat. Actually, everything except view and viewmodel will be xplat

MikhailArkhipov: package index is editor specific so it goes to R.Editor

AlexanderSher: Yeah, in this case we need some common interface that can be accessed by both R.Editor and R.Components

MikhailArkhipov: Added `IPackageInstallationNotifications`

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -69,15 +70,16 @@ public sealed class PackageIndex : IPackageIndex {
             }
         }
 
-        private Task OnBeforePackagesInstalledAsync(object sender, EventArgs e) {
+        public Task BeforePackagesInstalledAsync(CancellationToken ct) {
             try {
-                _host.StopSessionAsync().Wait(3000);
+                _host.StopSessionAsync(ct).Wait(3000);
```
AlexanderSher: `await _host.StopSessionAsync(ct)`

MikhailArkhipov: I didn't want to wait forever here. Just in case it doesn't stop.

MikhailArkhipov: Can bump timeout up

src/Host/Client/Impl/rtvs/R/overrides.R
```diff
@@ -55,8 +55,9 @@ show_file <- function(files, header, title, delete.file) {
 }
 
 install.packages <- function(...) {
+    invisible(rtvs:::send_notification('!BeforePackagesInstalled'))
```
int19h: Now that the callback is wired up async, this should be `send_request_and_wait_response`.

MikhailArkhipov: It is awaiting - does it still need response?

src/R/Support/Impl/Help/Packages/PackageIndex.cs
AlexanderSher: No, call should be async. If you need a timeout, you can make a `cancellationToken` with timeout:
```cs
var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
timeoutCts.CancelAfter(3000);
await _host.StopSessionAsync(timeoutCts.Token);
```

MikhailArkhipov: OK. Or just return `return _host.StopSessionAsync(timeoutCts.Token);`

src/Host/Client/Impl/rtvs/R/overrides.R
int19h: That C# code is awaiting doesn't have any effect on R code here - `send_notification` sends the message and immediately returns. So all that await in C# code does currently is prevent further handling of new incoming messages until that notification is handled. For R to wait until C# callback completes, it needs to be changed to request/response.

AlexanderSher:
👍

MikhailArkhipov:
@int19h checked in person

PR #3324, opened 2017-03-21
Fixing build of Release/1_0 in VS 2017.

mvacha:
1. The Microsoft.R.Host.vcxproj is missing from the repo, but referenced in Microsoft.VisualStudio.R.Package.csproj.
2. And the dlls: zip.dll and zlib.dll are also just referenced, but nowhere to be found.
Both of these make the build in VS 2017 fail.

### DIFF
--- a/src/Host/API/Impl/Microsoft.R.Host.Client.API.nuspec
+++ b/src/Host/API/Impl/Microsoft.R.Host.Client.API.nuspec
@@ -144,7 +144,7 @@
     <file src="System.Xml.XPath.dll" target="build" />
     <file src="System.Xml.XPath.XDocument.dll" target="build" />
     <file src="libuv.dll" target="build" />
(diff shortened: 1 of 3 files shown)


mvacha:
Oh, I see you haven't got a CI setup, so I'll test the build also in VS2015, run the tests and see if I didn't broke something.

MikhailArkhipov:
The C++ project is required it is host process for R. It is in a separate repo per GPL requrement. It is listed as submodule. https://github.com/Microsoft/R-Host. zlib comes with it. Package references it (although you get a warning since it is native) so the host is built before the package.

We build in a lab that does everything from clean slate.

int19h:
See the bit about submodules at https://github.com/Microsoft/RTVS/blob/master/CONTRIBUTING.md#getting-the-source-code

PR #3326, opened 2017-03-21
3325 (Release 1.0)

MikhailArkhipov:
The problem is that MRC installs over existing bits but does not remove old registry keys. So we end up seeing 3.2.2 and 3.3.2 pointing to the same path. Fetching first from the set take 3.2.2. Fixes

a) Stop special casing SQL. It is already in the registry anyway.
b). Take all found engines, select Microsoft ones, de-duplicate by install location and take highest version.
c). In connection manager always replace automatically detected engines since path can be the same but name can change after the upgrade.
d) Fix MRC download link.

### DIFF
--- a/src/R/Interpreters/Impl/Implementation/RInstallation.cs
+++ b/src/R/Interpreters/Impl/Implementation/RInstallation.cs
@@ -68,7 +68,11 @@ namespace Microsoft.R.Interpreters {
             var list = new List<IRInterpreterInfo>();
 
             foreach (var path in installLocations) {
-                var value = microsoftEngines.Where(e => e.InstallPath.EqualsIgnoreCase(path)).OrderByDescending(e => e.Version).First();
+                var value = microsoftEngines
+                    .Where(e => e.InstallPath.TrimTrailingSlash().EqualsIgnoreCase(path))
+                    .OrderByDescending(e => e.Version)
+                    .First();
+
                 list.Add(value);
             }
 
--- a/src/R/Interpreters/Impl/Implementation/RInstallation.cs
+++ b/src/R/Interpreters/Impl/Implementation/RInstallation.cs
@@ -77,21 +77,6 @@ namespace Microsoft.R.Interpreters {
         }
 
         /// <summary>
-        /// Attempts to extract version from registry key.For example '3.3.2 Microsoft R Client'. 
-        /// Returns null if first part of the key name does not appear to be the version.
-        /// </summary>
-        /// <param name="key"></param>
-        /// <returns>Version or null if key doesn't represent version</returns>
-        private Version VersionFromKey(string key) {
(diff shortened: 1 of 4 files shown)


MikhailArkhipov:
@AlexanderSher @int19h - please review. Thanks.

src/R/Interpreters/Impl/Implementation/RInstallation.cs
```diff
@@ -82,13 +62,22 @@ public sealed class RInstallation {
...
             svr = svr ?? new SupportedRVersionRange();
             var engines = GetInstalledEnginesFromRegistry().Where(e => svr.IsCompatibleVersion(e.Version));
-            // Remove duplicates (MRC registers under multiple keys)
-            var mrc = engines.FirstOrDefault(e => e.Name.Contains("Microsoft"));
-            if(mrc != null) {
-                var dupes = engines.Where(e => e.InstallPath.EqualsIgnoreCase(mrc.InstallPath)).Except(new IRInterpreterInfo[] { mrc });
-                engines = engines.Except(dupes);
+            // Among duplicates by path (MRC registers under multiple keys) take the highest version
+            var microsoftEngines = engines.Where(e => e.Name.Contains("Microsoft") || e.InstallPath.Contains("Microsoft"));
+            var installLocations = microsoftEngines.Select(e => e.InstallPath.TrimTrailingSlash()).Distinct();
```
int19h: I think what you want to do here is group by path, then take highest in each group. Probably easier to write as a query; something like this:
```cs
var microsoftEngines =
  from e in engines
  where e.Name.Contains("Microsoft") || e.InstallPath.Contains("Microsoft")
  group e by e.InstallPath.TrimTrailingSlash() into g
  select g.OrderByDescending(e => e.Version).First();
```
Also, I think you might as well dedup all keys, not just MRC. If they never duplicate, it doesn't really do anything. But in case they do, this would take care of them all.

MikhailArkhipov: Yes, something like that. I unrolled for easier debugging. If CRAN duplicates paths, then its fine. There problem here is that MRC/MRS update in place without deleting old keys. Not sure if we want to extend heuristics (in fact, fixing their bug) to everyone.

int19h: Why would we want to display duplicate CRAN entries, if two keys are registered to the same path? It would seem that single path = single workspace, regardless of which interpreter that is.

Basically, the only case this would apply is if someone does end up creating such duplicate paths. If they do, there's no harm in cleaning that up. If they don't, then none of this logic will kick in.

MikhailArkhipov: OK

int19h:
👍

Issue #3319, opened 2017-03-20
Move suggest_mro from RHost to RSession


karthiknadig:
On linux, we cannot call evals before the main loop has started.

PR #3330, opened 2017-03-22
Moving suggest MRO from RHost to RTVS

karthiknadig:
#3319 . In linux, we cannot call eval before we start the main loop.


karthiknadig:
@AlexanderSher can you take a look at this?

AlexanderSher:
+1

Issue #3333, opened 2017-03-23
Acquire RInstallation and RInterpreterInfo through IServiceProvider


AlexanderSher:
Also, remove ICoreShell workarounds
Can be done after https://github.com/Microsoft/RTVS/pull/3320 is checked in

AlexanderSher:
Fixed

PR #3335, opened 2017-03-23
Split Microsoft.R.Interpreters and Microsoft.R.Host.Broker into xplat and windows-specific parts

AlexanderSher:
This is not a final change, but I'm submitting it early so that @karthiknadig can take a look.


karthiknadig:
>     public class AboutController : Controller {

I think this class should be in xplat. We should provide an interface to get the OS specific stats.

---
Refers to: src/Host/Broker/Impl/About/AboutController.cs:20 in 3a6c48e. [](commit_id = 3a6c48e7591af250af4dbe20003d9bd1f6de27b8, deletion_comment = True)

karthiknadig:
>     public class Session {

Should we just provide a OS specific session factory and get session instances like that. Since most of the code for SessionManager and SessionController did not change between windows and linux

---
Refers to: src/Host/Broker/Impl/Sessions/Session.cs:23 in 3a6c48e. [](commit_id = 3a6c48e7591af250af4dbe20003d9bd1f6de27b8, deletion_comment = True)

src/Host/Client/Impl/Host/LocalBrokerClient.cs
```diff
@@ -20,7 +20,7 @@
 
 namespace Microsoft.R.Host.Client.Host {
     internal sealed class LocalBrokerClient : BrokerClient {
-        private const string RHostBrokerExe = "Microsoft.R.Host.Broker.exe";
+        private const string RHostBrokerExe = "Microsoft.R.Host.Broker.Windows.exe";
```
MikhailArkhipov: What is different in Windows broker?

src/R/Interpreters/Impl/IMicrosoftRClientInstaller.cs
```diff
@@ -10,6 +11,6 @@ namespace Microsoft.R.Interpreters {
     /// of the Microsoft R Client. Exported via MEF.
     /// </summary>
     public interface IMicrosoftRClientInstaller {
-        void LaunchRClientSetup(ICoreShell coreShell, IFileDownloader downloader = null);
+        void LaunchRClientSetup(IServiceContainer coreShell, IFileDownloader downloader = null);
```
MikhailArkhipov: This really should go to package and then Interpreters and broker are always x-plat

MikhailArkhipov:
I think we wanted to have broker and dependencies x-plat. If anything there is not (such as RInstallation or MRC) we should turn them into services and move implementation to the package or into some other supporting assembly (components?).

src/Host/Client/Impl/Host/LocalBrokerClient.cs
AlexanderSher: It's the same broker, just different name  cause xplat assembly is `Microsoft.R.Host.Broker.dll`

src/R/Interpreters/Impl/IMicrosoftRClientInstaller.cs
AlexanderSher: This is xplat already. If you mean the part that is in the `Microsoft.R.Interpreters.Windows`, broker either need to load it dynamically or reference directly, but in both cases I'd rather leave it in a small separate assembly rather than make it part of a `Microsoft.VisualStudio.R.Package`.

AlexanderSher:
I don't know if we can get rid of platform-specific executables. @karthiknadig , what do you think is better in our case - have two separate executable broker projects for windows and for Linux that reference same shared xplat part, or have one executable broker project that is compiled for all platforms and loads platform-specific part depending on the platform in runtime?

Reversing dependencies isn't a big deal, cause on the type level those are services already, but it won't change the amount of assemblies. Broker-specific platform-dependent code (i.e., starting broker as a service or starting RHost process) isn't needed it the RTVS. On the other hand, `RInstallation` is needed everywhere. So I think it's better to split them.

MikhailArkhipov:
So what is the platform specific part in the broker?

int19h:
I'd rather push xplat conditionals as far down into the libraries as possible.

That said, broker might still have some platform-specific code because it's a service. This will look differently elsewhere.

karthiknadig:
The following are the platform specific parts in broker:
1. Interpreters search/ path validation etc.
2. User profile management
3. User authentication
4. RHost session management (.net Process in linux vs Win32 Process in Windows)
5. Operating system info (RAM, VRAM, etc)
6. OS load and performance info

MikhailArkhipov:
These should be provided as services implemented elsewhere. I.e. RInstallation -> IRInstallation (see #3333 for example). We already have portable service container. So basically there should be a platform-specific supporting assembly providing platform services in a generic way.

src/Host/Broker/Impl/About/LoadController.cs
```diff
@@ -37,62 +32,30 @@ public class LoadController : Controller {
...
 
             // Select compatible active network interfaces
             var interfaces = NetworkInterface.GetAllNetworkInterfaces()
-                                             .Where(x => x.Speed > 0 &&
-                                                         x.Supports(NetworkInterfaceComponent.IPv4) &&
-                                                         x.Supports(NetworkInterfaceComponent.IPv6) &&
-                                                         x.OperationalStatus == OperationalStatus.Up &&
-                                                         x.Speed > 0 &&
-                                                         IsCompatibleInterface(x.NetworkInterfaceType)).ToArray();
+                .Where(x => x.Speed > 0
```
karthiknadig: This throws Platform Not supported exception on linux. Sadly there is no way to provide this information in a generic way.

AlexanderSher: Ok, I'll move it to `SystemInfoService` then.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -4,196 +4,78 @@
...
                     }
-                } catch (WebException wex) when (wex.Status == WebExceptionStatus.ProtocolError) {
-                    response = wex.Response as HttpWebResponse;
-                } finally {
-                    response?.Close();
                 }
             }
         }
-
-        private static void SetRequestHeaders(HttpWebRequest request, IHeaderDictionary requestHeaders) {
```
karthiknadig: Nice

src/Host/Broker/Impl/Services/IExitService.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.R.Host.Broker.Services {
+    public interface IExitService {
+        void Exit();
```
karthiknadig: We will also need a way to Exit with Error code.

src/Host/Broker/Impl/Services/ISystemInfoService.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.R.Host.Broker.Services {
+    public interface ISystemInfoService {
+        double GetCpuLoad();
+        (long TotalVirtualMemory, long FreeVirtualMemory, long TotalPhysicalMemory, long FreePhysicalMemory) GetMemoryInformation();
+        (string VideoCardName, long VideoRAM, string VideoProcessor) GetVideoControllerInformation();
```
karthiknadig: This may have to be a collection since there can be more than one video card

src/Host/Broker/Impl/Services/IExitService.cs
AlexanderSher: This is a temporary solution. I've created a task https://github.com/Microsoft/RTVS/issues/3411 to track refactoring of `CommonStartup`

src/Host/Broker/Impl/Services/ISystemInfoService.cs
AlexanderSher: Well, right now we take care only about the first one.

AlexanderSher:
Ping

karthiknadig:
:shipit:

Issue #3337, opened 2017-03-24
On localized VS and OS messages in rtvs package are converted incorrectly


MikhailArkhipov:
CHT UI. I guess VS locale is CHT while R is not. Should we be automatically setting R code page?

![image](https://cloud.githubusercontent.com/assets/12820357/24275306/be3bee7e-0feb-11e7-9b4c-311497f644b1.png)

MikhailArkhipov:
VSTS 399410

MikhailArkhipov:
Also relates to #3317

MikhailArkhipov:
Reproes if R code page is set to Chinese Tranditional (CNS) 

![image](https://cloud.githubusercontent.com/assets/12820357/24309138/e1b21f12-1087-11e7-83e4-b0cd3441b8d1.png)

Does NOT repro when set to Big5

MikhailArkhipov:
UI = CHT, OS = ENU, R Code Page = OS Default (437)

![image](https://cloud.githubusercontent.com/assets/12820357/24311564/f8f2c46a-1091-11e7-9e64-e8f694401d98.png)

MikhailArkhipov:
Logs appear to be correct 

```
  "rtvs:::query_reload_autosave()"

]

	#20# = rtvs:::query_reload_autosave()

	<== #99# ?LocStr ["rtvs_SessionTerminatedUnexpectedly"]

		==> #22# :LocStr #99# [

  "已终止上一个 R 会话，并且已将它的全局工作区保存到图像“%s”。是否要重新加载它?"

]

	<== #101# !! ["Error in winDialog(\"yesno\", sprintf(msg, autosave_filename)) : \n  invalid 'message' argument\n"]

		<== #103# :=@ #20# ["OK","Error in winDialog(\"yesno\", sprintf(msg, autosave_filename)) : \n  invalid 'message' argument\n",null]

```

and manual code execution produces correct result

![image](https://cloud.githubusercontent.com/assets/12820357/24312111/5ef599ac-1094-11e7-9702-51979fc711cb.png)

MikhailArkhipov:
Looks like `winDialog` does not like UTF-8

![image](https://cloud.githubusercontent.com/assets/12820357/24312200/b597901c-1094-11e7-8495-863f4513bbfe.png)

MikhailArkhipov:
And double quotes turn into fancy quotes which causes trouble
```
	#20# = rtvs:::query_reload_autosave()
[...]

PR #3350, opened 2017-03-26
#3337 On localized VS and OS messages in rtvs package are converted incorrectly 

MikhailArkhipov:
**Fix for VS 2017 1.0**

Long story in #3337. 

Basically we can't be using `message`, `warning`, `sprintf`, `winDialog` and friends in R to output localized content. They depend on currently set R locale. `winDialog` doesn't seem to be able to handle UTF-8 and tries to convert to OEM. `sprintf` appears to need proper C++ locale set (probably cp65001) to be able to handle UTF-8.

So the fix is to handle everything at the client.
- locstr removed
- added `loc_message`, `loc_warning`, `loc_yesno` and the respective parts in host
- host now supports `string.format` like functionality
- patched lcl files since `%s` changed to .NET formatting `{0}`

image

image

### DIFF
--- a/src/Host/Client/Impl/Host/RHost.cs
+++ b/src/Host/Client/Impl/Host/RHost.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
@@ -571,7 +572,7 @@ namespace Microsoft.R.Host.Client {
                                 break;
 
                             case "?LocStr":
-                                await RespondAsync(message, ct, _callbacks.GetLocalizedString(message.GetString(0, "id")));
+                                await RespondAsync(message, ct, GetLocalizedString(message));
                                 break;
 
                             default:
@@ -594,6 +595,15 @@ namespace Microsoft.R.Host.Client {
             }
 
             return null;
+        }
+
+        private string GetLocalizedString(Message message) {
+            var s = _callbacks.GetLocalizedString(message.GetString(0, "id"));
+            if (message.ArgumentCount == 2) {
+                var args = message.GetArgument(1, "a", JTokenType.Array).Select(o => o.Value<object>()).ToArray();
+                s = string.Format(CultureInfo.CurrentCulture, s, args);
+            }
+            return s;
         }
 
         private CancellationToken UpdateCancelAllCtsLink(ref CancellationTokenSource cancelAllCtsLink, CancellationToken loopCt) {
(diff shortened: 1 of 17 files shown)


src/Host/Client/Impl/Host/RHost.cs
```diff
@@ -558,8 +590,12 @@ public sealed partial class RHost : IDisposable, IRExpressionEvaluator, IRBlobSe
                                 }).DoNotWait();
                                 break;
 
-                            case "?LocStr":
-                                await RespondAsync(message, ct, _callbacks.GetLocalizedString(message.GetString(0, "id")));
+                            case "!LocMessage":
```
int19h: Should these also be formatting, for flexibility?

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -283,36 +290,35 @@ query_reload_autosave <- function() {
...
     }
 }
 
 save_state <- function() {
-    # This function runs when client is already disconnected, so locstr() cannot be used, since it requires
+    # This function runs when client is already disconnected, so loc_message() cannot be used, since it requires
     # a connected client to provide translated strings. However, since messages are not user-visible and are
     # here for logging purposes only, they don't need to be localized in the first place.
-    message(sprintf('Autosaving workspace to image "%s" ...', autosave_filename), appendLF = FALSE);
+    #message(sprintf('Autosaving workspace to image "%s" ...', autosave_filename), appendLF = FALSE);
```
int19h: Why comment these out?

src/Host/Client/Impl/Host/RHost.cs
int19h: Oh, never mind, they are, I didn't read it right.

src/Host/Client/Impl/rtvs/R/util.R
MikhailArkhipov: They are sometimes visible. Do we need them? Just to avoid loc bugs.

int19h: What's the case where they are visible? If so, the comment above is wrong and should be corrected.

MikhailArkhipov: I'll try to repro again - but - they are not supposed to be visible... when do user see them? I can leave them alone, after all, there is no bug really,

int19h: To clarify. They are *not* supposed to be user visible - that's why they weren't localized. But, they get printed out to the host log, along with other traffic. That can be helpful sometimes.

int19h:
👍

Issue #3331, opened 2017-03-22
Add option in broker to use PFX files


karthiknadig:
This method of aquiring the x509 certificate is preferred on linux.

int19h:
There's usually a centralized certificate registry in `/etc/ssl`. Ideally we should just reference stuff from there.

karthiknadig:
Apparently the localmachine store on linux is read only and does not support private keys. I was not able to get it to work using the local machine store.

PR #3351, opened 2017-03-27
Adds option to use certificate file for use with TLS (SSL)

karthiknadig:
#3331

### DIFF
--- a/src/Host/Broker/Impl/Security/Certificates.cs
+++ b/src/Host/Broker/Impl/Security/Certificates.cs
@@ -1,16 +1,30 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.Common.Core;
+using static System.FormattableString;
 
 namespace Microsoft.R.Host.Broker.Security {
     internal static class Certificates {
-        public static X509Certificate2 GetCertificateForEncryption(string certName) {
-            return FindCertificate(certName);
-        }
+		public static X509Certificate2 GetCertificateForEncryption(SecurityOptions securityOptions) {
+			X509Certificate2 certificate = null;
+			if (string.IsNullOrWhiteSpace(securityOptions.X509CertificateFile)) {
+				var certName = securityOptions.X509CertificateName ?? Invariant($"CN={Environment.MachineName}");
+				certificate = FindCertificate(certName);
+			} else {
+				if (securityOptions.X509CertificatePassword != null) {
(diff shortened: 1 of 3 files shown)


src/Host/Broker/Impl/Security/SecurityOptions.cs
```diff
@@ -14,5 +14,15 @@ public class SecurityOptions {
         /// Friendly name of the certificate installed for the TLS (SSL)
         /// </summary>
         public string X509CertificateName { get; set; }
+
+		/// <summary>
+		/// Path to the certificate for use with TLS (SSL)
+		/// </summary>
+		public string X509CertificateFile { get; set; }
```
AlexanderSher: Formatting

src/Host/Broker/Impl/Security/SecurityOptions.cs
```diff
@@ -14,5 +14,15 @@ public class SecurityOptions {
...
+
+		/// <summary>
+		/// Password to access private key in the certificate <see cref="X509CertificateFile"/>
+		/// </summary>
+		public string X509CertificatePassword { get; set; }
```
AlexanderSher: Plain string?

src/Host/Broker/Impl/Security/Certificates.cs
```diff
@@ -1,16 +1,30 @@
...
 namespace Microsoft.R.Host.Broker.Security {
     internal static class Certificates {
-        public static X509Certificate2 GetCertificateForEncryption(string certName) {
-            return FindCertificate(certName);
-        }
+		public static X509Certificate2 GetCertificateForEncryption(SecurityOptions securityOptions) {
+			X509Certificate2 certificate = null;
+			if (string.IsNullOrWhiteSpace(securityOptions.X509CertificateFile)) {
+				var certName = securityOptions.X509CertificateName ?? Invariant($"CN={Environment.MachineName}");
+				certificate = FindCertificate(certName);
```
AlexanderSher: `return FindCertificate(certName)`

src/Host/Broker/Impl/Security/SecurityOptions.cs
karthiknadig: This will be plain string stored in the config json file. It is a passphrase that protects the private key in a PFX cert file. The issue with setting a pass-phrase with the PFX file is that each time the server restarts some one will have to enter this pass-phrase. So storing it in the config file is one way to make this easy. The other option is to use a PFX file without a pass-phrase. The security here boils down to the integrity of the system itself. If a hacker can get access to the config file, the system is already compromised. 

Now, as for this property. I could not find a way to get configuration binding to bind to a SecureString. If there is a way let me know and I can make the change.

MikhailArkhipov: Yeah, similar to storing passwords in web.config

AlexanderSher:
👍

Issue #3356, opened 2017-03-28
Crash on create new remote workspace


tkkasimov:
I'm have remote Windows Server 2012 with SQL Server 2016 and installed R Services and I'm want to connect to this from my notebook over vpn. Notebook environment: Windows 10, Visual Studio 2015 with update 3, .NET Framework 4.6.2. When I try add new workspace https://"remoteip":5444, visual studio crashes. 

Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ComponentModel.Win32Exception
   at Microsoft.Common.Core.CredentialHandle.ReadFromCredentialManager(System.String)
   at Microsoft.Common.Core.Security.SecurityUtilities.GetUserName(System.String)
   at Microsoft.R.Host.Client.Host.BrokerConnectionInfo.Create(System.String, System.String, System.String)
   at Microsoft.R.Components.ConnectionManager.Connection..ctor(Microsoft.R.Components.ConnectionManager.IConnectionInfo)
   at Microsoft.R.Components.ConnectionManager.Implementation.ConnectionManager.AddOrUpdateConnection(Microsoft.R.Components.ConnectionManager.IConnectionInfo)
   at Microsoft.R.Components.ConnectionManager.Implementation.ViewModel.ConnectionManagerViewModel.Save(Microsoft.R.Components.ConnectionManager.ViewModel.IConnectionViewModel)
   at Microsoft.R.Components.ConnectionManager.Implementation.View.ConnectionManagerControl.ButtonSave_Click(System.Object, System.Windows.RoutedEventArgs)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(System.Object, System.Windows.RoutedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(System.Object, System.Windows.RoutedEventArgs, Boolean)
   at System.Windows.UIElement.RaiseEventImpl(System.Windows.DependencyObject, System.Windows.RoutedEventArgs)
   at System.Windows.UIElement.RaiseEvent(System.Windows.RoutedEventArgs)
   at System.Windows.Controls.Primitives.ButtonBase.OnClick()
   at System.Windows.Controls.Button.OnClick()
   at System.Windows.Controls.Primitives.ButtonBase.OnMouseLeftButtonUp(System.Windows.Input.MouseButtonEventArgs)
[...]

Issue #2610, opened 2016-10-31
Snippets are not respecting trailing spaces


mungojam:
As an example, type =, then tab to produce the <- snippet and the cursor ends up after the - with no trailing space.

The snippet correctly uses $end$ which should make the whitespace get noticed. 

```
<Code Language="R">
        <![CDATA[ <- $end$]]>
 </Code>
```

jflam:
@MikhailArkhipov - any update on this bug? Thx.

MikhailArkhipov:
It is basically formatting. After snippet insertion VS IDE calls format range so snippet would appear according to the selected formatting options rather than how it specified in the snippet file. Formatter trims trailing spaces by default.  It is technically possible to special case standalone operators and suppress formatting in this particular case.

MikhailArkhipov:
So question is: is this specifically about R operator snippets or is it general request to respect custom formatting with possible trailing spaces in all snippets. The former is doable, the latter is unclear how to handle since reformatting of snippets at insertion is standard VS IDE behavior.

mungojam:
Thanks for explaining the cause of the issue. I've not had the same issue
with C# but perhaps that's just because I haven't needed trailing spaces in
a snippet.

Could there be a way of altering the formatting rules so that it doesn't
get rid of the space after <-  ? It's rare to put a line break right after
that token so it seems to be a reasonable exception for R code
soecifically.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/7ecfc09d1653ce711bcea3e71b30b8972016edda

mungojam:
Thanks, just re-read your comment and realise what you meant. Your fix is a very sensible balance as presumably it will work for magrittr lines too if somebody has a snippet for that like: 

blah %>% 
select

The main motivation was specifically for <- though, so great that it works for that

PR #3360, opened 2017-03-28
#2610 #3356

MikhailArkhipov:
#3356 Crash on create new remote workspace 
- We really should not be throwing Win32 exceptions. Complete rework is out of question in Release/1_0 so I plumbed console output in one place and direct message box in another. Unfortunately, passing down ICoreShell is way too big change. Also, UI is frozen for localization, so reusing message from Core.

#2610 Snippets are not respecting trailing spaces 
- just check for standalone operator and bypass formatting


karthiknadig:
:shipit:

Issue #3352, opened 2017-03-27
Edit(file) opens file in Notepad


MikhailArkhipov:
Should be opening inside VS like View does.

![image](https://cloud.githubusercontent.com/assets/12820357/24372803/07108bf0-12e4-11e7-918d-4cf081788171.png)

MikhailArkhipov:
Correct command `edit(file = "./script.R")`

MikhailArkhipov:
Implemented, but change is fairly big although does not touch other areas, just plumbing up new command override.

MikhailArkhipov:
The implementation opens in the editor. VS UI is fully available but REPL is blocked since `edit` must return value (contents of the editor). REPL is unblocked when editor tab is closed. RStudio opens modal dialog, but  this is not something VS normally would do. Need testing since operation on RTVS UI may deadlock as REPL is blocked on the editor...

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/740694ef3fa7d1ad53f94d00872eac58e99f6f4c

PR #3362, opened 2017-03-29
#3352 Edit(file) opens file in Notepad 

MikhailArkhipov:
Special thanks to @AlexanderSher 
Reported at http://www.codinggeekette.com/why-does-rtvs-open-notepad

Implement override for `edit` command to open internal VS editor. REPL is blocked but VS UI is available for editing code in the regular editor. Tested cancellation, closing VS while command is in progress, resetting REPL. Package manager and VE are blocked until REPL is released.


src/Host/Client/Impl/rtvs/R/overrides.R
```diff
@@ -54,6 +54,20 @@ show_file <- function(files, header, title, delete.file) {
...
+# Override for edit(...). Opens file in VS editor and blocks until the file is closed.
+edit_file <- function(name, file, title) {
+    if(!is.null(name) && is.function(name)) {
+        source <- send_request_and_get_response("?EditFile", paste0(deparse(name), collapse = '\n'), NULL);
+        parsed <- try(eval.parent(parse(text = source)));
+        return(parsed);
+    } else if(!is.null(file)) {
+        send_request_and_get_response("?EditFile", NULL, file);
+    } else {
+        utils:::edit(name, file, title);
```
int19h: No need for triple colon, it's public, so double works.

src/Host/Client/Impl/rtvs/R/overrides.R
```diff
@@ -54,6 +54,20 @@ show_file <- function(files, header, title, delete.file) {
...
 
+
+# Override for edit(...). Opens file in VS editor and blocks until the file is closed.
+edit_file <- function(name, file, title) {
+    if(!is.null(name) && is.function(name)) {
+        source <- send_request_and_get_response("?EditFile", paste0(deparse(name), collapse = '\n'), NULL);
+        parsed <- try(eval.parent(parse(text = source)));
+        return(parsed);
+    } else if(!is.null(file)) {
+        send_request_and_get_response("?EditFile", NULL, file);
```
int19h: This will implicitly return the result of this function, whereas if I understand correctly it should really be `NULL`. Needs an explicit `return`.

src/Host/Client/Impl/rtvs/R/overrides.R
int19h: Actually, isn't this recursive now? `edit()` will eventually invoke `options('editor')`, which is this function. If this function then invokes `edit()` again, it's just going to loop.

int19h: In fact, this function should only be called on objects that are functions. On everything else, `edit()` just goes directly to the built-in editor. Here's what `edit.default` looks like:
```r
function (name = NULL, file = "", title = NULL, editor = getOption("editor"), ...) 
{
    if (is.null(title)) 
        title <- deparse(substitute(name))
    if (is.function(editor)) 
        invisible(editor(name = name, file = file, title = title))
    else .External2(C_edit, name, file, title, editor)
}
```

int19h: So I think the third branch is one of those "it should never happen" cases, and it is safe to just return `NULL` from it and do nothing.

src/Host/Client/Impl/rtvs/R/overrides.R
int19h: I think `edit` will handle that part, `editor` is just supposed to return the new content of it.

int19h: Although no, you're right, it needs to eval it as an expression and return the value. If you use different variable names in REPL and in the file, you'll see that variables set in the file do not escape. What escapes is the result (because assignment is an operator, and returns the assigned value).

src/Host/Client/Impl/rtvs/R/overrides.R
MikhailArkhipov: For whatever reason it works as intended...

image

MikhailArkhipov: Folks want files opened on VS too

---
In reply to: [108759364](https://github.com/Microsoft/RTVS/pull/3362#discussion_r108759364) [](ancestors = 108759364)

int19h: It will work on dataframes, because `edit` is actually an S3 method, and it has overload specifically for datraframes. What I posted above is `edit.default`, which is for everything else. Try doing something like `edit(123)`.

MikhailArkhipov: Currently, nothing. RGui opens scratch file with said content. We could do the same.

int19h: That's because when it's a number, it doesn't even hit our `editor` function. That's what I meant when I said that the third branch is a it-never-happens case. We shouldn't be trying to call `utils::edit` in it - if it *does* actually ever get invoked, that would be wrong (recursion), and if it doesn't, then that code doesn't do anything useful (but implies that it does).

src/Host/Client/Impl/rtvs/R/overrides.R
```diff
@@ -54,18 +54,29 @@ show_file <- function(files, header, title, delete.file) {
...
-    } else {
-        utils:::edit(name, file, title);
+    } else if(!is.null(file) && is.character(file) && file != "") {
+        source <- send_request_and_get_response("?EditFile", NULL, file);
+    }
+    if(!is.null(source)) {
+        result <- try(eval.parent(parse(text = source)));
+        return(result);
+    }
+    editor <- getOption("externalEditor");
```
int19h: Why do we need any of this? If we got here, it was neither `edit(func)` nor `edit(file=...)`. We should just bail out. Or is there some case you were hitting when the input was something else, and we do end up in this function anyway?

src/Host/Client/Impl/rtvs/R/overrides.R
```diff
@@ -54,18 +54,29 @@ show_file <- function(files, header, title, delete.file) {
...
+    source <- NULL;
+    if(is.function(name)) {
         source <- send_request_and_get_response("?EditFile", paste0(deparse(name), collapse = '\n'), NULL);
-        parsed <- try(eval.parent(parse(text = source)));
-        return(parsed);
-    } else if(!is.null(file)) {
-        send_request_and_get_response("?EditFile", NULL, file);
-    } else {
-        utils:::edit(name, file, title);
+    } else if(!is.null(file) && is.character(file) && file != "") {
```
int19h: `!identical(file, "")`

(since it can be a vector of more than one name, in which case using `!=` will give you a vector of booleans)

src/Host/Client/Impl/rtvs/R/overrides.R
MikhailArkhipov: Yes. It does call use back on editor(123). We are overriding editor and need to pass something to External2. It does not accepts nulls or missing arguments. I don't want to blindly pass 'notepad', perhaps user will want to provide something different via externalEditor option.

---
In reply to: [108814244](https://github.com/Microsoft/RTVS/pull/3362#discussion_r108814244) [](ancestors = 108814244)

MikhailArkhipov: FYI CF comments do not appear on GH

---
In reply to: [108815251](https://github.com/Microsoft/RTVS/pull/3362#discussion_r108815251) [](ancestors = 108815251,108814244)

src/Host/Client/Impl/rtvs/R/overrides.R
MikhailArkhipov: OK.

---
In reply to: [108814768](https://github.com/Microsoft/RTVS/pull/3362#discussion_r108814768) [](ancestors = 108814768)

int19h:
👍

PR #3367, opened 2017-03-30
Updating RHost, RTVS was not pointing to latest RHost

karthiknadig:
Looks like RHost submodule was not updated in one of the recent checkins.


MikhailArkhipov:
👍

Issue #3371, opened 2017-03-31
Active startup project is calculated incorrectly


MikhailArkhipov:
1. Create solution
2. Add another R project
3. Set second one as active
4. F5

No matter what you do always first project runs. Looks like code always takes first project...

```
       public EnvDTE.Project GetActiveProject() {
            DTE dte = VsAppShell.Current.GetGlobalService<DTE>();
            if (dte.Solution.Projects.Count > 0) {
                try {
                    return dte.Solution?.Projects?.Cast<EnvDTE.Project>()?.First();
                } catch (COMException) { }
            }
            return null;
        }
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/88d29bfd5198ec58edd1a7c3503c08ddf2b75a24

PR #3372, opened 2017-03-31
#3371 Active startup project is calculated incorrectly 

MikhailArkhipov:
Properly figure out startup project

### DIFF
--- a/src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
+++ b/src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
@@ -27,13 +27,14 @@ namespace Microsoft.VisualStudio.R.Package.ProjectSystem {
 
         public EnvDTE.Project GetActiveProject() {
             DTE dte = VsAppShell.Current.GetGlobalService<DTE>();
-            if (dte.Solution.Projects.Count > 0) {
+            var projects = dte?.Solution?.Projects;
+            if (projects != null && projects.Count > 0) {
                 try {
-                    var projects = dte.Solution?.SolutionBuild?.StartupProjects as IEnumerable;
-                    var e = projects?.GetEnumerator();
+                    var startupProjects = dte.Solution.SolutionBuild?.StartupProjects as IEnumerable;
+                    var e = startupProjects?.GetEnumerator();
                     e?.MoveNext();
                     var projectName = e?.Current as string;
-                    return !string.IsNullOrEmpty(projectName) ? dte.Solution.Projects.Item(projectName) : null;
+                    return !string.IsNullOrEmpty(projectName) ? projects.Item(projectName) : null;
                 } catch (COMException) { }
             }
             return null;
--- a/src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
+++ b/src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
@@ -6,7 +6,6 @@ using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.IO;
-using System.IO.Compression;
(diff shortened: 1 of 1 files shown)


src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
```diff
@@ -28,7 +29,11 @@ internal sealed class ProjectSystemServices : IProjectSystemServices {
             DTE dte = VsAppShell.Current.GetGlobalService<DTE>();
             if (dte.Solution.Projects.Count > 0) {
                 try {
-                    return dte.Solution?.Projects?.Cast<EnvDTE.Project>()?.First();
+                    var projects = dte.Solution?.SolutionBuild?.StartupProjects as IEnumerable;
```
AlexanderSher: `var projectName = dte.Solution?.SolutionBuild?.StartupProjects?.FirstOrDefault() as string;`

AlexanderSher: `var projectName = dte.Solution?.SolutionBuild?.StartupProjects?.Cast<string>().FirstOrDefault();`

src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
```diff
@@ -28,7 +29,11 @@ internal sealed class ProjectSystemServices : IProjectSystemServices {
...
+                    var e = projects?.GetEnumerator();
+                    e?.MoveNext();
+                    var projectName = e?.Current as string;
+                    return !string.IsNullOrEmpty(projectName) ? dte.Solution.Projects.Item(projectName) : null;
```
AlexanderSher: In old code, both `Solution` and `Projects` properties had null checks

MikhailArkhipov: if it DID find project name, they are not null. But we can be paranoid :-)

src/Package/Impl/ProjectSystem/ProjectSystemServices.cs
MikhailArkhipov: StartupProjects is `object`.

MikhailArkhipov: Samples only allow `foreach(string s in startupProjects)`

int19h: If it's actually `IEnumerable` (which your change implies by casting it to one), then you can do `Cast` to get a typed enumerable out of it.

AlexanderSher: From https://msdn.microsoft.com/en-us/library/envdte.solutionbuild.startupprojects.aspx:
```cs
        SolutionBuild sb = dte.Solution.SolutionBuild;
        SolutionConfiguration sc = sb.ActiveConfiguration;
        vsBuildState vsBS;
        string msg = "Return relative path to startup projects: ";
        foreach (String s in (Array)sb.StartupProjects) {
            msg += "\n   " + s;
        }
```
So it is an array. In this case,
```cs
var projectName = (dte.Solution?.SolutionBuild?.StartupProjects as IEnumerable)?.Cast<string>().FirstOrDefault();
```

MikhailArkhipov: OK. As long as it works on native COM `IEnum***`

MikhailArkhipov: Speaking of casting to `Array` I wouldn't be so sure. Implementation dependent.

AlexanderSher:
👍

PR #3382, opened 2017-04-03
Added executable_exists_on_path with platform specific implementation.

### LINKED ISSUE #3376: R Markdown publish command fails on linux with pandoc not available.
This is after installing pandoc and pdflatex. The check in the command specifically looks for "pandoc.exe"  and "pdflatex.exe". On linux it should look for "pandoc" and "pdflatex"

karthiknadig:
Fixes #3376

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/util.R
+++ b/src/Host/Client/Impl/rtvs/R/util.R
@@ -350,13 +350,7 @@ exists_on_path_posix <- function(filename) {
     return(FALSE);
 }
 
-exists_on_path <- function(filename) {
-    if(.Platform$OS.type == "windows"){
-        return(exists_on_path_windows(filename));
-    } else {
-        return(exists_on_path_posix(filename));
-    }
-}
+exists_on_path <- if(.Platform$OS.type == "windows") exists_on_path_windows else exists_on_path_posix;
 
 executable_exists_on_path <- function(filename_no_extension) {
     filename <- filename_no_extension;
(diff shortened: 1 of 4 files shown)


src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -338,3 +338,30 @@ exists_on_path <- function(filename) {
...
+    for (f in folders) {
+        if (file.exists(f)) {
+            return(TRUE);
+        }
+    }
+    return(FALSE);
+}
+
+exists_on_path <- function(filename) {
+    if(.Platform$OS.type == "windows"){
```
int19h: Rather than doing the check inside the function, better do it outside, i.e.:
```r
exists_on_path <- if (.Platform$OS.type == "windows") {
  function(filename) {
    ...
``

File: src/Package/Impl/Publishing/Commands/PreviewPdfCommand.cs [hunk 2]
```diff
@@ -24,11 +24,10 @@ internal sealed class PreviewPdfCommand : PreviewCommand {
             if (!await base.CheckPrerequisitesAsync()) {
                 return false;
             }
-            if (!await CheckExistsOnPathAsync("pdflatex.exe")) {
+            if (!await CheckExecutableExistsOnPathAsync("pdflatex")) {
                 var session = _workflowProvider.GetOrCreate().RSession;
                 var message = session.IsRemote ? Resources.Error_PdfLatexMissingRemote : Resources.Error_PdfLatexMissingLocal;
                 await Services.ShowErrorMessageAsync(message);
-                Process.Start("https://miktex.org/2.9/setup");
```
karthiknadig: We should add platform specific instructions here. Session should have platform info in it.

int19h:
👍

Issue #3377, opened 2017-04-01
Add database connection doesn't handle AAD 


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/3899f4e70e0ce43eb4468d86de940aec2c0f3c4d

Issue #3354, opened 2017-03-27
Graphic window appears every time I start Visual Studio


lebelinoz:
I created a new R project in **Visual Studio with R Tools 1.0**.  Now, every time I open the solution from Visual Studio, a separate "Graphics" window appears.  How do I stop this from happening?

Projects created with **Visual Studio with R Tools 0.5** do not have this problem. _(erratum: yes it does)_

huguesv:
Hi Alain, could you please paste a screenshot that shows this window?

lebelinoz:
![image](https://cloud.githubusercontent.com/assets/10016385/24379722/0886c9bc-1394-11e7-88dc-44c90a8ce800.png)

huguesv:
I don't really know what that thing is. I suppose it reappears when you click Reset on the Interactive Window?  Could it be part of your R profile startup script?
http://www.onthelambda.com/2014/09/17/fun-with-rprofile-and-customizing-r-startup/

huguesv:
You would see it in R-Gui too if it's from your startup script.

lebelinoz:
No, there's no .Rprofile file.  None was created when I let R Tools 1.0 create a new R project for me, and I didn't add one.

MikhailArkhipov:
Could you post high res image? I can't tell hat are menu items there. Does the window close when you close VS or does it stay? If it stays, is there a process in task manager that can be associated with it? I.e. task that has same icon, for example?

lebelinoz:
The menu items are

- Deducer
- Data
- Analysis
- Plots

There is something in Task Manager:  it's **Microsoft R Host**.  Ending that task makes the window go away without any consequences (though, to be fair, so does clicking the **X** in the corner of the window.

The window disappears on its own the instant I start closing Visual Studio.  In fact, it (and its task in Task Manager) disappears _before_ I get the "Save changes?" dialog.

lebelinoz:
**Erratum:**  Actually, that Graphics window appears whenever I open an R project in Visual Studio, regardless of whether I created the project with R Tools 0.5 or 1.0

MikhailArkhipov:
R Host is RTVS process which loads R.dll. The window appears to be something that may be coming from one of R packages automatically loaded by R. http://www.deducer.org/pmwiki/pmwiki.php?n=Main.DeducerManual perhaps?

lebelinoz:
[...]

PR #3388, opened 2017-04-04
#3354 #3377

MikhailArkhipov:
#3354 Graphic window appears every time I start Visual Studio
- suppress `utils::win*` on remote and non-interactive

#3377 Add database connection doesn't handle AAD
- improve SQL/ODBC conversion
- detect AAD on Azure
- Use ODBC 13 driver for AAD/Azure
- prompt to install new ODBC driver if missing
- tests


src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -488,6 +496,19 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
 "));
         }
 
+        private static Task SuppressUI(IRExpressionEvaluator eval) {
+            // # Suppress Windows UI 
+            // http://astrostatistics.psu.edu/datasets/R/html/utils/html/winMenus.html
+            return eval.ExecuteAsync(@"
+                rtvs:::suppress_function('bringToTop', 'grDevices');
```
karthiknadig: Can we do this entirely in R? as in rtvs:::suppress_all(), this way we don'y need to eval the replacement function each time.

int19h: There's only one eval here, it just calls the function multiple times.

karthiknadig: i meant this eval in the R code:
 replacement <- eval(parse(text = "function(...) { stop('Not supported') }"));

src/Host/Client/Impl/rtvs/R/overrides.R
```diff
@@ -82,3 +82,25 @@ remove.packages <- function(...) {
     utils::remove.packages(...)
     invisible(rtvs:::send_notification('!PackagesRemoved'))
 }
+
+suppress_function <- function(function_name, package_name) {
+    package_spec <- paste("package:", package_name, sep='');
+    # The message may only become visible in remote session.
+    # TODO: provide localized replacement.
+    replacement <- eval(parse(text = "function(...) { stop('Not supported') }"));
```
int19h: I don't think you need eval here, and you definitely don't need parse. If eval is needed, you can use quote:
```r
replacement <- eval(quote(function(...) { stop('Not supported') }))
```
But I don't see the reason for eval, either. You should just be able to use the function literal directly.

int19h:
👍

PR #3392, opened 2017-04-05
VSTS 402127 - 'Name' is missing in accessible properties

MikhailArkhipov:
Give name matching label of the text and combo boxes.


int19h:
👍

AlexanderSher:
👍 , though I don't believe anyone who requires narrator would use those windows - it is simpler to just type the command in REPL

Issue #3400, opened 2017-04-07
Update package installed in Program Files crashed RTVS


huguesv:
Package `survival` had an update available.  That package was installed in program files.  I clicked update and VS crashed.
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.R.Host.Client.REvaluationException
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions.ThrowOnError(System.String, Microsoft.R.Host.Client.REvaluationResult)
   at Microsoft.R.Host.Client.RExpressionEvaluatorExtensions+<EvaluateAsync>d__1`1[[Microsoft.R.Components.PackageManager.Model.PackageLockState, Microsoft.R.Components, Version=1.0.30405.1526, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a]].MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Components.PackageManager.Implementation.ViewModel.RPackageManagerViewModel+<UpdateImplAsync>d__43.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.R.Components.PackageManager.Implementation.ViewModel.RPackageManagerViewModel+<UpdateAsync>d__41.MoveNext()
   at Microsoft.Common.Core.TaskExtensions.ReThrowTaskException(System.Object)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(System.Delegate, System.Object, Int32)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(System.Object, System.Delegate, System.Object, Int32, System.Delegate)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(System.Object)
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(System.Object)
[...]

Issue #3399, opened 2017-04-07
edit(file = '') is broken on VS 2017


MikhailArkhipov:
![image](https://cloud.githubusercontent.com/assets/12820357/24820659/02308144-1b9f-11e7-98cf-5f4f27550a96.png)

MikhailArkhipov:
Also, issue opening markdown

![image](https://cloud.githubusercontent.com/assets/12820357/24820711/45732d44-1b9f-11e7-830b-71dc6f7d5e1a.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/49c60b01937430798d1f6218947741d875be2189

Issue #3398, opened 2017-04-07
Get tooltip for nested function call


Hong-Revo:
In the screenshot below, my mouse pointer is on top of `cos`. Is there a way to make the tooltip for the inner function show up, instead of the outer one?

![image](https://cloud.githubusercontent.com/assets/11328890/24812024/d38dbee8-1c0b-11e7-833e-071c1b19d877.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/49c60b01937430798d1f6218947741d875be2189

PR #3402, opened 2017-04-09
#3398 #3399 #3400

MikhailArkhipov:
#3399 edit(file = '') is broken on VS 2017
- not specific to 2017. 
- Fixed incorrect calculation of ~/ 
- Made reliable opening of files since some 'traditional' methods fail on CPS
- If file is already opened, just activate

#3398 Get tooltip for nested function call 
 - Use AST rather than signature help utilities

#3400 Update package installed in Program Files crashed RTVS 
- Hide button as we do with uninstall. Packages in Program Files location cannot be uninstalled or updated
- Added exception handling just in case


MikhailArkhipov:
Wrong branch

PR #3403, opened 2017-04-09
#3398 #3399 #3400

MikhailArkhipov:
#3399 edit(file = '') is broken on VS 2017
- not specific to 2017.
- Fixed incorrect calculation of ~/
- Made reliable opening of files since some 'traditional' methods fail on CPS
- If file is already opened, just activate

#3398 Get tooltip for nested function call
- Use AST rather than signature help utilities

#3400 Update package installed in Program Files crashed RTVS
- Hide button as we do with uninstall. Packages in Program Files location cannot be uninstalled or updated
- Added exception handling just in case


src/Package/Impl/DataInspect/Viewers/FileEditor.cs
```diff
@@ -58,31 +58,32 @@ internal sealed class FileEditor : IFileEditor {
 
             if (!string.IsNullOrEmpty(fileName)) {
                 try {
-                    if (!Path.IsPathRooted(fileName)) {
+                    if (fileName.StartsWithOrdinal("~\\")) {
```
int19h: Shouldn't this (also) check for `~/`?

MikhailArkhipov: It is converted `fileName?.FromRPath();` above

src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -135,5 +135,15 @@ public static class IOExtensions {
             }
             return allowWrite && !denyWrite;
         }
+
+        public static string NormalizePath(this string path) {
```
AlexanderSher: How is it different from `System.IO.Path.GetFullPath`?

MikhailArkhipov: Does `GetFullPath` normalize? I thought it does nothing to rooted ie `c:\foo\.\..\` remains. But if it indeed does the right thing, we dont need `NormalizePath`.

AlexanderSher: At least `System.IO.Path.GetFullPath(@"c:\FusionLog/..\NVIDIA/DisplayDriver\")` produced `"c:\\NVIDIA\\DisplayDriver\\"`

src/R/Editor/Test/QuickInfo/QuickInfoSourceTest.cs
```diff
@@ -66,16 +65,26 @@ class Session {
                 .Which.ToString().Should().StartWith("as.Date.character(x, ...)");
         }
 
+        [CompositeTest]
+        [InlineData(2, "sin(")]
+        [InlineData(5, "cos(")]
+        [InlineData(8, "cos(")]
+        [InlineData(11, "sin(")]
+        public async Task NestedCallTest(int position, string name) {
```
AlexanderSher: Do you want to check for second nested parameter and for multiple nested?

src/R/Editor/Impl/QuickInfo/QuickInfoSource.cs
```diff
@@ -57,61 +59,82 @@ internal sealed class QuickInfoSource : IQuickInfoSource {
...
+                int wrapLength = Math.Min(SignatureInfo.MaxSignatureLength, signatureString.Length);
+                string text;
+
+                if (string.IsNullOrWhiteSpace(functionInfo.Description)) {
+                    text = string.Empty;
+                } else {
+                    // VS may end showing very long tooltip so we need to keep 
+                    // description reasonably short: typically about
+                    // same length as the function signature.
+                    text = signatureString + "\r\n" + functionInfo.Description.Wrap(wrapLength);
```
AlexanderSher: `FormattableString`?

MikhailArkhipov: Yeah, I will look into cleaning this up in the editor split. There is bunch of old code here.

src/R/Editor/Test/QuickInfo/QuickInfoSourceTest.cs
MikhailArkhipov: Sure we can do that

src/Common/Core/Impl/Extensions/IOExtensions.cs
MikhailArkhipov: OK.

AlexanderSher:
👍 
@int19h , can you please take a look one more time?

int19h:
👍

Issue #3401, opened 2017-04-07
Replace static REditorSettings by service


MikhailArkhipov:

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/db7ef914b2a07fdb032a9f4fb7df3629f5114bf6

Issue #3288, opened 2017-03-10
Split Microsoft.R.Editor into portable and Windows-specific parts (Part I)


MikhailArkhipov:

MikhailArkhipov:
OK, split is done, closing. Needs PR and testing, will open new task for the next sprint.

PR #3405, opened 2017-04-10
#3401 Replace static REditorSettings by service

MikhailArkhipov:
Prequel to #3288

- Eliminated REditorSettings static
- Moved some common useful things like Idle to Core
- Provided fixture that provides `ICoreShell` rather than `IExportProvider`
- Updated tests to use `ICoreShell.GetService` rather than MEF
- Provide portable `IImageService`
- Added portable Languages.Editor project
- Moved some editor things to portable


src/Common/Core/Impl/Idle/IdleTimeAction.cs
```diff
@@ -4,10 +4,8 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.Common.Core.Shell;
-using Microsoft.Languages.Core.Utility;
-using Microsoft.Languages.Editor.Shell;
 
-namespace Microsoft.Languages.Editor.Tasks {
+namespace Microsoft.Common.Core.Idle {
```
AlexanderSher: I'm not fond of moving this class to the `Common`. It is not just some idle, it is a very specific event that happens only in application with UI thread when there are no scheduled tasks and there is no user interaction for some period of time. It doesn't have any meaning in Broker or in the out-of-proc parts of the application.

Why do you need it in Common?

src/Common/Core/Impl/Imaging/IImageService.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <returns>Platform-specific image source</returns>
+        object GetImage(ImageType imageType);
+
+        /// <summary>
+        /// Returns image source given name of the image moniker
+        /// such as name from http://glyphlist.azurewebsites.net/knownmonikers
+        /// when running in Visual Studio.
+        /// </summary>
+        /// <returns>Platform-specific image source</returns>
+        object GetImage(string name);
```
AlexanderSher: What's the difference between this method and the previous one? Those names are constant and we also can make them part of the enum

MikhailArkhipov: `GetImage(ImageType imageType)` returngs predefined icons. Like 'function', 'variable' etc such as from editor `IGlyphService`. So it is consistent and the same. `GetImage(string name)` can be used to return custom `foo.png` or VS moniker based image.

src/Common/Core/Impl/Idle/IdleTimeAction.cs
MikhailArkhipov: I don't technically need it in common but it is not tied to the editor either. Common is not just for UI-less apps anymore, it contains plenty of UI definitions and concept of the main thread and locale.

MikhailArkhipov: It may actually come useful in the broker for updating host load which is currently on timer. the idle does not rely necessarily on main thread activity, it is tied to 'idle service' which can be implemented as desired - such as timer.

src/Common/Core/Test/Fakes/Shell/TestImageService.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.IO;
+using System.Windows.Media;
+using System.Windows.Media.Imaging;
+using Microsoft.Common.Core.Imaging;
+
+namespace Microsoft.Common.Core.Test.Fakes.Shell {
+    internal sealed class TestImageService : IImageService {
+        private readonly ImageSource _image;
```
AlexanderSher: Something like 
```cs
_image = new DrawingImage(new GeometryDrawing(
    Brushes.Blue,
    new Pen(Brushes.Transparent, 0),
    new RectangleGeometry(new Rect(0, 0, 16, 16))
));
```
should be enough

src/Common/Core/Impl/Imaging/IImageService.cs
AlexanderSher: `foo.png` case is represented by the next method, when we get the file path. But VS moniker is just another predefined image name, code that uses the service shouldn't care where that image comes from.

src/Common/Core/Impl/Imaging/IImageService.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Core.Imaging {
+    /// <summary>
+    /// Provides access to images and icons
+    /// </summary>
+    public interface IImageService {
```
AlexanderSher: This service looks like 100% platform-specific, you can't use it without type cast. Do you need it anywhere in xplat code?

src/Common/Core/Impl/Idle/IdleTimeAction.cs
MikhailArkhipov: Come to think of this, we may do another pass on core at some point and split it into real core and IDE stuff. For example, task list support is currently in the editor but task list is typically a global service. Anything can push items there: project, extensions, configuration management...

src/Common/Core/Impl/Imaging/IImageService.cs
MikhailArkhipov: Yes, anywhere icons are needed. Which is anywhere in editor for example. BTW, same happens to Components. In plots there won't be 'BitmapSource' anymore, just object.

src/Common/Core/Test/Fakes/Shell/TestImageService.cs
MikhailArkhipov: OK. As long as we are happy with blue square ;-)

src/Common/Core/Impl/Imaging/IImageService.cs
MikhailArkhipov: There is preliminary split of Components in my Components branch. Waiting for @karthiknadig to complete Host.Client split since there are Shiny etc supporting code that uses objects not available in .NET Core

src/Common/Core/Impl/Imaging/IImageService.cs
MikhailArkhipov: Problem is - we can prefedine names as const strings, but they can end up clashing with moniker names.

src/Common/Core/Impl/Idle/IdleTimeAction.cs
MikhailArkhipov: I'll move idle etc to Components next round.

PR #3413, opened 2017-04-14
#3401 Replace static REditorSettings by service

MikhailArkhipov:
Prequel to #3288

Eliminated REditorSettings static
Moved some common useful things like Idle to Core
Provided fixture that provides ICoreShell rather than IExportProvider
Updated tests to use ICoreShell.GetService rather than MEF
Provide portable IImageService
Added portable Languages.Editor project
Moved some editor things to portable


src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
```diff
@@ -44,6 +44,10 @@
       <Project>{01E3D8C2-9C24-492B-B13B-C6D7068B6BF8}</Project>
       <Name>Microsoft.Rtvs.Core.Windows</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\Windows\R\Interpreters\Impl\Microsoft.R.Interpreters.Windows.csproj">
+      <Project>{DE1EC50A-C588-45A9-B22C-981419CD7CD4}</Project>
+      <Name>Microsoft.R.Interpreters.Windows</Name>
```
AlexanderSher: That is the change I'd like to avoid. It is better when test assemblies follow the reference hierarchy of the product assemblies. Why do you need it?

MikhailArkhipov: Actually it is merge. Not sure it is needed.

MikhailArkhipov: Actually, TestCoreShell needs RInstallation service for tests. Generally speaking any assembly that preps test core shell will need a bunch.

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Common.Core.Services;
+using Microsoft.Common.Core.Shell;
+using Microsoft.Common.Core.Test.Fakes.Shell;
+using Microsoft.UnitTests.Core.Shell;
+using Microsoft.UnitTests.Core.XUnit;
+using Microsoft.UnitTests.Core.XUnit.MethodFixtures;
+using Xunit.Sdk;
+
+namespace Microsoft.Common.Core.Test.Fixtures {
+    public abstract class CoreShellProviderFixture : MethodFixtureBase, ICoreShellProvider {
```
AlexanderSher: This is the change that I don't understand. `IExportProvider` isn't bound to the MEF, it is pretty much the same `IServiceProvider` that supports some of the MEF specifics.

Since you don't need `T GetExportedValue<T>(string metadataKey, params object[] metadataValues) where T : class;` and `IEnumerable<Lazy<T, TMetadataView>> GetExports<T, TMetadataView>() where T : class;` anymore, you can rename it into `ITestServices` with the only method `T Get<T>()`, without any other changes. Many of the tests (72 out of 106 files) that use `IExportProvider` do not require `ICoreShell`

MikhailArkhipov: Tests are unified to use core shell (as the rest of the code). So it is simpler to have shell fixture rather than one method or providing services there and another here. Get core shell, get stuff from it. 

Same for the rest of the code (and in the upcoming editor split). I mostly pass core shell around, makes it a lot simpler that passing multiple separate parameters or service container. Makes it easier to make tactical fixes in the end phase too. Adding service and consuming it makes 2 line change, plumbing new argument through 20 nested calls is much harder. I already had to hack fix for Win32Exception where there is no core shell - was not possible to show user message/prompt.

AlexanderSher: `ICoreShell` isn't a service provider, `IServiceContainer` is. On some levels, `ICoreShell` is actually undesired. For example, `RSession` shouldn't have access not only to the `ICoreShell`, but to the `IUIService` as well. And so are the tests.

It it good to have `IServiceContainer` as an entry point all around, not only because it allows us to access any required service, but also because it allow scoping and doesn't bound existing code to a set of services that is defined by `IServiceContainer` implementation. And `ICoreShell` is actually just one of the services. It is an entry point for the application when it starts, but not for the rest of the code.

src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
AlexanderSher: We will have many more of those services when we get to the `RComponents`. We probably need to specify some common entry points in each assembly that provides services, so it can be accessed by both test fixtures and RPackage.

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
MikhailArkhipov: Well it is. Actually, it was reduced pretty much to services, app name and locale id. doesn't matter much what to pass anymore.

src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
MikhailArkhipov: I guess we could add AddServices(IServiceContainer) to the service manager so code can provide additional sets when needed

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
MikhailArkhipov: I think we can move remaining fields to some `IApplicationService` and then only raw 'IServiceContainer` will remain and then we won't need `ICoreShell` at all. This will take lots of renames though so perhaps next time around.

AlexanderSher: That's what I've been thinking about as well.
I just wanted to minimize the amount of changes, cause honestly, I can't fit the whole 400+ file changes into my mind.

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
```diff
@@ -0,0 +1,42 @@
...
+        private CompositionContainer _compositionContainer;
+        public ICoreShell CoreShell { get; private set; }
+
+        public override Task<Task<RunSummary>> InitializeAsync(ITestInput testInput, IMessageBus messageBus) {
+            _compositionContainer = CreateCompositionContainer();
```
AlexanderSher: That can drop test speed significantly. Test fixtures can reference one another, so you can get `IExportProvider` from constructor and just add it to the `TestCoreShell` without adding whole new hierarchy of fixtures.

src/R/Editor/Impl/Completions/Providers/PackagesCompletionProvider.cs
```diff
@@ -21,9 +20,9 @@ public class PackagesCompletionProvider : IRCompletionListProvider, IRHelpSearch
         private readonly IPackageIndex _packageIndex;
 
         [ImportingConstructor]
-        public PackagesCompletionProvider(IPackageIndex packageIndex, IGlyphService glyphService) {
+        public PackagesCompletionProvider(IPackageIndex packageIndex, IImageService imageService) {
             _packageIndex = packageIndex;
-            _glyph = glyphService.GetGlyphThreadSafe(StandardGlyphGroup.GlyphLibrary, StandardGlyphItem.GlyphItemPublic);
+            _glyph = imageService.GetImage(ImageType.Library) as ImageSource;
```
AlexanderSher: Will this code remain in xplat?

src/Common/Core/Impl/Imaging/IImageService.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <returns>Platform-specific image source</returns>
+        object GetImage(ImageType imageType);
+
+        /// <summary>
+        /// Returns image source given name of the image moniker
+        /// such as name from http://glyphlist.azurewebsites.net/knownmonikers
+        /// when running in Visual Studio.
+        /// </summary>
+        /// <returns>Platform-specific image source</returns>
+        object GetImage(string name);
```
AlexanderSher: These comments have been left in the previous PR, but I still don't understand why we need to differentiate `object GetImage(string name)` and `object GetImage(ImageType imageType)`. Even in `ImageService` implementation, there is this part:
```cs
                case ImageType.File:
                case ImageType.Document:
                    return GetImage("Document");
```

To me it looks like some VS-specific differentiation that matters only to the implementation, but not to the contract. If there is a collision of names, we can always resolve this explicitly, but I don't think there should be two different results for `GetImage(ImageType.Keyword)` and `GetImage("Keyword")` - this is quite confusing

src/Common/Core/Test/Fakes/Shell/TestCoreShell.cs
```diff
@@ -56,6 +52,23 @@ public static ICoreShell CreateFromCompositionContainer(CompositionContainer com
             return shell;
         }
 
+        /// <summary>
+        /// Creates test core shell with basic services and delegation
+        /// to the supplied export provider for additional services.
+        /// </summary>
+        /// <param name="exportProvider"></param>
+        public TestCoreShell(IExportProvider exportProvider) : this(new TestServiceManager(exportProvider)) {
```
AlexanderSher: Do you use it anywhere?

src/Common/Core/Test/Fakes/Shell/TestImageService.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Windows.Media.Imaging;
+using Microsoft.Common.Core.Imaging;
+
+namespace Microsoft.Common.Core.Test.Fakes.Shell {
+    public sealed class TestImageService : IImageService {
+        private readonly ImageSource _image;
+
+        public TestImageService() {
+            using (var memory = new MemoryStream()) {
+                Resources.BlueSquare.Save(memory, System.Drawing.Imaging.ImageFormat.Png);
```
AlexanderSher: Didn't we decided to use geometry here instead of a resource?

src/Common/Core/Test/Fakes/Shell/TestServiceManager.cs
```diff
@@ -2,48 +2,34 @@
...
 using System.ComponentModel.Composition.Hosting;
-using System.Linq;
 using Microsoft.Common.Core.Services;
+using Microsoft.UnitTests.Core.Mef;
 
 namespace Microsoft.Common.Core.Test.Fakes.Shell {
-    public class TestServiceManager : IServiceManager {
-        private readonly ExportProvider _compositionContainer;
-        private readonly IServiceManager _services;
+    public class TestServiceManager: ServiceManager {
```
AlexanderSher: I haven't derived `TestServiceManager` from `ServiceManager` so that it can wrap another existing `IServiceManager` without transferring all service instances. This is useful in fixtures when one provides services for the other

src/EditorTestApp/Packages/SettingsStorage.cs
```diff
@@ -4,77 +4,27 @@
...
-
-        public virtual bool GetBoolean(string name, bool defaultValue = true) {
-            if (_settings.ContainsKey(name))
-                return (bool)_settings[name];
-
-            return defaultValue;
+        #region IEditorSettingsStorage Members
+        public T Get<T>(string name, T defaultValue = default(T)) {
+            object value;
+            return _settings.TryGetValue(name, out value) ? (T)value : defaultValue;
```
AlexanderSher:
```cs
public T Get<T>(string name, T defaultValue = default(T)) 
    => _settings.TryGetValue(name, out object value) ? (T)value : defaultValue;
```

Welcome to C# 7 :)

src/EditorTestApp/Packages/SettingsStorage.cs
```diff
@@ -4,77 +4,27 @@
...
-        }
-
-        public void ResetSettings() {
-        }
-
-        public void LoadFromStorage() {
-        }
+        #region IWritableEditorSettingsStorage Members
+        public void Set<T>(string name, T value)=> _settings[name] = value;
+        public void ResetSettings() { }
```
AlexanderSher: Shouldn't this erase all the settings from `_settings`?

src/Languages/Editor/Impl/Microsoft.Languages.Editor.csproj
```diff
@@ -1,173 +1,29 @@
...
-    <None Include="project.json" />
+    <ProjectReference Include="..\..\..\Common\Core\Impl\Microsoft.Common.Core.csproj" />
+    <ProjectReference Include="..\..\Core\Impl\Microsoft.Languages.Core.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <FilesToSign Include="$(OutputPath)\$(AssemblyName).dll">
-      <Authenticode>Microsoft</Authenticode>
-      <StrongName>StrongName</StrongName>
-    </FilesToSign>
+    <Folder Include="Text\" />
```
AlexanderSher: Do you need this folder?

src/Languages/Editor/Impl/Definitions/IEditorBuffer.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using Microsoft.Common.Core.Services;
+using Microsoft.Languages.Core.Text;
+
+namespace Microsoft.Languages.Editor {
+    /// <summary>
+    /// Represents editor code (text) buffer
+    /// </summary>
+    public interface IEditorBuffer {
+        IServiceManager Services { get; }
```
AlexanderSher: Those are the services that are attached to the `IEditorBuffer`, correct?

src/Languages/Editor/Impl/Definitions/IEditorBufferSnapshot.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.Languages.Core.Text;
+
+namespace Microsoft.Languages.Editor {
+    /// <summary>
+    /// Represents immutable snapshot of the editor buffer
+    /// </summary>
+    public interface IBufferSnapshot : ITextProvider {
+        T ToTextSnapshot<T>() where T : class;
```
AlexanderSher: Aren't there any base type for TextSnapshot?

src/Languages/Editor/Impl/Definitions/IEditorView.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// Retrieves caret position for the given buffer
+        /// </summary>
+        ISnapshotPoint GetCaretPosition(IEditorBuffer buffer = null);
+
+        /// <summary>
+        /// Current selection in the view
+        /// </summary>
+        IEditorSelection Selection { get; }
+
+        T As<T>() where T : class;
```
AlexanderSher: Any constraints on `T`?

src/Languages/Editor/Impl/Settings/EditorSettings.cs
```diff
@@ -0,0 +1,76 @@
...
+            WritableStorage = storage as IWritableEditorSettingsStorage;
+            Storage.SettingsChanged += (s, e) => SettingsChanged?.Invoke(this, EventArgs.Empty);
+        }
+
+        public event EventHandler<EventArgs> SettingsChanged;
+        public virtual void ResetSettings() => WritableStorage?.ResetSettings();
+
+        public bool AutoFormat {
+            get { return Storage.Get(AutoFormatKey, true); }
+            set { WritableStorage?.Set(AutoFormatKey, value); }
```
AlexanderSher: Shouldn't we throw an exception if `WritableStorage` is null?

src/Languages/Editor/Impl/Settings/EditorSettings.cs
```diff
@@ -0,0 +1,76 @@
...
+        public const string InsertMatchingBracesKey = "InsertMatchingBraces";
+        public const string SignatureHelpEnabledKey = "SignatureHelpEnabled";
+
+        protected IEditorSettingsStorage Storage { get; }
+        protected IWritableEditorSettingsStorage WritableStorage { get; }
+
+        public EditorSettings(IEditorSettingsStorage storage) {
+            Storage = storage;
+            WritableStorage = storage as IWritableEditorSettingsStorage;
+            Storage.SettingsChanged += (s, e) => SettingsChanged?.Invoke(this, EventArgs.Empty);
```
AlexanderSher: Event handler chain is usually an anti-pattern. If SettingsChanged is just a proxy, it would be better to change it to `add/remove`:
```cs
        public event EventHandler<EventArgs> SettingsChanged {
            add => Storage.SettingsChanged += value;
            remove => Storage.SettingsChanged -= value;
        }
```

src/Languages/Editor/Impl/Settings/EditorSettings.cs
```diff
@@ -0,0 +1,76 @@
...
+            set { WritableStorage?.Set(CompletionEnabledKey, value); }
+        }
+
+        public int IndentSize {
+            get { return Storage.Get(FormatterIndentSizeKey, 4); }
+            set { WritableStorage?.Set(FormatterIndentSizeKey, value); }
+        }
+
+        public IndentType IndentType {
+            get { return (IndentType)Storage.Get(FormatterIndentTypeKey, (int)IndentType.Spaces); }
```
AlexanderSher: Why do you convert it to `int`?

src/Languages/Editor/Impl/Settings/EditorSettings.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Microsoft.Languages.Editor.Settings {
+    public abstract class EditorSettings : IWritableEditorSettings, IEditorSettings {
+        public const string AutoFormatKey = "AutoFormat";
+        public const string CompletionEnabledKey = "CompletionEnabled";
+        public const string FormatterIndentSizeKey = "FormatterIndentSize";
+        public const string FormatterTabSizeKey = "FormatterTabSize";
+        public const string FormatterIndentTypeKey = "FormatterIndentType";
+        public const string IndentStyleKey = "IndentStyle";
+        public const string SyntaxCheckKey = "SyntaxCheckEnabled";
+        public const string InsertMatchingBracesKey = "InsertMatchingBraces";
```
AlexanderSher: Are these keys required anywhere outside the storage itself?

src/Package/Impl/Editors/LanguageSettingsStorage.cs
```diff
@@ -18,108 +17,94 @@ namespace Microsoft.VisualStudio.R.Package.Editors {
     /// application-agnostic editor code. This class also provides tracking of tab/indent settings via VS
     /// text manager (<seealso cref="IVsTextManager"/>) and language preferences (<seealso cref="LANGPREFERENCES"/>).
     /// </summary>
-    public abstract class LanguageSettingsStorage
-        : IVsTextManagerEvents4
-        , IWritableEditorSettingsStorage
-        , IDisposable {
+    public sealed class LanguageSettingsStorage : IVsTextManagerEvents4, IWritableEditorSettingsStorage, IDisposable {
+        private readonly Guid _languageServiceId;
+        private readonly Dictionary<string, bool> _booleanSettings = new Dictionary<string, bool>();
```
AlexanderSher: Why do you split `LanguageSettingsStorage` storage into several dictionaries?

src/Package/Impl/TaskList/VsTaskItem.cs
```diff
@@ -53,8 +54,8 @@ internal sealed class VsTaskItem : ErrorTask {
                     break;
             }
 
-            IVsHierarchy hierarchy = _source.TextBuffer.GetHierarchy();
-            base.HierarchyItem = hierarchy;
+            var textBuffer = _source.EditorBuffer as ITextBuffer;
```
AlexanderSher: `(ITextBuffer)_source.EditorBuffer`

src/Package/Impl/TaskList/VsTaskItem.cs
```diff
@@ -88,8 +89,8 @@ internal sealed class VsTaskItem : ErrorTask {
         }
 
         protected override void OnNavigate(EventArgs e) {
-            if (_source.TextBuffer != null && _item.Line > 0 && _item.Column > 0) {
-                var textView = TextViewConnectionListener.GetFirstViewForBuffer(_source.TextBuffer);
+            if (_source.EditorBuffer != null && _item.Line > 0 && _item.Column > 0) {
+                var textView = TextViewConnectionListener.GetFirstViewForBuffer(_source.EditorBuffer as ITextBuffer);
```
AlexanderSher: `GetFirstViewForBuffer` doesn't allow nulls, so `(ITextBuffer)_source.EditorBuffer`

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionStatusBaseViewModel.cs
```diff
@@ -30,19 +30,13 @@ internal abstract class ConnectionStatusBaseViewModel : BindableBase, IDisposabl
...
         }
 
-        public virtual void Dispose(bool disposing) {
-            _disposableBag.TryDispose();
-        }
-
-        public void Dispose() {
-            Dispose(true);
-        }
+        public virtual void Dispose(bool disposing) => _disposableBag.TryDispose();
```
AlexanderSher: Class has no destructor, so `public void Dispose() => _disposableBag.TryDispose();` is enough

src/EditorTestApp/Packages/SettingsStorage.cs
MikhailArkhipov: This is test only, not used

src/Common/Core/Test/Fakes/Shell/TestImageService.cs
MikhailArkhipov: Wait, I fixed this... Let me check branches. Could have happened elsewhere.

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
```diff
@@ -373,35 +373,28 @@ public class IntellisenseTest : FunctionIndexBasedTest {
...
                 script.DoIdle(300);
 
-                UIThreadHelper.Instance.Invoke(() => {
-                    var session = script.GetCompletionSession();
-                    session.Should().NotBeNull();
-
-                    var infoSourceProvider = ExportProvider.GetExportedValue<ISnippetInformationSourceProvider>();
-                    var infoSource = infoSourceProvider.InformationSource;
-                    var completion = session.SelectedCompletionSet.SelectionStatus.Completion;
+                var session = script.GetCompletionSession();
```
AlexanderSher: I thought `script.DoIdle(300);` doesn't work if invoked from UI thread

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
MikhailArkhipov: Agreed. However, if we turn it into services now, there will be more :-) and I'll have to merge into 2 more branches. I think when we done, we'll just do global search/replace and fix it in one sweep.

AlexanderSher: Ok

src/R/Editor/Application.Test/Fixtures/EditorHostMethodFixture.cs
```diff
@@ -2,50 +2,47 @@
...
 using Microsoft.R.Host.Client;
 using Microsoft.R.Host.Client.Test.Script;
 using Microsoft.R.Support.Help;
-using Microsoft.UnitTests.Core.Mef;
+using Microsoft.UnitTests.Core.XUnit;
 using Microsoft.UnitTests.Core.XUnit.MethodFixtures;
 using static Microsoft.UnitTests.Core.Threading.UIThreadTools;
 
 namespace Microsoft.R.Editor.Application.Test {
+    [AssemblyFixture]
```
AlexanderSher: `EditorHostMethodFixture` isn't an assembly-level fixture - it is method-level fixture, this attribute should be removed

src/Languages/Editor/Impl/Settings/EditorSettings.cs
MikhailArkhipov: VS storage is old COM and only understands basic types. I can alternatively convert in VS implementation at the package level but there may be more code.

src/Languages/Editor/Impl/Settings/EditorSettings.cs
MikhailArkhipov: Apparently not anymore. I'll make them private.

MikhailArkhipov: Actually, package uses them when sending data to `IVsTextManager4.LANGPREFERENCES3`

AlexanderSher: Ok

src/Languages/Editor/Impl/Settings/EditorSettings.cs
AlexanderSher: But why do you need it for `Get` method? I believe it only reads, so VS storage types shouldn't matter.

src/R/Editor/Impl/Settings/REditorSettings.cs
```diff
@@ -22,125 +16,73 @@ public static class REditorSettings {
...
-        public static RFormatOptions FormatOptions {
+        public RFormatOptions FormatOptions {
             get {
-                _formatOptions.IndentSize = REditorSettings.IndentSize;
-                _formatOptions.IndentType = REditorSettings.IndentType;
-                _formatOptions.TabSize = REditorSettings.TabSize;
-
+                _formatOptions.IndentSize = IndentSize;
+                _formatOptions.IndentType = IndentType;
+                _formatOptions.TabSize = TabSize;
```
AlexanderSher: Is it intentional that you return one and the same instance of `RFormatOptions`?

src/R/Editor/Test/Fixtures/REditorShellProviderFixture.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.ComponentModel.Composition.Hosting;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.R.Support.Test;
+using Microsoft.UnitTests.Core.XUnit;
+
+namespace Microsoft.R.Editor.Test {
+    [AssemblyFixture]
```
AlexanderSher: It is already marked in `AssemblyInfo` - this attribute isn't needed.

src/R/Editor/Test/Signatures/ComputeCurrentParameterTest.cs
```diff
@@ -117,15 +121,15 @@ public class ComputeCurrentParameterTest : FunctionIndexBasedTest {
         public async Task ParameterTest_ComputeCurrentParameter03() {
             await PackageIndexUtility.GetFunctionInfoAsync(FunctionIndex, "legend");
 
-            REditorSettings.PartialArgumentNameMatch = false;
+            //REditorSettings.PartialArgumentNameMatch = false;
```
AlexanderSher: Why?

src/R/Support/Test/RSupportShellProviderFixture.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.Common.Core.Extensions;
+using Microsoft.Languages.Editor.Test;
+using Microsoft.R.Components.Settings;
+using Microsoft.R.Support.Test.Utility;
+using Microsoft.UnitTests.Core.XUnit;
+using Xunit.Sdk;
+
+namespace Microsoft.R.Support.Test {
+    // Fixture doesn't import itself. Use AssemblyFixtureImportAttribute
+    [AssemblyFixture]
```
AlexanderSher: [AssemblyFixture] attribute isn't needed

src/UnitTests/Core/Impl/Mef/IExportProvider.cs
```diff
@@ -3,10 +3,9 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.UnitTests.Core.XUnit;
 
 namespace Microsoft.UnitTests.Core.Mef {
-    public interface IExportProvider : IMethodFixture {
+    public interface IExportProvider {
         T GetExportedValue<T>() where T : class;
```
AlexanderSher: You can leave only `T GetExportedValue<T>() where T : class;` - all other methods can be removed

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
```diff
@@ -56,12 +56,18 @@ public class UIThreadHelper : IMainThread {
...
-        private UIThreadHelper() {}
+        private UIThreadHelper() { }
 
         public Thread Thread => _thread;
         public SynchronizationContext SyncContext => _syncContext;
         public ControlledTaskScheduler TaskScheduler => _taskScheduler;
 
+        #region IMainThread
+        int IMainThread.ThreadId => Thread.ManagedThreadId;
+        void IMainThread.Post(Action action, CancellationToken cancellationToken)
```
AlexanderSher: This change will turn many tests into false positive.

src/UnitTests/Core/Impl/XUnit/TestMefCatalog.cs
```diff
@@ -38,18 +32,7 @@ public abstract class MefCatalogFixture : IMethodFixtureFactory<IExportProvider>
             }
         }
 
-        private class NullExportProvider : IExportProvider {
-            public Task<Task<RunSummary>> InitializeAsync(ITestInput testInput, IMessageBus messageBus)
```
AlexanderSher: `InitializeAsync` is part of `IMethodFixture` interface. How does it compile?

src/R/Editor/Test/Signatures/ComputeCurrentParameterTest.cs
MikhailArkhipov: Will remove comment. false is default and now that it is not static no need to set it.

src/R/Editor/Impl/Settings/REditorSettings.cs
MikhailArkhipov: There should be one instance of format options per REditorSettings

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
MikhailArkhipov: Hm, `Thread => _thread` so it should the one that is considered main, no?

src/UnitTests/Core/Impl/XUnit/TestMefCatalog.cs
MikhailArkhipov: Whole class removed, it does compile... and tests pass apart from known few

src/UnitTests/Core/Impl/Threading/UIThreadHelper.cs
AlexanderSher: No, that isn't what broken. The old code used `_blockingLoop`:
```cs
void IMainThread.Post(Action action, CancellationToken cancellationToken) {
    var bl = _blockingLoop.Value;
    if (bl != null) {
        bl.Post(action);
    } else {
        InvokeAsync(action, cancellationToken).DoNotWait();
    }
}
```

`_blockingLoop` is used to simulate CPS behavior when it blocks UI thread.

src/UnitTests/Core/Impl/XUnit/TestMefCatalog.cs
AlexanderSher: Ok

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
MikhailArkhipov: About same speed as before

src/R/Editor/Impl/Completions/Providers/PackagesCompletionProvider.cs
MikhailArkhipov: Yes, but it is another change

src/Common/Core/Impl/Imaging/IImageService.cs
MikhailArkhipov: Actually this is the point. One place may want `"Keyword"` from monikers and another icon from intellisense glyph service. How would we resolve which is which? These map to different services but use same names.

src/Languages/Editor/Impl/Definitions/IEditorBuffer.cs
MikhailArkhipov: Yes

src/Languages/Editor/Impl/Definitions/IEditorView.cs
MikhailArkhipov: Not really. `ITextView` but it is not a x-plat type

src/Languages/Editor/Impl/Definitions/IEditorBufferSnapshot.cs
MikhailArkhipov: `ITextSnapshot`, but it is not x-plat.

src/Languages/Editor/Impl/Settings/EditorSettings.cs
MikhailArkhipov: Nope. Allowed in test apps or other containers that don't have means to change options.

src/Languages/Editor/Impl/Microsoft.Languages.Editor.csproj
MikhailArkhipov: Removed

src/EditorTestApp/Packages/SettingsStorage.cs
MikhailArkhipov: More useless syntax sugar

src/Common/Core/Test/Fakes/Shell/TestServiceManager.cs
MikhailArkhipov: Maybe create another class - reimplementing whole thing is work

src/Package/Impl/Editors/LanguageSettingsStorage.cs
MikhailArkhipov: Its old code, don't remember.

src/Common/Core/Test/Fakes/Shell/TestServiceManager.cs
AlexanderSher: Let's leave it as it is, I can always revive it from history if needed.

src/Common/Core/Impl/Imaging/IImageService.cs
AlexanderSher: Ok, I still don't understand it, and I have a feeling that I wouldn't, so let's leave it as it is.

src/R/Editor/Impl/Completions/Providers/PackagesCompletionProvider.cs
AlexanderSher: ok

src/Common/Core/Test/Fixtures/CoreShellProviderFixture.cs
AlexanderSher: Ok then.

src/Package/Impl/Editors/LanguageSettingsStorage.cs
AlexanderSher: Ah, ok

src/R/Editor/Impl/Settings/REditorSettings.cs
AlexanderSher: ok

src/Package/Impl/TaskList/VsTaskItem.cs
MikhailArkhipov: This will turn into 'IEditorBuffer` soon rather than the object so doesn't matter much at the moment. However, it should do null checks since task list source is not necessarily an editor document.

src/Package/Impl/TaskList/VsTaskItem.cs
MikhailArkhipov: As above, I will make it into `IEditorBuffer` soon

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionStatusBaseViewModel.cs
MikhailArkhipov: FxCop doesn't like non-virtual Dispose in non-sealed classes

src/R/Editor/Application.Test/Completion/IntellisenseTest.cs
MikhailArkhipov: Technically any test that uses some StartScript or editor host should be UI only.

src/Package/Impl/Editors/LanguageSettingsStorage.cs
MikhailArkhipov: Probably to avoid boxing/unboxing and extra casting

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionStatusBaseViewModel.cs
AlexanderSher: I know, and it is wrong. This rule assumes that derived class may have both managed and unmanaged resources, which is:
- Wrong by design, and .net framework itself follows the rule that only single unmanaged resource should be wrapped into a managed for at least 10 years (look at any `*SafeHandle` type)
- Won't happen in our application, because we don't acquire unmanaged handles in ViewModel

Can we just disable this rule?

src/Package/Impl/Editors/LanguageSettingsStorage.cs
MikhailArkhipov: I'll look at it in the editor pass

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionStatusBaseViewModel.cs
MikhailArkhipov: It is a DevDiv rule

AlexanderSher: image

Ok then.

src/Languages/Editor/Impl/Definitions/IEditorBufferSnapshot.cs
AlexanderSher: Ok

src/Languages/Editor/Impl/Definitions/IEditorView.cs
AlexanderSher: Ok

src/Languages/Editor/Impl/Definitions/IEditorBuffer.cs
AlexanderSher: Ok

AlexanderSher:
👍 
I think we should merge the PR. I have some changes that I want to be on top of the `ICoreShell` refactoring.

src/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionStatusBaseViewModel.cs
MikhailArkhipov: Actually, you can look into DevDiv.ruleset we have checked in, maybe it is gone?

MikhailArkhipov:
We should make couple more passes over code when all split is done. Clean up old stuff too. Will be easier when everything is in place and fully works.

PR #3418, opened 2017-04-17
Finalize Broker split

AlexanderSher:
- Move xplat initialization code into Broker
- Remove static code from CommonStartup
- Remove IExitService that has been creating duplicated messages in the log
- Use `IWebHost.Start` instead of `IWebHost.Run` in service
- Use `IApplicationLifetime` cancellation tokens instead of custom ones
- Use configuration after it has been read from config file
- Use `urls` startup parameter instead of legacy `server.urls`
- Replace `start.as` parameter with `startup:isService`


src/Host/Broker/Impl/Extensions/ExceptionExtensions.cs
```diff
@@ -2,15 +2,47 @@
...
+using System.Runtime.ExceptionServices;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.R.Host.Protocol;
 
-namespace Microsoft.R.Host.Broker {
+namespace Microsoft.R.Host.Broker.Extensions {
```
MikhailArkhipov: I prefer to avoid 'Extensions' in namespace so all methods naturally appear in the completion list as long as main definition is available.

src/Host/Broker/Impl/Resources.resx
```diff
@@ -117,18 +117,39 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="Critical_InvalidPipeHandle" xml:space="preserve">
+    <value>Requested to write urls to pipe '{0}', but it is not a valid pipe handle</value>
```
MikhailArkhipov: URLs

src/Host/Broker/Impl/Resources.resx
```diff
@@ -117,18 +117,39 @@
...
+    <value>TLS certificate not found: {0}</value>
+  </data>
   <data name="Critical_ParentProcessNotFound" xml:space="preserve">
     <value>Designated parent process {0} not found</value>
   </data>
   <data name="Critical_PingTimeOut" xml:space="preserve">
     <value>Ping timed out, terminating</value>
   </data>
+  <data name="Critical_PipeConnectTimeOut" xml:space="preserve">
+    <value>Requested to write urls to pipe '{0}', but timed out while trying to connect to pipe</value>
```
MikhailArkhipov: As above

src/Host/Broker/Impl/Resources.resx
```diff
@@ -117,18 +117,39 @@
...
+    <value>Config file load failed: {0}</value>
+  </data>
+  <data name="Error_ConfigParseFailed" xml:space="preserve">
+    <value>Config file parse failed: {0}</value>
+  </data>
+  <data name="Error_ConfiguredPortNotAvailable" xml:space="preserve">
+    <value>Server port {0} is already in use.</value>
+  </data>
   <data name="Error_ExitRHost" xml:space="preserve">
     <value>R Host Exited with Win32 error : {0}</value>
```
MikhailArkhipov: Win32 specific?

src/Host/Broker/Impl/Resources.resx
```diff
@@ -180,4 +201,10 @@
   <data name="Trace_ErrorDataReceived" xml:space="preserve">
     <value>|{0}|: {1}</value>
   </data>
+  <data name="Trace_ServerUrlsToPipeBegin" xml:space="preserve">
+    <value>Writing urls to pipe '{0}':{1}{2}</value>
```
MikhailArkhipov: URLs

src/Host/Broker/Impl/Extensions/ExceptionExtensions.cs
AlexanderSher: Ok, fixed

src/Host/Broker/Impl/Resources.resx
AlexanderSher: It is actually `urls` parameter that is passed. Need to put it in quotes.

src/Host/Broker/Impl/Resources.resx
AlexanderSher: No, I believe it applies to unix as well. Fixed the message.

src/Host/Broker/Impl/Resources.resx
MikhailArkhipov: Ah OK

MikhailArkhipov:
👍

AlexanderSher:
@karthiknadig , can you also please take a look?

src/Windows/Host/Broker/Impl/Microsoft.R.Host.Broker.Config.json
```diff
@@ -1,3 +1,3 @@
 ﻿{
-  "server.urls": "https://0.0.0.0:5444"
+  "urls": "http://0.0.0.0:5444"
```
karthiknadig: Why is this http? the default for remote scenarios should be https.

karthiknadig:
One minor change.
👍

Issue #3424, opened 2017-04-18
VS hang on exit


huguesv:
Using this build, on VS 2017: `RTVS\Build\Release\1_0\2017-04-13.2`

I used interactive window, package manager and debugger.  When I close VS, it blocks there, using full CPU.

Main thread:

```
>	WindowsBase.dll!System.Windows.Threading.DispatcherSynchronizationContext.Wait(System.IntPtr[] waitHandles, bool waitAll, int millisecondsTimeout)	Unknown
 	mscorlib.dll!System.Threading.SynchronizationContext.InvokeWaitMethodHelper(System.Threading.SynchronizationContext syncContext, System.IntPtr[] waitHandles, bool waitAll, int millisecondsTimeout)	Unknown
 	[Native to Managed Transition]	
 	[Managed to Native Transition]	
 	mscorlib.dll!System.Threading.Monitor.Wait(object obj, int millisecondsTimeout, bool exitContext)	Unknown
 	mscorlib.dll!System.Threading.Monitor.Wait(object obj, int millisecondsTimeout)	Unknown
 	mscorlib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.SpinThenBlockingWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.InternalWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.WaitAny(System.Threading.Tasks.Task[] tasks, int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	Microsoft.VisualStudio.Shell.15.0.dll!Microsoft.VisualStudio.Shell.UIThreadReentrancyScope.WaitOnTaskCompleteInternal(System.Threading.Tasks.Task task, System.Threading.CancellationToken cancel, int ms)	Unknown
[...]

PR #3429, opened 2017-04-19
#3424 VS hang on exit

MikhailArkhipov:


src/R/Support/Test/RD/Help/GetFunctionInfoTest.cs
```diff
@@ -70,6 +70,18 @@ public class GetFunctionInfoTest {
 
         [Test]
         [Category.R.Signatures]
+        public void GetRdFunctionArgumentsTest03() {
```
AlexanderSher: Do you want to add any tests for parser itself?

MikhailArkhipov: It is basically covered by quick info and signature tests.

AlexanderSher:
👍

src/R/Support/Test/RD/Help/GetFunctionInfoTest.cs
MikhailArkhipov: But i guess i could add some more granular going forward

PR #3430, opened 2017-04-19
Split Components

MikhailArkhipov:
Here is proposed split. Not 100% sure about interface names, make suggestions.


src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -23,16 +22,14 @@
 using Microsoft.VisualStudio.R.Package.Repl;
 
 namespace Microsoft.R.Components.InteractiveWorkflow.Implementation {
-    public sealed class RInteractiveWorkflow : IRInteractiveWorkflow {
+    public sealed class RInteractiveWorkflow : IRInteractiveWorkflowVisual {
```
AlexanderSher: I don't think we should rename this interface at all, cause there is nothing "visual" in this class, even in its implementation. It references `IInteractiveWindowVisualComponent` in these 3 methods:
```cs
        IInteractiveWindowVisualComponent ActiveWindow { get; }
        event EventHandler<ActiveWindowChangedEventArgs> ActiveWindowChanged;
        Task<IInteractiveWindowVisualComponent> GetOrCreateVisualComponentAsync(int instanceId = 0);
```
but it can be decoupled later if needed.

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
```diff
@@ -105,7 +105,7 @@ public Task<PlotDeviceProperties> PlotDeviceCreate(Guid deviceId, CancellationTo
         public async Task ViewLibraryAsync(CancellationToken cancellationToken = default(CancellationToken)) {
             await _coreShell.SwitchToMainThreadAsync(cancellationToken);
             var containerFactory = _coreShell.GetService<IRPackageManagerVisualComponentContainerFactory>();
-            _workflow.Packages.GetOrCreateVisualComponent(containerFactory).Container.Show(focus: true, immediate: false);
+            (_workflow.Packages as IRPackageManagerVisual).GetOrCreateVisualComponent(containerFactory).Container.Show(focus: true, immediate: false);
```
AlexanderSher: I'd like to avoid such downcasting if possible. Since we now have IServiceContainer in place, we can always resolve it as `_services.GetService<IRPackageManagerVisualComponentContainerFactory>().Get(_workflow.Packages)`.
What I would rename though is this factory interface. `IRPackageManagerVisualComponent` to `IRPackageManagerVisual` and `IRPackageManagerVisualComponentContainerFactory` to `IRPackageManagerVisualFactory`.

AlexanderSher: And later, when we add more features to `IServiceContainer` we can even get something like `_services.GetServiceFromFactory<IPackageManager, IRPackageManagerVisual>(_workflow.Packages)`

AlexanderSher: Important part here is that `IRPackageManagerVisualProvider` may return null, so there is no `GetOrCreate`, we don't guarantee that `IRPackageManagerVisual` is implemented.

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
MikhailArkhipov: Sure. The problem is in `IInteractiveWindowVisualComponent` which references `IWpfTextView`, `ITextBuffer` and `IInteractiveWindow`. I could technically make first two into new `IEditorView` and `IEditorBuffer` but generally UI-related stuff is not x-plat anyway. So question is how to decouple. In CM, PM, Workflow and plots interfaces there x-plat parts that manage infrastructure and UI parts.

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
MikhailArkhipov: OK

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
MikhailArkhipov: In VS there is `IVsShell` and `IVsUIShell`. We could make separate `IRInteractiveWorkflow` and `IRInteractiveUIWorkflow` rather than derive

AlexanderSher: But we don't have to define `IInteractiveWindowVisualComponent` in xplat. `ActiveWindow` is used only in `RHistory` (which we won't convert into xplat right now) and in one place in RPlotManager, where it can be replaced with `IUIService.UpdateCommandStatus`. The rest are either Interactive Window bits (they can't be xplat), or code in Package project.

MikhailArkhipov: Right. But currently `IRInteractiveWorkflow` has `Task<IInteractiveWindowVisualComponent> GetOrCreateVisualComponentAsync`.  Where do we move it.

MikhailArkhipov: CM and PM have similar issues so it would be great to come up with consistent plan that addresses those as well.

AlexanderSher: `GetOrCreateVisualComponentAsync` goes to the `IInteractiveWindowVisualProvider`. So it becomes:
```cs
public interface IInteractiveWindowVisualProvider {
    IInteractiveWindowVisual GetAsync(IRInteractiveWorkflow workflow)
    IInteractiveWindowVisual Current { get; }
    event EventHandler<ActiveWindowChangedEventArgs> CurrentChanged;
}
```

In CM and PM, there will be only `Get` method.

AlexanderSher: Also, it will help us to decouple interactive window functionality from `IRInteractiveWorkflow`

MikhailArkhipov: OK will do

MikhailArkhipov: I spent a day on this and they will be huge. It will take another couple of days + test fixes + review. It entails removal of `GetOrCreateVisualComponent` or equivalent from Workflow, PM, CM, History and Plots. Moving visual-handling code to a separate class and exporting/adding as service. Changing all commands and related code that assume they can get active window etc from workflow, active device from plot manager and so on. Adding test classes (or splitting existing ones), updating tests.

MikhailArkhipov: `Components2` branch is partial change

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RSessionCallback.cs
MikhailArkhipov: So this is simpler split, unblocks further work. Theses areas are stable and there is no immediate work planned. There is only a week left and we should not be spilling refactoring into May. Casting is not that of a big deal, it basically same concept as `QueryInterface` for higher order interface when getting base.

MikhailArkhipov:
So this is simpler split, unblocks further work. Theses areas are stable and there is no immediate work planned. There is only a week left and we should not be spilling refactoring into May. Casting is not that of a big deal, it basically same concept as QueryInterface for higher order interface when getting base.

AlexanderSher:
Ok, let's leave it as it is for now, I'll take a look into it when we have time.

AlexanderSher:
👍

AlexanderSher:
It is broker exe file, it isn't referenced by Package.

MikhailArkhipov:
But is it needed to *run* locally. It's not reference, it is deployment to extension folder. .NET core stuff is in the project file

AlexanderSher:
Ok

Issue #3426, opened 2017-04-19
VS2015+RTVS on Win 10 not setting working directory of a project


virusme:
I have VS2015(Enterprise) + RTVS installation on a Win7 and a Win10 machine.

I have an r-project with a main RMD file that needs to be converted to HTML (using "preview HTML"). RMD file has "source()"commands to source multiple R files.  When i "preview HTML" on WIn7, it works fine. But on Win10, it gives "cannot open connection" error.  After lots of head scratching, it boils down to - on Win10 somehow opening a RPROJECT does not set the working directory to the project directory when processing RMD file!  It always reverts to "ÄppData/Local/Temp/xxxxxx" as working directory!

To re-create this error:
1. create a /YOUR_LOCATION/RPROJ with a "test.RMD" inside the project, on Win 7 machine
2. inside "test.RMD" include "getwd()" (to display working directory)
3. run "preview HTML" 
4. generated HTML file will show "YOUR_LOCATION"

1. Open /YOUR_LOCATION/RPROJ on Win10 machine
2. run "preview HTML"
4. generated HTML file will show "ÄppData/Local/Temp/xxxxx"

Happy to elaborate further, if required.

Note: I have tried contacting Microsoft Visual Studio Enterprise support, got kicked around like a football   for just creating the ticket (which by the way has not been created yet!) and they had not even started with the actual problem. Hopeless!

smortaz:
Sorry about that! For Python and R you should use github to file bugs and
you will be directly taken care of by the development team :-)

virusme:
I am under the impression that with bulk licensing+support agreement, we will get premium support irrespective of the product.

smortaz:
It would eventually route to us. That's why in the product's R menu we
direct you straight to github to avoid getting passed around. I will check
into why it did not route to us quickly. Thanks for the heads up

MikhailArkhipov:
Also #3404

virusme:
Thanks @smortaz and @MikhailArkhipov for including this in your next sprint.

MikhailArkhipov:
It is actually current one :-) We hope to fix it for when 1.1 goes out with 2015 ans 2017. Which is very soon.

virusme:
@MikhailArkhipov. Thank you. Do you have a date for v2017?

MikhailArkhipov:
Data Science workload should be in 2017 early May. About same time we'll update 2015 to 1.1

MikhailArkhipov:
[...]

Issue #3404, opened 2017-04-10
rmarkdown and relative paths?


tja2000:
I reference files using a relative path, e.g. read.csv2("../data/file.csv"). It works interactively but when using the code in r markdown I always get Error in file(file, "rt") : cannot open the connection.
When I execute using ctrl+enter it does work fine, the working directory is correctly set. Is there a specific setting for the rmarkdown startup directory?

Hong-Revo:
I guess the error occurs when you try to Preview HTML/PDF/Word? If so, that's because RTVS runs the knitting from a temporary directory. It should work if you knit the document manually with `rmarkdown::render()`.

tja2000:
Yes, this works. Is there a way to default knitr to the current workdirectory?
Thanks!

MikhailArkhipov:
You can try and change working directory explicitly before reading file with 'setwd'

PR #3438, opened 2017-04-21
#3426 VS2015+RTVS on Win 10 not setting working directory of a project 

MikhailArkhipov:
The problem: we use TEMP path in local and remote. This breaks workflow when markdown has R code that sources files or reads CSV - basically any operation that involves paths.

Fix: provide work folder to the publishing code. 

1. Standalone file: use TEMP as before.
2. Project: work folder path is relative to the current working directory. Handle remote mapping as well.

```
source('./script.R') # x <- c(1:10)
```

```
plot(x)
```

work correctly in local and remote as long as file is in project.

### DIFF
--- a/src/Common/Core/Impl/Extensions/IOExtensions.cs
+++ b/src/Common/Core/Impl/Extensions/IOExtensions.cs
@@ -7,6 +7,7 @@ using System.Globalization;
 using System.IO;
 using System.Security.AccessControl;
 using System.Security.Principal;
+using Microsoft.Common.Core.Diagnostics;
 using Microsoft.Common.Core.IO;
 
 namespace Microsoft.Common.Core {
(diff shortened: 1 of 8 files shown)


src/Common/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -22,6 +23,20 @@ public static class IOExtensions {
             return path.StartsWithIgnoreCase(bp) ? path.Substring(bp.Length) : path;
         }
 
+        // TODO: merge with above
+        public static string MakeCompleteRelativePath(this string path, string basePath) {
```
MikhailArkhipov: Not too happy with the name, but don't want to change behaviour of the existing `MakeRelativePath`

src/Host/Client/Impl/rtvs/R/util.R
```diff
@@ -204,11 +204,17 @@ export_to_pdf <- function(device_id, plot_id, width, height) {
...
-    writeBin(get_blob(blob_id), rmdpath);
+    if(identical(work_folder, '')) {
+        work_folder <- tempdir();
+    } else {
+        work_folder <- normalizePath(work_folder);
+    }
+
+    input_path <- tempfile('rmd_', tmpdir = work_folder, fileext = '.rmd');
+    writeBin(get_blob(blob_id), input_path);
+    on.exit(file.remove(input_path), add = TRUE);
```
MikhailArkhipov: This also fixes problem that one of the temp files was not deleted. `on.exit` without `add = TRUE` overwrites previously set actions.

AlexanderSher:
So what was the problem with TEMP on remote - we don't have permissions for it?

MikhailArkhipov:
`source('./script.R')` or `read.csv("..\..\foo\myfile.csv") failed because relative path was no longer valid since file was actually in TEMP and not where user expected.

MikhailArkhipov:
See also #3404 and #3426

MikhailArkhipov:
Ping. Need to kick signed build and publish along with #3441

karthiknadig:
👍

PR #3439, opened 2017-04-22
Intermediate cleanup

AlexanderSher:
- Added `ServiceExtensions` to `Host.Client` and `R.Interpreters`. They are used in both test fixtures and VsAppShell
- Simplified test fixtures:
    - `IServiceContainer` can be accessed directly from test constructor. No specific fixture type is required, `IServiceManager` can be used instead
    - Service imports are declared in fixtures derived from `ServiceManagerFixture`, all other service-related fixtures are removed
- Fixed most of the tests


MikhailArkhipov:
👍

PR #3445, opened 2017-04-24
Accessibility & localization fixes

### LINKED ISSUE #3416: Sorting indicator in Grid View is wrong
```r
View(mtcars)
```
then click on `mpg` column:
image

Note that the arrow indicates that sorting is descending, but it's actually sorting ascending. If you click on the column again, the arrow will show ascending sort, but actual sorting is descending. Basically, it's always the opposite of what it should be.

In addition, I would expect the first click to sort the rows ascending, rather than descending (as is the usual behavior for sortable lists and tables). It might be that this is the actual root cause here - that the UI flag is simply in the incorrect initial state.

### LINKED ISSUE #3417: Allow mouse wheel scroll in any area of View() window
mouse wheel scroll only  works when the mouse is positioned over the horizontal or vertical scroll bars. Ideally, this scrolling would work if the mouse is positioned over any area of the grid window produced by a call to `View()`. It appears that the scrolling also works if the mouse is over a grid line, but not if the mouse is over the data area of a cell in the grid

Also, it would be nice to be able to select/highlight a row or multiple rows in the grid

### LINKED ISSUE #3434: Parts of project settings not localized
Tested with French language.

First project settings page is missing localized text (second page is all ok).

Startup file
Remote Project Path
Transfer files on run
Files to transfer

image

AlexanderSher:
- Fix #3434: Parts of project settings not localized
- Fix #3417: Allow mouse wheel scroll in any area of View() window
- Fix #3416: Sorting indicator in Grid View is wrong
- Fix expand/collapse element accessibility properties

### DIFF
--- a/src/Common/Wpf/Impl/Controls/ExpandCollapseMode.cs
+++ b/src/Common/Wpf/Impl/Controls/ExpandCollapseMode.cs
@@ -1,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Wpf.Controls {
+    public enum ExpandCollapseMode {
+        Click,
+        LeftRigthArrows
+    }
+}
--- a/src/Common/Wpf/Impl/Controls/ExpandCollapseMode.cs
+++ b/src/Common/Wpf/Impl/Controls/ExpandCollapseMode.cs
@@ -4,6 +4,6 @@
 namespace Microsoft.Common.Wpf.Controls {
     public enum ExpandCollapseMode {
         Click,
-        LeftRigthArrows
+        LeftRightArrows
     }
 }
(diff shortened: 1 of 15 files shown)


AlexanderSher:
@zooba , you may want to grab the `ExpandCollapseButton` for PTVS

src/Common/Wpf/Impl/Controls/ExpandCollapseMode.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Common.Wpf.Controls {
+    public enum ExpandCollapseMode {
+        Click,
+        LeftRigthArrows
```
zooba: Left*Right*Arrows

zooba:
@AlexanderSher This is the one for groups within a list view? I don't think we do that anywhere, and as far as I know the Expander control works fine as a standalone thing.

src/Common/Wpf/Impl/Controls/ExpandCollapseMode.cs
AlexanderSher: Thanks!

MikhailArkhipov:
👍

PR #3450, opened 2017-04-26
Fix Variable Explorer accessibility

AlexanderSher:
- Variable tree view is accessible by tab navigation, but inside it can be navigated only by arrows (left/right arrow expands/collapses the record). This is a common behavior for many tree views across windows and VS. Tab navigation switches between environments combobox and list of variables
- Buttons on variable records aren't accessible by keyboard anymore. Making them a tab stop creates an additional mess and makes it really hard to get out of the list. Buttons are available on the toolbar, which is accessible by Shift+Alt. Also, "Show details" can be invoked by hitting "Enter"
- Variable explorer focus is preserved when user switches to the toolbar or to the environments combobox

### DIFF
--- a/src/Package/Impl/DataInspect/VariableView.xaml
+++ b/src/Package/Impl/DataInspect/VariableView.xaml
@@ -20,7 +20,7 @@
     <UserControl.Resources>
         <ResourceDictionary>
             <ResourceDictionary.MergedDictionaries>
-                <ResourceDictionary Source="pack://application:,,,/Microsoft.VisualStudio.R.Package;component/Wpf/Controls.xaml" />
+                <ResourceDictionary Source="pack://application:,,,/Microsoft.R.Wpf;component/CommonResources.xaml " />
                 <ResourceDictionary Source="DataGridStyle.xaml" />
             </ResourceDictionary.MergedDictionaries>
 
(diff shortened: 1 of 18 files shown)


AlexanderSher:
Fixed issues:
- 402111: Accessibility: MAS42A : High Contrast: Environment selection from combo box is not in compliance with HC themes
- 402112: Accessibility: MAS42A : High Contrast: Select Environment combo box when focused using keyboard doesn't get a highlight
- 402114: Accessibility : MAS05A : Keyboard : Focus is not visible on the cells present in the 'Variable Explorer' window
- 402116: Accessibility : MAS36 : Keyboard : Options in the 'Selected Environment' combo box in Variable Explorer window is not invoked using Space/Enter key
- 402117: Accessibility : MAS05 : Keyboard : Focus order is incorrect in Variable Explorer window
    *There is no tab navigation inside list of variables anymore, tab enters in and exits out without changing selection*
- 402124: Accessibility: MAS40B: Inspect: Name property is "null" for "search" button and "csv file" button for variable records in variable explorer pane
    *Those buttons aren't accessible by keyboard anymore*

MikhailArkhipov:
👍

AlexanderSher:
@huguesv , can you please also take a look?

src/Package/Impl/DataInspect/VariableView.xaml
```diff
@@ -154,16 +154,14 @@
         </ResourceDictionary>
     </UserControl.Resources>
 
-    <Grid>
+    <Grid KeyboardNavigation.TabNavigation="Cycle">
```
huguesv: Is this to prevent tabbing to the toolbar? (just curious)

huguesv:
👍

src/Package/Impl/DataInspect/VariableView.xaml
AlexanderSher: Yes. And Shift+Alt is used to switch between toolbar and the main window part

Issue #3446, opened 2017-04-24
Remove internal browser and options 


MikhailArkhipov:
Per incoming deprecation in VS 2017

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/dcc1db3f4d94a5e172e55fff3cf7b16e8b2f6c37

PR #3471, opened 2017-05-01
#3446 Remove internal browser and options

MikhailArkhipov:
Keeping browser service for now to a) minimize changes and b) in case we get internal Edge or Webkit at some point.


src/Package/Test/Browsers/WebBrowserServicesTest.cs
```diff
@@ -28,49 +25,5 @@ public class WebBrowserServicesTest {
...
-            var vswbs = Substitute.For<IVsWebBrowsingService>();
-            var ps = Substitute.For<IProcessServices>();
-
-            var wbs = new WebBrowserServices(vswbs, ps, externalSettings);
-            wbs.OpenBrowser(role, _url);
-            ps.Received().Start(_url);
-
-            ps.ClearReceivedCalls();
-            wbs = new WebBrowserServices(vswbs, ps, internalSettings);
-            wbs.OpenBrowser(role, _url);
```
AlexanderSher: I thought this test creates an external browser

MikhailArkhipov: Since there is only one now, there is not much to test. However, I can keep this just in case

AlexanderSher:
👍

PR #3473, opened 2017-05-01
R environment variables 

karthiknadig:
R environment variables  were not passed on to the R session.

### DIFF
--- a/src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
+++ b/src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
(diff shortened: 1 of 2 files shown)


src/Common/Core/Impl/OS/Win32EnvironmentBlock.cs
```diff
@@ -16,6 +17,12 @@ public class Win32EnvironmentBlock {
 
         private ConcurrentDictionary<string, string> _environment;
 
+        public IEnumerable<string> EnvironmentVariableNames {
```
int19h: It seems like it would be better to have the class itself enumerable, and delegate to `_environment.GetEnumerator()`. Then the loop below doesn't have to do lookup for every variable to get the value for its name, it can just iterate over key-value pairs.

MikhailArkhipov:
👍

Issue #3474, opened 2017-05-01
ESC key in Interactive Window cancels session


jflam:
We think that we'll need to handle waiting for input differently and prevent ESC from binding to Visual Studio Cancel command which is its usual behavior.

Repro:

Microsoft R Open 3.3.3
The enhanced R distribution from Microsoft
Microsoft packages Copyright (C) 2017 Microsoft

Loading Microsoft R Client packages, version 3.3.3.0059. 
Microsoft R Client limits some functions to available memory.
See: https://msdn.microsoft.com/en-us/microsoft-r-client-windows for information
about additional features.

Type 'readme()' for release notes, privacy() for privacy policy, or
'RevoLicense()' for licensing information.

Using the Intel MKL for parallel mathematical computing(using 2 cores).
Default CRAN mirror snapshot taken on 2017-03-15.
See: https://mran.microsoft.com/.

Deleting autosaved workspace image '~/.Autosave.RData'.
> library(swirl)

| Hi! Type swirl() when you are ready to begin.
> swirl()

| Welcome to swirl! Please sign in. If you've been here before, use the same name as you did
| then. If you are new, call yourself something unique.
What shall I call you? 
Tim

| Would you like to continue with one of these lessons?

1: R Programming Sequences of Numbers
2: No. Let me start something new.

Selection: Hit <Esc>

| Leaving swirl now. Type swirl() to resume.

Hong-Revo:
I'm not sure if this is the same underlying bug, but for me, pressing Esc at an interactive prompt leaves the REPL in a wedged state:
```
m <- lm(mpg ~ ., mtcars)
termplot(m)
# press Esc
```
Resetting the workspace is the only way to recover. This is RTVS build 1.1.30423.1752.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/b4dc0762f6f834a72930255df499d7197ba264f3

Issue #3467, opened 2017-05-01
RTVS locks package dlls, interferes with package (re)installation


Hong-Revo:
I think there was an existing bug report on this, but I can't find it. RTVS seems to lock a bunch of dlls, which stops R from overwriting them as part of package reinstallation. I just ran into this when trying to install RSQLServer from Github. The process broke on installing the rlang dependency:

```
Installing rlang
"C:/PROGRA~1/MICROS~3/RSERVE~1/R_SERVER//bin/x64/R" --no-site-file --no-environ --no-save --no-restore --quiet CMD INSTALL  \
  "C:/Users/hong/AppData/Local/Temp/RtmpgDMvve/devtools118c13d86b3b/hadley-rlang-0756eaf" --library="c:/Rlib" --install-tests 

* installing *source* package 'rlang' ...
** libs
c:/Rtools/mingw_64/bin/gcc -m64 -I"C:/PROGRA~1/MICROS~3/RSERVE~1/R_SERVER/include" -DNDEBUG     -I"C:/swarm/workspace/External-R-3.3.3/vendor/extsoft/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c attrs.c -o attrs.o
c:/Rtools/mingw_64/bin/gcc -m64 -I"C:/PROGRA~1/MICROS~3/RSERVE~1/R_SERVER/include" -DNDEBUG     -I"C:/swarm/workspace/External-R-3.3.3/vendor/extsoft/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c capture.c -o capture.o
c:/Rtools/mingw_64/bin/gcc -m64 -I"C:/PROGRA~1/MICROS~3/RSERVE~1/R_SERVER/include" -DNDEBUG     -I"C:/swarm/workspace/External-R-3.3.3/vendor/extsoft/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c eval.c -o eval.o
c:/Rtools/mingw_64/bin/gcc -m64 -I"C:/PROGRA~1/MICROS~3/RSERVE~1/R_SERVER/include" -DNDEBUG     -I"C:/swarm/workspace/External-R-3.3.3/vendor/extsoft/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c internals.c -o internals.o
[...]

PR #3478, opened 2017-05-02
#3474 #3467

MikhailArkhipov:
#3467 RTVS locks package dlls, interferes with package (re)installation 
- track installation requests better. Throttle index update if multiple commands are processed in quick succession

#3474 ESC key in Interactive Window cancels session 
- track readline() and similar. If session is waiting for the user input, do not handle ESC as cancel/interrupt. This also alleviates issues that interrupting R while readline() is active never returns to the prompt and REPL has to be reset.


int19h:
> This also alleviates issues that interrupting R while readline() is active never returns to the prompt and REPL has to be reset.

Can you file a separate bug for this? Interrupt R should work when waiting on `ReadConsole`.

MikhailArkhipov:
Sure. It interrupts, but interactiveWindow.ReadStandardInput is not cancelleable. However, it does return on, say, Enter.

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -551,7 +554,12 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
 
             var callback = _callback;
             if (!addToHistory && callback != null) {
-                return await callback.ReadUserInput(prompt, len, ct);
+                try {
+                    IsReadingUserInput = true;
+                    return await callback.ReadUserInput(prompt, len, ct);
```
AlexanderSher: @int19h , is it possible to have one input inside another?

src/R/Components/Impl/InteractiveWorkflow/Commands/InterruptRCommand.cs
```diff
@@ -11,47 +11,33 @@ public sealed class InterruptRCommand : IAsyncCommand {
...
-            if (_enabled) {
-                _enabled = false;
+            if (CanInterrupt()) {
                 _interactiveWorkflow.Operations.ClearPendingInputs();
                 await _session.CancelAllAsync();
             }
         }
+
+        private bool CanInterrupt()
+            => _session.IsHostRunning && _session.IsProcessing && !_session.IsReadingUserInput && !_debuggerModeTracker.IsInBreakMode;
```
int19h: We should allow to interrupt during readline etc (and fix it if it doesn't working). I think it's only Esc as the shortcut that's problematic in that case... can we handle that specifically? Or maybe just rebind it entirely actually - for something lacking confirmation, it's too easy to e.g. terminate a long-running computation accidentally and have to restart.

src/Host/Client/Impl/Session/RSession.cs
```diff
@@ -560,6 +568,7 @@ public Task StartHostAsync(RHostStartupInfo startupInfo, IRSessionCallback callb
             Prompt = GetDefaultPrompt(prompt);
             MaxLength = len;
 
+            IsProcessing = contexts.Count != 1;
```
AlexanderSher: I'm not sure that it is enough. If there are any inputs in the queue, `IsProcessing` may be switched back to `true` after index has started update. Also, is it possible that any of our evaluation commands also blocks dll's?

MikhailArkhipov: This is the check that InterruptR command has been performing. I am open to suggestions. Process can start and may continue as long as user input does not require more package installs. I.e. x <- 1 is perfectly safe. If it does, the session will be canceled before package installation begins.

Not sure about evals, I guess it may be possible to do something like `eval("install.packages").

src/Host/Client/Impl/Session/RSession.cs
int19h: Yes, if you do a reentrable eval while on `ReadConsole`, and that eval involves some code that asks for input (e.g. another `readline`).

src/R/Components/Impl/InteractiveWorkflow/Commands/InterruptRCommand.cs
MikhailArkhipov: Yes, we can fix this, but it is not strictly necessary IMO in the release branch. We can fix #3479 in master.

src/Host/Client/Impl/Session/RSession.cs
MikhailArkhipov: the callback ends in `interactiveWindows.ReadStandardInput()` which is not cancellable AFAICT.

MikhailArkhipov: Introduce nesting counter perhaps?

AlexanderSher: You can use `CountdownDisposable`

MikhailArkhipov: OK

src/R/Components/Impl/InteractiveWorkflow/Commands/InterruptRCommand.cs
int19h: Thing is, it might actually be an easier fix. It's something that's supposed to work, and there's code in the host to handle it; I'm pretty sure it worked before, too. So if it's a one-line fix in the host that would eliminate the problem entirely (and only leave the inconvenience of Esc), wouldn't that be preferable for release?

int19h: Wait, I missed the other comment. So the real problem isn't that the host can't handle the interrupt, it's that the Interactive Window doesn't support it when reading input? Can we send it dummy input (just plain Enter) to get past the prompt it's stuck on?

MikhailArkhipov: Yes, of course smaller fix is preferable. There are two problems. First, IW does not return from `ReadStandardInput` unless user types something. Second, if user interrupts and then types something, `ReadStandardInput` returns but REPL never gets prompt back. Second part is the biggest deal IMO.

src/R/Support/Impl/Help/Packages/PackageIndex.cs
```diff
@@ -70,14 +70,22 @@ public sealed class PackageIndex : IPackageIndex {
...
-        public Task BeforePackagesInstalledAsync(CancellationToken ct) {
+        public async Task BeforePackagesInstalledAsync(CancellationToken ct) {
+            // Package is about to be installed. Stop intellisense session
+            // so loaded packages are released and new one will not be locked.
+            // If update is pending, cancel it.
+            CancelPendingIndexUpdate();
             var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
             timeoutCts.CancelAfter(5000);
-            return _host.StopSessionAsync(timeoutCts.Token);
+            await _host.StopSessionAsync(timeoutCts.Token);
```
AlexanderSher: Since there is no code running after `StopSessionAsync`, there is no difference in behavior with or without `await`.

src/R/Components/Impl/InteractiveWorkflow/Commands/InterruptRCommand.cs
int19h: Can we do `SubmitInput` (with `\n') to terminate the prompt *before* sending cancellation?

src/R/Support/Impl/Help/Packages/PackageIndex.cs
MikhailArkhipov: Yes, remainder of deleted code. Will restore.

AlexanderSher:
```cs
        public async Task<string> ReadUserInput(string prompt, int maximumLength, CancellationToken ct) {
            _coreShell.DispatchOnUIThread(() => _interactiveWindow.Write(prompt));
            var tcs = new TaskCompletionSource<string>();

            Task.Run(() => {
                using (var reader = _interactiveWindow.ReadStandardInput()) {
                    tcs.TrySetResult(reader?.ReadToEnd() ?? "\n");
                }
            }).DoNotWait();

            tcs.RegisterForCancellation(ct).UnregisterOnCompletion(tcs.Task);

            try {
                return await tcs.Task;
                _coreShell.DispatchOnUIThread(() => _interactiveWindow.Operations.TrySubmitStandardInput());
                throw;
            }
        }
```

int19h:
👍

Issue #3455, opened 2017-04-27
Intellisense for Parameters not working?


vykhand:
Hi! I am trying to use intellisence to autocomplete function parameters and it seems it is not working.
For example, I am typing:

`rxImport(inData = infile, outFile = "airExample.xdf", strin`

and it does not propose stringsAsFactors as a parameter. 

Screenshot is attached.
![2017-04-27 23_43_10-rproject1 - microsoft visual studio](https://cloud.githubusercontent.com/assets/498481/25505770/a0657202-2ba3-11e7-989a-a7fb5198430b.png)

MikhailArkhipov:
Looks like it is not working when close brace is missing. If it is auto-inserted, it does. We'll see if this can be improved.

![image](https://cloud.githubusercontent.com/assets/12820357/25630026/87233e48-2f20-11e7-848e-6148908e2139.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/a58b77c543da89568f119d66266f924e75c520b0

PR #3490, opened 2017-05-03
#3455 Intellisense for Parameters not working

MikhailArkhipov:
Handle case when caret is right at the end and there is no closing brace in the function call.

### DIFF
--- a/src/R/Editor/Test/Completions/RCompletionSourceTest.cs
+++ b/src/R/Editor/Test/Completions/RCompletionSourceTest.cs
@@ -298,5 +298,21 @@ aaa(a
             completionSets[0].Completions.Should().NotBeEmpty()
                 .And.Contain(c => c.DisplayText == "a =");
         }
+
+        [Test]
+        public void UserFunctionArguments02() {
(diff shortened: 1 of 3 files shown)


src/R/Editor/Test/Completions/RCompletionSourceTest.cs
```diff
@@ -298,5 +298,21 @@ public class RCompletionSourceTest {
...
                 .And.Contain(c => c.DisplayText == "a =");
         }
+
+        [Test]
+        public void UserFunctionArguments02() {
+            var completionSets = new List<CompletionSet>();
+            string content =
+                @"
+aaa <- function(a, b, c) { }
+aaa(x, ";
```
AlexanderSher: Do you care about the following cases:
`aaa(x, aaa(y, `
`aaa(x, (1+2`

MikhailArkhipov: First one, yes, second - not so much. But I'll add few more tests

AlexanderSher:
👍

PR #3499, opened 2017-05-04
Fix #3231: View() does not show row names for ts and zoo

### LINKED ISSUE #3231: View() does not show row names for ts and zoo
View now support 2d data structures. ty very much. unfortunately, it does not show row names and instead just shows a row ordinal.  This is particularly horrible if the rownames contain useful information.

Example:
capture

int19h:
Use time() for ts, and zoo::index() for zoo, in lieu of row.names() when the latter is absent.


MikhailArkhipov:
👍

int19h:
Do we need two sign-offs for Release/1_0, or is it good to go?

MikhailArkhipov:
@alexandersher is probably offline so go ahead, he can review post mortem

PR #3516, opened 2017-05-10
Linux broker: part 2

karthiknadig:
1. Video card information is extracted in linux
2. PAM authentication support using RunAsUser binary
3. implementation of IAuthenticationService on linux

### DIFF
--- a/src/Host/Broker/Impl/About/AboutController.cs
+++ b/src/Host/Broker/Impl/About/AboutController.cs
@@ -50,17 +50,12 @@ namespace Microsoft.R.Host.Broker.About {
             a.TotalPhysicalMemory = memoryInfo.TotalPhysicalMemory;
             a.FreePhysicalMemory = memoryInfo.FreePhysicalMemory;
 
-            var videoInfo = _systemInfo.GetVideoControllerInformation();
-            List<VideoCardInfo> videoCards = new List<VideoCardInfo>();
-            foreach(var cardInfo in videoInfo) {
-                var card = new VideoCardInfo() {
-                    VideoCardName = cardInfo.VideoCardName,
-                    VideoRAM = cardInfo.VideoRAM,
-                    VideoProcessor = cardInfo.VideoProcessor
-                };
-                videoCards.Add(card);
-            }
-            a.VideoCards = videoCards.ToArray();
+            a.VideoCards = _systemInfo.GetVideoControllerInformation().Select(ci => 
+                new VideoCardInfo() {
+                    VideoCardName = ci.VideoCardName,
+                    VideoRAM = ci.VideoRAM,
+                    VideoProcessor = ci.VideoProcessor
+                }).ToArray();
 
             a.Interpreters = _interpManager.Interpreters.Select(x => Invariant($"[{x.Id}] {x.Name}")).ToArray();
             if(a.Interpreters.Length > 0) {
--- a/src/Host/Broker/Impl/About/AboutController.cs
+++ b/src/Host/Broker/Impl/About/AboutController.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
@@ -50,9 +51,16 @@ namespace Microsoft.R.Host.Broker.About {
             a.FreePhysicalMemory = memoryInfo.FreePhysicalMemory;
 
             var videoInfo = _systemInfo.GetVideoControllerInformation();
-            a.VideoCardName = videoInfo.VideoCardName;
-            a.VideoRAM = videoInfo.VideoRAM;
-            a.VideoProcessor = videoInfo.VideoProcessor;
+            List<VideoCardInfo> videoCards = new List<VideoCardInfo>();
+            foreach(var cardInfo in videoInfo) {
+                var card = new VideoCardInfo() {
+                    VideoCardName = cardInfo.VideoCardName,
+                    VideoRAM = cardInfo.VideoRAM,
+                    VideoProcessor = cardInfo.VideoProcessor
+                };
+                videoCards.Add(card);
+            }
+            a.VideoCards = videoCards.ToArray();
 
             a.Interpreters = _interpManager.Interpreters.Select(x => Invariant($"[{x.Id}] {x.Name}")).ToArray();
             if(a.Interpreters.Length > 0) {
--- a/src/Linux/Core/Impl/OS/UnixProcessServices.cs
+++ b/src/Linux/Core/Impl/OS/UnixProcessServices.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.Common.Core.OS {
+    public class UnixProcessServices : IProcessServices {
+        public string MessageFromExitCode(int processExitCode) {
+            throw new NotImplementedException();
+        }
+
+        public Process Start(ProcessStartInfo psi) {
+            return Process.Start(psi);
+        }
+
+        public Process Start(string path) {
+            return Process.Start(path);
+        }
+    }
+}
--- a/src/Linux/Core/Impl/OS/UnixProcessServices.cs
+++ b/src/Linux/Core/Impl/OS/UnixProcessServices.cs
@@ -1,12 +1,17 @@
-﻿using System;
-using System.Collections.Generic;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.ComponentModel;
 using System.Diagnostics;
-using System.Text;
 
 namespace Microsoft.Common.Core.OS {
     public class UnixProcessServices : IProcessServices {
         public string MessageFromExitCode(int processExitCode) {
-            throw new NotImplementedException();
+            // on linux this calls the native strerror_r to get the message
+            // see: https://github.com/dotnet/corefx/blob/master/src/Microsoft.Win32.Primitives/src/System/ComponentModel/Win32Exception.Unix.cs
+            Win32Exception ex = new Win32Exception(processExitCode);
+            return ex.Message;
         }
 
         public Process Start(ProcessStartInfo psi) {
--- a/src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
+++ b/src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
@@ -14,20 +14,35 @@ using Microsoft.Common.Core;
 using Microsoft.Common.Core.IO;
 using Microsoft.Common.Core.OS;
 using Microsoft.R.Host.Protocol;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.R.Host.Broker.Services {
     public class LinuxSystemInfoService : ISystemInfoService {
+        private readonly ILogger<ISystemInfoService> _logger;
         private readonly IFileSystem _fs;
         private readonly IProcessServices _ps;
-        private static char[] _memInfoSplitter = new char[] { ' ', ':', '\t' };
-        private static Dictionary<string, long> _sizeLUT = new Dictionary<string, long> {
+        private readonly static char[] _memInfoSplitter = new char[] { ' ', ':', '\t' };
+        private readonly static Dictionary<string, long> _sizeLUT = new Dictionary<string, long> {
             {"KB", 1000L},
             {"MB", 1000L * 1000L},
             {"GB", 1000L * 1000L * 1000L},
             {"TB", 1000L * 1000L * 1000L * 1000L},
         };
 
-        public LinuxSystemInfoService(IFileSystem fs, IProcessServices ps) {
+        // Match anything that looks like 1024M or 1024MB
+        private readonly static Regex _ramPattern = new Regex(@"[\d]+[GM][B]?", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+        // Match numbrs in a string
+        private readonly static Regex _numPattern = new Regex(@"[\d]+", RegexOptions.Compiled);
+
+        // Match 'whitespace' at the begining of a string
+        private readonly static Regex _indentPattern = new Regex(@"(?<indent>[\s]*).*", RegexOptions.Compiled);
+
+        // Match '<whitespace><key>:<value>' in a string
+        private readonly static Regex _keyValuePairsPattern = new Regex(@"[\s]*(?<key>[ \w]*):(?<value>.*)", RegexOptions.Compiled);
+
+        public LinuxSystemInfoService(ILogger<ISystemInfoService> logger, IFileSystem fs, IProcessServices ps) {
+            _logger = logger;
             _fs = fs;
             _ps = ps;
         }
@@ -58,10 +73,8 @@ namespace Microsoft.R.Host.Broker.Services {
                 foreach (string line in data) {
                     string[] split = line.Split(_memInfoSplitter, StringSplitOptions.RemoveEmptyEntries);
                     if (split.Length == 3) {
-                        long value;
-                        if (split.Length > 2 && long.TryParse(split[1], out value)) {
-                            long multiplier;
-                            if (!_sizeLUT.TryGetValue(split[2].ToUpper(), out multiplier)) {
+                        if (split.Length > 2 && long.TryParse(split[1], out long value)) {
+                            if (!_sizeLUT.TryGetValue(split[2].ToUpper(), out long multiplier)) {
                                 multiplier = 1;
                             }
                             meminfo.Add(split[0], value * multiplier);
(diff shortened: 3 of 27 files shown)


src/Linux/Core/Impl/OS/UnixProcessServices.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.Common.Core.OS {
+    public class UnixProcessServices : IProcessServices {
+        public string MessageFromExitCode(int processExitCode) {
+            throw new NotImplementedException();
```
AlexanderSher: Will we implement it in future?

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
```diff
@@ -111,8 +118,103 @@ public class LinuxSystemInfoService : ISystemInfoService {
...
+
+                    _videoCardInfo.Add(new VideoCardInfo() { VideoCardName = productName, VideoRAM = vram, VideoProcessor = vendorName });
+                    ++count;
+                }
+            }
+        }
+
+        private static long GetRamValueMB(string vramStr) {
+            long ram = 0;
+            Regex numPattern = new Regex("[0-9]+");
```
AlexanderSher: `new Regex("[0-9]+",  RegexOptions.Compiled);`, and let's make it a static field.

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
```diff
@@ -111,8 +118,103 @@ public class LinuxSystemInfoService : ISystemInfoService {
...
+                psi.RedirectStandardOutput = true;
+                psi.RedirectStandardError = true;
+                proc = _ps.Start(psi);
+
+                using (StreamReader reader = new StreamReader(proc.StandardOutput.BaseStream)) {
+                    while (!reader.EndOfStream) {
+                        standardOutData.Add(reader.ReadLine());
+                    }
+                }
+            } catch (Exception) {
```
AlexanderSher: Why do we swallow all errors?

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
```diff
@@ -111,8 +118,103 @@ public class LinuxSystemInfoService : ISystemInfoService {
...
+
+            string productPart = "product: ";
+            string vendorPart = "vendor: ";
+            string busPart = "bus info: ";
+
+            // Match anything that looks like 1024M or 1024MB
+            Regex pattern = new Regex("[0-9]+[GM][B]?", RegexOptions.IgnoreCase);
+
+            int count = 1;
+            for (int i=0;  i < lshwDisplayData.Length; ++i) {
```
int19h: This whole thing feels like it could be a single regex with named matches to extract values that we need.

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
int19h: Also, `[0-9]` is `\d`.

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
```diff
@@ -111,8 +118,103 @@ public class LinuxSystemInfoService : ISystemInfoService {
...
+            try {
+                ProcessStartInfo psi = new ProcessStartInfo();
+                psi.Arguments = arguments;
+                psi.FileName = command;
+                psi.RedirectStandardInput = true;
+                psi.RedirectStandardOutput = true;
+                psi.RedirectStandardError = true;
+                proc = _ps.Start(psi);
+
+                using (StreamReader reader = new StreamReader(proc.StandardOutput.BaseStream)) {
```
int19h: `proc.StandardOutput` is already a reader - why do we need to create another one here?

Also, it seems that it would be easier to just do `ReadToEnd` to read everything at once?

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Microsoft.R.Host.Broker.Services {
+    public class Utility {
+        public static bool AuthenticateUser(ILogger<IAuthenticationService> logger, IProcessServices ps,  string username, string password, out string profileDir) {
+            bool retval = false;
+            Process proc = null;
+            string userDir = string.Empty;
+            try {
+                proc = CreateRLaunchProcess(ps, true);
+                using (BinaryWriter writer = new BinaryWriter(proc.StandardInput.BaseStream))
```
int19h: Need to use the constructor that doesn't take ownership (`leaveOpen: true`), since we don't own the streams here.

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+                    writer.Flush();
+
+                    // wait for the process to exit;
+                    while (!proc.HasExited) {
+                        Thread.Sleep(100);
+                    }
+
+                    if (proc.ExitCode == 0) {
+                        retval = true;
+                        int size = (int)reader.ReadUInt32();
```
int19h: `ReadInt32`.

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+                    string json = JsonConvert.SerializeObject(message);
+                    byte[] jsonBytes = Encoding.UTF8.GetBytes(json);
+
+                    writer.Write(jsonBytes.Length);
+                    writer.Write(jsonBytes);
+                    writer.Flush();
+
+                    // wait for the process to exit;
+                    while (!proc.HasExited) {
+                        Thread.Sleep(100);
```
AlexanderSher: Any change we hang here?

src/Linux/Host/Broker/Impl/Startup/LinuxStartup.cs
```diff
@@ -17,11 +18,13 @@ public sealed class LinuxStartup : Startup {
...
 
             var fs = new UnixFileSystem();
+            var ps = new UnixProcessServices();
             services.AddSingleton<IFileSystem>(fs)
+                .AddSingleton<IProcessServices>(ps)
                 .AddSingleton<IAuthenticationService, LinuxAuthenticationService>()
                 .AddSingleton<IRHostProcessService, LinuxRHostProcessService>()
                 .AddSingleton<IRInstallationService, RInstallation>()
-                .AddSingleton<ISystemInfoService>(new LinuxSystemInfoService(fs));
+                .AddSingleton<ISystemInfoService>(new LinuxSystemInfoService(fs, ps));
```
AlexanderSher: You don't need to instantiate it explicitly - `DI` will handle it for you:
```cs
services.AddSingleton<IFileSystem>(fs)
                 .AddSingleton<IProcessServices, UnixFileSystem>()
                 .AddSingleton<IAuthenticationService, LinuxAuthenticationService>()
                 .AddSingleton<IRHostProcessService, LinuxRHostProcessService>()
                 .AddSingleton<IRInstallationService, RInstallation>()
                 .AddSingleton<ISystemInfoService, LinuxSystemInfoService>();
```

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+                    }
+
+                    if (proc.ExitCode == 0) {
+                        retval = true;
+                        int size = (int)reader.ReadUInt32();
+                        byte[] bytes = new byte[size];
+                        if (reader.Read(bytes, 0, size) == size) {
+                            userDir = new string(Encoding.UTF8.GetChars(bytes));
```
int19h: `Encoding.UTF8.GetString`

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+                    // wait for the process to exit;
+                    while (!proc.HasExited) {
+                        Thread.Sleep(100);
+                    }
+
+                    if (proc.ExitCode == 0) {
+                        retval = true;
+                        int size = (int)reader.ReadUInt32();
+                        byte[] bytes = new byte[size];
+                        if (reader.Read(bytes, 0, size) == size) {
```
int19h: You can do `ReadBytes`, then you don't need to alloc the array.

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+                    } else {
+                        logger.LogCritical(Resources.Error_AuthFailed, GetRLaunchExitCodeMessage(proc.ExitCode));
+                        retval = false;
+                    }
+                }
+            } catch (Exception ex) {
+                logger.LogCritical(Resources.Error_AuthFailed, ex.Message);
+            } finally {
+                if(proc != null && !proc.HasExited) {
+                    proc.Kill();
```
int19h: This can also throw, so there needs to be another try/catch around it.

src/Windows/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
```diff
@@ -88,16 +88,21 @@ public sealed class SessionInformationCommand : IAsyncCommand {
             sb.AppendLine("\t" + Resources.PhysicalMemory.FormatInvariant(aboutHost.TotalPhysicalMemory, aboutHost.FreePhysicalMemory));
             sb.AppendLine("\t" + Resources.VirtualMemory.FormatInvariant(aboutHost.TotalVirtualMemory, aboutHost.FreeVirtualMemory));
 
-            if (!string.IsNullOrEmpty(aboutHost.VideoCardName)) {
-                sb.AppendLine("\t" + Resources.VideoCardName.FormatInvariant(aboutHost.VideoCardName));
-            }
+            int count = 1;
+            foreach(var cardInfo in aboutHost.VideoCards) {
```
AlexanderSher:
```cs
var videoCardsInfos = aboutHost.VideoCards.SelectMany((vc, i) => new [] {
    Resources.VideoCardName.FormatInvariant(i+1, cardInfo.VideoCardName),
    Resources.VideoProcessor.FormatInvariant(i+1, cardInfo.VideoProcessor),
    Resources.VideoRAM.FormatInvariant(i+1, cardInfo.VideoRAM)
});

foreach (var cardInfo in videoCardsInfos) {
    sb.AppendLine("\t" + cardInfo );
}
```

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (!str.empty()) {
+        if (fread(&str[0], str.size(), 1, stream) != 1) {
+            return std::string();
+        }
+    }
+
+    return str;
+}
+
+std::mutex output_lock;
```
int19h: Why do we need to synchronize here? There should be only one thread, no?

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+void write_string(FILE* stream, std::string &data) {
+    boost::endian::little_uint32_buf_t data_size(static_cast<uint32_t>(data.size()));
+
+    std::lock_guard<std::mutex> lock(output_lock);
+    if (fwrite(&data_size, sizeof data_size, 1, stream) == 1) {
```
int19h: There should probably be some error handling here. Since we can't really do much if the writes failed, I think you can just call std::terminate to fail fast.

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if (fwrite(data.data(), data.size(), 1, stream) == 1) {
+            fflush(stream);
+        }
+    }
+}
+
+void write_string(FILE* stream, const char *data) {
```
int19h: This shouldn't be necessary, since `std::string` has an implicit constructor from `const char*`.

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if (fread(&str[0], str.size(), 1, stream) != 1) {
+            return std::string();
+        }
+    }
+
+    return str;
+}
+
+std::mutex output_lock;
+void write_string(FILE* stream, std::string &data) {
```
int19h: `const std::string& data`

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+void write_string(FILE* stream, const char *data) {
+    std::string str_data(data);
+    return write_string(stream, str_data);
+}
+
+int rtvs_conv(int num_msg, const struct pam_message **msgm, struct pam_response **response, void *appdata_ptr)
+{
```
int19h: Formatting.

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+}
+
+void write_string(FILE* stream, const char *data) {
+    std::string str_data(data);
+    return write_string(stream, str_data);
+}
+
+int rtvs_conv(int num_msg, const struct pam_message **msgm, struct pam_response **response, void *appdata_ptr)
```
int19h: Since this is C++ rather than C, you don't have to spell `struct` every time.

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+            break;
+        case PAM_TEXT_INFO:
+            write_string(output, msgm[count]->msg);
+            break;
+        }
+
+        if (str) {
+            reply[count].resp_retcode = 0;
+            reply[count].resp = str;
+            str = nullptr;
```
int19h: No need to reset it, since it's local and initialized on every iteration.

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    if ((pam_err = pam_authenticate(pamh, 0)) != PAM_SUCCESS) {
+        return RTVS_AUTH_BAD_INPUT;
+    }
+
+    return RTVS_AUTH_OK;
+}
+
+std::string get_user_home(std::string &username) {
```
int19h: `const std::string& username`

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+    struct passwd *pw = getpwnam(username.c_str());
+    if (pw && pw->pw_dir && pw->pw_dir[0] != '\0') {
+        return std::string(strdup(pw->pw_dir));
+    }
+    return std::string();
+}
+
+int authenticate_only(picojson::object& json) {
```
int19h: `const picojson::object`

src/Linux/Native/RunAsUser/readme.txt
```diff
@@ -0,0 +1,11 @@
+========================================================================
+    MAKEFILE PROJECT : Microsoft.R.Host.RunAsUser Project Overview
+========================================================================
+
+Build instructions:
+Compile: g++ -std=c++14 -fexceptions -fpermissive -O0 -ggdb -I../src -I../lib/picojson -c ../src/*.c*
```
int19h: Why `-O0`? Doesn't this disable all optimizations?

src/Linux/Native/RunAsUser/readme.txt
```diff
@@ -0,0 +1,11 @@
...
+Link: g++ -g -o Microsoft.R.Host.RunAsUser.out ./*.o -lpthread -L/usr/lib/x86_64-linux-gnu -lpam
```
int19h: The output binary shouldn't have an extension - `a.out` is a very old convention that's no longer in use. So this should just be `Microsoft.R.Host.RunAsUser`.

src/Linux/Native/RunAsUser/util.h
```diff
@@ -0,0 +1,60 @@
+/* ****************************************************************************
```
int19h: We should figure out a way to share these headers between the C++ projects that we have, so that we don't have multiple copies floating around. Probably a good idea to share libraries like picojson as well.

src/Linux/Core/Impl/OS/UnixProcessServices.cs
karthiknadig: Yes. We should be able to provide some form of text error message here.

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -0,0 +1,99 @@
...
+            string userDir = string.Empty;
+            try {
+                proc = CreateRLaunchProcess(ps, true);
+                using (BinaryWriter writer = new BinaryWriter(proc.StandardInput.BaseStream))
+                using (BinaryReader reader = new BinaryReader(proc.StandardOutput.BaseStream)) {
+                    AuthenticationOnlyMessage message = new AuthenticationOnlyMessage();
+                    message.Username = username;
+                    message.Password = password;
+
+                    string json = JsonConvert.SerializeObject(message);
```
int19h: The usual naming convention for fields in JSON is `camelCase` - can you enable that here?
```cs
SerializeObject(... ,new JsonSerializerSettings { 
        ContractResolver = new CamelCasePropertyNamesContractResolver() 
    })
```
(and adjust the C++ side accordingly)

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
karthiknadig: I don't think we should kill broker incase of error. I added logging to track failures.

---
In reply to: [115842875](https://github.com/Microsoft/RTVS/pull/3516#discussion_r115842875) [](ancestors = 115842875)

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
karthiknadig: It was lot simpler to do it line by line.

---
In reply to: [115844910](https://github.com/Microsoft/RTVS/pull/3516#discussion_r115844910) [](ancestors = 115844910)

src/Windows/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
karthiknadig: That won't skip a value if it is not present.

---
In reply to: [115848894](https://github.com/Microsoft/RTVS/pull/3516#discussion_r115848894) [](ancestors = 115848894)

src/Linux/Native/RunAsUser/util.h
karthiknadig: Agreed. But not is this change.

---
In reply to: [115851705](https://github.com/Microsoft/RTVS/pull/3516#discussion_r115851705) [](ancestors = 115851705)

src/Windows/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
AlexanderSher: Is it possible that there is no value?

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -1,61 +1,98 @@
...
-                    AuthenticationOnlyMessage message = new AuthenticationOnlyMessage();
-                    message.Username = username;
-                    message.Password = password;
-
-                    string json = JsonConvert.SerializeObject(message);
-                    byte[] jsonBytes = Encoding.UTF8.GetBytes(json);
-
+                using (BinaryWriter writer = new BinaryWriter(proc.StandardInput.BaseStream, Encoding.UTF8, true))
+                using (BinaryReader reader = new BinaryReader(proc.StandardOutput.BaseStream, Encoding.UTF8, true)) {
+                    AuthenticationOnlyMessage message = new AuthenticationOnlyMessage() { Username = username, Password = password };
```
AlexanderSher: `var message`

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -1,61 +1,98 @@
...
+                    var jsonBytes = Encoding.UTF8.GetBytes(json);
                     writer.Write(jsonBytes.Length);
                     writer.Write(jsonBytes);
                     writer.Flush();
 
-                    // wait for the process to exit;
-                    while (!proc.HasExited) {
-                        Thread.Sleep(100);
-                    }
+                    var waitForExit = Task.Run(() => {
```
AlexanderSher: Is there any problem with `Process.WaitForExit(3000)`?

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
```diff
@@ -58,10 +73,8 @@ public class LinuxSystemInfoService : ISystemInfoService {
                 foreach (string line in data) {
                     string[] split = line.Split(_memInfoSplitter, StringSplitOptions.RemoveEmptyEntries);
```
int19h: Can't values have spaces in them (not numbers, obviously, but in general?). It feels like this should really just split by `:`, and use the `String.Split` overload that specifies the maximum number of subitems (so it only splits on the first colon, not any subsequent ones). Then `Trim` spaces on the values, and parse them if they're numbers.

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
```diff
@@ -135,50 +148,80 @@ public class LinuxSystemInfoService : ISystemInfoService {
...
                             }
                         }
                     }
+                }
+                _videoCardInfo.Add(new VideoCardInfo() { VideoCardName = productName, VideoRAM = vram, VideoProcessor = vendorName });
+            }
+        }
+
+        private static IEnumerable<IDictionary<string, string>> ParseDisplayData(IEnumerable<string> lines) {
+            List<Dictionary<string, string>> parsedData = new List<Dictionary<string, string>>();
```
int19h: This thing will be slightly simpler if, instead of creating a list and filling it, you'll use `yield break` and `yield return` to generate results as you're parsing.

src/Linux/Host/Broker/Impl/Services/LinuxSystemInfoService.cs
int19h: Ah, nevermind, I confused that with video info.

src/Linux/Native/RunAsUser/main.cpp
```diff
@@ -41,6 +40,16 @@ FILE *input = nullptr, *output = nullptr, *error = nullptr;
 #define RTVS_AUTH_BAD_INPUT   201
 #define RTVS_AUTH_NO_INPUT    202
 
+#define RTVS_JSON_MSG_NAME      "name"
```
int19h: Can you make these all constants rather than defines? `constexpr int` for numbers, `constexpr char[]` for strings.

src/Windows/R/Components/Impl/InteractiveWorkflow/Commands/SessionInformationCommand.cs
karthiknadig: yes. WSL for example does not have display device. 'lshw -c display' will return nothing.

---
In reply to: [116114194](https://github.com/Microsoft/RTVS/pull/3516#discussion_r116114194) [](ancestors = 116114194)

src/Linux/Host/Broker/Impl/Services/Utility.cs
karthiknadig: oops. forgot about that one.

---
In reply to: [116115166](https://github.com/Microsoft/RTVS/pull/3516#discussion_r116115166) [](ancestors = 116115166)

karthiknadig:
Ping.

src/Host/Broker/Impl/About/AboutController.cs
```diff
@@ -50,9 +51,16 @@ public class AboutController : Controller {
...
+            List<VideoCardInfo> videoCards = new List<VideoCardInfo>();
+            foreach(var cardInfo in videoInfo) {
+                var card = new VideoCardInfo() {
+                    VideoCardName = cardInfo.VideoCardName,
+                    VideoRAM = cardInfo.VideoRAM,
+                    VideoProcessor = cardInfo.VideoProcessor
+                };
+                videoCards.Add(card);
+            }
+            a.VideoCards = videoCards.ToArray();
```
AlexanderSher:
```cs
a.VideoCards = _systemInfo.GetVideoControllerInformation()
    .Select(ci => new VideoCardInfo {
        VideoCardName = ci.VideoCardName,
        VideoRAM = ci.VideoRAM,
        VideoProcessor = ci.VideoProcessor
    }).ToArray();
```

Though up to you

PR #3518, opened 2017-05-10
Fix PM accessability issues

### LINKED ISSUE #3348: Make package manager message more friendly
Something like 'Open R interactive window or restart R session...`

### LINKED ISSUE #3458: Use VS infobar instead of custom implementation in Package Manager

AlexanderSher:
- Fix #3348: Make package manager message more friendly
- Fix #3458: Use VS infobar instead of custom implementation in Package Manager


MikhailArkhipov:
👍

AlexanderSher:

PR #3539, opened 2017-05-16
Fix #3523: Find out why fixtures get disposed before test is competed

### LINKED ISSUE #3523: Find out why fixtures get disposed before test is competed

AlexanderSher:
@MikhailArkhipov , @int19h , @karthiknadig , I'd like to hear your opinion about `TestEnvironment` singleton. In production, it will always be null. It tests, it will be initialized once per each AppDomain and then utilized by test runners. While it is a static entry point, it doesn't create a concurrency issues inside tests cause all its functionality is delegated to `AsyncLocal` instances.


MikhailArkhipov:
I think its fine. Actually it may also help eliminate `IsUnitTestEnvironment` from `ICoreShell` and effectively reduce it to the `IServiceContainer`

src/Common/Core/Impl/Testing/TestEnvironment.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Common.Core.Testing {
+    public static class TestEnvironment {
+        private static ITestEnvironment _current;
+
+        public static ITestEnvironment Current {
+            get => _current;
+            set {
+                var oldValue = Interlocked.CompareExchange(ref _current, value, null);
+                if (oldValue != null) {
```
MikhailArkhipov: `Check.InvalidOperation`, up to you

MikhailArkhipov:
👍

PR #3544, opened 2017-05-16
Fix couple more issues with TaskObserver

AlexanderSher:
- Allow UIThreadHelper to handle tasks that throw exceptions to UI thread
- Fail TaskObserver with inner exception if possible

### DIFF
--- a/src/Common/Core/Impl/Threading/MainThreadExtensions.cs
+++ b/src/Common/Core/Impl/Threading/MainThreadExtensions.cs
@@ -7,5 +7,8 @@ namespace Microsoft.Common.Core.Threading {
     public static class MainThreadExtensions {
         public static MainThreadAwaitable SwitchToAsync(this IMainThread mainThread, CancellationToken cancellationToken = default(CancellationToken))
             => new MainThreadAwaitable(mainThread, cancellationToken);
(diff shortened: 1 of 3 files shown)


src/Common/Core/Impl/Threading/MainThreadExtensions.cs
```diff
@@ -7,5 +7,8 @@ namespace Microsoft.Common.Core.Threading {
     public static class MainThreadExtensions {
         public static MainThreadAwaitable SwitchToAsync(this IMainThread mainThread, CancellationToken cancellationToken = default(CancellationToken))
             => new MainThreadAwaitable(mainThread, cancellationToken);
+
+        public static bool CheckAccess(this IMainThread mainThread)
```
int19h: Should this be something like `IsOnMainThread`? It's not clear what access is being checked here...

int19h:
👍

src/Common/Core/Impl/Threading/MainThreadExtensions.cs
AlexanderSher: `IMainThread.IsOnMainThread()` is a bit tautological

PR #3551, opened 2017-05-17
Fix #3545: Add test fixture that will preserve broker between tests

### LINKED ISSUE #3545: Add test fixture that will preserve broker between tests
- Fixture will implement `IMethodFixture` and will start broker process at first usage. 
- Broker process should be terminated as a child process to test runner, though we can terminate it manually

AlexanderSher:
@int19h , can you please take a look at broker startup code (`RemoteBrokerProcess` class)?

Current implementation required changes in product code so that security service can be injected.
There is an alternative way to implement it: `RemoteBrokerFixture` can reference `IServiceContainer` in its `Create` method and change `SecurityServiceStub` right in it, but that will affect the test itself.

@MikhailArkhipov , @int19h , @karthiknadig , which one do you think is better?

### DIFF
--- a/src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
+++ b/src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
@@ -14,19 +14,17 @@ using Microsoft.R.Host.Client.BrokerServices;
 
 namespace Microsoft.R.Host.Client.Host {
     internal class RemoteCredentialsDecorator : ICredentialsDecorator {
+        private readonly IServiceContainer _services;
         private volatile Credentials _credentials;
         private readonly AsyncReaderWriterLock _lock;
         private readonly string _authority;
         private readonly string _workspaceName;
-        private readonly IMainThread _mainThread;
-        private readonly ISecurityService _security;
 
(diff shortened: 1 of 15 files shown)


src/Host/Client/Impl/Host/RemoteCredentialsDecorator.cs
```diff
@@ -14,17 +14,19 @@
...
         private volatile Credentials _credentials;
         private readonly AsyncReaderWriterLock _lock;
         private readonly string _authority;
         private readonly string _workspaceName;
+        private readonly IMainThread _mainThread;
+        private readonly ISecurityService _security;
 
-        public RemoteCredentialsDecorator(string credentialAuthority, string workspaceName, IServiceContainer services) {
-            _services = services;
+        public RemoteCredentialsDecorator(string credentialAuthority, string workspaceName, IMainThread mainThread, ISecurityService security) {
```
MikhailArkhipov: Intent to pass `IServiceContainer` here was to be able to handle that `//TODO` below when cred storage cannot be accessed (we had actual bug on it). We may need to show some UI or log stuff.

AlexanderSher: Ok then, I've reverted product changes

int19h:
👍

PR #3554, opened 2017-05-17
Linux broker: Part 3

karthiknadig:


src/Host/Broker/Impl/Interpreters/Interpreter.cs
```diff
@@ -26,13 +27,16 @@ public class Interpreter {
             Version = Version
         };
 
-        public Interpreter(InterpreterManager manager, string id, string name, string path, string binPath, Version version) {
+        public IRInterpreterInfo RInterpreterInfo { get; }
+
+        public Interpreter(InterpreterManager manager, string id, string name, string path, string binPath, Version version, IRInterpreterInfo rInterpreterInfo) {
```
int19h: If we have `rInterpreterInfo` directly, then a bunch of arguments become redundant, since they're properties of that instance. As do the corresponding properties on this class.

src/Linux/R/Interpreters/Impl/RInterpreterInfo.cs
```diff
@@ -26,21 +29,34 @@ public sealed class RInterpreterInfo : IRInterpreterInfo {
...
+
+        public string RShareDir { get; private set; }
+
+        public string[] SiteLibraryDirs { get; }
 
         /// <summary>
         /// 
         /// </summary>
         /// <param name="name">Name of the R interpreter</param>
         /// <param name="path">Path to the /R folder</param>
```
int19h: Docstring needs to be fixed to match the new args.

src/Linux/R/Interpreters/Impl/RInterpreterInfo.cs
```diff
@@ -7,6 +7,9 @@
 using Microsoft.Common.Core.IO;
 using Microsoft.Common.Core.Services;
 using Microsoft.Common.Core.UI;
+using Microsoft.Common.Core.OS;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.R.Interpreters {
     public sealed class RInterpreterInfo : IRInterpreterInfo {
```
int19h: I'm somewhat confused - why do we have two different implementations of `RInterpreterInfo` in two different namespaces, with seemingly both used at the same time?

src/Linux/R/Interpreters/Impl/RInstallation.cs
```diff
@@ -36,9 +36,9 @@ public sealed class RInstallation : IRInstallationService {
             }
 
             if(package != null) {
-                return new RInterpreterInfo(name, path, package.Version, package.GetVersion(), _fileSystem);
+                return new RInterpreterInfo(name, package, package.Version, package.GetVersion(), _fileSystem);
             }
-            return new RInterpreterInfo(name, path, string.Empty, null, _fileSystem);
+            return new RInterpreterInfo(name, InstalledPackageInfo.EmptyPackage, string.Empty, null, _fileSystem);
```
int19h: Given that the constructor extracts all paths from the package now, is this a valid case? i.e. would it be a usable interpreter if we couldn't find the package?

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -131,5 +145,13 @@ public class Utility {
                 ContractResolver = new CamelCasePropertyNamesContractResolver()
             };
         }
+
+        private static IEnumerable<string> GetEnvironmentForForkAndExec(IDictionary<string,string> environment){
```
int19h: This method can be replaced (where it's called, since it's only called once) with:
```cs
environment.Select(e => $"{e.Key}={e.Value}")
```

src/Linux/R/Interpreters/Impl/RInterpreterInfo.cs
```diff
@@ -26,21 +29,34 @@ public sealed class RInterpreterInfo : IRInterpreterInfo {
...
+        public string[] SiteLibraryDirs { get; }
 
         /// <summary>
         /// 
         /// </summary>
         /// <param name="name">Name of the R interpreter</param>
         /// <param name="path">Path to the /R folder</param>
         /// <param name="fileSystem"></param>
-        public RInterpreterInfo(string name, string path, string version, Version parsedVersion, IFileSystem fileSystem) {
+        public RInterpreterInfo(string name, InstalledPackageInfo package, string version, Version parsedVersion, IFileSystem fileSystem) {
```
int19h: Can you rename `package` here to something like `corePackage`, to make it clear that it's not just some random package, but the one representing the whole interpreter?

src/Linux/R/Interpreters/Impl/RInstallation.cs
```diff
@@ -55,34 +55,18 @@ public sealed class RInstallation : IRInstallationService {
             var selectedPackages = packagesInfo.Where(p => p.PackageName.StartsWithIgnoreCase("microsoft-r-open-mro") && svl.IsCompatibleVersion(p.GetVersion()));
             foreach (var package in selectedPackages) {
                 var files = package.GetPackageFiles(_fileSystem);
-                string rInstallPath = GetRInstallPath(files, _fileSystem);
-                list.Add(new RInterpreterInfo($"Microsoft R Open '{package.Version}'", rInstallPath, package.Version, package.GetVersion(), _fileSystem));
+                list.Add(RInterpreterInfo.CreateFromPackage(package, "Microsoft R Open", _fileSystem));
```
int19h: You can do `yield return` here, and drop the list altogether.

int19h: (and the same in the method below)

src/Linux/R/Interpreters/Impl/RInterpreterInfo.cs
karthiknadig: Do you mean `InterpreterInfo` and `RInterpreterInfo`?

int19h: Ah, nevermind, I didn't notice that the other `RInterpreterInfo` is in the `Windows` namespace.

PR #3562, opened 2017-05-19
Fix part of the issue with main thread tasks in tests

### LINKED ISSUE #3009: Some tests always fail in a group
image

image

Pass on rerun

AlexanderSher:
Also fixes #3009: Some tests always fail in a group


src/UnitTests/Core/Impl/Threading/TestMainThread.cs
```diff
@@ -0,0 +1,149 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Common.Core;
+using Microsoft.Common.Core.Threading;
+using Microsoft.Common.Core.UI;
+
+namespace Microsoft.UnitTests.Core.Threading {
+    public class TestMainThread : IProgressDialog, IMainThread, IDisposable {
```
MikhailArkhipov: So main thread fights with progress in tests?

MikhailArkhipov:
👍

src/UnitTests/Core/Impl/Threading/TestMainThread.cs
AlexanderSher: There were 3 types of issues:
- Progress is shown by one test and other test tries to access UI thread at the same time. This is fixed by using `AsyncLocal`, so now each test has its own instance of `_blockingLoop`
- Test sends the task to UI thread after it is actually finished. This is fixed by adding `CancellationToken` which is cancelled after fixtures are disposed
- ServiceManager is disposed from background thread while task on UI thread is running. Not fixed yet, probably will require some product code changes.

Issue #3387, opened 2017-04-04
setLocale fails on linux


karthiknadig:
```
In Sys.setlocale("LC_CTYPE", ".437") :
 
 OS reports request to set locale to ".437" cannot be honored
```

int19h:
Unsurprising, since 437 is a Windows codepage, which has no meaning on Linux. On MinGW, you can use codepages as encodings, but on Unix systems we have to use proper identifiers like `.UTF-8`.

MikhailArkhipov:
We will be removing LC_CTYPE setting and rather replace it with setting environment variable `LANGUAGE=fr` since this is how R sets the locale in RGui.

int19h:
`LC_CTYPE` does not have anything to do with UI language, it's about encoding.

int19h:
But for Linux, we should just always set it to UTF-8 - it's strictly better than having any national encoding. We'd do it on Win32 if we could, but Windows (and hence MinGW) doesn't support UTF-8 as a codepage.

MikhailArkhipov:
Yes, should be `UTF-8` on Linux, I just mean that R sets locale internally by looking into the env variable. We don't have to set `LC_*` explicitly.

MikhailArkhipov:
Basically same as calling `Sys.setlocale('french')`

int19h:
`LC_CTYPE` is also an environment variable actually. But they set different aspects of the locale. `LC_CTYPE` specifies classification (e.g. what's considered a letter) and encoding/conversion. `LANGUAGE` is the UI language (specifically, it's what gettext uses to look strings up) - it doesn't specify the encoding. But we do want to work with a specific encoding, so we need to set both of these separately.

MikhailArkhipov:
I think if we just set the language. R will set encoding as per platform. I.e. utf-8 on Linux and appropriate codepage on Win32. Try `Sys.setlocale('russian')`, see what `Sys.getlocale()` reports.

int19h:
To clarify, `LANGUAGE` is not actually part of the locale (as C/POSIX defines it) at all. It's a gettext-specific variable, and it can only be set to a language identifier like `en` or `en-US`. It does not include the encoding, and does not say anything about encodings. It also doesn't affect the `LC_*` categories in any way. 

[...]

PR #3567, opened 2017-05-23
Fixes setLocale issue on Linux

karthiknadig:
#3387

### DIFF
--- a/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
+++ b/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
@@ -210,13 +210,18 @@ grDevices::deviceIsInteractive('ide')
             return evaluation.ExecuteAsync(script);
         }
 
-        public static Task SetCodePageAsync(this IRExpressionEvaluator evaluation, int codePage, CancellationToken cancellationToken = default(CancellationToken)) {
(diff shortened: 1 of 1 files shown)


src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -210,13 +210,18 @@ public static class RSessionEvaluationCommands {
             return evaluation.ExecuteAsync(script);
         }
 
-        public static Task SetCodePageAsync(this IRExpressionEvaluator evaluation, int codePage, CancellationToken cancellationToken = default(CancellationToken)) {
+        public static async Task SetCodePageAsync(this IRExpressionEvaluator evaluation, int codePage, CancellationToken cancellationToken = default(CancellationToken)) {
+            string cp = $".{codePage}";
             if (codePage == 0) {
-                // TODO: update this to get the codepage via platform agnostic service
-                codePage = 437;
+                var isWindows = await evaluation.EvaluateAsync<bool>(".Platform$OS.type == 'windows'", REvaluationKind.Normal, cancellationToken);
```
int19h: I suspect we'll be doing this often enough in the future, it might be worth making it an extension method on `IRExpressionEvaluator`.

PR #3568, opened 2017-05-23
Adding files needed for linux service

karthiknadig:


src/Linux/Setup/rservice.sh
```diff
@@ -0,0 +1,43 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root for license information.
+#
+# Description: Remote R Server 
+
+. /etc/init.d/functions
+
+start() {
+        initlog -c "echo -n Starting Remote R Server: "
+        dotnet /usr/bin/Microsoft.R.Host.Broker.dll --config "/usr/lib/rtvs/Microsoft.R.Host.Broker.config.json" &
```
int19h: We should put the config into `/etc/rtvs` (not necessarily in this PR, but at some point).

int19h:
👍

Issue #3024, opened 2017-01-20
Auto Completion of Roxygen Tags


Neil-Schneider:
When writing Roxygen documentation, it would be helpful to have autocompletion (tab completion) of Roxygen tags. 

Likely the trigger for this would be the `@` symbol in a roxygen comment line.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f654cfd573f15c2fbf477393bfb6675359d9422e

Issue #2767, opened 2016-11-29
Consider adding support for lintr


jflam:
lintr provides static code analysis for R - https://github.com/jimhester/lintr

There is already integration for a number of different editors and IDEs, including vim, emacs, Sublime, Atom, and RStudio.

![image](https://cloud.githubusercontent.com/assets/8877/20698248/905b551c-b5b4-11e6-8cb7-2e327ccd0c90.png)

MikhailArkhipov:
It may be relatively simple to add this diagnostics to our AST. We already have async processing as you type (rather than running package command in R in sync on the entire file) and full integration with error list etc. It is simple to output these as warnings or information. Multi-file handling is doable as well.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f654cfd573f15c2fbf477393bfb6675359d9422e

Issue #1317, opened 2016-03-22
Colorize inside Roxygen block


MikhailArkhipov:
It would be nice to colorize parameter names and other related items differently in 

```
#' @param ...
```

similar to what C# does in /// blocks

Neil-Schneider:
Currently, I use extra white space to make roxygen keywords stand out more. Colorization of the `@` parameters would be a nice addition.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/f654cfd573f15c2fbf477393bfb6675359d9422e

PR #3573, opened 2017-05-25
Roxygen colors, completion + LintR

MikhailArkhipov:
#1317 Colorize inside Roxygen block 
#3024 Auto Completion of Roxygen Tags 
#2767 Consider adding support for lintr 

+ tests


src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -54,7 +54,19 @@ public class ServiceManager : IServiceManager {
         public virtual IServiceManager AddService(Type type) {
             _disposeToken.ThrowIfDisposed();
 
-            var lazy = new Lazy<object>(() => Activator.CreateInstance(type));
+            object parameter = null;
```
AlexanderSher: There is an extension method that uses compiled expression tree: `ServiceManagerExtensions.AddService`. I'm actually not sure if we need `ServiceManager.AddService(Type type)` at all

MikhailArkhipov: OK, I'll see if that one is sufficient

src/Package/Impl/Editors/LanguageSettingsStorage.cs
```diff
@@ -24,27 +24,37 @@ public sealed class LanguageSettingsStorage : IVsTextManagerEvents4, IWritableEd
...
 
         /// <summary>
+        /// Loads settings via language (editor) tools options page
+        /// </summary>
+        public void Load() {
+            LoadLanguagePreferences();
+
+            var vsShell = _services.GetService<IVsShell>(typeof(SVsShell));
+            foreach (var curAutomationObjectName in _automationObjectNames) {
+                _package.GetAutomationObject(curAutomationObjectName, out object automationObject);
```
AlexanderSher: So `automationObject` isn't used, correct?

src/Package/Impl/Options/R/Editor/RLintOptionsDialog.cs
```diff
@@ -0,0 +1,250 @@
...
+using Microsoft.VisualStudio.R.Package.Shell;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.R.Package.Options.R.Editor {
+    public class RLintOptionsDialog: DialogPage {
+        private readonly IWritableREditorSettings _settings;
+
+        public RLintOptionsDialog() {
+            SettingsRegistryPath = @"UserSettings\R_Lint";
+            _settings = VsAppShell.Current.GetService<IWritableREditorSettings>();
```
AlexanderSher: Can we reference `VsAppShell.Current.GetService<IWritableREditorSettings>().LintOptions` directly?

src/Package/Impl/Editors/LanguageSettingsStorage.cs
MikhailArkhipov: Nope, this is just some acrobatics to cause VS to load settings.

src/Package/Impl/Options/R/Editor/RLintOptionsDialog.cs
MikhailArkhipov: Sure, was just simpler to debug

src/R/Editor/Impl/Document/REditorDocument.cs
```diff
@@ -16,6 +16,7 @@ namespace Microsoft.R.Editor.Document {
     /// </summary>
     public sealed class REditorDocument : IREditorDocument {
         private readonly IServiceContainer _services;
+        private readonly TreeValidator _validator;
```
AlexanderSher: Do you need to dispose this instance?

src/Windows/R/Editor/Impl/Settings/REditorSettings.cs
```diff
@@ -90,5 +93,7 @@ public sealed class REditorSettings: EditorSettings, IWritableREditorSettings, I
                 return _formatOptions;
             }
         }
+
+        public LintOptions LintOptions => _lintOptions;
```
AlexanderSher: `public LintOptions LintOptions { get; } = new LintOptions();`

src/R/Editor/Impl/Document/REditorDocument.cs
MikhailArkhipov: It is not disposable. Used to be when validator/syntax checkers could be native

Issue #3549, opened 2017-05-17
Logging for RunAsUser


karthiknadig:
Microsoft.R.Host.RunAsUser binary currently does not support logging.

PR #3588, opened 2017-06-02
Few more fixes

### LINKED ISSUE #3576: Linux: R Host process crashes on start.
RHost instance starts and immediately crashes when connecting to linux remote.

karthiknadig:
Fixes #3576
Implements #3549
Adds current working directory when RHost starts.

### DIFF
--- a/src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
+++ b/src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
@@ -45,37 +45,25 @@ namespace Microsoft.R.Host.Broker.Services {
             return new UnixProcess(process);
         }
 
-        private string GetUsrLibPath() {
-            const string x64Lib = "/usr/lib/x86_64-linux-gnu";
-            const string x86Lib = "/usr/lib";
-            return _fs.DirectoryExists(x64Lib) ? x64Lib : x86Lib;
-        }
-
-        private static string[] _standardPaths = { "/bin", "/usr/bin", "/usr/local/bin" };
-        private static string[] _pathSplitter = { ":" };
-        private IEnumerable<string> GetAllPaths() {
-            foreach(string s in _standardPaths) {
-                yield return s;
+        private string GetLdLibraryPath(string rBinPath) {
+            string ldLibPath = Environment.GetEnvironmentVariable("LD_LIBRARY_PATH");
+            if (string.IsNullOrEmpty(ldLibPath)) {
+                return rBinPath;
             }
-
-            string[] paths = Environment.GetEnvironmentVariable("PATH").Split(_pathSplitter, StringSplitOptions.RemoveEmptyEntries);
-            foreach(string p in paths) {
-                if (_fs.DirectoryExists(p)) {
-                    yield return p;
-                }
-            }
+            return $"{rBinPath}:{ldLibPath}";
         }
 
         private IDictionary<string, string> GetHostEnvironment(Interpreter interpreter, string profilePath, string userName) {
             string siteLibrary = string.Join(":", interpreter.RInterpreterInfo.SiteLibraryDirs);
-            string loadLibraryPath = string.Join(":", new string[] { interpreter.RInterpreterInfo.BinPath, GetUsrLibPath() });
-            string envPath = string.Join(":", GetAllPaths().Distinct());
+
+            // TODO: LD_LIBRARY_PATH should be set in R-Host where we call dlopen.
+            string loadLibraryPath = GetLdLibraryPath(interpreter.RInterpreterInfo.BinPath);
 
             Dictionary<string, string> environment = new Dictionary<string, string>() {
                 { "HOME"                    , profilePath},
                 { "LD_LIBRARY_PATH"         , loadLibraryPath},
                 { "LN_S"                    , GetCurrentOrDefault("LN_S")},
-                { "PATH"                    , envPath},
+                { "PATH"                    , Environment.GetEnvironmentVariable("PATH")},
                 { "PWD"                     , profilePath},
                 { "R_ARCH"                  , GetCurrentOrDefault("R_ARCH")},
                 { "R_BROWSER"               , GetCurrentOrDefault("R_BROWSER")},
--- a/src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
+++ b/src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
@@ -5,8 +5,10 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Linq;
 using System.Security.Claims;
 using System.Text;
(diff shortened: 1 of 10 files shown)


src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
```diff
@@ -41,15 +45,37 @@ class LinuxRHostProcessService : IRHostProcessService {
             return new UnixProcess(process);
         }
 
+        private string GetUsrLibPath() {
+            const string x64Lib = "/usr/lib/x86_64-linux-gnu";
```
int19h: Not sure how portable this is. IIRC some distros use `lib64`, and I don't know if there's any reliable standard.

In general, why do we need to try to detect this, instead of using the preset `LD_LIBRARY_PATH` and appending our own value as before? And ditto for `PATH`?

src/Linux/Native/RunAsUser/log.cpp
```diff
@@ -0,0 +1,164 @@
...
+* along with Microsoft R Host.  If not, see <http://www.gnu.org/licenses/>.
+*
+* ***************************************************************************/
+
+#include "stdafx.h"
+#include "log.h"
+
+using namespace std::literals;
+
+namespace rau {
```
int19h: Most of this is identical to R-Host code, right? Can we just share this file between the two?

src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
karthiknadig: when i was testing from VS Code this was the value for `PATH=
/home/user1/bin:/home/user1/.local/bin:/bin:/usr/bin:/usr/X11R6/bin:/usr/local/bin:/snap/bin`. I was trying to only include paths that exist. I don't think this is needed. Will remove.

For, `LD_LIBRARY_PATH', I wanted to make sure it works if a user decides to start broker from a command line. Again not absolutely needed.

int19h: For the latter, though, we only need to add the R directory to `LD_LIBRARY_PATH`, right? The standard `/usr/lib` etc entries should already be there - or should be implied.

int19h: Actually, come to think of it, does it need to be in the path at all? can't we just `dlopen` with full path to binary?

src/Linux/Native/RunAsUser/log.cpp
karthiknadig: Agreed. May be we can move these to a share common folder? or just change the build command to include the RHost log file?

karthiknadig: Actually currently the master RHost is not the merged one from msys. so this can't be done until we merge that.

src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
karthiknadig: MRO has other binaries e.g., libRBlas.so that gets loaded by libR.so. since MRO is in this location `/usr/lib64/microsoft-r/3.3/lib64/R/lib`. I am not sure what is the search rule for dlopen is, and also the EFL for libR.so does not specify where to find the lib relative to libR.so. It just specifies that it is needed.
```
user1@ub-dev:/usr/lib64/microsoft-r/3.3/lib64/R/lib$ readelf -d libR.so

Dynamic section at offset 0x1d6e328 contains 30 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libRblas.so]
 0x0000000000000001 (NEEDED)             Shared library: [libgfortran.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgomp.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000c (INIT)               0xda0b0
 0x000000000000000d (FINI)               0x3cde58
 0x000000006ffffef5 (GNU_HASH)           0x158
 0x0000000000000005 (STRTAB)             0x276e0
 0x0000000000000006 (SYMTAB)             0x9b90
 0x000000000000000a (STRSZ)              148201 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x1f6ff48
 0x0000000000000002 (PLTRELSZ)           63072 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0xcaa50
 0x0000000000000007 (RELA)               0x4e2c8
 0x0000000000000008 (RELASZ)             509832 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffe (VERNEED)            0x4e168
 0x000000006fffffff (VERNEEDNUM)         9
 0x000000006ffffff0 (VERSYM)             0x4b9ca
 0x000000006ffffff9 (RELACOUNT)          18466
 0x0000000000000000 (NULL)               0x0

```

src/Linux/Native/RunAsUser/log.cpp
int19h: Yep. Is this file different between the two branches?

If yes, then let's do the copy now, but put a TODO comment there saying that it's copy/paste and needs to be unified, and let's have a separate issue open for that as well.

src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
int19h: Ah, I see. The host itself now accepts the argument which is path to R install, right? I think it can do `putenv` to set `LD_LIBRARY_PATH` variable inside the host, immediately before `dlopen`, and it should respect that. That way we don't have to specify this path in two different places (arguments and environment) when launching the host.

int19h: (or rather `setenv`, because `putenv` has some really weird lifetime semantics)

karthiknadig: I will create a task for this and do it after the RHost merge.

src/Linux/Native/RunAsUser/log.cpp
karthiknadig: created Task #3590

PR #3591, opened 2017-06-02
Remove IServiceManager.AddService(Type)

AlexanderSher:
Remove `IServiceManager` method that almost not used and can be replaced with existing extension method


MikhailArkhipov:
👍

PR #3595, opened 2017-06-04
MinGW host

int19h:
These are the changes necessary on RTVS side to use the new MinGW-based R host.

Note that after this is merged, you will need to follow [these instructions](https://github.com/Microsoft/R-Host/blob/5c737e84cc800ebf7710ef38fe016e930e3f4409/BUILDING-WIN32.md) in order to set up the build environment for the host.


src/Package/Impl/Microsoft.VisualStudio.R.Package.csproj
```diff
@@ -1604,20 +1605,15 @@
       <VSIXSourceItem Include="$(OutputPath)\Microsoft.CodeAnalysis.CSharp.dll" />
       <VSIXSourceItem Include="$(OutputPath)\Microsoft.CodeAnalysis.dll" />
       <VSIXSourceItem Include="$(OutputPath)\libuv.dll" />
-      <VSIXSourceItem Include="$(OutputPath)\zip.dll" />
+      <VSIXSourceItem Include="$(OutputPath)\libzip-5.dll" />
+      <VSIXSourceItem Include="$(OutputPath)\libwinpthread-1.dll" />
```
int19h: @MikhailArkhipov We also need the corresponding change in setup.

int19h:
Should I go ahead and merge?

AlexanderSher:
👍

Issue #150, opened 2015-10-09
R Environment window stops working after you reset your session


huguesv:
You won't see any variables displayed after you've reset.  Restarting VS makes it work again.

PR #3604, opened 2017-06-06
Bug fixes

MikhailArkhipov:
- Fix F1 in `Tools | Options | R Tools | Advanced` - `#150` should be actual English string
- Fix occasional exception when  roaming settings are incompatible
- Remove `Install MRC` and 'Check for Updates` from 2017 as former is part of VS setup and latter doesn't make sense
- Update VS SDK tools to shipped


AlexanderSher:
👍

PR #3629, opened 2017-06-14
Client side fixes needed for help.

karthiknadig:


src/Linux/Host/Broker/Impl/Services/LinuxRHostProcessService.cs
```diff
@@ -45,78 +45,32 @@ class LinuxRHostProcessService : IRHostProcessService {
...
+        private void SetEnvironmentVaraibleIfAvailable(Dictionary<string,string> environment, string key) {
             var value = Environment.GetEnvironmentVariable(key);
-            if (string.IsNullOrEmpty(value) && !_defaultEnvironment.TryGetValue(key, out value)) {
-                return string.Empty;
+            if (!string.IsNullOrEmpty(value)) {
+                environment.Add(key, value);
             }
-            return value;
         }
-        private static readonly IDictionary<string, string> _defaultEnvironment = new Dictionary<string,string>(){
```
AlexanderSher: So there will be no defaults?

int19h: We shouldn't be in the business of providing any of these - if they're not inherited from the parent environment, R itself chooses the appropriate defaults (which may well be different).

PR #3635, opened 2017-06-16
Bunch of fixes

### LINKED ISSUE #3349: Help/Shiny does not work on linux
I had to change the HTTP request handler. Something might have broken during the change.

karthiknadig:
fixes #3349.
Few other bugs discovered while testing.

### DIFF
--- a/src/Host/Broker/Impl/Microsoft.R.Host.Broker.csproj
+++ b/src/Host/Broker/Impl/Microsoft.R.Host.Broker.csproj
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="1.1.1" />
     <PackageReference Include="System.Net.NetworkInformation" Version="4.3.0" />
     <PackageReference Include="System.Net.Requests" Version="4.3.0" />
+    <PackageReference Include="System.Net.WebSockets.Client" Version="4.3.1" />
     <PackageReference Include="System.ValueTuple" Version="4.3.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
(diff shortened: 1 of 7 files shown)


src/Host/Broker/Impl/Microsoft.R.Host.Broker.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="1.1.1" />
     <PackageReference Include="System.Net.NetworkInformation" Version="4.3.0" />
     <PackageReference Include="System.Net.Requests" Version="4.3.0" />
+    <PackageReference Include="System.Net.WebSockets.Client" Version="4.3.1" />
```
MikhailArkhipov: Does it need to be added to `<VSIXSourceItem>` in package as well as to Wix (remote setup) and `swix` in RTVS-pr?

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
```diff
@@ -22,27 +23,17 @@ public class RemoteUriHelper {
...
                     }
 
-                    var response = await client.SendAsync(request);
-                    HttpStatusCode statusCode = response.StatusCode;
-                    if (statusCode != HttpStatusCode.SwitchingProtocols) {
-                        response.Dispose();
-                    } else {
-                        string respSubProtocol = response.Headers.GetValues(Constants.Headers.SecWebSocketProtocol).FirstOrDefault();
-                        // TODO: match sub protocols.
+                    clientWebsocket.Options.KeepAliveInterval = TimeSpan.FromMinutes(10);
```
AlexanderSher: Can we write a test for this, or it will be too complicated?

src/Host/Broker/Impl/Microsoft.R.Host.Broker.csproj
karthiknadig: Doesn't look like it for this one. However, i did find one that was not added.

src/Host/Broker/Impl/RemoteUri/RemoteUriHelper.cs
karthiknadig: Work item to write tests for help and shiny: #3640

PR #3647, opened 2017-06-23
Fix crash and accessibility issue

MikhailArkhipov:
User and Watson reported.
Add accessible name to R argument field


karthiknadig:
👍

AlexanderSher:
👍

PR #3651, opened 2017-06-24
Signed .NET core build

MikhailArkhipov:
- obj path must be specified BEFORE inclusion of SDK targets
- rest should be specified AFTER SDK targets
- obj path depends on assembly name hence it needs to be specified on top
- assembly name must patch project file name
- organized loc/sign targets


src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
```diff
@@ -44,7 +44,7 @@
       <Project>{01e3d8c2-9c24-492b-b13b-c6d7068b6bf8}</Project>
       <Name>Microsoft.R.Common.Core.Windows</Name>
     </ProjectReference>
-    <ProjectReference Include="..\Impl\Microsoft.Common.Core.csproj">
+    <ProjectReference Include="..\Impl\Microsoft.R.Common.Core.csproj">
```
AlexanderSher: Maybe call it `Microsoft.Rtvs.Common.Core.csproj`, cause we use `.R.` for projects that have code related to R language.

src/R.Settings.Paths.props
```diff
@@ -0,0 +1,10 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <!-- Common Build Settings and Configurations -->
+    <PropertyGroup>
+        <RootDirectory Condition="'$(RootDirectory)' == ''">$(MSBuildThisFileDirectory)..\</RootDirectory>
```
AlexanderSher: Since `AssemblyName` is critical, we can add validation target. Something like:
```cs
    <Target Name = "ValidateRSettingsPaths" >
        <Error Condition = " '$(AssemblyName)'=='' " Text = "AssemblyName property must be specified before R.Settings.Paths.props is referenced" />
    </Target>
    <PropertyGroup>
        <BuildDependsOn>
            ValidateRSettingsPaths;
            $(BuildDependsOn)
        </BuildDependsOn>
        <BaseIntermediateOutputPath>$(ObjDirectory)\$(AssemblyName)\</BaseIntermediateOutputPath>
```

src/R.Settings.Paths.props
```diff
@@ -0,0 +1,10 @@
...
+        <ObjDirectory>$(RootDirectory)obj\</ObjDirectory>
```
AlexanderSher: `<ObjDirectory Condition="'$(ObjDirectory)' == ''">$(RootDirectory)obj\</ObjDirectory>`

src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
MikhailArkhipov: It is common to VSC so I'd avoid RTVS in names

AlexanderSher: Ok

src/R.Settings.Paths.props
MikhailArkhipov: OK

src/R.Settings.Paths.props
MikhailArkhipov: Done

AlexanderSher:
👍

PR #3662, opened 2017-06-29
Fix #3019: Way to navigate through history

### LINKED ISSUE #3019: Way to navigate through history
Hi!  I hate to say copy but please have a look at the history feature in RStudio.  It is vastly superior to having a separate history window.

It works within the R Interactive Prompt window.  Just type a few letters of a prior command, then press Ctrl-UpArrow.  A live popup appears with the history in order **and filtered** to commands that begin with the typed letters.  The last item highlighted ready to select.

This is a mouse-free solution, about 2 keystrokes, and all the action is from within the interactive window.  

I can only get about 30% of the same feeling in MSVS, I press Ctrl-4, and then tab 5 times to get the search box.  Type some letters. Press tab again only to land on the first item instead of the last.  Thenplay with down arrow or end.  Select it by pressing return (and then can't see it run) and then press Ctrl-2 to hide the history.  The next time the Ctrl-4 is pressed the tab sequence changes because it remembers the position in the list. 

If this is already available please provide the hotkey.   Thanks for your consideration!

P.S. I see Alt-UpArrow would work great if it would simply match against any already-typed letters.

AlexanderSher:


src/Common/Core/Impl/Extensions/StringExtensions.cs
```diff
@@ -22,6 +23,43 @@ public static class StringExtensions {
...
+                if (i < 0) {
+                    break;
+                }
+
+                yield return i;
+                i = allowOverlap ? i + value.Length : i + 1;
+            }
+        }
+
+        public static bool IsStartOfNewLine(this string s, int index, bool ignoreWhitespaces = false) {
```
MikhailArkhipov: Whitespace.IsNewLineBeforePosition might work too (class can be moved to Core if needed)

MikhailArkhipov:
👍

src/Common/Core/Impl/Extensions/StringExtensions.cs
AlexanderSher: There is a difference between those two methods for `position == 0`. `IsStartOfNewLine` will return `true`, while `IsNewLineBeforePosition` will return `false`.

Issue #3670, opened 2017-06-30
Replace standard ping with port ping


karthiknadig:
https://blogs.msdn.microsoft.com/mast/2014/06/22/use-port-pings-instead-of-icmp-to-test-azure-vm-connectivity/

Issue #3668, opened 2017-06-29
Unicode output appears broken


MikhailArkhipov:
See `UnicodeOutput` test

```
{#101# ! [
  " \"日本語\"\u0000\u0000\u0000"
] <blob content="" />}
```

Should not be sending trailing zeroes

karthiknadig:
@int19h I think doing a resize with the value returned by RString2Unicode should fix this. Any comments on that?
```
if (cch > 0) {
    ws.resize(cch);
    int nc = RString2Unicode(&ws[0], (char*)buf, len);
    ws.resize(nc);
}
```

int19h:
Yep, that sounds right.

PR #3672, opened 2017-06-30
Bunch of fixes.

karthiknadig:
#3670, #3668


src/Common/Core/Impl/Microsoft.R.Common.Core.csproj
```diff
@@ -36,6 +36,19 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Update="Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+  </ItemGroup>
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
   <Import Project="$(SourceDirectory)R.Build.Version.targets" />
   <Import Project="$(SourceDirectory)R.Build.Sign.targets" />
```
MikhailArkhipov: Should be Build.Loc target then since you've added resources

src/Common/Core/Impl/Net/Network.cs
```diff
@@ -18,20 +18,18 @@ public static class NetworkExtensions {
...
-                if (reply.Status != IPStatus.Success) {
-                    return reply.Status.ToString();
-                }
-            } catch (PingException pex) {
-                var pingMessage = pex.InnerException?.Message ?? pex.Message;
-                if (!string.IsNullOrEmpty(pingMessage)) {
-                    return pingMessage;
+                using (TcpClient pingClient = new TcpClient()) {
+                    var pingTask = pingClient.ConnectAsync(url.Host, url.Port);
+                    if (await Task.WhenAny(pingTask, Task.Delay(timeout)) == pingTask) {
```
MikhailArkhipov: Should we port this to Release/1_0

src/Common/Core/Impl/Microsoft.R.Common.Core.csproj
karthiknadig: Actually, I think i should move this to Client. It is used only there. and avoid creating a new Resources file.

MikhailArkhipov:
Let's port ping fix to Release
👍

src/Common/Core/Impl/Net/Network.cs
karthiknadig: Yes.

Issue #3621, opened 2017-06-13
r plot history multiselect support


myusrn:
Please briefly describe what you were doing that led to the issue, if applicable:

I find ability to scrub specific plots from r plot history view [ of "ide" graphics device ] very helpful as opposed to entering the dev.off() command in r interactive window approach that drops everything. Currently r plot history view doesn't support range multiselect, using shift+click, and arbitrary set selection, using ctrl+click, gestures for selecting set of plots to delete in one step. No idea whether enabling that involves use of some boiler plate UI code or complex view pane specific addition. Either way if others also feel it would be a nice addition to variable explorer interactions then perhaps worth considering.

If you have any screenshots demonstrating the issue, please include them as well to help us diagnose it better.

Additional information:

OS Information
    Version:       Microsoft Windows NT 10.0.15063.0

RTVS Information:
    Assembly: Microsoft.VisualStudio.R.Package, Version=1.1.30523.1233, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a

Active R URI:
    R 3.3.2: C:\Program Files\R\R-3.3.2

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/85238e4dd8a124f39674a19a60075d7b10da00bc

myusrn:
thanks, look forward to trying this out when next update is pushed with this change in it.

PR #3683, opened 2017-07-05
Plot history multiselect

MikhailArkhipov:
#3621


src/Windows/R/Components/Impl/Plots/Implementation/Commands/PlotHistoryCommand.cs
```diff
@@ -2,17 +2,50 @@
...
-        public PlotHistoryCommand(IRInteractiveWorkflow interactiveWorkflow, IRPlotHistoryVisualComponent visualComponent) :
+        protected PlotHistoryCommand(IRInteractiveWorkflow interactiveWorkflow, IRPlotHistoryVisualComponent visualComponent) :
             base(interactiveWorkflow) {
             Check.ArgumentNull(nameof(visualComponent), visualComponent);
             VisualComponent = visualComponent;
         }
+
+        public virtual CommandStatus Status {
+            get {
+                var selection = VisualComponent.SelectedPlots.ToArray();
```
AlexanderSher: `PlotHistoryAutoHideCommand`, `PlotHistoryZoomInCommand` and `PlotHistoryZoomOutCommand` have no dependency on selection. I think there should be a separate base class for `PlotHistoryActivateCommand`, `PlotHistoryRemoveCommand` and `PlotHistoryCutCopyCommand`. Something like `PlotHistorySelectionCommand`.

src/Windows/R/Components/Impl/Plots/Implementation/Commands/PlotHistoryCommand.cs
```diff
@@ -2,17 +2,50 @@
...
+
+        protected virtual bool CanInvoke() => true;
+
+        protected virtual Task InvokeAsync(IRPlot plot) => Task.CompletedTask;
+
+        public virtual async Task InvokeAsync() {
+            var selection = VisualComponent.SelectedPlots.ToArray();
+            if (selection.Length > 0 && CanInvoke()) {
+                try {
+                    foreach (var plot in selection) {
```
AlexanderSher: All these 3 commands (`PlotHistoryActivateCommand`, `PlotHistoryRemoveCommand` and `PlotHistoryCutCopyCommand`) call our own functions in R: `rtvs:::graphics.ide.selectplot`, `rtvs:::graphics.ide.removeplot`, `rtvs:::graphics.ide.copyplot`. Instead of iterating through them inside command, I'd rather change signature of those functions so they can accept arrays.

MikhailArkhipov: Change would be much more extensive including bunch of tests in `IdeGraphicsDeviceTest`. Performance is probably not an issue here.

src/Windows/R/Components/Impl/Plots/Implementation/Commands/PlotHistoryCommand.cs
MikhailArkhipov: OK

src/Common/Core/Impl/Microsoft.R.Common.Core.csproj
```diff
@@ -8,6 +8,7 @@
   <Import Project="..\..\..\R.Settings.props" />
   <ItemGroup>
     <Compile Include="$(SourceDirectory)GlobalAssemblyInfo.cs" Link="Properties\GlobalAssemblyInfo.cs" />
+    <Compile Include="..\..\..\AssemblyVersionInfo.cs" Link="Properties\AssemblyVersionInfo.cs" />
```
AlexanderSher: Why `$(SourceDirectory)GlobalAssemblyInfo.cs` doesn't work?

src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
```diff
@@ -50,6 +50,9 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\..\..\AssemblyVersionInfo.cs">
+      <Link>Properties\AssemblyVersionInfo.cs</Link>
```
AlexanderSher: Single-line declaration (like in `Microsoft.R.Common.Core.csproj`) looks better

MikhailArkhipov: Sure. This is just what VS generated

src/Common/Core/Impl/Microsoft.R.Common.Core.csproj
MikhailArkhipov: I'll replace manually. This generated part. Currently we generate for every assembly but it is really unnecessary. Also, adding it to `Compile` works in regular project but produces `File specified twice` in .NET Core since latter picks up all files by default.

src/Package/Impl/Packages/R/RPackage.cs
```diff
@@ -140,7 +140,10 @@ internal class RPackage : BasePackage<RLanguageService>, IRPackage {
 
             LogCleanup.Cancel();
             ProjectIconProvider.Close();
-            CsvAppFileIO.Close(Services.FileSystem());
+
+            if (Services?.FileSystem() != null) {
```
AlexanderSher: I think it is better to store reference to `Services?.FileSystem()` inside `Initialize`. `Services` may be disposed here already.

MikhailArkhipov: This was one-off crash when package failed to load and got disposed before services were actually populated. Sure, I can store it.

src/Common/Core/Test/Microsoft.Common.Core.Test.csproj
MikhailArkhipov: There is like 50 places, and no guarantee VS won't change it back

AlexanderSher:
👍

Issue #3639, opened 2017-06-19
Linux: Restrict users by Group


karthiknadig:
We do this check currently on windows and not on Linux.

Issue #3661, opened 2017-06-28
object 'bringToTop' of mode 'function' was not found


karthiknadig:
Error message seen when connected to a linux machine. 

```
Error in get(function_name, package_spec, mode = "function") : 
  object 'bringToTop' of mode 'function' was not found
```

This does not block working with linux.

PR #3684, opened 2017-07-05
Check user group in linux and minor fixes

karthiknadig:
Fixes: #3639, #3661

### DIFF
--- a/src/Linux/Host/Broker/Impl/Services/Utility.cs
+++ b/src/Linux/Host/Broker/Impl/Services/Utility.cs
@@ -46,7 +46,7 @@ namespace Microsoft.R.Host.Broker.Services {
             return proc;
         }
 
-        public static bool AuthenticateUser(ILogger<IAuthenticationService> logger, IProcessServices ps,  string username, string password, out string profileDir) {
+        public static bool AuthenticateUser(ILogger<IAuthenticationService> logger, IProcessServices ps,  string username, string password, string allowedGroup, out string profileDir) {
             bool retval = false;
             Process proc = null;
             string userDir = string.Empty;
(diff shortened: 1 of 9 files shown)


src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -46,15 +46,15 @@ public class Utility {
             return proc;
         }
 
-        public static bool AuthenticateUser(ILogger<IAuthenticationService> logger, IProcessServices ps,  string username, string password, out string profileDir) {
+        public static bool AuthenticateUser(ILogger<IAuthenticationService> logger, IProcessServices ps,  string username, string password, string allowedGroup, out string profileDir) {
```
MikhailArkhipov: Looks like it could belong to `IAuthenticationService`

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -31,13 +31,15 @@ static constexpr int RTVS_AUTH_OK           = 0;
 static constexpr int RTVS_AUTH_INIT_FAILED = 200;
 static constexpr int RTVS_AUTH_BAD_INPUT   = 201;
 static constexpr int RTVS_AUTH_NO_INPUT    = 202;
+static constexpr int RTVS_AUTH_NOT_ALLOWED = 203;
```
MikhailArkhipov: Could we use there constants up above in the switch instead of `case 203:`

src/Linux/Host/Broker/Impl/Services/Utility.cs
karthiknadig: That is actually called by `IAuthenticationService` implementation for Linux, i.e., `LinuxAuthenticationService`.

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -438,8 +434,55 @@ int authenticate_and_run(const picojson::object& json) {
...
+            if (!gp) {
+                err = errno;
+                logf(log_verbosity::minimal, "Error [getgrnam]:[%d] %s\n", err, strerror(err));
+                return err;
+            }
+
+            gid_t allowed_gid = gp->gr_gid;
+
+            int ngroups = 1000;
+            gid_t *user_groups = calloc_or_exit<gid_t*>(ngroups, sizeof(gid_t));
```
int19h: It looks like this is freed immediately in the same block, and no other function takes ownership, so this should probably just be `std::vector<gid_t*>`.

PR #3685, opened 2017-07-05
Fixes issue where broker user cannot kill process running as another user.

### LINKED ISSUE #3600: Linux: RHost process does not get killed after disconnect

karthiknadig:
fixes #3600

### DIFF
--- a/src/Linux/Core/Impl/OS/UnixProcess.cs
+++ b/src/Linux/Core/Impl/OS/UnixProcess.cs
@@ -4,6 +4,9 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Common.Core.OS {
     public class UnixProcess : IProcess {
@@ -41,11 +44,16 @@ namespace Microsoft.Common.Core.OS {
         }
 
         private static void KillProcess(IProcessServices ps, int pid) {
-            ProcessStartInfo psi = new ProcessStartInfo();
-            psi.FileName = UnixProcessServices.RunAsUserBinPath;
-            psi.Arguments = "-k {pid}";
-            Process proc = ps.Start(psi);
-            proc.WaitForExit(1000);
+            KillProcessMessage kpm = new KillProcessMessage() { ProcessId = pid };
+            string json = JsonConvert.SerializeObject(kpm, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+            Process proc = UnixProcessServices.CreateRunAsUserProcess(ps, true);
+            using (BinaryWriter writer = new BinaryWriter(proc.StandardInput.BaseStream, Encoding.UTF8, true)) {
+                var jsonBytes = Encoding.UTF8.GetBytes(json);
+                writer.Write(jsonBytes.Length);
+                writer.Write(jsonBytes);
+                writer.Flush();
+                proc.WaitForExit(1000);
+            }
         }
     }
 }
--- a/src/Linux/Core/Impl/OS/UnixProcess.cs
+++ b/src/Linux/Core/Impl/OS/UnixProcess.cs
@@ -2,16 +2,15 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
-using System.Text;
 
 namespace Microsoft.Common.Core.OS {
     public class UnixProcess : IProcess {
         private readonly Process _process;
-        public UnixProcess(Process process) {
+        private readonly IProcessServices _ps;
+        public UnixProcess(IProcessServices ps, Process process) {
+            _ps = ps;
             _process = process;
         }
 
@@ -36,9 +35,17 @@ namespace Microsoft.Common.Core.OS {
             }
         }
 
-
         public void Kill() {
-            _process.Kill();
+            // This is needed because broker user cannot kill process running as another user.
+            KillProcess(_ps, _process.Id);
+        }
+
+        private static void KillProcess(IProcessServices ps, int pid) {
+            ProcessStartInfo psi = new ProcessStartInfo();
+            psi.FileName = UnixProcessServices.RunAsUserBinPath;
+            psi.Arguments = "-k {pid}";
+            Process proc = ps.Start(psi);
+            proc.WaitForExit(1000);
         }
     }
 }
--- a/src/Linux/Core/Impl/OS/UnixProcess.cs
+++ b/src/Linux/Core/Impl/OS/UnixProcess.cs
@@ -4,9 +4,6 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Text;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Common.Core.OS {
     public class UnixProcess : IProcess {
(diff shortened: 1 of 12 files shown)


src/Linux/Core/Impl/OS/UnixProcessServices.cs
```diff
@@ -20,5 +20,12 @@ public class UnixProcessServices : IProcessServices {
         public Process Start(string path) {
             return Process.Start(path);
         }
+
+        // usage:
+        // Microsoft.R.Host.RunAsUser [-q]
+        //    -q: Quiet
+        //    -k <pid> : kill process
+        private const string runAsUserPath = "/usr/lib/rtvs/Microsoft.R.Host.RunAsUser";
```
int19h: Why not just `public const`?

src/Linux/Core/Impl/OS/UnixProcess.cs
```diff
@@ -36,9 +35,17 @@ public class UnixProcess : IProcess {
...
         public void Kill() {
-            _process.Kill();
+            // This is needed because broker user cannot kill process running as another user.
+            KillProcess(_ps, _process.Id);
+        }
+
+        private static void KillProcess(IProcessServices ps, int pid) {
+            ProcessStartInfo psi = new ProcessStartInfo();
+            psi.FileName = UnixProcessServices.RunAsUserBinPath;
+            psi.Arguments = "-k {pid}";
```
int19h: It sounds like it could use RAU to run `kill -9 <pid>`?

karthiknadig: It could. Is there any reason it should not call `kill(3)`?

int19h: What I meant is that then we won't need the `-k` switch at all, so RAU code remains single-purpose (launching elevated processes).

karthiknadig: Ahh! Got it. I will update this.

karthiknadig:
ping

src/Linux/Core/Impl/OS/UnixProcess.cs
```diff
@@ -36,9 +38,22 @@ public class UnixProcess : IProcess {
...
 
-
         public void Kill() {
-            _process.Kill();
+            // This is needed because broker user cannot kill process running as another user.
+            KillProcess(_ps, _process.Id);
+        }
+
+        private static void KillProcess(IProcessServices ps, int pid) {
+            KillProcessMessage kpm = new KillProcessMessage() { ProcessId = pid };
```
MikhailArkhipov: You may want to try-catch IOException here

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -294,13 +297,15 @@ int run_rhost(const picojson::object& json, const char* user, const gid_t gid, c
             if (err) {
                 logf(log_verbosity::minimal, "Error rhost exited:[%d] %s\n", err, strerror(err));
             } else {
-                logf(log_verbosity::minimal, "RHost exited.\n");
+                logf(log_verbosity::minimal, "RHost exited normally.\n");
             }
         } else if (WIFSIGNALED(ws)) {
             logf(log_verbosity::minimal, "Error rhost terminated by a signal: %d\n", WTERMSIG(ws));
```
MikhailArkhipov: Not related to this change, but in log messages accessible to the user I'd put `Microsoft.R.Host.exe` rather than `rhost` since no one out there knows what `rhost` means

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -475,7 +480,60 @@ int authenticate_and_run(const picojson::object& json) {
...
+    int err = 0;
+    char** args = calloc_or_exit<char**>(4, sizeof(char*));
+    std::string arg1("-9"); // SIGKILL
+    char carg2[100] = {}; // pid
+    sprintf(carg2, "%d", pid);
+    std::string arg2(carg2); // SIGKILL
+
+    std::string kill_path(RTVS_KILL_PATH);
+    args[0] = calloc_or_exit<char*>(kill_path.size() + 1, sizeof(char));
+    memcpy(args[0], kill_path.c_str(), kill_path.size());
```
MikhailArkhipov: no strdup equivalent?

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -475,7 +480,60 @@ int authenticate_and_run(const picojson::object& json) {
...
+    memcpy(args[2], arg2.c_str(), arg2.size());
+
+    args[3] = NULL;
+
+    int pid = fork();
+    if (pid == -1) {
+        err = errno;
+        logf(log_verbosity::minimal, "Error [fork]: %s\n", explain_errno_fork(err));
+        return err;
+    } else if (pid == 0) {
```
MikhailArkhipov: `else` is unnecessary

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -475,7 +480,60 @@ int authenticate_and_run(const picojson::object& json) {
...
+            logf(log_verbosity::minimal, "Error kill terminated by a signal: %d\n", WTERMSIG(ws));
+            err = ws;
+        }
+    }
+
+    return err;
+}
 int main(int argc, char **argv) {
     int option = 0;
     bool quiet = false;
```
MikhailArkhipov: just `quiet = option == 'q'` ?

src/Linux/Native/RunAsUser/src/main.cpp
karthiknadig: Will try with `strdup`. I think it should work.

src/Linux/Core/Impl/OS/UnixProcessServices.cs
```diff
@@ -20,5 +20,21 @@ public class UnixProcessServices : IProcessServices {
         public Process Start(string path) {
             return Process.Start(path);
         }
+
+        // usage:
+        // Microsoft.R.Host.RunAsUser [-q]
+        //    -q: Quiet
+        private const string RunAsUserBinPath = "/usr/lib/rtvs/Microsoft.R.Host.RunAsUser";
+
+        public static Process CreateRunAsUserProcess(IProcessServices ps, bool quietMode) {
```
AlexanderSher: Why is it static?

src/Linux/Core/Impl/OS/UnixProcessServices.cs
```diff
@@ -20,5 +20,21 @@ public class UnixProcessServices : IProcessServices {
         public Process Start(string path) {
             return Process.Start(path);
         }
+
+        // usage:
+        // Microsoft.R.Host.RunAsUser [-q]
+        //    -q: Quiet
+        private const string RunAsUserBinPath = "/usr/lib/rtvs/Microsoft.R.Host.RunAsUser";
```
AlexanderSher: Can this path be used with other processes?

karthiknadig: This path has meaning only to broker and not other processes.

AlexanderSher: Then we should move it back to the Linux Broker assembly.

src/Linux/Native/RunAsUser/src/main.cpp
```diff
@@ -49,8 +50,10 @@ static constexpr char RTVS_RESPONSE_TYPE_RTVS_ERROR[] = "rtvs-error";
 
 static constexpr char RTVS_MSG_AUTH_ONLY[] = "AuthOnly";
 static constexpr char RTVS_MSG_AUTH_AND_RUN[] = "AuthAndRun";
+static constexpr char RTVS_MSG_KILL_PROCESS[] = "KillProcess";
 
 static constexpr char RTVS_RHOST_PATH[] = "/usr/lib/rtvs/Microsoft.R.Host";
+static constexpr char RTVS_KILL_PATH[] = "/bin/kill";
```
int19h: Sorry, I should have clarified what I meant here.

If we pass the binary as an argument in the JSON message, instead of hardcoding, then this thing doesn't need to know what it is starting. And then you can pass the host binary or kill or whatever from the managed side, and there's no special code for kill.

(And if this doesn't work for some reason, then there's no point using kill as a process instead of the corresponding API.)

karthiknadig: That was intentional. I wanted to restrict it to specific processes that we use, rather than make it general purpose.

int19h: It feels like it would be better as a more generic API, so that the broker can use it more flexibly as and when it needs to. What's the downside? If it's mainly security, then, given that the host can run arbitrary code (including spawning arbitrary child processes), I think it's moot anyway.

karthiknadig: I had Created https://github.com/Microsoft/RTVS/issues/3702 to track that.

PR #3689, opened 2017-07-06
Default code page fix +  crash fix

MikhailArkhipov:
Leave code page alone, R sets it correctly on its own (tried on Spanish Windows Language pack)
Fix crash in parameter-less signatures

### DIFF
--- a/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
+++ b/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
@@ -222,11 +222,13 @@ grDevices::deviceIsInteractive('ide')
 
         public static async Task SetCodePageAsync(this IRExpressionEvaluator evaluation, int codePage, CancellationToken cancellationToken = default(CancellationToken)) {
             var cp = $".{codePage}";
-            if (codePage == 0) {
-                cp = await evaluation.IsRSessionPlatformWindowsAsync() ? ".437" : "en_US.UTF-8";
+            if (codePage == 0 && !await evaluation.IsRSessionPlatformWindowsAsync(cancellationToken)) {
+                cp = "en_US.UTF-8";
             }
-            var script = Invariant($"Sys.setlocale('LC_CTYPE', '{cp}')");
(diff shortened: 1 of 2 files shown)


src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -222,11 +222,13 @@ public static Task OptionsSetWidthAsync(this IRExpressionEvaluator evaluation, i
 
         public static async Task SetCodePageAsync(this IRExpressionEvaluator evaluation, int codePage, CancellationToken cancellationToken = default(CancellationToken)) {
             var cp = $".{codePage}";
-            if (codePage == 0) {
-                cp = await evaluation.IsRSessionPlatformWindowsAsync() ? ".437" : "en_US.UTF-8";
+            if (codePage == 0 && !await evaluation.IsRSessionPlatformWindowsAsync(cancellationToken)) {
+                cp = "en_US.UTF-8";
```
int19h: I'm confused about how this works. So now if this function gets called with codePage==0, we try to set LC_CTYPE to `".0"` - but that's not a valid codepage, is it?

MikhailArkhipov: Whoops, wrong check...

int19h:
I wonder why we we changing the default in the first place - surely that was for some encoding-related reason. Does it do the right thing now with e.g. Russian?

MikhailArkhipov:
I set Spanish language pack and R locale was as user wanted. Tried setwd-getwd with Unicode path (Russian) on Spanish UI? worked fine. I think we used to set OEM b/c technically R is non-Unicode program. OEM should work most of the time, I think. I can try Russian lang pack too.

MikhailArkhipov:
Russian works fine, publishing markdown to folder named in Unicode works fine too (we switched to short paths anyways a while ago). I think the only think OEM might affect is reading/writing files without specifying encoding, but it is up to the user to specify correct one.

Issue #3695, opened 2017-07-07
Drag-drop fails with: Cannot complete the drag and drop operation


karthiknadig:
---------------------------
Microsoft Visual Studio
---------------------------
Cannot complete the drag and drop operation: Invalid FORMATETC structure (Exception from HRESULT: 0x80040064 (DV_E_FORMATETC)).
---------------------------
OK   
---------------------------

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/333d10287a8d7cebdbec9b2da4824a2727027054

PR #3700, opened 2017-07-10
 #3695 Drag-drop fails with: Cannot complete the drag and drop operation

MikhailArkhipov:
The issue was that some code still assumed `string` rather than `string[]`

Consolidated data object operations in one place so they are consistent across code and use same data representation. Also fixed multiselect (right click unselected elements)

### DIFF
--- a/src/Windows/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
+++ b/src/Windows/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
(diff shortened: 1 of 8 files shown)


src/Windows/R/Components/Impl/Plots/Implementation/View/RPlotHistoryControl.xaml.cs
```diff
@@ -61,24 +62,36 @@ private void ActivatePlot(IRPlotHistoryEntryViewModel entry)
...
+            }
+
+            var control = Keyboard.IsKeyDown(Key.LeftCtrl) || Keyboard.IsKeyDown(Key.RightCtrl);
+            if (!control) {
+                // Regular click clear selection then reselects
+                HistoryListView.SelectedItems.Clear();
+            }
+        }
+
+        private void HistoryListView_PreviewMouseLeftButtonUp(object sender, MouseButtonEventArgs e) => e.Handled = true;
```
AlexanderSher: Why do you need this?

MikhailArkhipov: List view has this odd 'I am a checkbox' behavior. Click selects element and then more elements (without Control pressed). One of the fixes is to make sure plain click selects primary element and ctrl+click adds/removes. So LBUp behavior is overridden just in case.

AlexanderSher:
👍

Issue #3693, opened 2017-07-07
Live Markdown: pasting over R blocks disconnects session


MikhailArkhipov:
Default RMD
Select first block and paste over with

```{r}
boxplot(1:10~rep(1:2,5))
plot(x, y)
```

![image](https://user-images.githubusercontent.com/12820357/27976162-300cd7d0-631a-11e7-8a97-93da09d18a31.png)

MikhailArkhipov:
May be related to #3694

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/a41b8bdafe14223acd780a936a23ca6c2345f964

Issue #3692, opened 2017-07-07
Some Lint errors are reported twice


MikhailArkhipov:
![image](https://user-images.githubusercontent.com/12820357/27939575-a141fba6-627a-11e7-99b1-ed84e81929eb.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/a41b8bdafe14223acd780a936a23ca6c2345f964

PR #3704, opened 2017-07-10
#3692, #3693

MikhailArkhipov:
#3693 Live Markdown: pasting over R blocks disconnects session 
- call `Execute` rather than `BeginInteraction`

#3692 Some Lint errors are reported twice 
- changes in lint options were not triggering re-validation
- tagger assumed that changes to whitespace cannot change squiggles

Apparently we had two `BindableBase`, removed one and moved another down to Core since it is not really WPF related and is useful in other scenarios.

The change is on top on https://github.com/Microsoft/RTVS/pull/3700 so ignore plotting changes


AlexanderSher:
> it is not really WPF related and is useful in other scenarios

Hm, I didn't know we use `INotifyPropertyChanged` in non-WPF scenarios.

src/Windows/R/Editor/Impl/Settings/REditorSettings.cs
```diff
@@ -24,15 +26,28 @@ public sealed class REditorSettings: EditorSettings, IWritableREditorSettings, I
...
         public override void ResetSettings() {
             _formatOptions = new RFormatOptions();
-            LintOptions = new LintOptions();
+            CreateLintOptions();
             base.ResetSettings();
         }
 
+        private void CreateLintOptions() {
+            LintOptions = new LintOptions();
+            LintOptions.PropertyChanged += _OnLintOptionsPropertyChanged;
```
MikhailArkhipov: I didn't want to use that storage since we need to clean it up. The problem is that in VS setting storage is a bit of a mess. Editor settings (at least some of them) have to use language storage since things like tab/space/indent/enable completion are provided by VS and have to be read from specific VS storage. However, options like Lint are stored in separately via DialogPage. Now, connections are stored in the roaming store... 

At some point I'd like to reduce language storage use (aka WritableStorage) to only settings that are provided/controlled by VS editor tab.

AlexanderSher: I thought that `ISettingStorage`/`IWritableSettingsStorage` is an abstraction that has several implementation, each of those works with specific type of storing: local storage, roaming storage, etc. And types like `REditorSettings` use them to save/load/provide access to a specific settings.
If you say that `IWritableSettingsStorage` is for editor only, how do we save/load settings from DialogPage?

MikhailArkhipov: There are three...

`ISettingsStorage` (which is actually writable) is implemented over roaming store (`SVsSettingsPersistenceManager`). This is what we *partially* use for `IRSettings`. Mostly for settings we store manually (without dialog page) such as connections, most recently used settings in SQL publishing etc.

There is `IEditorSettingsStorage` which is implemented by `LanguageSettingsStorage` over `SVsTextManager` and `LANGPREFERENCES3`

DialogPage settings are loaded by package by calling `GetAutomationObject` on every option page we have. This loads settings from virtual registry key specified in the dialog page and sets appropriate fields in IRSettings.


It is actually `IWritableEditorSettingsStorage`

MikhailArkhipov: Now, we could map everything to `ISettingsStorage` (roaming) except few editor settings which could filter by name (indent, tabs, syntax check etc - see 

image

But IDE folks don't like extensions to roam on their own. Also, we'd need to participate in import/export settings manually - DialogPage implementation does it for us currently.

AlexanderSher: Ok, so there are three different interfaces which can't be abstracted by one because of the way data is saved and loaded.

MikhailArkhipov: Well anything can be abstracted :-) question is now much work is there...

MikhailArkhipov: I do want to abstract this properly, we need to discuss exactly how.

AlexanderSher: I still can't find how properties from `RSettingsImplementation` are saved and restored. Can you point to the code?

MikhailArkhipov: Sure. Anything exposed to the user in Tools | Options (ex `LoadRDataOnProjectLoad` goes via `ISettingsStorage` (roaming) - see `RSettingsImplementation` ctor and `_settingStorage`. Package calls `LoadSettings` in `ConfigurePackageServices`

Tools | Options dialog page overrides `LoadSettingsFromStorage` and `SaveSettingsToStorage` so `DialogPage` code is not used.

MikhailArkhipov: So theoretically storages can be reduced to two: lang preferences for the common editor stuff and roaming. Basically override DialogPage storage and map it to roaming. Import/Export settings won't work but since it is a roaming storage settings will travel. But requires logging on to VS account since otherwise roaming won't work...

AlexanderSher: So right now it looks like this:

| Storage | Value saved | Value loaded |
| --- | --- | --- |
| Editor  | Every time property value changed | When `LanguageSettingsStorage.Load` is called |
| DialogPage | When `SettingsHolder.Apply` is called from `RToolsOptionsPage.OnApply` | When `RToolsOptionsPage` instance is created |
| Roaming | When `VsSettingsStorage.PersistAsync` is called from `RSettingsImplementation.SaveSettingsAsync` | Every time property value requested and not cached |

Is it correct?

MikhailArkhipov: Sounds about right

AlexanderSher: Ok, first of all, I think I've found a bug in `SettingsHolder`:

```cs
            public T GetValue<T>(T defaultValue, [CallerMemberName] string name = null) {
                object value;
                return _dict.TryGetValue(name, out value) ? (T)value : default(T);
            }
```

`defaultValue` isn't used, it should probably be 

```cs
            public T GetValue<T>(T defaultValue, [CallerMemberName] string name = null) 
                => _dict.TryGetValue(name, out object value) ? (T)value : defaultValue;
```

AlexanderSher: Back to storages (I'm still trying to understand the whole picture). `IRSettings` implements `INotifyPropertyChanged`, which is used to update `RSession` code page and CRAN mirror, but isn't used to trigger property saving. `DialogPage` doesn't save anything directly, it saves values through `IRSettings.SaveSettingsAsync`, so in fact we have two different places to save values: `ISettingsStorage` (implemented by `VsSettingsStorage`) and `IWritableEditorSettingsStorage` (implemented by `LanguageSettingsStorage`).

So where do you want to store `LintOptions` - in `ISettingsStorage` or in `IWritableEditorSettingsStorage`?

MikhailArkhipov: Doesn't matter really but since it is editor-related, may map to the `LanguageSettingsStorage` then

AlexanderSher: Then `LintOptions` can follow the same pattern as `REditorSettings`. `RLintOptionsDialog` already follows `RLintOptionsDialog` pattern, so `LintOptions` can use `Storage` and `WritableStorage`, something like this:
```cs
    public sealed class LintOptions {
        public const string EnabledKey = nameof(LintOptions) + ":" + nameof(Enabled);

        private readonly IEditorSettingsStorage _storage;
        private readonly IWritableEditorSettingsStorage _writableStorage;

        public class LintOptions (IEditorSettingsStorage storage) {
            _storage = storage;
            _writableStorage = storage as IWritableEditorSettingsStorage;
        }

        public bool Enabled {
            get => Storage.Get(EnabledKey, true);
            set => WritableStorage?.Set(EnabledKey, value);
        }
```

MikhailArkhipov: Sure, OK

src/R/Editor/Impl/Validation/Lint/LintOptions.cs
```diff
@@ -1,60 +1,109 @@
...
 
+using System;
+using Microsoft.Common.Core.Diagnostics;
+using Microsoft.Languages.Editor.Settings;
+using Microsoft.R.Common.Core.UI;
+using static System.FormattableString;
+
 namespace Microsoft.R.Editor.Validation.Lint {
-    public sealed class LintOptions {
+    public sealed class LintOptions: BindableBase {
```
AlexanderSher: I think it doesn't have to be `BindableBase`  anymore

src/Common/Core/Impl/UI/BindableBase.cs
```diff
@@ -5,7 +5,7 @@
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 
-namespace Microsoft.Common.Wpf {
+namespace Microsoft.R.Common.Core.UI {
```
AlexanderSher: I'm not a big fan of using `INotifyPropertyChanged` in non-UI context, but if you want it to be in `Common.Core`, I think we should rename it into `PropertyChangedBase` or something like that and change its namespace.

src/R/Editor/Impl/Validation/Lint/LintOptions.cs
```diff
@@ -1,60 +1,109 @@
...
+
+        private IEditorSettingsStorage Storage => _storage ?? (_storage = _storageAccess());
+        private IWritableEditorSettingsStorage WritableStorage => Storage as IWritableEditorSettingsStorage;
+
+        public LintOptions(Func<IEditorSettingsStorage> storageAccess) {
+            Check.ArgumentNull(nameof(storageAccess), storageAccess);
+            _storageAccess = storageAccess; // Storage is delay-created after package loads
+        }
+
+        private string GetKey(string optionName) => Invariant($"LintR_{optionName}");
```
AlexanderSher: If you want you can define two methods (`Get` and `Set`) and in this case you don't have to use `nameof`:

```cs
        private T Get<T>(T defaultValue, [CallerMemberName] string name = null) => Storage.Get(GetKey(name), defaultValue);
        private void Set<T>(T value, [CallerMemberName] string name = null) => WritableStorage?.Set(GetKey(name), value);

        public bool Enabled {
            get => Get(false);
            set => Set(value);
        }
```

Up to you though

Issue #3703, opened 2017-07-10
Disable network load on WSL


karthiknadig:
This issue in dot net core causes a crash on WSL when we do network load check.
https://github.com/dotnet/corefx/issues/22048

PR #3705, opened 2017-07-10
Fix for WSL specific .net core issue.

karthiknadig:
#3703


MikhailArkhipov:
👍 

May want to open a task to restore the code when the said bug is fixed and reference the task in the comment.

Issue #3717, opened 2017-07-17
Lint complains about ::


MikhailArkhipov:
Need to special-case certain operators and also turn off lint in REPL

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/bb097aa92ed78b001d738ab73fac190c976d0d51

Issue #3715, opened 2017-07-15
Vs Hangs When streaming tweets


Crackz:
it hangs forever i'm using vs2017 latest version

![capture](https://user-images.githubusercontent.com/12835589/28243079-c0094e22-69be-11e7-9ca7-f13676917a40.PNG)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/bb097aa92ed78b001d738ab73fac190c976d0d51

Issue #3691, opened 2017-07-07
Lint complaints about braces in markdown block parameters


MikhailArkhipov:
![image](https://user-images.githubusercontent.com/12820357/27939555-8867248a-627a-11e7-9642-2091c3620481.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/bb097aa92ed78b001d738ab73fac190c976d0d51

PR #3721, opened 2017-07-19
3691, 3715, 3717

MikhailArkhipov:
#3715 Vs Hangs When streaming tweets 
 - rewrote carriage return processor to use queue rather than `IMainThread.Post`

#3717 Lint complains about `::`
- Added special case for operators that do not need spaces around them 

#3691 Lint complaints about braces in markdown block parameters 
- Special cased `{r` block
- Plumbed through projection status to suppress trailing blank lines check

Added tests


src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
```diff
@@ -0,0 +1,198 @@
...
+        private readonly IInteractiveWindow _interactiveWindow;
+        private volatile bool _disposed;
+
+        private int _currentLineStart = -1;
+        private string _newText;
+
+        public InteractiveWindowWriter(IMainThread mainThread, IInteractiveWindow interactiveWindow) {
+            _mainThread = mainThread;
+            _interactiveWindow = interactiveWindow;
+            Task.Run(async () => await OutputProcessingTask()).DoNotWait();
```
AlexanderSher: `OutputProcessingTask().DoNotWait();` should be enough

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
```diff
@@ -0,0 +1,198 @@
...
+            private readonly List<Message> _messages = new List<Message>();
+            private readonly ManualResetEventSlim _messagesAvailable = new ManualResetEventSlim(false);
+            private readonly object _lock = new object();
+            private int _lastCRMessageIndex = -1;
+
+            public void Enqueue(string text, bool isError) {
+                lock (_lock) {
+                    var m = new Message { Text = text, IsError = isError };
+                    // Try merging text with the last CR message, if any
+                    if (_messages.Count > 0) {
```
AlexanderSher: Why do you need this merge?

MikhailArkhipov: Less flicker and fewer changes to text buffer.

AlexanderSher: I'm not entirely sure, but it looks like this merge can be moved into consumer, and in that case you can use `BufferBlock<Message>` instead of custom queue.

MikhailArkhipov: I wanted to merge in the background thread so UI thread would have minimal work to do. Previous code did way too much in UI thread overloading `Post`. I need to be able to replace elements so plain `Queue` or concurrent queue-type varieties do not fit.

MikhailArkhipov: `rtweet::stream_tweets` produces literally thousands of messages so I didn't want to process them in UI thread

AlexanderSher: Ah, so you actually remove some of the messages before they become part of the buffer.

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
```diff
@@ -0,0 +1,198 @@
...
+                        if (lastMessage.MovesCaretBack() && m.IsPlainText() && lastMessage.IsError == m.IsError) {
+                            lastMessage.Text += text;
+                            return;
+                        }
+                    }
+
+                    if (m.MovesCaretBack()) {
+                        // CR moves caret back effectively erasing previous message at the line
+                        if (_lastCRMessageIndex >= 0) {
+                            _messages.RemoveRange(_lastCRMessageIndex, _messages.Count - _lastCRMessageIndex);
```
AlexanderSher: It looks like `_lastCRMessageIndex` in this place is always equal to `_messages.Count - 1`.

AlexanderSher: Also, shouldn't we remove only part of message if it is longer than one character?

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
```diff
@@ -0,0 +1,198 @@
...
+                    }
+
+                    _messages.Add(m);
+                    _messagesAvailable.Set();
+                }
+            }
+
+            public async Task<IEnumerable<Message>> WaitForMessagesAsync() {
+                await TaskUtilities.SwitchToBackgroundThread();
+                _messagesAvailable.Wait();
```
AlexanderSher: If you use `AsyncManualResetEvent`, you can use cancellation token.

src/Windows/R/Components/Test/InteractiveWorkflow/InteractiveWindowWriterTest.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        [Test]
+        public async Task MessageQueueTest02() {
+            using (var mq = new InteractiveWindowWriter.MessageQueue()) {
+                mq.Enqueue("start1", false);
+                mq.Enqueue("start2\n", false);
+                mq.Enqueue("\rtest1", false);
+                mq.Enqueue(" test2", false);
+                mq.Enqueue("test3\n", false);
```
AlexanderSher: I believe we can also have messages like "abc\rdef" and "abc\ndef"

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
MikhailArkhipov: Yep

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
MikhailArkhipov: Yes, it is generally Count-1 due to the merging. Could use bool technically, but code will look about the same :-)

About removal: in general case, perhaps. Assumption is that ASCII progress always looks like
`\rDownloading... 000% done`. The code obviously doesn't handle `Down\rloading...` or `abc\r` and similar cases. Basically we erase everything up to and including complete message with CR.

src/Windows/R/Components/Test/InteractiveWorkflow/InteractiveWindowWriterTest.cs
MikhailArkhipov: Former is handled as `\rabcdef` and latter contains LF so it resets CR index. But I can add both here.

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
AlexanderSher: Do you need to distinct error and non-error messages here?

src/Windows/R/Components/Test/InteractiveWorkflow/InteractiveWindowWriterTest.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+
+        [CompositeTest]
+        [InlineData(new[] { "start1", "start2\n", "\rtest1", " test2", "test3\n", "\rtest4", "test5", "test6", "\rtest4", " test8" },
+                    new[] { "start1", "start2\n", "\rtest1 test2", "test3\n", "\rtest4 test8" })]
+        [InlineData(new[] { "st\rart1", "start2", "\rtest1", " test2"},
+                    new[] { "\rtest1 test2" })]
+        [InlineData(new[] { "st\rart1", "st\nart2", "\rtest1", " test2" },
+                    new[] { "st\rart1", "st\nart2", "\rtest1 test2" })]
```
AlexanderSher: Why we don't erase what is after "\n" in the middle of the message? I mean, why expected isn't `new[] { "st\rart1", "st\n", "\rtest1 test2" }`?

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/InteractiveWindowWriter.cs
MikhailArkhipov: It is rather odd to output errors with CR... Change error/regular works like LF, breaking CR sequence. So if error happens output would look like

```
Downloading.... 35%
Error!
Downloading.... 45%
```

src/Windows/R/Components/Test/InteractiveWorkflow/InteractiveWindowWriterTest.cs
MikhailArkhipov: We technically could, quite simple. Probably rather unusual to do this anyway.

AlexanderSher:
👍

PR #3766, opened 2017-07-27
MatrixView accessibility

AlexanderSher:
image


src/Common/Wpf/Impl/Extensions/DoubleExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public static bool GreaterThan(this double value1, double value2) 
+            => value1 > value2 && !value1.IsCloseTo(value2);
+
+        public static bool GreaterOrCloseTo(this double value1, double value2)
+            => value1 >= value2 || value1.IsCloseTo(value2);
+
+        public static bool IsCloseTo(this double value1, double value2) => value1.IsNonreal() || value2.IsNonreal()
+            ? value1.CompareTo(value2) == 0
+            : Math.Abs(value1 - value2) < 1.53E-06;
```
MikhailArkhipov: How did you come with this specific number? :-)

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
```diff
@@ -0,0 +1,145 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
```
MikhailArkhipov: I am wondering if we should pull grid out and into a separate assembly like `Wpf.Controls` or something. Might come useful to PTVS. Probably not now, just saying.

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+            }
+        }
+
+        public bool IsInVisibleRange(int row, int column) => _visibleRange.Contains(row, column);
+
+        protected override List<AutomationPeer> GetChildrenCore() {
+            var children = new List<AutomationPeer>();
+
+            for (var row = _visibleRange.Rows.Start.ReduceToInt(); row < _visibleRange.Rows.End.ReduceToInt(); row++) {
```
MikhailArkhipov: Can Rows/Cols be `IEnumerable`?

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
```diff
@@ -0,0 +1,145 @@
...
+            return children;
+        }
+        
+        public IRawElementProviderSimple[] GetRowHeaders() => null;
+
+        public IRawElementProviderSimple[] GetColumnHeaders() {
+            var start = _visibleRange.Columns.Start.ReduceToInt();
+            var count = _visibleRange.Columns.End.ReduceToInt() - start;
+            var headers = new IRawElementProviderSimple[count];
+            for (var i = 0; i < headers.Length; i++) {
```
MikhailArkhipov: Could be list Linq if `Columns` was `enumerable`

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeerScrollProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        public MatrixViewAutomationPeerScrollProvider(MatrixView owner) : base(owner) {
+            _owner = owner;
+        }
+
+        public void UpdateValues() {
+            var points = _owner.Points;
+            HorizontallyScrollable = points.HorizontalExtent.GreaterThan(points.ViewportWidth);
+            VerticallyScrollable = points.VerticalExtent.GreaterThan(points.ViewportHeight);
+            HorizontalViewSize = HorizontallyScrollable ? 100 * points.ViewportWidth / points.HorizontalExtent : 100;
```
MikhailArkhipov: May want to `Math.Round`  so you actually get to 100

src/Common/Wpf/Impl/Extensions/DoubleExtensions.cs
AlexanderSher: It is used in several places in WPF, but for some reason it is never exposed

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
AlexanderSher: We certainly should, but for that we need to unbind it from R specifics.

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeerScrollProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+                    _owner.Scroller.EnqueueCommand(GridUpdateType.PageRight, 1);
+                    break;
+            }
+
+            switch (verticalAmount) {
+                case ScrollAmount.LargeDecrement:
+                    _owner.Scroller.EnqueueCommand(GridUpdateType.PageUp, 1);
+                    break;
+                case ScrollAmount.SmallDecrement:
+                    _owner.Scroller.EnqueueCommand(GridUpdateType.LineUp, 1);
```
MikhailArkhipov: Code below does `Scroller?` do we care or it is never actually null?

MikhailArkhipov:
👍

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
AlexanderSher: It can, but there is no value in it, cause it will be an `IEnumerable<int>`.

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
AlexanderSher: In LINQ, it would be something like:
```cs
_visibleRange.Columns.GetEnumerable()
    .Where(i => i <= int.MaxValue)
    .Cast<int>()
    .Select(i => ProviderFromPeer(GetOrCreateHeader(i)))
    .ToArray();
```

Can't say I like it more than existing code

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeerScrollProvider.cs
AlexanderSher: You mean `Math.Max(100, 100 * points.ViewportHeight / points.VerticalExtent)`?

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeerScrollProvider.cs
AlexanderSher: It is `null` if user changes variable to something that we can't display, for example:
1. `x <- data.frame(replicate(1000, sample(0:1000, 1000, rep = TRUE)))`
2. Open variable view
3. `x <- function() {}`

src/Common/Wpf/Impl/Extensions/DoubleExtensions.cs
karthiknadig: Is this supposed to be numerical closeness or visual (like thickness of line, points in space etc)?

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeerScrollProvider.cs
MikhailArkhipov: `Math.Round(100 * points.ViewportHeight / points.VerticalExtent)` assuming this is percentage.

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeer.cs
MikhailArkhipov: OK

src/Common/Wpf/Impl/Extensions/DoubleExtensions.cs
AlexanderSher: Not sure I understand your question. It is used in WPF, so it is related to lines and points

src/Package/Impl/DataInspect/VisualGrid/MatrixViewAutomationPeerScrollProvider.cs
AlexanderSher: It is a percentage, but it can be a fraction of percent (not documented though)

src/Common/Wpf/Impl/Extensions/DoubleExtensions.cs
karthiknadig: If it is for lines and points , then it might be visual closeness. That is the smallest value that has to be added/subtracted to make a visual change in WPF.

MikhailArkhipov:
Merge and build for BB?

AlexanderSher:
Not yet. I've found an issue that sometimes "ScrollIntoViewAsync" ends up in the infinite loop. Need to fix it first.

Issue #3770, opened 2017-07-27
Format document works oddly in R Markdown


MikhailArkhipov:
Repositions trailing separator, changes formatting of `{r ...}` blocks etc.

Issue #3764, opened 2017-07-27
R Markdown indent should default to spaces


MikhailArkhipov:
![image](https://user-images.githubusercontent.com/12820357/28676933-edb914fc-72a0-11e7-94f7-ee590579406b.png)

MikhailArkhipov:
Also, need adjust R formatting for projections since `^K^D` doesn't quite work right in RMD

Issue #3763, opened 2017-07-27
Smart indent in markdown


MikhailArkhipov:
Seems to be having issues

![image](https://user-images.githubusercontent.com/12820357/28676808-832caa72-72a0-11e7-8103-5a1ac00a6eb6.png)

MikhailArkhipov:
Needs proper implementation of contained language support since core editor only asks for smart indenter from the top-level language.

Issue #3550, opened 2017-05-17
Hyperlinks messed up in tooltips


Hong-Revo:
This is with the rlang package, but it happens with any help file that has links in the text. Here's the rendered tooltip:

![image](https://cloud.githubusercontent.com/assets/11328890/26154022/35b5f34c-3b51-11e7-9f79-4c1561606ddc.png)

And the actual help text:

![image](https://cloud.githubusercontent.com/assets/11328890/26154032/3d6d33c0-3b51-11e7-9bee-7c4515a66e20.png)

The links appear as doubled text.

MikhailArkhipov:
We filter some data in the documentation since it doesn't make sense in the tooltip. I guess we could drop urls as well since standard VS tooltip is not clickable.

MikhailArkhipov:
Need to handle
```
\href{{http://rlang.tidyverse.org/articles/tidy-evaluation.html}{tidy evaluationframework}}
```
[eval_tidy.rd.txt](https://github.com/Microsoft/RTVS/files/1178260/eval_tidy.rd.txt)

PR #3774, opened 2017-07-28
Editor and Service Manager fixes

MikhailArkhipov:
Fix NRE in editor and object leaks:
- Change ServiceManager to handle services as existing code expects. Specifically, if `A` implements `I1` and `I2` both should be available as services no matter how exactly `A` was added. Also, if `A` was added but removed in the base class `Dispose` it should be removed.

#3763 Smart indent in markdown 
- Delegate to R in code blocks

#3764 R Markdown indent should default to spaces 
- Set package attributes appropriately

#3770 Format document works oddly in R Markdown 
- Add Format Document command and format blocks one by one

#3550 Hyperlinks messed up in tooltips 
- Partial fix, to be continued

Also
- Better detect REPL in validation since view is not yet available when document is created


src/Common/Core/Impl/Services/IServiceManager.cs
```diff
@@ -20,6 +20,13 @@ public interface IServiceManager : IServiceContainer, IDisposable {
         /// <summary>
         /// Adds on-demand created service
         /// </summary>
+        IServiceManager AddService<TService, TImplementation>()
```
AlexanderSher: What was the problem with `ServiceManagerExtensions.AddService`?

src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -28,22 +28,33 @@ public class ServiceManager : IServiceManager {
         public virtual IServiceManager AddService<T>(T service, Type type = null) where T : class {
             _disposeToken.ThrowIfDisposed();
 
-            type = type ?? typeof(T);
+            type = type ?? service.GetType();
```
AlexanderSher: This change creates several problems:
1. It pretty much removes all the benefits of using the dictionary, cause most of the times service is requested by interface, not by type itself, so to find required instance `ServiceManager` will have to iterate through collection all the time.
2. Inconsistency. For `class C : I1, I2` `ServiceManager.AddService<I1>(new C())` and `ServiceManager.AddService<I1>(() => new C1())` will create different results.

It looks like this method has to be changed to:
```cs
IServiceManager AddService(object instance, Type type = null)
```
which will match `void RemoveService(object service)`. And then we may have an extension:
```cs
IServiceManager AddService<T>(this IServiceManager serviceManager, T instance) where T : class => serviceManager.AddService(instance, typeof(T));
```

src/Common/Core/Impl/Services/ServiceManager.cs
```diff
@@ -28,22 +28,33 @@ public class ServiceManager : IServiceManager {
...
             Check.ArgumentNull(nameof(service), service);
-            Check.InvalidOperation(() => _s.TryAdd(type, service), $"Service of type {type} already exists");
+           _s.TryAdd(type, service); // Ignore if service added second time
```
AlexanderSher: > Ignore if service added second time

Ok, what if it a different instance?

AlexanderSher:
> Specifically, if `A` implements `I1` and `I2` both should be available as services no matter how exactly `A` was added. 

We really have a case like:
```cs
serviceManager.AddService<I1, A>();
var s = serviceManager.GetService<I2>();
```

If service is added as something that should be accessed by any interface it implements, why not just add it as `serviceManager.AddService<A>();`?

src/Common/Core/Test/Services/ServiceManagerTests.cs
```diff
@@ -42,26 +36,74 @@ public class ServiceManagerTests {
...
+            _serviceManager.RemoveService<I1>();
+
+            _serviceManager.GetService<I1>().Should().BeNull();
+            _serviceManager.GetService<I2>().Should().BeNull();
+        }
+
+        [Test]
+        public void AddRemoveLazy03() {
+            _serviceManager.AddService<I2, Derived>();
+            _serviceManager.RemoveService<Base>();
```
AlexanderSher: I don't like this idea, do we really need it? It makes service removal operation more accessible than adding: to add a service, code needs access to the type itself or one of its constructors. To remove a type, code needs access to any of the interfaces this type implements. And in case when several types implement one and the same interface, there is no way to find out what service will be removed. 

This can be an unpleasant bug prone. For example, imagine the following code:
```cs
public class Service : IDisposable { /*...*/ }

public class ServiceProvider : IDisposable {
    private IServiceManager _services;
    private Service _service;
    public Client(IServiceManager services) {
        _services = services;
        _service = new Service();
        services.AddService(_service);
    }

    public void Dispose() {
        _services.Remove(_service);
        _service.Dispose();
    }
}
```

It will work as expected only until `_service` field is of `Service`. If someone changes it to `IDisposable`, behavior will be broken (cause there is more than one service that implements `IDisposable`) and it will be hard to find out why.

MikhailArkhipov: Yes. However, it is pretty unusual to so it. I'd be more concerned if this was some public API surface. The problem is there is a lot of code that does `Add(this)` while `Remove(that)` and that started causing trouble. Ex: document closes, some services are removed and some not, so it can't be re-created. Format RMD document and then you'll lose R blocks and eventually crash. I'd rather not chase all these cases now. So change is effectively backward compat.

MikhailArkhipov: To the above - for lazy case there is `AddService<I1, A1>` (if someone looks for `I1`, instantiate `A1`) and generally `AddService<A>` is not very useful.

src/Common/Core/Impl/Services/ServiceManager.cs
MikhailArkhipov: May be worth adding a check. Same instance is OK, new instance is probably a bug.

AlexanderSher: Something like this:
```cs
Check.InvalidOperation(() => _s.GetOrAdd(type, service) == service, $"Another instance of service of type {type} already added");
```

src/Common/Core/Test/Services/ServiceManagerTests.cs
AlexanderSher: The example that I've provided is actually wrong - `_services.Remove(_service)` will work because it checks for instance type, not for the type of variable.
Generally, we have two different cases:
- Service is added as instance or as a concrete type, which is backward compatible with editor code. For that, we have `_serviceManager.AddService(service)`, `_serviceManager.Remove(service)` and `_serviceManager.AddService<ServiceType>()`. These services can be found by any interface they implement as well as type itself.
- Service is especially added by its interface or some other key. For that, we call `_serviceManager.AddService(service, typeof(InterfaceType))` and `_serviceManager.AddService<InterfaceType, ServiceType>()`. This provides better code decoupling and testability and shouldn't provide access to service by anything except the specified interface.

`IServiceManager.RemoveService<T>` doesn't fall into any of these cases. Since we don't use it right now, maybe we can just remove it?

Issue #3773, opened 2017-07-28
Grid row names display is incorrect


MikhailArkhipov:
`MatrixNATest` is broken and it is product issue. `MatrixOnlyRowNameTest` most probably has similar issue.

![image](https://user-images.githubusercontent.com/12820357/28737025-8d7c0e58-73a1-11e7-9435-fc2c0cfc17cd.png)

int19h:
RStudio is not doing the correct thing here either, because it converts matrix to a dataframe first, which breaks NA columns.

The correct behavior is to show numeric indices. If the entry in `colnames()` or `rownames()` for a particular row or column is `NA`, that row/column does not have a name, meaning that you cannot use e.g. `m[NA,]` or `m[,NA]` to slice to just that row/column - you'll need to use `m[2,]` or `m[,3]`. Row and column labels in the grid correspond to the slicing syntax, so they should use numeric indices accordingly when those are the only valid way to address that row/column.

So on RTVS screenshot, the `[,3]` label is actually correct, and the `NA` label is wrong, and should be `[2]`. On the RS screenshot, they do this correctly for the rows, but then fail for columns. We should do it correctly for both.

int19h:
The unit tests are written to reflect the behavior described above, so this just needs to fix the product code to match what the tests expect.

This appears to be a regression from adding formatting to row names a while ago. This is the bit that breaks it, which is done only for rownames and not colnames, which is why columns are labeled correctly:
```r
        if (is.atomic(rn)) {
            # For atomic vectors, we want to apply format() to the whole thing at once,
            # so that it can determine the number of decimal places accordingly - e.g.
            # for c(1.5, 2, 3.04), we want the output to be "1.50 2.00 3.04".
            rn <- format(rn, trim = TRUE, justify = "none")
        }
```
Calling `format` like this makes `NA` become string `"NA"`, and that then becomes the label.

Just checking for "NA" is not good enough, because that is also a valid string label. However, the code above is really only useful when labels are numeric. So it should probably check for `is.numeric` rather than `is.atomic` before it applies `format`; and then after applying it, it can be certain that any `"NA"` value in the result was originally `NA`, since inputs weren't strings, and replace all such with NULLs (which will cause the grid to use numeric indices for that row).

int19h:
Also, this should probably be a test in `GridDataTest` (and then the one in `EvaluationWrapperTests` can be removed, since it has a bunch of unnecessary dependencies for this).

PR #3776, opened 2017-07-31
Fix for grid row names.

karthiknadig:
#3773

### DIFF
--- a/src/Host/Client/Impl/rtvs/R/grid.R
+++ b/src/Host/Client/Impl/rtvs/R/grid.R
@@ -136,7 +136,7 @@ grid_data <- function(x, rows, cols, row_selector) {
             # For numeric vectors, we want to apply format() to the whole thing at once,
             # so that it can determine the number of decimal places accordingly - e.g.
             # for c(1.5, 2, 3.04), we want the output to be "1.50 2.00 3.04".
-            rn <- format(rn, trim = TRUE, justify = "none")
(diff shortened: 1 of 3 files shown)


AlexanderSher:

int19h:
This is not sufficient, because it will still mangle NA row names when they're numeric. 

To fix that, after checking for `is.numeric` and doing `format`, it should look for string `"NA"` in the resulting vector, and replace all instances of that with `NULL`. 

Note that regular vectors and lists can't have numeric rownames, which is why it's a bit tricky to test. If I remember correctly, the case where this happens is when you work with time series, or `quantmod` objects.

karthiknadig:
@int19h NULL cannot be assigned to a vector.
```
> x <- c(1,2,3,4,NA,6, NA, NaN, 9)
> length(x)
[1] 9
> y<-format(x, trim = TRUE, justify = "none")
> y[y=="NA"] <- NULL
Error in y[y == "NA"] <- NULL : replacement has length zero
```
Attempting to convert to list, then assigning NULL removes it 
```
> z <- as.list(y)
> z[z=="NA"] <- NULL
> length(z)
[1] 7
```
Any suggestions on this?

int19h:
Makes sense with vectors, since NULL is not an atomic type.

With lists, I think that's because your second assignment is a slicing assignment. You should be able to assign individual elements via `[[...]] <-`.

Also, perhaps `lapply` would work better here?

PR #3788, opened 2017-08-02
Removing unnecessary dependencies.

karthiknadig:
These binaries clobbered the product binaries which caused RTVS to fail after install.


AlexanderSher:
:+1

Issue #3671, opened 2017-06-30
Remote support for static html pages


karthiknadig:
Currently R Remote services do not support static html pages, as produced by highcharter, ggviz, etc. They generate html files and content stored in %TMP% or /tmp. On the client side, we receive a path with file protocol (file:///). We could start a web server and use blob stream/file fetch to pull the files as requested from the remote machine. This will also handle some local scenarios where using web server to render the HTML is better.

PR #3817, opened 2017-08-07
Adds local and remote static file web server.

karthiknadig:
#3671

### DIFF
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -13,6 +13,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Common.Core;
 using Microsoft.Common.Core.Disposables;
+using Microsoft.Common.Core.IO;
 using Microsoft.Common.Core.Json;
 using Microsoft.Common.Core.Logging;
 using Microsoft.Common.Core.Net;
@@ -231,6 +232,15 @@ namespace Microsoft.R.Host.Client.Host {
             return ex.ApiError.ToString();
         }
 
-        public virtual Task<string> HandleUrlAsync(string url, CancellationToken cancellationToken)  => Task.FromResult(url);
+        public virtual Task<string> HandleUrlAsync(string url, CancellationToken cancellationToken) {
+            UriBuilder ub = new UriBuilder(url);
+            if (ub.Scheme.StartsWithIgnoreCase("file")) {
+                var remotingService = _services.GetService<IRemotingWebServer>();
+                var fs = _services.GetService<IFileSystem>();
+                return remotingService.CreateLocalStaticFileServerAsync(url, fs, Log, _console, cancellationToken);
+            }
+
+            return Task.FromResult(url);
+        }
     }
 }
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -237,7 +237,7 @@ namespace Microsoft.R.Host.Client.Host {
             if (ub.Scheme.StartsWithIgnoreCase("file")) {
                 var remotingService = _services.GetService<IRemotingWebServer>();
                 var fs = _services.GetService<IFileSystem>();
-                return remotingService.CreateLocalStaticFileServerAsync(url, fs, Log, _console, cancellationToken);
+                return remotingService.HandleLocalStaticFileUrlAsync(url, _console, cancellationToken);
             }
 
             return Task.FromResult(url);
--- a/src/Host/Client/Impl/Host/BrokerClient.cs
+++ b/src/Host/Client/Impl/Host/BrokerClient.cs
@@ -91,8 +91,7 @@ namespace Microsoft.R.Host.Client.Host {
         public async Task<T> GetHostInformationAsync<T>(CancellationToken cancellationToken) {
             string result = null;
             try {
-                string endpoint;
-                if (!_typeToEndpointMap.TryGetValue(typeof(T), out endpoint)) {
+                if (!_typeToEndpointMap.TryGetValue(typeof(T), out string endpoint)) {
                     throw new ArgumentException($"There is no endpoint for type {typeof(T)}");
                 }
 
@@ -233,7 +232,7 @@ namespace Microsoft.R.Host.Client.Host {
         }
 
         public virtual Task<string> HandleUrlAsync(string url, CancellationToken cancellationToken) {
-            UriBuilder ub = new UriBuilder(url);
+            var ub = new UriBuilder(url);
             if (ub.Scheme.StartsWithIgnoreCase("file")) {
                 var remotingService = _services.GetService<IRemotingWebServer>();
                 var fs = _services.GetService<IFileSystem>();
--- a/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
+++ b/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
@@ -266,5 +266,8 @@ grDevices::deviceIsInteractive('ide')
 
         public static Task EnableAutosaveAsync(this IRExpressionEvaluator evaluation, bool deleteExisting) =>
             evaluation.ExecuteAsync(Invariant($"rtvs:::enable_autosave({deleteExisting.ToRBooleanLiteral()})"));
+
+        public static Task<bool> FileExistsAsync(this IRExpressionEvaluator evaluation, string path) =>
+            evaluation.EvaluateAsync<bool>(Invariant($"file.exists({path.ToRPath().ToRStringLiteral()})"), REvaluationKind.Normal);
     }
 }
--- a/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
+++ b/src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
@@ -267,7 +267,7 @@ grDevices::deviceIsInteractive('ide')
         public static Task EnableAutosaveAsync(this IRExpressionEvaluator evaluation, bool deleteExisting) =>
             evaluation.ExecuteAsync(Invariant($"rtvs:::enable_autosave({deleteExisting.ToRBooleanLiteral()})"));
 
-        public static Task<bool> FileExistsAsync(this IRExpressionEvaluator evaluation, string path) =>
-            evaluation.EvaluateAsync<bool>(Invariant($"file.exists({path.ToRPath().ToRStringLiteral()})"), REvaluationKind.Normal);
+        public static Task<bool> FileExistsAsync(this IRExpressionEvaluator evaluation, string path, CancellationToken cancellationToken = default(CancellationToken)) =>
+            evaluation.EvaluateAsync<bool>(Invariant($"file.exists({path.ToRPath().ToRStringLiteral()})"), REvaluationKind.Normal, cancellationToken);
     }
 }
--- a/src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
+++ b/src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
@@ -12,10 +12,7 @@ using Microsoft.Common.Core.IO;
 using Microsoft.Common.Core.Logging;
 
 namespace Microsoft.R.Host.Client.BrokerServices {
-    public class LocalStaticFileServer : StaticFileServerBase {
-        private static LocalStaticFileServer _server;
-        private static object _serverLock = new object();
-
+    internal class LocalStaticFileServer : StaticFileServerBase {
         public LocalStaticFileServer(IFileSystem fs, IActionLog log, IConsole console) : base(fs, log, console) {
         }
 
@@ -25,24 +22,17 @@ namespace Microsoft.R.Host.Client.BrokerServices {
             Log.WriteLine(LogVerbosity.Minimal, MessageCategory.General, Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
             Console.WriteLine(Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
 
-            UriBuilder ub = new UriBuilder(urlStr);
-            return GetFileServerPath(ub.Path);
-        }
-
-        public static Task<string> CreateAsync(string url, IFileSystem fs, IActionLog log, IConsole console, CancellationToken ct = default(CancellationToken)) {
-            lock (_serverLock) {
-                _server = _server ?? new LocalStaticFileServer(fs, log, console);
-            }
-
-            return _server.HandleUrlAsync(url);
+            return GetFileServerPath(urlStr);
         }
 
         public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
-            var uri = context.Request.Url.PathAndQuery;
-            if (!FileSystem.FileExists(uri) && uri.StartsWith("/")) {
-                uri = uri.Substring(1);
-            }
+            UriBuilder ub = new UriBuilder();
+            ub.Scheme = "file";
+            ub.Host = "";
+            ub.Path = context.Request.Url.LocalPath;
+            ub.Query = context.Request.Url.Query;
 
+            var uri = ub.Uri.LocalPath;
             if (FileSystem.FileExists(uri)) {
                 using (var stream = FileSystem.FileOpen(uri, FileMode.Open)) {
                     await stream.CopyToAsync(context.Response.OutputStream, null, ct);
--- a/src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
+++ b/src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
@@ -26,19 +26,19 @@ namespace Microsoft.R.Host.Client.BrokerServices {
         }
 
         public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
-            UriBuilder ub = new UriBuilder();
-            ub.Scheme = "file";
-            ub.Host = "";
-            ub.Path = context.Request.Url.LocalPath;
-            ub.Query = context.Request.Url.Query;
+            UriBuilder ub = new UriBuilder() {
+                Scheme = "file",
+                Host = "",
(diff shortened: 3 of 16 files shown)


src/Windows/Host/Client/Impl/BrokerServices/StaticFileServerBase.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (_listener != null && _listener.IsListening) {
+                return true;
+            }
+
+            while (true) {
+                if (ct.IsCancellationRequested) {
+                    return false;
+                }
+                _listener.Prefixes.Clear();
+                var lport = _random.Next(49152, 65535);
```
int19h: Since this piece of code is proliferating, can you put it into one of the utility assemblies?

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,54 @@
...
+            await InitializeAsync(ct);
+
+            Log.WriteLine(LogVerbosity.Minimal, MessageCategory.General, Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+            Console.WriteLine(Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+
+            UriBuilder ub = new UriBuilder(urlStr);
+            return GetFileServerPath(ub.Path);
+        }
+
+        public static Task<string> CreateAsync(string url, IFileSystem fs, IActionLog log, IConsole console, CancellationToken ct = default(CancellationToken)) {
```
int19h: The method is misleading - from the name and signature, it would seem that it creates a new instance every time, but in practice it's a singleton accessor (and worse yet, it will silently ignore the arguments if singleton is already there, potentially giving you an instance with wrong services... at the very least, this might break tests).

I think it's better to manage the lifetime of that instance in the place where it is used, i.e. `RemotingWebServer`.

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,54 @@
...
+        private static object _serverLock = new object();
+
+        public LocalStaticFileServer(IFileSystem fs, IActionLog log, IConsole console) : base(fs, log, console) {
+        }
+
+        public async Task<string> HandleUrlAsync(string urlStr, CancellationToken ct = default(CancellationToken)) {
+            await InitializeAsync(ct);
+
+            Log.WriteLine(LogVerbosity.Minimal, MessageCategory.General, Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+            Console.WriteLine(Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
```
int19h: Should this be doing console writes? Where do they go?

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,54 @@
...
+            lock (_serverLock) {
+                _server = _server ?? new LocalStaticFileServer(fs, log, console);
+            }
+
+            return _server.HandleUrlAsync(url);
+        }
+
+        public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
+            var uri = context.Request.Url.PathAndQuery;
+            if (!FileSystem.FileExists(uri) && uri.StartsWith("/")) {
```
int19h: I'm confused about this bit - what is it trying to achieve? It would seem that if I'm requesting, say, `/home/foo/bar.txt`, it will happily return `home/foo/bar.txt`, which might actually be a different file. But if the URI starts with `/`, it's specifically requesting an absolute file path. What's the scenario where a relative one is needed instead?

src/Windows/Host/Client/Impl/BrokerServices/RemoteStaticFileServer.cs
```diff
@@ -0,0 +1,79 @@
...
+            string path = urlStr;
+            if (urlStr.StartsWithIgnoreCase("file://")) {
+                UriBuilder ub = new UriBuilder(urlStr);
+                path = ub.Path;
+            }
+            
+            return GetFileServerPath(path);
+        }
+
+        public static Task<string> CreateAsync(string url, IRSessionProvider sp, IFileSystem fs, IActionLog log, IConsole console, CancellationToken ct = default(CancellationToken)) {
```
int19h: Same thing wrt lifetime management - let's move it to the invoking code.

src/Windows/Host/Client/Impl/BrokerServices/RemoteStaticFileServer.cs
```diff
@@ -0,0 +1,79 @@
...
+            IRSession session = _sessionProvider.GetOrCreate(remoteFileSession);
+            await session.EnsureHostStartedAsync(new RHostStartupInfo(), null, 3000, ct);
+            _session = session;
+
+            Log.WriteLine(LogVerbosity.Minimal, MessageCategory.General, Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+            Console.WriteLine(Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+
+            string path = urlStr;
+            if (urlStr.StartsWithIgnoreCase("file://")) {
+                UriBuilder ub = new UriBuilder(urlStr);
```
int19h: Since you don't modify anything here, this should just use `Uri`, not `UriBuilder`.

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,54 @@
...
+        public LocalStaticFileServer(IFileSystem fs, IActionLog log, IConsole console) : base(fs, log, console) {
+        }
+
+        public async Task<string> HandleUrlAsync(string urlStr, CancellationToken ct = default(CancellationToken)) {
+            await InitializeAsync(ct);
+
+            Log.WriteLine(LogVerbosity.Minimal, MessageCategory.General, Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+            Console.WriteLine(Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+
+            UriBuilder ub = new UriBuilder(urlStr);
```
int19h: `Uri`

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
int19h: Ah, so it comes from the request. But doesn't this imply that there's basically no way to distinguish between relative and absolute file access in the protocol as used here? It feels like it should either always be stripping it, or never stripping it, but not trying to check if file exists or not first. Otherwise you get this weird "file at this absolute path if it exists, otherwise file at this relative path" semantics that can turn error conditions into invalid data being returned quietly.

src/Windows/Host/Client/Impl/BrokerServices/RemoteStaticFileServer.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+
+        public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
+            if (!_session.IsHostRunning) {
+                await _session.EnsureHostStartedAsync(new RHostStartupInfo(), null, 3000, ct);
+            }
+
+            var uri = context.Request.Url.PathAndQuery;
+            if (await _session.FileExistsAsync(uri)) {
```
int19h: Same thing here wrt absolute and relative paths.

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
int19h: While we're at it - the name also implies that this creates the server and returns it. But it actually creates the server, and returns the result of handling the URL. It definitely needs a better name.

src/Windows/Host/Client/Impl/BrokerServices/RemotingWebServer.cs
```diff
@@ -11,5 +13,13 @@ public class RemotingWebServer : IRemotingWebServer {
         public Task<string> CreateWebServerAsync(string remoteUrl, string baseAddress, string name, IActionLog log, IConsole console, CancellationToken ct = default(CancellationToken)) {
             return WebServer.CreateWebServerAsync(remoteUrl, baseAddress, name, log, console, ct);
         }
+
+        public Task<string> CreateLocalStaticFileServerAsync(string url, IFileSystem fs, IActionLog log, IConsole console, CancellationToken ct = default(CancellationToken)) {
```
int19h: These also need better names, I think (currently they imply that they are factory methods for servers, but they don't return a server instance...).

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
int19h: If this is mainly about Windows vs Linux paths (C:\Foo vs /foo), then I think the proper way to handle these is to create a new `Uri` instance using the path part from `Request.Url`, but with `file:` scheme. Then you can use `Uri.LocalPath` to convert that to path, and it should take care of all platform-specific idiosyncrasies.

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
karthiknadig: Console here is a property that provides the saved IConsole reference. In this case it goes to REPL.

src/Host/Client/Impl/Host/BrokerClient.cs
```diff
@@ -231,6 +232,15 @@ protected virtual Task<Exception> HandleHttpRequestExceptionAsync(HttpRequestExc
             return ex.ApiError.ToString();
         }
 
-        public virtual Task<string> HandleUrlAsync(string url, CancellationToken cancellationToken)  => Task.FromResult(url);
+        public virtual Task<string> HandleUrlAsync(string url, CancellationToken cancellationToken) {
+            UriBuilder ub = new UriBuilder(url);
```
AlexanderSher: `var`

src/Host/Client/Impl/Session/RSessionEvaluationCommands.cs
```diff
@@ -266,5 +266,8 @@ public static Task OptionsSetWidthAsync(this IRExpressionEvaluator evaluation, i
 
         public static Task EnableAutosaveAsync(this IRExpressionEvaluator evaluation, bool deleteExisting) =>
             evaluation.ExecuteAsync(Invariant($"rtvs:::enable_autosave({deleteExisting.ToRBooleanLiteral()})"));
+
+        public static Task<bool> FileExistsAsync(this IRExpressionEvaluator evaluation, string path) =>
```
AlexanderSher: Please add `, CancellationToken cancellationToken = default(CancellationToken)`

src/Windows/Host/Client/Impl/BrokerServices/RemoteStaticFileServer.cs
```diff
@@ -0,0 +1,79 @@
...
+        private IRSession _session;
+
+        public RemoteStaticFileServer(IRSessionProvider sessionProvider, IFileSystem fs, IActionLog log, IConsole console) : base(fs, log, console) {
+            _sessionProvider = sessionProvider;
+        }
+
+        public async Task<string> HandleUrlAsync(string urlStr, CancellationToken ct = default(CancellationToken)) {
+            await InitializeAsync(ct);
+
+            IRSession session = _sessionProvider.GetOrCreate(remoteFileSession);
```
AlexanderSher: `var`

src/Windows/Host/Client/Impl/BrokerServices/StaticFileServerBase.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+                break;
+            }
+
+            DoWorkAsync(ct).DoNotWait();
+            return true;
+        }
+
+
+        private async Task DoWorkAsync(CancellationToken ct = default(CancellationToken)) {
```
AlexanderSher: Do you use `ct` at all?

src/Windows/Host/Client/Impl/BrokerServices/WebServer.cs
```diff
@@ -135,8 +134,7 @@ public class WebServer {
             var remoteUri = new Uri(remoteUrl);
             var localUri = new UriBuilder(remoteUri);
 
-            WebServer server;
-            if(!Servers.TryGetValue(remoteUri.Port, out server)) {
+            if (!Servers.TryGetValue(remoteUri.Port, out WebServer server)) {
```
AlexanderSher: Is it possible that because of concurrent access we end up with two `WebServer` instances on the same port?

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
karthiknadig: When i try to build the path using Local Path and "file" scheme. It gives me this as a path:
\\localhost\C:\Users\kanadig\AppData\Local\Temp\RtmpCM0Fz3\viewhtml3d984f3728f4\index.html
File.exists check fails on that.

karthiknadig: Ah. I see i need to set the host as string.Empty. Otherwise it picks up localhost as default host.

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            return GetFileServerPath(urlStr);
+        }
+
+        public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
+            UriBuilder ub = new UriBuilder();
+            ub.Scheme = "file";
+            ub.Host = "";
+            ub.Path = context.Request.Url.LocalPath;
+            ub.Query = context.Request.Url.Query;
```
int19h: I think you can skip the query, since it's meaningless for local paths even when present (I'm actually not sure what will happen if you specify it as non-empty - might even be an error).

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,44 @@
...
+            Console.WriteLine(Resources.Info_StaticFileServerStarted.FormatInvariant(Listener.Prefixes.FirstOrDefault()));
+
+            return GetFileServerPath(urlStr);
+        }
+
+        public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
+            UriBuilder ub = new UriBuilder();
+            ub.Scheme = "file";
+            ub.Host = "";
+            ub.Path = context.Request.Url.LocalPath;
```
int19h: Should this be `Path` rather than `LocalPath`?

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -0,0 +1,44 @@
...
+            ub.Query = context.Request.Url.Query;
+
+            var uri = ub.Uri.LocalPath;
+            if (FileSystem.FileExists(uri)) {
+                using (var stream = FileSystem.FileOpen(uri, FileMode.Open)) {
+                    await stream.CopyToAsync(context.Response.OutputStream, null, ct);
+                }
+                context.Response.OutputStream.Close();
```
int19h: If the handler is required to close the stream, this should probably be outside of `if`.

src/Windows/Host/Client/Impl/BrokerServices/RemotingWebServer.cs
```diff
@@ -1,15 +1,55 @@
...
+            CreateLocalStaticFileServer(console);
+            return _localStaticFileServer.HandleUrlAsync(url, ct);
+        }
+
+        public Task<string> HandleRemoteStaticFileUrlAsync(string url,  IRSessionProvider rSessionProvider, IConsole console, CancellationToken ct = default(CancellationToken)) {
+            CreateRemoteStaticFileServerAsync(rSessionProvider, console);
+            return _remoteStaticFileServer.HandleUrlAsync(url, ct);
+        }
+
+        private void CreateLocalStaticFileServer(IConsole console) {
```
int19h: This should probably be called something like `EnsureLocalStaticFileServer`.

Better yet, make it `Get...` and return the instance (whether created or not). Then the methods above can be one-liners, just using the returned value.

src/Windows/Host/Client/Impl/BrokerServices/StaticFileServerBase.cs
```diff
@@ -0,0 +1,91 @@
...
+            FileSystem = fs;
+            Log = log;
+            Console = console;
+        }
+
+        protected async Task<bool> InitializeAsync(CancellationToken ct) {
+            await TaskUtilities.SwitchToBackgroundThread();
+
+            _listener = _listener ?? new HttpListener();
+            if (_listener != null && _listener.IsListening) {
```
int19h: `_listener != null` is always going to be true, since a new instance is created on the line above if it was null when this method was entered.

src/Windows/Host/Client/Impl/BrokerServices/LocalStaticFileServer.cs
```diff
@@ -26,19 +26,17 @@ internal class LocalStaticFileServer : StaticFileServerBase {
         }
 
         public override async Task HandleRequestAsync(HttpListenerContext context, CancellationToken ct) {
-            UriBuilder ub = new UriBuilder();
+            UriBuilder ub = new UriBuilder(context.Request.Url);
```
int19h: This will likely copy more than you want - e.g. port and username/password. Safer to build it only from the components that you want there, like you were doing before. I think the only thing that you do need from the original URL is `Path`, and then set scheme and host to fixed values.

Issue #3818, opened 2017-08-09
Intellisense doesn't like ! in arithmetic expr


Hong-Revo:
![image](https://user-images.githubusercontent.com/11328890/29121014-242fab9a-7d50-11e7-84ac-f8f3049e4762.png)

MikhailArkhipov:
This comes from the fact that

http://stat.ethz.ch/R-manual/R-patched/library/base/html/Syntax.html
https://www.programiz.com/r-programming/precedence-associativity 

list `!` as _lower_ precedence to compared to `*` However, when converted to _value_ type, `!` has higher precedence than `*` (similar unary `+` or `-`) BUT it applies to the entire expression to the right. I.e. 

```
x <- 1
y <- 1
2 + !x+y
```

yields 2. This is different from typical C case (see http://www.swansontec.com/sopc.html) where `!` is plain vanilla unary.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ea512475a4f87015c7a49fe2ab94f42f2b05f4ee

Issue #3816, opened 2017-08-07
IntelliSense not prompt global variable in [ ] 


Victsz:
![image](https://user-images.githubusercontent.com/8154646/29011228-f02f49e2-7b63-11e7-9949-a50e657e69af.png)

...

If you have any screenshots demonstrating the issue, please include them as well to help us diagnose it better.

Additional information:

OS Information
    Version:       Microsoft Windows NT 6.3.9600.0

RTVS Information:
    Assembly: Microsoft.VisualStudio.R.Package, Version=1.1.30523.1233, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a

Active R URI:
    R 3.3.2: C:\Program Files\R\R-3.3.2

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ea512475a4f87015c7a49fe2ab94f42f2b05f4ee

Issue #3814, opened 2017-08-06
Annoying behaviour with Rmd files when code is in last line in file


Hong-Revo:
This applies when

1. The cursor is at the bottom of the editor window (or near the bottom)
2. The cursor is on the last nonempty line in the file
2. That line contains code (either a code chunk, or `inline`)

When these conditions are satisfied, the window scrolls up unexpectedly, hiding the line the cursor is on. The only way to stop this is to insert dummy lines at the end.

To reproduce:

1. Make a new Rmd file
2. Press Enter enough times so that you're at the bottom of the window
3. Type 

> This is a `code` line

You should see the behaviour described.

MikhailArkhipov:
This happens when there is a projecting block partially visible at the top. Doesn't have to be the last line.

![image](https://user-images.githubusercontent.com/12820357/29182699-bae0dc94-7db4-11e7-9652-f767399e4f36.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/ea512475a4f87015c7a49fe2ab94f42f2b05f4ee

Issue #3767, opened 2017-07-27
Closing VS throws: Connection point was not released


karthiknadig:
```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Connection point was not released. Interface Microsoft.VisualStudio.TextManager.Interop.IVsTextManagerEvents4, 
Call stack: 
   at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
   at System.Environment.get_StackTrace()
   at Microsoft.VisualStudio.R.Package.Interop.ConnectionPointCookie..ctor(Object source, Object sink, Type eventInterface, Boolean throwException)
   at Microsoft.VisualStudio.R.Package.Interop.ConnectionPointCookie..ctor(Object source, Object sink, Type eventInterface)
   at Microsoft.VisualStudio.R.Package.Editors.LanguageSettingsStorage..ctor(IVsPackage package, IServiceContainer services, Guid languageServiceId, IEnumerable`1 automationObjectNames)
   at Microsoft.VisualStudio.R.Packages.R.RPackage.LoadEditorSettings()
   at Microsoft.VisualStudio.R.Packages.R.RPackage.Initialize()
   at Microsoft.VisualStudio.Shell.Package.Microsoft.VisualStudio.Shell.Interop.IVsPackage.SetSite(IServiceProvider sp)
   at Microsoft.VisualStudio.Shell.Interop.IVsShell5.LoadPackageWithContext(Guid& packageGuid, Int32 reason, Guid& context)
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.GetPackage()
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.ConstructContent()
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.EnsureContentConstructed()
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.PreloadOnScreenContent(View view)
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.set_FrameView(View value)
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowManagerService.ConnectViewToFrame(View shownView)
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowManagerService.viewManager_WindowProfileChanging(Object sender, WindowProfileChangingEventArgs e)
   at System.EventHandler`1.Invoke(Object sender, TEventArgs e)
[...]

PR #3822, opened 2017-08-10
Bug fixes

### LINKED ISSUE #3761: Live markdown fails with NRE
Repro markdown:
https://raw.githubusercontent.com/Microsoft/acceleratoRs/master/CreditRiskPrediction/Code/CreditRiskDeploy.Rmd

image

MikhailArkhipov:
#3761 Live markdown fails with NRE
- Check that lines collection is not null

#3818 Intellisense doesn't like ! in arithmetic expr
- Special case `!` since it applies to the entire expression that follows and is not like C unary `!`
- Update tests

#3767 Closing VS throws: Connection point was not released
- Dispose language settings storage on package unload

#3816 IntelliSense not prompt global variable in [ ]
- Make sure we do include global variables in completion inside indexer

#3814 Annoying behavior with Rmd files when code is in last line
- Preserve viewport position along with the caret position
- Do not replace spans if they are identical

Also, exclude templates from VSIX since they are in the extension manifest.

### DIFF
--- a/src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
+++ b/src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.Languages.Core.Text;
-using Microsoft.Languages.Core.Tokens;
 using Microsoft.R.Core.AST.Functions;
 using Microsoft.R.Core.AST.Operands;
 using Microsoft.R.Core.AST.Operators;
@@ -383,7 +382,7 @@ namespace Microsoft.R.Core.AST.Expressions {
         }
 
         private OperationType HandleOperator(ParseContext context, out ParseErrorType errorType) {
-            errorType = this.HandleOperator(context, out bool isUnary);
+            errorType = HandleOperator(context, out bool isUnary);
             return isUnary ? OperationType.UnaryOperator : OperationType.BinaryOperator;
         }
 
--- a/src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
+++ b/src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
@@ -383,7 +383,7 @@ namespace Microsoft.R.Core.AST.Expressions {
         }
 
         private OperationType HandleOperator(ParseContext context, out ParseErrorType errorType) {
-            errorType = this.HandleOperator(context, null, out bool isUnary);
+            errorType = this.HandleOperator(context, out bool isUnary);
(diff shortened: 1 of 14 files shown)


src/R/Core/Impl/AST/Expressions/ExpressionParser.cs
```diff
@@ -383,7 +383,7 @@ enum OperationType {
         }
 
         private OperationType HandleOperator(ParseContext context, out ParseErrorType errorType) {
-            errorType = this.HandleOperator(context, null, out bool isUnary);
+            errorType = this.HandleOperator(context, out bool isUnary);
```
AlexanderSher: You don't need `this`

MikhailArkhipov: Lol I've read I don't need this `code` ;-)

src/Windows/Editor/Impl/Projection/ProjectionBufferManager.cs
```diff
@@ -17,7 +18,12 @@ namespace Microsoft.Languages.Editor.Projection {
     /// Manages the projection buffer for the primary language
     /// </summary>
     public sealed class ProjectionBufferManager : IProjectionBufferManager {
-        private int? _savedCaretPosition;
+        private class ViewPosition {
```
AlexanderSher: you can use tuples for this

MikhailArkhipov: Kind of looks cleaner.

src/Windows/Editor/Impl/Projection/ProjectionBufferManager.cs
```diff
@@ -151,17 +161,68 @@ public sealed class ProjectionBufferManager : IProjectionBufferManager {
...
+            _savedViewPosition = new ViewPosition {
+                CaretPosition = GetCaretPosition(),
+                ViewportTop = DiskBuffer.GetFirstView()?.ViewportTop
+            };
+        }
+
+        private void RestoreViewPosition() {
+            var textView = DiskBuffer.GetFirstView();
+            if(textView == null) {
+                return;
```
AlexanderSher: Shouldn't we set `_savedViewPosition = null;` here as well?

karthiknadig:
:shipit:

src/Windows/Editor/Impl/Projection/ProjectionBufferManager.cs
MikhailArkhipov: OK. Generally this is an odd case though.

AlexanderSher:
👍

Issue #3821, opened 2017-08-10
Switching from remote to local occasionally throws 404 Not found.


karthiknadig:
Response status code does not indicate success: 404 (Not Found).
```
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at Microsoft.R.Host.Client.BrokerServices.WebService.EnsureSuccessStatusCode(HttpResponseMessage response)
   at Microsoft.R.Host.Client.BrokerServices.WebService.<>c__DisplayClass14_0.<<HttpDeleteAsync>b__0>d.MoveNext()
```

karthiknadig:
This is when we send HTTP request DELETE to Broker.

PR #3824, opened 2017-08-10
Ignore delete if session is not found

karthiknadig:
#3821

### DIFF
--- a/src/Host/Client/Impl/Extensions/AboutHostExtensions.cs
+++ b/src/Host/Client/Impl/Extensions/AboutHostExtensions.cs
@@ -15,6 +15,7 @@ namespace Microsoft.R.Host.Client {
         }
 
         public static string IsHostVersionCompatible(this AboutHost aboutHost) {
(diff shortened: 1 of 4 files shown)


src/Host/Client/Impl/Extensions/AboutHostExtensions.cs
```diff
@@ -15,6 +15,7 @@ public static class AboutHostExtensions {
         }
 
         public static string IsHostVersionCompatible(this AboutHost aboutHost) {
+#if !DEBUG
```
AlexanderSher: ?

karthiknadig: If you don't have this you cannot connect the debug version (i.e when you F5) to a installed remote. The debug client will be 1.0.0.0 but the remote server could be 1.2.x.x. makes it hard to debug issues.

MikhailArkhipov:
👍

Issue #3832, opened 2017-08-11
Occasionally Plot History toolbar buttons are disabled


MikhailArkhipov:
Context menu appears to be OK. Does not consistently.

![image](https://user-images.githubusercontent.com/12820357/29230713-fde6582c-7e98-11e7-91db-e2b4884b1622.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/9043163d3cfe13c0fcf40b88e80efc60bcd21765

Issue #3636, opened 2017-06-16
Cannot change working directory while switching between linux remote and windows local


karthiknadig:
```
Error in setwd("/home/kanadig") : cannot change working directory
```

MikhailArkhipov:
Not a bug. What we should do is to store WD per connection rather than as single value. Single value is a remnant of the local-only R case.

MikhailArkhipov:
![image](https://user-images.githubusercontent.com/12820357/29188768-1115787c-7dc9-11e7-8847-1e08744d5432.png)

MikhailArkhipov:
Option was removed
https://github.com/Microsoft/RTVS/commit/9043163d3cfe13c0fcf40b88e80efc60bcd21765

PR #3841, opened 2017-08-14
#3636  #3832

MikhailArkhipov:
#3832 Occasionally Plot History toolbar buttons are disabled
 - notify container about selection change

#3636 Cannot change working directory while switching between linux remote and windows local 
- remove directory from options and don't set it on the host.

+ couple minor fixes

We can discuss if it is worth storing directory per connection. Not sure if it necessary....

### DIFF
--- a/src/Windows/R/Components/Impl/Plots/Implementation/RPlotHistoryVisualComponent.cs
+++ b/src/Windows/R/Components/Impl/Plots/Implementation/RPlotHistoryVisualComponent.cs
@@ -28,10 +28,12 @@ namespace Microsoft.R.Components.Plots.Implementation {
             control.DataContext = _viewModel;
 
             _disposableBag = DisposableBag.Create<RPlotDeviceVisualComponent>()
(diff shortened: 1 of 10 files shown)


src/Windows/R/Components/Impl/Plots/Implementation/RPlotHistoryVisualComponent.cs
```diff
@@ -30,7 +31,7 @@ public class RPlotHistoryVisualComponent : IRPlotHistoryVisualComponent {
                 .Add(() => control.ContextMenuRequested -= Control_ContextMenuRequested);
 
             control.ContextMenuRequested += Control_ContextMenuRequested;
-
+            control.SelectionChanged += Control_SelectionChanged;
```
AlexanderSher: `_disposableBag.Add(() => control.SelectionChanged -= Control_SelectionChanged)`

karthiknadig:
:shipit:

PR #3851, opened 2017-08-16
Adds host port property to containers.

karthiknadig:
#3839

### DIFF
--- a/src/Containers/Impl/Docker/LocalDockerContainer.cs
+++ b/src/Containers/Impl/Docker/LocalDockerContainer.cs
@@ -2,7 +2,10 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Text.RegularExpressions;
+using Microsoft.Common.Core;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.R.Containers.Docker {
     public class LocalDockerContainer : IContainer {
@@ -14,9 +17,32 @@ namespace Microsoft.R.Containers.Docker {
 
         public bool IsShortId => Id.Length < 64;
 
-        public LocalDockerContainer(string id, string name) {
-            Id = id;
-            Name = name;
+        public IEnumerable<int> HostPorts { get; }
+
+        internal LocalDockerContainer(JToken containerObject) {
+            Id = (string)containerObject["Id"];
+            Name = GetContainerName(containerObject);
+            HostPorts = GetHostPorts(containerObject);
+        }
+
+        private IEnumerable<int> GetHostPorts(dynamic containerObject) {
+            var hostPorts = new List<int>();
+            try {
+                dynamic portMap = containerObject.NetworkSettings.Ports["5444/tcp"];
+                foreach (dynamic pm in portMap) {
+                    if (int.TryParse(pm.HostPort.Value, out int port)) {
+                        hostPorts.Add(port);
+                    }
+                }
+            } catch (Exception ex) when (ex.IsCriticalException()) {
+                // RuntimeBinderException can occur if a container was not port mapped.
+            }
+            return hostPorts;
+        }
+
+        private string GetContainerName(JToken containerObj) {
+            var name = (string)containerObj["Name"];
+            return (name.StartsWithIgnoreCase("/") ? name.Substring(1) : name);
         }
     }
 }
--- a/src/Containers/Impl/Docker/LocalDockerContainer.cs
+++ b/src/Containers/Impl/Docker/LocalDockerContainer.cs
@@ -25,16 +25,16 @@ namespace Microsoft.R.Containers.Docker {
             HostPorts = GetHostPorts(containerObject);
         }
 
-        private IEnumerable<int> GetHostPorts(dynamic containerObject) {
+        private IEnumerable<int> GetHostPorts(JToken containerObject) {
             var hostPorts = new List<int>();
             try {
-                dynamic portMap = containerObject.NetworkSettings.Ports["5444/tcp"];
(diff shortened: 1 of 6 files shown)


src/Containers/Impl/Docker/LocalDockerContainer.cs
```diff
@@ -14,9 +17,32 @@ public class LocalDockerContainer : IContainer {
...
+        internal LocalDockerContainer(JToken containerObject) {
+            Id = (string)containerObject["Id"];
+            Name = GetContainerName(containerObject);
+            HostPorts = GetHostPorts(containerObject);
+        }
+
+        private IEnumerable<int> GetHostPorts(dynamic containerObject) {
+            var hostPorts = new List<int>();
+            try {
+                dynamic portMap = containerObject.NetworkSettings.Ports["5444/tcp"];
```
AlexanderSher: What does the `5444/tcp` mean?

src/Containers/Impl/Docker/LocalDockerContainer.cs
```diff
@@ -14,9 +17,32 @@ public class LocalDockerContainer : IContainer {
...
-            Name = name;
+        public IEnumerable<int> HostPorts { get; }
+
+        internal LocalDockerContainer(JToken containerObject) {
+            Id = (string)containerObject["Id"];
+            Name = GetContainerName(containerObject);
+            HostPorts = GetHostPorts(containerObject);
+        }
+
+        private IEnumerable<int> GetHostPorts(dynamic containerObject) {
```
AlexanderSher: Why dynamic? Type of the object is well known.

src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -29,32 +29,32 @@ public abstract class LocalDockerService : IDockerService {
...
-        public async Task<IEnumerable<string>> ListContainersAsync(bool getAll = true, CancellationToken ct = default(CancellationToken)) {
+        public async Task<IEnumerable<IContainer>> ListContainersAsync(bool getAll = true, CancellationToken ct = default(CancellationToken)) {
             TaskUtilities.AssertIsOnBackgroundThread();
             var command = "ps";
             var commandOptions = getAll ? "-a -q" : "-q";
             var output = await ExecuteCommandAsync(Invariant($"{command} {commandOptions}"), _defaultTimeout, ct);
             var ids = output.Split(new string[] { "\n", "\r\n" }, StringSplitOptions.RemoveEmptyEntries);
-            return ids.Where(id => _containerIdMatcher12.IsMatch(id) || _containerIdMatcher64.IsMatch(id));
+            var arr = await InspectContainerAsync(ids.Where(id => _containerIdMatcher12.IsMatch(id) || _containerIdMatcher64.IsMatch(id)), ct);
+            return arr.Select((c) => new LocalDockerContainer(c));
```
AlexanderSher: Braces around `c` aren't required

src/Windows/Containers/Test/WindowsContainerServiceTest.cs
```diff
@@ -64,29 +65,29 @@ public class WindowsContainerTests {
...
             await svc.StopContainerAsync(container, CancellationToken.None);
 
             var runningContainers2 = await svc.ListContainersAsync(false, CancellationToken.None);
-            runningContainers2.Should().NotContain(container.Id.Substring(0, 12));
+            runningContainers2.Select((c) => c.Id).Should().NotContain(container.Id);
 
             await svc.DeleteContainerAsync(container, CancellationToken.None);
             var allContainers = await svc.ListContainersAsync(true, CancellationToken.None);
-            allContainers.Should().NotContain(container.Id.Substring(0, 12));
+            allContainers.Select((c) => c.Id).Should().NotContain(container.Id);
```
AlexanderSher: If you want to see details about containers in test results, it is better to use `allContainers.Should().NotContain(c => c.Id == container.Id);`

src/Containers/Impl/Docker/LocalDockerContainer.cs
karthiknadig: This is the port inside the container that we care about. Since we use the default port inside the container, for now we only need to look for the mapping for that port.

AlexanderSher: Does it mean that it is impossible to have more than one running container at the same time?

karthiknadig: No, you can have multiple containers running. The mapped host port will be different. for example:
Container1 5444/tcp -> 0.0.0.0:5444
Container2 5444/tcp -> 0.0.0.0:6464
Container3 5444/tcp -> 0.0.0.0:6897
and so on...
the 5444/tcp part remains constant for all containers that use the default configuration file. But on the container host (your computer) side they have to be on unique ports.

AlexanderSher: Ok

src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -53,10 +54,14 @@ public abstract class LocalDockerService : IDockerService {
 
         public async Task<JArray> InspectContainerAsync(IEnumerable<string> containerIds, CancellationToken ct) {
             TaskUtilities.AssertIsOnBackgroundThread();
-            var command = "container inspect";
-            var commandOptions = string.Join(" ", containerIds);
-            var result = await ExecuteCommandAsync(Invariant($"{command} {commandOptions}"), _defaultTimeout, ct);
-            return JArray.Parse(result);
+
+            if (containerIds.Count() > 0) {
```
AlexanderSher: `containerIds.Any()`

AlexanderSher:
👍

PR #3858, opened 2017-08-16
Update references to 15.3

MikhailArkhipov:
And fix few tests


src/UnitTests/Core/Impl/Threading/TestMainThread.cs
```diff
@@ -8,10 +12,11 @@
 using Microsoft.Common.Core.UI;
 
 namespace Microsoft.UnitTests.Core.Threading {
+    [ExcludeFromCodeCoverage]
     public class TestMainThread : IProgressDialog, IMainThread, IDisposable {
         private readonly Action _onDispose;
         private readonly CancellationTokenSource _cts = new CancellationTokenSource();
-        private readonly AsyncLocal<BlockingLoop> _blockingLoop = new AsyncLocal<BlockingLoop>();
+        private readonly AsyncLocal<BlockingLoop> _blockingLoop = new System.Threading.AsyncLocal<BlockingLoop>();
```
AlexanderSher: ?

src/Windows/R/Components/Test/Threading/JoinableTaskContextProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.ComponentModel.Composition;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.UnitTests.Core.Threading;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.R.Components.Test.Threading {
+    [Export]
+    [ExcludeFromCodeCoverage]
+    public sealed class JoinableTaskContextProvider {
```
AlexanderSher: None of the Microsoft.R.Components.dll types require VS Threading Types.

AlexanderSher:
From the email thread, I think we need to add `Microsoft.VisualStudio.Shell.UI.Internal.dll` to the `RPackageServicesFixture`:

```cs
    public class RPackageServicesFixture : MarkdownEditorServicesFixture {
        protected override IEnumerable<string> GetAssemblyNames() => base.GetAssemblyNames().Concat(new[] {
            "Microsoft.VisualStudio.Shell.Mocks.dll",
            "Microsoft.VisualStudio.R.Package.dll",
            "Microsoft.VisualStudio.R.Package.Test.dll",
            "Microsoft.VisualStudio.ProjectSystem.dll",
            "Microsoft.VisualStudio.ProjectSystem.VS.dll",
            "Microsoft.VisualStudio.ProjectSystem.Implementation.dll",
            "Microsoft.VisualStudio.ProjectSystem.VS.Implementation.dll",
            "Microsoft.VisualStudio.Shell.UI.Internal.dll"
        });
```

src/Windows/R/Components/Test/Threading/JoinableTaskContextProvider.cs
MikhailArkhipov: This is the lowest level of test assemblies that a) exports components and b) has dependencies on core text editor. Text editor needs this export otherwise view/buffer creation fails.

MikhailArkhipov:
IDE/Editor folks recommended exporting separate test implementation as they do in their tests rather than depend on internal assembly. I am not sure what else shell/internal might be exporting or what else it might depend on.

src/UnitTests/Core/Impl/Threading/TestMainThread.cs
MikhailArkhipov: Remnant of change where I tried to provide joinable context from here. Will remove.

src/Windows/R/Components/Test/Threading/JoinableTaskContextProvider.cs
AlexanderSher: Ok

AlexanderSher:
👍

PR #3867, opened 2017-08-17
Fix various issues in container-related code

AlexanderSher:


src/Common/Core/Impl/Collections/DictionaryExtensions.cs
```diff
@@ -8,6 +8,11 @@
 
 namespace Microsoft.Common.Core {
     public static class DictionaryExtensions {
+        public static void Deconstruct<TK, TV>(this KeyValuePair<TK, TV> kvp, out TK key, out TV value) {
```
MikhailArkhipov: GetParts? `Deconstruct` sounds a bit like it is deleting the pair, but up to you.

MikhailArkhipov:
👍

src/Common/Core/Impl/Collections/DictionaryExtensions.cs
AlexanderSher: It is a duct typing notation. With this extension method, you can use tuples instead of `KeyValuePair`:
```cs
foreach (var (key, value) in dictionary) {

}
```

AlexanderSher:
@karthiknadig , can you please take a look as well?

src/Common/Core/Impl/OS/ProcessExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Common.Core.Tasks;
+
+namespace Microsoft.Common.Core.OS {
+    public static class ProcessExtensions {
+        public static Task WaitForExitAsync(this IProcess process, int milliseconds, CancellationToken cancellationToken = default(CancellationToken)) {
+            var tcs = new TaskCompletionSource<int>();
+            process.Exited += (o, e) => tcs.TrySetResult(0);
```
karthiknadig: Should't this be the process exit code?

AlexanderSher: Do you want `WaitForExitAsync` to return exit code?

karthiknadig: Is there a way to know if this returned due to time out or because the process exited?

---
In reply to: [133770027](https://github.com/Microsoft/RTVS/pull/3867#discussion_r133770027) [](ancestors = 133770027)

AlexanderSher: > Is there a way to know if this returned due to time out or because the process exited?

Async methods throw `OperationCanceledException` when they fail by timeout, so if task is completed, it has actually exited.

karthiknadig: Got it. Then this is fine.

---
In reply to: [133773230](https://github.com/Microsoft/RTVS/pull/3867#discussion_r133773230) [](ancestors = 133773230)

Issue #3865, opened 2017-08-17
Pressing Esc to dismiss tooltip in REPL also clears commandline


Hong-Revo:
I think this is new in VS 2017 update 3; I don't recall seeing it in update 2.

To repro: in the REPL, type anything that will pop up a tooltip, eg `dir()`. Press Esc. The commandline is cleared.

MikhailArkhipov:
Reproes in PTVS too. Does not repro in 2015. Might be some subtle interactive window change.

MikhailArkhipov:
Will try and get it to 15.3.2

https://github.com/Microsoft/RTVS/commit/379c11ee640511132851b7b2fb9f3f5b91a3b2d2

Issue #3862, opened 2017-08-17
Performance issue on vs 2017 v15.3


eightrivers:
We upgraded VS 2017 from v15.2 to v15.3, and the speed of typing R code in R interactive became some slowly. It always takes about 2 or more seconds to show the intellisense panel when typing some codes, such as ls(), dir(), options(), or custom variables. The same thing also happened when trying to edit a R code file. 

We had tried in several physical machines, and all machines installed the v15.3 have the same issue, while the machines with v15.2 are OK. We even try the way to uninstall the components related to R (R tools, MS R Client) and reinstall it again, but nothing happened.

The environment we tried on is:

- OS: Win 10 Enterprise x64 CHS, version: 10.0.15063
- CPU: i7-6820HQ 2.70GHz
- Mem: 64G
- Visual Studio: Enterprise 2017 v15.3.0
- R tool: version: 1.1.30623.0940
- R interpreter : Microsoft R Client 3.3.2.0, in local machine

Hong-Revo:
I think I've seen this as well, but only in the case where I use `::` to look inside a package. Eg typing `devtools::install_github` slows down dramatically after the `::` is typed.

eightrivers:
In my side, it is some slow when trying anything in R interactive window or code editor, Eg, typing ls(), it will be about 1-2 seconds the character 's' will be shown after typing the character 'l'.

MikhailArkhipov:
Found it

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/379c11ee640511132851b7b2fb9f3f5b91a3b2d2

Will try and get it to 15.3.2

eightrivers:
Today we installed a new vs.net upgrade package, and the version of RTVS has been upgraded to 1.1.30804.1005. Seems the issue is still in there. So would you please kindly let us now when the vs.net 2017 v15.3.2 could be available? Thanks!

MikhailArkhipov:
It is available now as far as I can see.

eightrivers:
Get it, already updated, the issue has been solved. Thanks!

PR #3869, opened 2017-08-17
3862, 3865

MikhailArkhipov:
This is Release/1_0 aka 15.3

#3862 Performance issue on vs 2017 v15.3 
- Do not wait for the function description (same as in master/15.4)
- Simplify completion list filtering

#3865 Pressing Esc to dismiss tooltip in REPL also clears commandline
- Reproes in PTVS as well. Fix is to handle ESC manually and just dismiss completion if it is up as F# and C# do.

Build is in progress. Will be merging after testing/validation.


src/R/Editor/Impl/Completion/RFunctionCompletion.cs
```diff
@@ -36,9 +35,11 @@ public class RFunctionCompletion : RCompletion {
         }
 
         private void TryFetchDescription() {
-            Task.Run(async () => {
-                SetDescription(await _functionIndex.GetFunctionInfoAsync(this.DisplayText));
-            }).Wait(500);
+            _functionIndex.GetFunctionInfoAsync(this.DisplayText).ContinueWith(t => {
+                if (t.IsCompleted) {
```
AlexanderSher: It will always be completed at this point. Did you want to check `Status == RunToCompletion`?

src/R/Editor/Impl/Completion/RCompletionSet.cs
```diff
@@ -46,49 +46,11 @@ internal sealed class RCompletionSet : CompletionSet {
...
-            // Match at least something
-            int i = 0;
-            for (i = 0; i < Math.Min(typedText.Length, compText.Length); i++) {
-                if (char.ToLowerInvariant(typedText[i]) != char.ToLowerInvariant(compText[i])) {
-                    return i;
-                }
-            }
-
-            return i;
+            _completions.ForEach(x => ((RCompletion)x).IsVisible = x.DisplayText.StartsWithIgnoreCase(typedText));
```
AlexanderSher: Did this really affect performance? It looks like the same linear complexity as before.

src/R/Editor/Impl/Completion/RFunctionCompletion.cs
MikhailArkhipov: So `IsFaulted` and `IsCanceled` are set along with completion?

src/R/Editor/Impl/Completion/RCompletionSet.cs
MikhailArkhipov: ~24% total. It is still linear but does not attempt to build dictionaries. Dictionaries I believe were added on HTML times to do completion for `a-b-c` attributes.

src/R/Editor/Impl/Completion/RFunctionCompletion.cs
AlexanderSher: `Completed` means that it is either `IsFaulted` or `IsCanceled` or `Status == RunToCompletion`

MikhailArkhipov: OK

AlexanderSher:
👍

MikhailArkhipov:
Merging for signed build since it takes a while

PR #3885, opened 2017-08-23
Adds tests and fixes local docker finder.

karthiknadig:


src/Linux/Containers/Impl/Docker/LinuxDockerService.cs
```diff
@@ -85,46 +72,12 @@ public class LinuxDockerService : LocalDockerService, IContainerService {
         }
 
         protected override LocalDocker GetLocalDocker() {
-            _docker = _docker ?? GetLocalDocker(_services);
-            CheckIfServiceIsRunning();
+            _docker = _docker ?? _dockerFinder.GetLocalDocker();
+            _dockerFinder.CheckIfServiceIsRunning();
```
AlexanderSher: Shouldn't we do this before `_dockerFinder.GetLocalDocker();`?

src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -29,6 +29,11 @@ public abstract class LocalDockerService : IDockerService {
             // TODO: No instance of IActionLogWriter is exported in default IServiceContainer. Need scope support.
         }
 
+        public Task<string> BuildImageAsync(string buildOptions, CancellationToken ct) {
```
AlexanderSher: Should it be `public` or `protected`?

src/Windows/Containers/Test/RegistryKeyMock.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Common.Core.OS;
+using Microsoft.Win32;
+
+namespace Microsoft.R.Containers.Windows.Test {
+    class RegistryKeyMock : IRegistryKey {
```
AlexanderSher: We have registry mocks in `Microsoft.Common.Core.Test` project. You can simply reference it.

src/Windows/Containers/Test/WindowsDockerFinderTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.R.Containers.Docker;
+using Microsoft.UnitTests.Core.XUnit;
+
+namespace Microsoft.R.Containers.Windows.Test {
+    [ExcludeFromCodeCoverage]
+    [Category.Threads]
+    public class WindowsDockerFinderTests {
+        [Test]
+        public void FindDockerInstallationTest() {
+            IServiceManager services= new ServiceManager()
```
AlexanderSher: `var`

src/Linux/Containers/Impl/Docker/LinuxDockerService.cs
karthiknadig: Then we will always get service not running error. Instead of service is not installed error when docker is not installed.

AlexanderSher: Ok

AlexanderSher:
👍

src/Windows/Containers/Test/RegistryKeyMock.cs
karthiknadig: Will switch to the one in Core.Test, that looks better.

Issue #3872, opened 2017-08-17
package::func() does not always bring signature help


MikhailArkhipov:
Try `ggplot2::aes(` - no signature. Now, load the package. Try `aes(` or `ggplot::aes(` - now it works.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/54426fb36461586ce1e7ca1e1466614ccdd69cee

Issue #3831, opened 2017-08-11
Unable to connect to Linux box


MikhailArkhipov:
System.Net.WebException: 'The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.'
AuthenticationException: The remote certificate is invalid according to the validation procedure.

Reason: HTTP error while creating session on the machine 'Linux'. 
Exception: The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/54426fb36461586ce1e7ca1e1466614ccdd69cee

Issue #3811, opened 2017-08-05
Live Markdown: R block does not execute if there are spaces between braces


karthiknadig:
if the R block starts with {r} and the cursor is at the end of the line pressing enter causes {r} to be changed to { r }. the block stops updating after this.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/54426fb36461586ce1e7ca1e1466614ccdd69cee

PR #3889, opened 2017-08-25
Bug fixes

### LINKED ISSUE #3888: Pressing Esc in REPL doesn't clear commandline
This is in update 3.2. Looks like the fix to #3865 broke something else.

To repro: type anything in the REPL. Press Esc. What was typed remains there.

MikhailArkhipov:
#3888 Pressing Esc in REPL doesn't clear commandline 
- Move 'CanInterrupt' to session extensions so it can be accessed from other places
- Move `IDebuggerModeTracker` down to Host.Client
- Fix `InterruptRCommand` accordingly

#3811 Live Markdown: R block does not execute if there are spaces between braces 
- Fix formatting so it does not format `{r` block in autoformat (call host to make sure range is formattable)

#3872 package::func() does not always bring signature help 
- Add async information fetching method to IFunctionIndex
- make sure function index data gets updated
- Add test

#3831 Unable to connect to Linux box 
- Simplify cert verification code, make sure we only return `false` when appropriate

Also:
- Add additional tests for `{r` block extraction
- In markdown R code formatting correct calculation of the formattable range


src/Host/Client/Impl/Debugging/IDebuggerModeTracker.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.R.Components.InteractiveWorkflow {
+namespace Microsoft.R.Host.Client.Debugging {
```
AlexanderSher: I'd rather define `CanInterrupt` in `Components` assembly and make it extension to `IDebuggerModeTracker`.

MikhailArkhipov: But it is session we are interrupting. The fact that it checks the debugger state is implementation detail.

src/Windows/R/Editor/Impl/Completions/RCompletionSet.cs
```diff
@@ -40,12 +40,48 @@ internal sealed class RCompletionSet : CompletionSet {
...
+            var matches = new Dictionary<int, List<Completion>>();
+            var maxKey = 0;
+
             var typedText = GetTypedText();
             if (typedText.Length == 0) {
                 return;
             }
 
-            _completions.ForEach(x => ((RCompletion)x).IsVisible = x.DisplayText.StartsWithIgnoreCase(typedText));
+            foreach (var c in _completions) {
```
MikhailArkhipov: Restore original code. 15.3 fix was too aggressive

src/Package/Impl/Repl/Commands/ReplCommandController.cs
```diff
@@ -159,18 +159,17 @@ public class ReplCommandController : ViewController {
...
-                    return true;
-                }
-            } else {
+            if (controller.HasActiveCompletionSession || controller.HasActiveSignatureSession(TextView)) {
                 controller.DismissAllSessions();
                 return true;
             }
+
+            // If session is reading user input, do not terminate it
+            if (!Workflow.RSession.IsReadingUserInput && Workflow.RSession.CanInterrupt(Services)) {
```
AlexanderSher: `CanInterrupt` is checked in `InterruptRCommand` already. Why do we need to check it here as well?

src/R/Editor/Impl/Extensions/StringExtensions.cs
```diff
@@ -16,5 +17,27 @@ public static class StringExtensions {
...
+            if (start < 0) {
+                return TextRange.FromBounds(0, 0);
+            }
+
+            var bc = new BraceCounter<char>('{', '}');
+            var end = start;
+            bc.CountBrace(text[end]);
+            while (bc.Count > 0 && end < text.Length - 1) {
+                end++;
+                bc.CountBrace(text[end]);
```
AlexanderSher: Don't you need to re-instantiate `BraceCounter` in the loop?

src/Package/Impl/Repl/Commands/ReplCommandController.cs
MikhailArkhipov: I guess we can try and post command anyway... Let me see.

src/R/Editor/Impl/Extensions/StringExtensions.cs
MikhailArkhipov: No, it counts braces in the loop (it is a stack effectively) `bc.Count == 0` when braces match.

AlexanderSher: Ok, do you care about cases like `}{}`?

src/R/Editor/Impl/Completions/Providers/PackageFunctionCompletionProvider.cs
```diff
@@ -45,9 +45,10 @@ public class PackageFunctionCompletionProvider : IRCompletionListProvider, IRHel
         public IReadOnlyCollection<ICompletionEntry> GetEntries(IRIntellisenseContext context) {
             var completions = new List<ICompletionEntry>();
             var infoSource = _snippetInformationSource?.InformationSource;
-            var packages = GetPackages(context);
+            var packages = GetPackages(context).ToList();
```
AlexanderSher: Can this method be called on UI thread?

src/R/Editor/Impl/Extensions/StringExtensions.cs
MikhailArkhipov: Not really. It is up to the caller. The code will detect `{}`

src/Package/Impl/Repl/Commands/ReplCommandController.cs
MikhailArkhipov: OK, simplified it back

src/R/Editor/Impl/Completions/Providers/PackageFunctionCompletionProvider.cs
MikhailArkhipov: It is actually only called on UI thread

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -72,16 +68,16 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflowVisual {
             History = historyProvider.CreateRHistory(this);
             Packages = packagesProvider.CreateRPackageManager(_settings, this);
             Plots = plotsProvider.CreatePlotManager(_settings, this, coreShell.FileSystem());
-            _operations = new RInteractiveWorkflowOperations(this, _debuggerModeTracker, Shell);
+            _operations = new RInteractiveWorkflowOperations(this, debuggerModeTracker, Shell);
 
-            _activeTextViewTracker.LastActiveTextViewChanged += LastActiveTextViewChanged;
+            activeTextViewTracker1.LastActiveTextViewChanged += LastActiveTextViewChanged;
```
AlexanderSher: Why do you need another variable instead of `activeTextViewTracker`?

src/R/Editor/Impl/Completions/Providers/PackageFunctionCompletionProvider.cs
AlexanderSher: Then you may want to use `ITaskService.Wait` for the `GetAllFilePackagesAsync`. That will prevent UI from freezing

AlexanderSher: It returns `false` if we stopped waiting by timeout.

src/Package/Impl/Repl/Commands/ReplCommandController.cs
```diff
@@ -14,6 +13,7 @@
 using Microsoft.R.Editor.Completions;
 using Microsoft.R.Editor.Document;
 using Microsoft.R.Editor.Formatting;
+using Microsoft.R.Host.Client.Session;
```
AlexanderSher: Probably not needed anymore

src/Package/Impl/Repl/VsRInteractiveWorkflowProvider.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.R.Components.InteractiveWorkflow.Implementation;
 using Microsoft.R.Components.PackageManager;
 using Microsoft.R.Components.Plots;
+using Microsoft.R.Host.Client.Debugging;
```
AlexanderSher: Same here

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
MikhailArkhipov: Resharper change lol

src/Host/Client/Impl/Microsoft.R.Host.Client.csproj
```diff
@@ -13,7 +13,7 @@
   <ItemGroup>
     <PackageReference Include="MicroBuild.Core" Version="0.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.WebSockets.Protocol" Version="0.1.0" />
-    <PackageReference Include="System.Net.Http.WinHttpHandler" Version="4.3.0" />
+    <PackageReference Include="System.Net.Http.WinHttpHandler" Version="4.4.0" />
```
AlexanderSher: I had type resolution problems because of this change. Do you need it?

MikhailArkhipov: Not really. Seems to work, but I can roll back if you had issues.

AlexanderSher:
👍

Issue #3898, opened 2017-08-28
NRE in CreateToolWindow


karthiknadig:
For toolWindowGuid = {99d2ea62-72f2-33be-afc8-b8ce6e43b5d0}
```
System.NullReferenceException occurred
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Microsoft.VisualStudio.R.Package
  StackTrace:
   at Microsoft.VisualStudio.R.Packages.R.RPackageToolWindowProvider.CreateToolWindow(Guid toolWindowGuid, Int32 id) in G:\GIT\RTVS\src\Package\Impl\Packages\R\RPackageToolWindowProvider.cs:line 29
   at Microsoft.VisualStudio.R.Packages.R.RPackage.CreateToolWindow(Guid& toolWindowType, Int32 id) in G:\GIT\RTVS\src\Package\Impl\Packages\R\RPackage.cs:line 186
   at Microsoft.VisualStudio.Shell.Package.Microsoft.VisualStudio.Shell.Interop.IVsToolWindowFactory.CreateToolWindow(Guid& toolWindowType, UInt32 id)
   at Microsoft.VisualStudio.Platform.WindowManagement.WindowFrame.ConstructContent()
```

MikhailArkhipov:
Yeah, I fixed it already as part of #3771

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/235d2372fc802648005bd71d2291feaf7c2f84f4

Issue #3772, opened 2017-07-28
Sorting tibble data_frames in View pane results in NAs


mkoohafkan:
When `View`ing a `data_frame` from (tidyverse package `tibble`, attempting to sort by a column in the View pane results in NAs. It does not change the values in the variable, just what is being populated in the View table.
    
    library(tibble)
    test = data_frame(x = sample(100, 26), y = sample(LETTERS, 26))
    View(test)

and then try to sort by a column.

**Version Info**

Visual Studio Professional 2015 v. 14.0.25431.01 Update 3
RTVS: 1.1.30627.1321

**SessionInfo**

R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252 LC_CTYPE=English_United States.437
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C
[5] LC_TIME=English_United States.1252

attached base packages:
[1] stats graphics grDevices utils datasets methods base

loaded via a namespace (and not attached):
[1] colorspace_1.3-2 scales_0.4.1 compiler_3.4.1 lazyeval_0.2.0
[5] plyr_1.8.4 gtable_0.2.0 tibble_1.3.3 Rcpp_0.12.12
[9] ggplot2_2.2.1 grid_3.4.1 rtvs_1.0.0.0 rlang_0.1.1
[13] munsell_0.4.3

MikhailArkhipov:
Related #3771

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/235d2372fc802648005bd71d2291feaf7c2f84f4

Issue #3771, opened 2017-07-28
Can't View .Last.value


mkoohafkan:
The `View` command doesn't work with `.Last.value`. Simple example:

    data.frame(x = sample(1:100, 26), y = sample(LETTERS, 26))
    View(.Last.value)    # nothing happens

    data.frame(x = sample(1:100, 26), y = sample(LETTERS, 26))
    test = .Last.value
    View(test)
  
This problem also occurs with Piping:

    library(magrittr)
    test %>% View()    # nothing happens

**Version Info**

Visual Studio Professional 2015 v. 14.0.25431.01 Update 3
RTVS: 1.1.30627.1321

**SessionInfo**

R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.437    
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] colorspace_1.3-2 scales_0.4.1     compiler_3.4.1   lazyeval_0.2.0  
 [5] plyr_1.8.4       gtable_0.2.0     tibble_1.3.3     Rcpp_0.12.12    
 [9] ggplot2_2.2.1    grid_3.4.1       rtvs_1.0.0.0     rlang_0.1.1     
[13] munsell_0.4.3

int19h:
I think `.Last.value` doesn't work, because `View()` itself changes its value (to `NULL`), and the grid dynamically tracks changes to the variable/expression it's opened for, and auto-refreshes itself.

The difference in observed behavior with RStudio seems to stem from the fact that RS only tracks changes done via explicit assignment, and so it doesn't notice the implicit change to `.Last.value` from every REPL eval, while RTVS tracks all changes, no matter how they're done.

mkoohafkan:
This was observed in RGui (not RStudio) but I presume the underlying difference is the same. This is a a bit of a limitation with respect to piping `%>%` operations though, since it can be very helpful to `View` the results when constructing/debugging a dplyr chain without assigning a variable every time.

int19h:
[...]

PR #3901, opened 2017-08-29
Data grid on snapshots

MikhailArkhipov:
#3772 Sorting tibble data_frames in View pane results in NAs 
#3771 Can't View .Last.value 
- introduce *default* mode that takes snapshot of data as data frame (also discussed in #3231)

#3898 NRE in CreateToolWindow 

Other
- Remove PII data from telemetry
- small simplification in caret position determination (prevents exceptions when editing in REPL - they get caught but anyway)

### DIFF
--- a/src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
+++ b/src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
@@ -6,13 +6,13 @@ using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Common.Core.Services;
-using Microsoft.Common.Core.Shell;
 using Microsoft.Common.Core.Threading;
+using Microsoft.R.Components.Settings;
 using Microsoft.R.DataInspection;
-using Microsoft.VisualStudio.R.Package.Shell;
 using Microsoft.VisualStudio.R.Package.Utilities;
 using Microsoft.VisualStudio.Shell.Interop;
 using static Microsoft.R.DataInspection.REvaluationResultProperties;
+using static System.FormattableString;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
     internal abstract class GridViewerBase : ViewerBase, IObjectDetailsViewer {
@@ -20,11 +20,8 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
         private const REvaluationResultProperties _properties =
            ClassesProperty | ExpressionProperty | TypeNameProperty | DimProperty | LengthProperty;
 
-        private readonly IServiceContainer _services;
-
-        public GridViewerBase(IServiceContainer services, IDataObjectEvaluator evaluator) : base(evaluator) {
-            _services = services;
-        }
+        protected GridViewerBase(IServiceContainer services, IDataObjectEvaluator evaluator) : 
+            base(services, evaluator) { }
 
         #region IObjectDetailsViewer
         public ViewerCapabilities Capabilities => ViewerCapabilities.List | ViewerCapabilities.Table;
@@ -32,9 +29,12 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
         public abstract bool CanView(IRValueInfo evaluation);
 
         public async Task ViewAsync(string expression, string title, CancellationToken cancellationToken = default(CancellationToken)) {
+            if (!Services.GetService<IRSettings>().GridDynamicEvaluation) {
+                expression = Invariant($"as.data.frame({expression})");
+            }
             var evaluation = await EvaluateAsync(expression, _properties, RValueRepresentations.Str(), cancellationToken);
             if (evaluation != null) {
-                await _services.MainThread().SwitchToAsync(cancellationToken);
+                await Services.MainThread().SwitchToAsync(cancellationToken);
                 var id = Math.Abs(_toolWindowIdBase + expression.GetHashCode() % (Int32.MaxValue - _toolWindowIdBase));
 
                 var pane = ToolWindowUtilities.FindWindowPane<VariableGridWindowPane>(id);
@@ -46,8 +46,8 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
                     frame?.Show();
                 }
 
-                title = !string.IsNullOrEmpty(title) ? title : evaluation.Expression;
-                pane.SetEvaluation(new VariableViewModel(evaluation, _services), title);
+                title = !string.IsNullOrEmpty(title) ? title : expression;
+                pane.SetEvaluation(new VariableViewModel(evaluation, Services), title);
             }
         }
         #endregion
--- a/src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
+++ b/src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
@@ -11,6 +11,7 @@ using Microsoft.R.DataInspection;
 using Microsoft.VisualStudio.R.Package.Utilities;
 using Microsoft.VisualStudio.Shell.Interop;
 using static Microsoft.R.DataInspection.REvaluationResultProperties;
+using static System.FormattableString;
 
 namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
     internal abstract class GridViewerBase : ViewerBase, IObjectDetailsViewer {
@@ -27,6 +28,7 @@ namespace Microsoft.VisualStudio.R.Package.DataInspect.Viewers {
         public abstract bool CanView(IRValueInfo evaluation);
 
         public async Task ViewAsync(string expression, string title, CancellationToken cancellationToken = default(CancellationToken)) {
+            expression = Invariant($"as.data.frame({expression})");
             var evaluation = await EvaluateAsync(expression, _properties, RValueRepresentations.Str(), cancellationToken);
             if (evaluation != null) {
                 await Services.MainThread().SwitchToAsync(cancellationToken);
--- a/src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
+++ b/src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.Common.Core.Services;
 using Microsoft.Common.Core.Threading;
 using Microsoft.R.DataInspection;
+using Microsoft.R.Editor.Data;
 using Microsoft.VisualStudio.R.Package.Utilities;
 using Microsoft.VisualStudio.Shell.Interop;
 using static Microsoft.R.DataInspection.REvaluationResultProperties;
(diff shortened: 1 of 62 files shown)


src/Package/Impl/DataInspect/Viewers/GridViewerBase.cs
```diff
@@ -6,35 +6,35 @@
...
+            base(services, evaluator) { }
 
         #region IObjectDetailsViewer
         public ViewerCapabilities Capabilities => ViewerCapabilities.List | ViewerCapabilities.Table;
 
         public abstract bool CanView(IRValueInfo evaluation);
 
         public async Task ViewAsync(string expression, string title, CancellationToken cancellationToken = default(CancellationToken)) {
+            if (!Services.GetService<IRSettings>().GridDynamicEvaluation) {
+                expression = Invariant($"as.data.frame({expression})");
```
int19h: Why convert to data frame?

MikhailArkhipov: Works for #3772

int19h: But then why do it conditional on `GridDynamicEvaluation`?

More importantly, converting to a data frame is a lossy operation for matrices and a bunch of other stuff. The code in `grid.R` actually goes to great lengths to avoid doing this for that exact reason (originally, it did such a conversion, and I kept finding corner cases where it would do the wrong thing). 

We need to figure out what exactly makes tibble different for sorting purposes, and adjust `grid.R` accordingly, to fix this right. The docs for tibble seem to indicate that `arrange()` is the right function for the job, but it would require substantial rework of our code to make use of it. 

If a quick fix is needed right now, then let's do `as.data.frame` in grid.R only for tibbles.

int19h: (It looks like the easiest way to do the type check without having to load the tibble library is to check for `"tbl"` in `class()`)

MikhailArkhipov: I wanted to avoid changing dynamic case since it is tailored to large sets specifically. We can try check for tibbles, it gets more complicated though since row selector expression is passed separately and just converting `x` in grid.R doesn't work.

MikhailArkhipov: One thing i want to check before investing more is to get telemetry on how many uses do we get on dynamic mode. RS seems to be happy without it?

int19h: RStudio data grid is dynamic by default, actually. It just doesn't capture all changes as reliably as we do - they seem to be trying to refresh it only when they see `<-` in REPL input, based on my observarions, while we refresh on eveery command. So if you define a function that changes your data, and then later invoke that function, the grid doesn't refresh. But e.g. try this:
```r
mtcars <- mtcars
View(mtcars)
mtcars[[1, 1]] <- mtcars[[1, 1]] + 1
```

int19h: Converting `x` first thing in `grid_data` ought to work. The row selector is passed as a function, and that function accepts the dataset as the argument, which is passed by `grid_data` (I believe we use `x` as the name for that argument when we generate the call to `grid_data`, but it's not the same `x` as `grid_data` itself accepts). So if you change `x`, the row selector should just get the changed value as the argument, and apply its filtering logic to that - which should work as expected with a data frame.

MikhailArkhipov: OK. This prob will render unnecessary first part specific to 1D

int19h:
Are you sure you're running all the tests? With matrices, one of the problems is that if they have unnamed columns, data frame assigns them names based on indices. As I recall, there were some tests for that in `EvaluationWrapperTest` that I didn't create equivalents for `GridDataTest` yet - `MatrixOnlyColumnNameTest` is one example. 

Yes, we do have to special-case a bunch of stuff, unfortunately. This would be easier if there was a common interface for 2D data that was guaranteed to work the same, but the whole point of tibble and zoo is that they're different in that regard. We already do something special for tibble, in fact (you can find the comment to that effect in grid.R), just not via class name check, so this isn't adding anything new. Luckily, both zoo and tibble are "foundational" packages in a sense that they provide the basic structure, and then others (like e.g. quantmod) build on them, or on data frames and matrices; so we don't have to support dozens of different types. 

When I have more time, I'd like to go back to the drawing board and see if I can derive a more fundamental interface for these that would not require explicit class checks (or `as.data.frame`), so that the code can be as uniform as possible, or at least checking for unexpected operation results, and not type names. But for a quick fix, doing a data frame conversion only for tibbles should do it while minimizing opportunity for regressions elsewhere.

MikhailArkhipov:
I don't agree here. We show matrix row/column names differently from what users are accustomed to (in both RS and RGui) - with changes display is now traditional. If we had conversion to data frame early we would not have to spend time chasing bug reports and having non-working grid until we manage to update it some time later. Same for taking snapshots. For 99% cases it gives what user expects and matches what they got used to.

int19h:
My philosophy has been that debugger should be as unobtrusive as possible. Meaning that it shouldn't quietly substitute row and column names, or in any other way inject pieces of data and metadata that were not there in the original stuff (e.g. I should be able to distinguish a matrix with unnamed columns from the one that has columns named "V1" etc, because they behave differently in the language itself). When it comes to column names, it deliberately shows them using the notation that can be applied to index the data in R itself, just like expanded children in Variable Explorer. That is the philosophy that all VS debuggers for other languages generally follow; RS does not. I personally prefer the VS approach, but it's certainly debatable.

But there are other reasons to avoid the conversion. When it happens, it by necessity has to copy the entire dataset - so if you have a 1Gb matrix, for example, you end up needing another 1Gb to hold that temporary dataframe copy (converting a non-dataframe to a dataframe is not copy-on-write, like simple assignments). And the copy is then immediately discarded as soon as it's sliced to retrieve the subset of data that is being fetched. This effectively doubles the amount of memory necessary to deal with large data sets - where before you could get away with 4Gb, you now need 8Gb etc. For large datasets, it also makes scrolling slow because that copying is pretty expensive, and it happens every time the viewport changes.

 Again, the code in grid.R is written with that in mind - it tries to avoid copying until it slices, so that afterwards it only has to deal with a small chunk, which it can copy to its heart's content. One case where it deviates from this is when it handles 1D data - it does it by converting them to a matrix, which causes a copy - but that also was a shortcut, and it's not on the "hot path" memory wise. So this change will introduce a memory use regression, as well.

To remind, snapshots are not the default behavior in RS. They are the default behavior in RGui only. Furthermore, true snapshots require caching the entire dataset - not refreshing the grid on updates doesn't actually produce a true snapshot, because scrolling around will still load slices of data dynamically, so it'll pick up new values if you e.g. scroll away and then back, or resize the grid; we end up with a dynamic grid still, but the one that is now inconsistent in its updating behavior in weird ways. That's why only RGui has snapshot behavior - because it's the only one that doesn't lazily load grid data.

int19h:
Also note that if dataframe conversion is actually desirable, or when it is necessary to work around some issue, it's trivial for the user to do `View(as.data.frame(x))` in lieu of `View(x)`, and opt-in into that behavior.

MikhailArkhipov:
+ @jflam 

Although it is simple to add 'as.data.frame' user don't have to do it in RS or RGui. We can discuss this internally but to a) better match existing products and b) reduce ongoing maintenance costs we want to keep DF conversion in all cases.

int19h:
The decision is yours. I just wanted to make it clear that there are important downsides there as well, not just benefits alone. If you feel that the benefits outweigh the downsides, then let's go ahead with that.

Hong-Revo:
Just a note: rather than `any(class(x) == "classname")`, the idiomatic way is `inherits(x, "classname")`. This is important particularly when working with S4 objects, where `class` always returns a length-1 vector (ie, you won't get the full inheritance list).

src/Windows/R/Components/Impl/InteractiveWorkflow/Implementation/RInteractiveWorkflow.cs
```diff
@@ -183,5 +185,11 @@ public sealed class RInteractiveWorkflow : IRInteractiveWorkflowVisual {
                 } catch (OperationCanceledException) { }
             }
         }
+
+        private async Task SetGridEvalMode() {
+            try {
+                await RSession.SetGridEvalModeAsync(_settings.GridDynamicEvaluation);
```
AlexanderSher: You don't need to catch `OperationCanceledException` if you use `DoNotWait`

src/Package/Impl/Telemetry/RtvsTelemetry.cs
```diff
@@ -171,7 +172,11 @@ internal class WindowEvents {
                                 CompletionEnabled = _editorSettings.CompletionEnabled,
                                 SyntaxCheckInRepl = _editorSettings.SyntaxCheckInRepl,
                                 PartialArgumentNameMatch = _editorSettings.PartialArgumentNameMatch,
-                                RCommandLineArguments = _settings.LastActiveConnection?.RCommandLineArguments ?? string.Empty,
+
+                                // Do not report command line arguments - they may contain 
+                                // PII information and they are not that interesting anyway.
+                                // RCommandLineArguments = _settings.LastActiveConnection?.RCommandLineArguments ?? string.Empty,
```
AlexanderSher: Shouldn't we just delete this line?

MikhailArkhipov: I guess. Added comment so no one would put it back :-)

src/Package/Impl/DataInspect/VisualGrid/VisualGridScroller.cs
```diff
@@ -106,18 +106,18 @@ internal sealed class VisualGridScroller {
...
 
-                            bool execute = true;
+                            var execute = true;
                             // if next command is same the current one, skip to next (new one) for optimization
                             if (i < (batch.Count - 1)) {
                                 if (IsRepeating(batch, i, RepeatSkip)) {
                                     execute = false;
                                 } else if (IsRepeating(batch, i, RepeatAccum)) {
-                                    batch[i + 1].Param = (double)batch[i + 1].Param + (double)batch[i].Param;
+                                    batch[i + 1].Param = Convert.ToDouble(batch[i + 1].Param) + Convert.ToDouble(batch[i].Param);
```
AlexanderSher: `double.ToDouble` is implemented like this:
```cs
    double IConvertible.ToDouble(IFormatProvider provider)
    {
      return this;
    }
```

On the other hand, if type doesn't match, it means that it is unsynchronized with code that post commands, so it should fail on casting.

MikhailArkhipov: Hmmm, I've seen object coming as `int` which caused `(double)o` cast to throw.

AlexanderSher: This can be the case. `int` usually means row or column index. `double` is usually a pixel. So if it happens, we may send, for instance, offset as row or column index instead of offset in pixels.

AlexanderSher: The whole system of untyped command arguments is a bug prone, but I had no time to change it into proper producer/consumer.

MikhailArkhipov: Mouse wheel scroll was int

Issue #3911, opened 2017-08-30
Intellisense shows irrelevant text for squiggles


karthiknadig:
![image](https://user-images.githubusercontent.com/3840081/29891182-1408c87a-8d7f-11e7-9cee-49a3fcec53bd.png)

```
permits <- read.csv("https://data.seattle.gov/api/views/mags-97de/rows.csv?accessType=DOWNLOAD", stringsAsFactors =FALSE )
permits$Status <- as.factor(permits$Status)

library(dplyr)
permits %>%
    group_by(Status) %>%
    summarise(count = length(Status)) %>%
```

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/eab57cd565512d22dce04fa67733bdc386591dd0

Issue #3899, opened 2017-08-28
Intellisense not shown for back-tick variables


karthiknadig:
Before pressing back-tick:
![image](https://user-images.githubusercontent.com/3840081/29794495-101b3c26-8bfd-11e7-8d61-c689ddade477.png)

After pressing back-tick (no intellisense):
![image](https://user-images.githubusercontent.com/3840081/29794600-807b00b4-8bfd-11e7-8b5b-25bb4ab3c9bd.png)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/eab57cd565512d22dce04fa67733bdc386591dd0

Issue #3379, opened 2017-04-01
Autocompletion for ::: should contain all of the functions from the package namespace environment, not just the exported ones.


mvacha:
I know it's not very common to use ::: for calling not exported functions but it still would be nice to show proper intellisense for it. RBox for Sublime "supports it" (on the other hand the autocomplete for :: is wrong in the RBox).

I have found relevant issue #2507, but it seems to me, that it got fixed only for the rtvs "special package".

RBox:
![namespace](https://cloud.githubusercontent.com/assets/2319571/24581495/ec4c830e-171c-11e7-9b38-70aacb9f1c1e.PNG)

RTVS:
![rtvs](https://cloud.githubusercontent.com/assets/2319571/24581500/f0cc7a24-171c-11e7-941a-c6b85e096e26.PNG)

MikhailArkhipov:
We currently use 

```
package.functions.names <- function(packageName) {
    as.list(union(
		tryCatch({
			getNamespaceExports(packageName)
		}, error = function(e) {
			return(c())
		}), 
		tryCatch({
			ls(paste0('package:', packageName))
		}, error = function(e) {
			return(c())
		})
	))
}
```
located in packages.R in C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\Extensions\Microsoft\R Tools....\rtvs

We could try

`ls(getNamespace("testthat"), all.names = TRUE)`

instead. You can change it directly in the RTVS package and try :-)

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/eab57cd565512d22dce04fa67733bdc386591dd0

PR #3927, opened 2017-09-05
#3379 #3911 #3899

MikhailArkhipov:
This is 15.5

#3379 Autocompletion for ::: should contain all of the functions from the package namespace environment, not just the exported ones. 

- Add ability to fetch all functions
- Separate caching of exported and internal functions
- Change disk cache to have version and store information on internal/external
- Make sure internal `foo` does not override exported `foo` in intellisense
- Show separate icon in completion for internal functions

#3911 Intellisense shows irrelevant text for squiggles 
- Use passed range in light bulb and quick info rather than the caret position

#3899 Intellisense not shown for back-tick variables 
- make it like RS and show entry without backticks but insert actual text

- Fix scrolling batch in grid
- Add tests for the above


src/Windows/R/Editor/Impl/SuggestedActions/RSuggestedActionsSource.cs
```diff
@@ -88,20 +82,26 @@ internal sealed class RSuggestedActionsSource : SuggestedActionsSourceBase, ISug
...
         #endregion
+
+        /// <remarks>
+        /// Picks position in range to fetch suggested acions. 
+        /// Ideally range content should be split and analyzed 
+        /// but this is TODO in the future.
+        /// </remarks>
+        private static int PositionFromRange(SnapshotSpan range) {
+            var text = range.Snapshot.GetText(range);
+            var nonWsOffset = text.Length - text.TrimStart().Length;
```
AlexanderSher: Looks like an overkill to create a new string just to find its index, but since there is a TODO, let's leave it as it is.

src/R/Editor/Impl/Extensions/StringExtensions.cs
```diff
@@ -18,6 +18,13 @@ public static class StringExtensions {
             return name;
         }
 
+        public static string RemoveBackticks(this string name) {
+            if (!string.IsNullOrEmpty(name) && name.Length >= 2 && name[0] == '`' && name[name.Length-1] == '`') {
```
karthiknadig: >name[0] == '`' && name[name.Length-1] == '`' [](start = 67, length = 44)

.Trim('`') ?

PR #3928, opened 2017-09-05
Minor fixes and improvements

karthiknadig:
1. RTVS daemon installation uses the snakeoil pem and key as the default self-signed SSL certificate.
2. Fix and updates to container build.

### DIFF
--- a/src/Containers/Impl/Docker/LocalDockerService.cs
+++ b/src/Containers/Impl/Docker/LocalDockerService.cs
@@ -138,7 +138,8 @@ namespace Microsoft.R.Containers.Docker {
                 RedirectStandardError = true,
                 RedirectStandardInput = true,
                 RedirectStandardOutput = true,
-                UseShellExecute = false
+                UseShellExecute = false,
+                CreateNoWindow = true
             };
 
             var process = _ps.Start(psi);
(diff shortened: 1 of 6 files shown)


src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -149,13 +150,24 @@ public Task<string> RepositoryLogoutAsync(RepositoryCredentials auth, Cancellati
 
             var output = await process.StandardOutput.ReadToEndAsync();
             var error = await process.StandardError.ReadToEndAsync();
-            if (!string.IsNullOrEmpty(error)) {
+            if (!string.IsNullOrEmpty(error)  && !IsSecurityWarning(error)) {
                 _outputLogWriter?.Write(MessageCategory.Error, error);
                 throw new ContainerException(error);
             }
 
+            if (!process.HasExited) {
```
AlexanderSher: How is this possible?

karthiknadig: I missed removing this code. I had added this while testing the "it leaves behind a window" problem.

AlexanderSher:

karthiknadig:
@int19h This is the same thing I discussed with you over email. Basically using the snakeoil pem and key to generate the PFX. This removes the need for the user to create their own SSL cert.

PR #3936, opened 2017-09-11
Containers Window

AlexanderSher:
- New Containers Window which contains list of available containers and allows creation of new local Docker containers
- `ContainerManager` component that monitors local containers status
- `IOutputService` and `IOutput` that unifies console output and provide access to the VS Output window
- `ServiceContainerExtensions.Extend` method which provides extension form of child containers
- Several refactorings related to the tool windows


src/Common/Core/Impl/Threading/MainThreadExtensions.cs
```diff
@@ -12,6 +14,14 @@ public static MainThreadAwaitable SwitchToAsync(this IMainThread mainThread, Can
         public static bool CheckAccess(this IMainThread mainThread)
             => Thread.CurrentThread.ManagedThreadId == mainThread.ThreadId;
 
+
+        [Conditional("TRACE")]
+        public static void Assert(this IMainThread mainThread, [CallerMemberName] string memberName = "", [CallerFilePath] string sourceFilePath = "", [CallerLineNumber] int sourceLineNumber = 0) {
+            if (mainThread.ThreadId != Thread.CurrentThread.ManagedThreadId) {
+                Debug.Fail(FormattableString.Invariant($"{memberName} at {sourceFilePath}:{sourceLineNumber} was incorrectly called from a background thread."));
```
karthiknadig: Should be in resource file?

AlexanderSher: Don't think so. It is debug-only information. @MikhailArkhipov , what do you think?

MikhailArkhipov: No, it is just debug string. Users won't see it. Besides, you don't want to see it in Chinese when you debug localized VS ;-)

src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -4,43 +4,57 @@
...
+            _outputService = services.GetService<IOutputService>();
+        }
+        
+        public async Task<IContainer> CreateContainerFromFileAsync(BuildImageParameters buildParams, CancellationToken ct) {
+            await TaskUtilities.SwitchToBackgroundThread();
+
+            var buildOptions = $"-t {buildParams.Image}:{buildParams.Tag} {Path.GetDirectoryName(buildParams.DockerfilePath)}";
+            await BuildImageAsync(buildOptions, ct);
+
+            var createOptions = Invariant($"-p 5444:5444 --name {buildParams.Name} {buildParams.Image}:{buildParams.Tag} rtvsd");
```
karthiknadig: Should we need to parameter-ize the host port. "-p {hostport}:5444"?

src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -4,43 +4,57 @@
...
 
         public async Task<IEnumerable<IContainer>> ListContainersAsync(bool getAll = true, CancellationToken ct = default(CancellationToken)) {
             await TaskUtilities.SwitchToBackgroundThread();
 
             var command = "ps";
             var commandOptions = getAll ? "-a -q" : "-q";
-            var output = await ExecuteCommandAsync(Invariant($"{command} {commandOptions}"), _defaultTimeout, true, ct);
-            var lines = output.Split(new string[] { "\n", "\r\n" }, StringSplitOptions.RemoveEmptyEntries);
+            var output = await ExecuteCommandAsync(Invariant($"{command} {commandOptions}"), null, _defaultTimeout, true, ct);
+            var lines = output.Split(new[] { "\n", "\r\n" }, StringSplitOptions.RemoveEmptyEntries);
```
MikhailArkhipov: I think it will split on `\n` even if it is `\r\n`, no ? Prob doesn't matter here though.

src/Containers/Impl/Docker/LocalDockerService.cs
```diff
@@ -77,86 +91,93 @@ public abstract class LocalDockerService : IDockerService {
...
+                while (!process.StandardOutput.EndOfStream) {
+                    var line = await process.StandardOutput.ReadLineAsync();
+                    if (printOutput) {
+                        _output.WriteLine(Invariant($"{outputPrefix}> {line}"));
+                    }
+                    result.AppendLine(line);
+                }
+
                 await process.WaitForExitAsync(timeoutms, ct);
             } catch(OperationCanceledException) when (!failOnTimeout && !ct.IsCancellationRequested){
```
MikhailArkhipov: You may want to catch `IOException` just in case `process.StandardOutput` reading fails.

AlexanderSher: Ok

src/Package/Impl/Shell/Services/VsOutputService.cs
```diff
@@ -0,0 +1,51 @@
...
+        private IMainThread _mainThread;
+
+        public VsOutputService(IServiceContainer services) {
+            _services = services;
+            _mainThread = services.MainThread();
+            _outputs = new Dictionary<string, IOutput>();
+        }
+
+        public async Task<IOutput> GetAsync(string name, CancellationToken cancellationToken) {
+            await _mainThread.SwitchToAsync(cancellationToken);
```
MikhailArkhipov: Does it need to be main thread? `OutputWindowLogWriter` needs main thread only when it needs to show pane for the first time in `EnsurePaneVisible`. Otherwise it is `OutputStringThreadSafe`

src/Containers/Impl/Docker/LocalDockerService.cs
AlexanderSher: @karthiknadig , what do you think?

src/Package/Impl/Shell/Services/VsOutputService.cs
AlexanderSher: Agreed.

src/R/Components/Impl/ConnectionManager/Connection.cs
```diff
@@ -14,21 +16,29 @@ public class Connection : ConnectionInfo, IConnection {
 
         public static Connection Create(ISecurityService securityService, string name, string path, string rCommandLineArguments, bool isUserCreated, bool fetchHostLoad) {
```
MikhailArkhipov: At some point it may be worth actually splitting class into base connection, `UrlConnection`, `ContainerConnection` and `LocalConnection` and have `Create` as factory since we have several `IsABC` properties which is generally a bad form.

src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
```diff
@@ -0,0 +1,143 @@
...
+                await UpdateContainersOnceAsync(cancellationToken);
+            }
+        }
+
+        public async Task<IContainer> CreateLocalDockerAsync(string name, string username, string password, string version, CancellationToken cancellationToken = default(CancellationToken)) {
+            if (string.IsNullOrWhiteSpace(version)) {
+                version = "3.4.1";
+            }
+
+            var dockerImageContent = $@"FROM kvnadig/rtvsd-ub1604:{version}
```
MikhailArkhipov: I'd put this in a file

src/Windows/Core/Impl/OS/RegistryKeyImpl.cs
```diff
@@ -17,7 +17,7 @@ internal sealed class RegistryKeyImpl : IRegistryKey {
         }
 
         public string[] GetSubKeyNames() => _key?.GetSubKeyNames() ?? new string[0];
-        public string[] GetValueNames() => _key.GetValueNames() ?? new string[0];
+        public string[] GetValueNames() => _key.GetValueNames();
```
MikhailArkhipov: `_key?.GetValueNames() ?? new string[0]` ?

AlexanderSher: Good catch

src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
AlexanderSher: Ok, should I just put it in the folder with RTVS assemblies?

src/Package/Impl/Options/R/Tools/RSettingsImplementation.cs
```diff
@@ -126,6 +128,16 @@ internal sealed class RSettingsImplementation : BindableBase, IRSettings {
             set => SetProperty(ref _lastActiveConnection, value);
         }
 
+        public string LastLocalDockerUsername {
+            get => _lastLocalDockerUsername;
+            set => SetProperty(ref _lastLocalDockerUsername, value);
+        }
+
+        public string LastLocalDockerPassword {
```
MikhailArkhipov: This appears to be per VS session as it is not persisted in the Tools | Options. Expected?

MikhailArkhipov: If it does need to be persisted it may need to be instead stored in the Windows cred storage.

MikhailArkhipov:
👍 

Can be polished in the next couple of weeks as we go to 15.5

src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
MikhailArkhipov: Yes, it will do. We (or user) can then edit it without touching the code.

src/Containers/Impl/Docker/LocalDockerService.cs
karthiknadig: The string actually splits on `\r\n`. The split algorithm compares the first character of the split list against the current character of the string. Then it compares the remaining to determine where it needs to cut. So, if the split list has both `\n` and `\r\n`, it will cut the following string, `sample = "abcd\r\nefgh"`, using `\r\n`, instead of `\n`. Since the current character match when it is at  `sample[4] == '\r'`.
http://referencesource.microsoft.com/#mscorlib/system/string.cs,95f22bdebc0c3230

src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
AlexanderSher: https://github.com/Microsoft/RTVS/issues/3947 - for some reason `CopyToOutputDirectory` doesn't work in this project (works perfectly in `Microsoft.UnitTests.Core.Linux`).

src/Package/Impl/Options/R/Tools/RSettingsImplementation.cs
AlexanderSher: https://github.com/Microsoft/RTVS/issues/3939
`it is not persisted in the Tools | Options` - it expected to be similar to `LastActiveConnection`. Did I miss something?

MikhailArkhipov: Won't be encrypted...

AlexanderSher: Ok, we need some place to store it. For #3939, I wanted to save credentials using container name (which is also a connection name) as an authority. I can create another authority (something like "RTVS:LastDockerCredentials") and get username from it. But I'll have to convert password into normal string to pass it to the docker.

MikhailArkhipov: https://technet.microsoft.com/library/jj554668.aspx

I believe we have `NativeMethods` for it

PR #3971, opened 2017-09-14
Added container image listing and updated tests.

karthiknadig:
#3969


src/Containers/Impl/Docker/LocalDockerImage.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.R.Containers.Docker {
+    public class LocalDockerImage : IContainerImage {
```
AlexanderSher: This class doesn't have any logic expect its own creation. I'd rather make `ContainerImage` class
```cs
public interface ContainerImage {
    public string Id { get; }
    public string Name { get; }
    public string Tag { get; }

    public ContainerImage (string id, string name, string tag) {
       Id = id;
       Name = name;
       Tag = tag;
    }

    public void Deconstruct(out string id, out string name, out string tag) {
       id = Id;
       name = Name;
       tag = Tag;
    }
}
```
and get input values from `JToken` in `LocalDockerService` itself.

src/Windows/Containers/Test/OutputMock.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.R.Common.Core.Output;
+
+namespace Microsoft.R.Containers.Windows.Test {
+    public class OutputMock : IOutput {
```
AlexanderSher: There is `Microsoft.Common.Core.Test.Fakes.Shell.TestOutputService` that writes output in log file that becomes test artifact.

Issue #3980, opened 2017-09-15
Show watermark that docker isn't installed on Containers window instead of UI


AlexanderSher:

MikhailArkhipov:
Does not show watermark. I don't have docker

![image](https://user-images.githubusercontent.com/12820357/30659346-59ebad34-9df2-11e7-8066-ca3d7af6dddb.png)

AlexanderSher:
Yes, I've forgotten that I didn't add an event that changes VM properties.

Issue #3978, opened 2017-09-15
UI issues in Create Container panel


MikhailArkhipov:
![image](https://user-images.githubusercontent.com/12820357/30504020-6f34213c-9a21-11e7-9709-a7f72acfa48e.png)

It says 'Workspaces' and 'Local'. The window already has 'Containers' title, so 'Workspaces' is not needed. 'Local' probably 'Local Docker'? Also, single letter name is not supported.

Entries to list of docker containers should have docker icon. Can we find small docker image or maybe just blue 'D' ? or boxes like in

![image](https://user-images.githubusercontent.com/12820357/30504188-3049342a-9a22-11e7-9799-402ed01b0d12.png)

![image](https://user-images.githubusercontent.com/12820357/30504207-4afe548a-9a22-11e7-8c6a-c64bc1f9388b.png)

AlexanderSher:
> Also, single letter name is not supported.

It isn't supported by docker: https://github.com/moby/moby/issues/3138

AlexanderSher:
- "Workspaces" button is removed
- "Local" is renamed to "Local Docker Containers"
- Single letter names aren't supported by docker (https://github.com/moby/moby/issues/3138 )
- Docker brand guidelines (https://www.docker.com/brand-guidelines) do not allow us to use Docker logo without 'docker' letters. With them, it looks quite ugly:
![image](https://user-images.githubusercontent.com/8933858/30663548-9d169484-9e10-11e7-8a71-b06e794a9d62.png)
Do we really need identical icon for every line in the list?

MikhailArkhipov:
[cid:image002.jpg@01D33213.421AB3D0]

MikhailArkhipov:
![image](https://user-images.githubusercontent.com/12820357/30665651-851d8de2-9e06-11e7-9738-82bef428b0bb.png)

Issue #3976, opened 2017-09-15
Container dialog 'Version' field should have tooltip/watermark with explanation


MikhailArkhipov:
What exactly user should type: R3.4? 3.4.0? Maybe pre-fill with 'latest'?

AlexanderSher:
Maybe just a drop-down?

MikhailArkhipov:
Maybe... But where do we get the list? what to do with new versions?

MikhailArkhipov:
Still reproes

![image](https://user-images.githubusercontent.com/12820357/30659090-abbd0d70-9df1-11e7-9442-27308c48a0d9.png)

AlexanderSher:
Tooltip was overridden by Watermark. Should work after #3996 is checked in.

Issue #3967, opened 2017-09-14
Use docker file hash as the image name


karthiknadig:
Currently we use a randomly generated Guid as the image name. That means a new image is created each time user tries to create a container.

AlexanderSher:
How can we find out that image with that name already exists?

karthiknadig:
I can provide the API for this.

PR #3992, opened 2017-09-19
Container manager bug fixes

### LINKED ISSUE #3939: Save credentials for created local docker container

AlexanderSher:
#3939: Save credentials for created local docker container
#3952: Container password field should show system password character
#3963: Delete should be disabled for running containers (it actually shows the popup)
#3964: Connection is not removed after the container is deleted
#3967: Use docker file hash as the image name
#3976: Container dialog 'Version' field should have tooltip/watermark with explanation
#3978: UI issues in Create Container panel
#3980: Show watermark on Containers window instead of UI that docker isn't installed

### DIFF
--- a/src/Windows/Core/Impl/Security/SecurityService.cs
+++ b/src/Windows/Core/Impl/Security/SecurityService.cs
@@ -24,7 +24,8 @@ namespace Microsoft.Common.Core.Security {
 
         public Credentials GetUserCredentials(string authority, string workspaceName) {
             _services.MainThread().CheckAccess();
-            return ReadSavedCredentials(authority) ?? PromptForWindowsCredentials(authority, workspaceName);
+            var (username, password) = ReadUserCredentials(authority);
+            return username != null ? Credentials.Create(username, password) : PromptForWindowsCredentials(authority, workspaceName);
         }
 
         public bool ValidateX509Certificate(X509Certificate certificate, string message) {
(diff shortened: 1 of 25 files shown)


src/Windows/Core/Impl/Security/SecurityService.cs
```diff
@@ -60,23 +61,38 @@ public class SecurityService : ISecurityService {
...
-        private Credentials ReadSavedCredentials(string authority) {
-            using (var ch = CredentialHandle.ReadFromCredentialManager(authority)) {
-                if (ch != null) {
-                    var credData = ch.GetCredentialData();
-                    return Credentials.Create(credData.UserName, SecurityUtilities.SecureStringFromNativeBuffer(credData.CredentialBlob));
+        public void SaveUserCredentials(string authority, string userName, SecureString password, bool save) {
+            var creds = default(NativeMethods.CredentialData);
+            try {
+                creds.TargetName = authority;
+                // We have to save the credentials even if user selected NOT to save. Otherwise, user will be asked to enter
```
MikhailArkhipov: This sounds insecure. Why are we not respecting user's intent not to save? Perhaps they want to put container somewhere or have encrypted disk and want data to be secure. This may not pass sec review.

AlexanderSher: This isn't related to the container. These credentials are used by connection to that container and aren't stored in the container itself. The same logic is used when we connect to any broker using URL.

src/Windows/R/Components/Impl/Resources.resx
```diff
@@ -442,12 +442,18 @@ If you want to delete all data on the remote machine you must delete your user p
   <data name="ConnectionManager_RemoveConnectionConfirmation" xml:space="preserve">
     <value>Are you sure you want to remove the '{0}' connection?</value>
   </data>
+  <data name="ContainerManager_ContainerServiceIsNotInstalled" xml:space="preserve">
+    <value>Docker service isn't installed.\nTo start working with containers, install Docker for Windows and restart VS.</value>
+  </data>
+  <data name="ContainerManager_ContainerServiceIsNotRunning" xml:space="preserve">
+    <value>Docker service isn't running.\nTo start working with containers, run Docker for Windows and restart VS.</value>
```
MikhailArkhipov: Restart Visual Studio

src/Windows/R/Components/Impl/Resources.resx
```diff
@@ -442,12 +442,18 @@ If you want to delete all data on the remote machine you must delete your user p
   <data name="ConnectionManager_RemoveConnectionConfirmation" xml:space="preserve">
     <value>Are you sure you want to remove the '{0}' connection?</value>
   </data>
+  <data name="ContainerManager_ContainerServiceIsNotInstalled" xml:space="preserve">
+    <value>Docker service isn't installed.\nTo start working with containers, install Docker for Windows and restart VS.</value>
```
MikhailArkhipov: We usually auto-open the download page. `ProcessServices.Start('https://www.docker.com/docker-windows')`

MikhailArkhipov: Does `\n` actually work here? since it is `xml:space="preserve"` it should be just regular formatting

src/Windows/Core/Impl/Security/SecurityService.cs
MikhailArkhipov: But then it is even less secure. We allow reconnection without re-entering creds?

src/Windows/R/Components/Impl/Resources.resx
AlexanderSher: Do we want to open it every time user clicks on the "Containers" button?

AlexanderSher: Good catch.

MikhailArkhipov: Only after reporting the error. Same as in say, markdown publishing and pandoc

src/Windows/Core/Impl/Security/SecurityService.cs
AlexanderSher: https://msdn.microsoft.com/en-us/library/windows/desktop/aa374788(v=vs.85).aspx


Value | Meaning
-- | --
CRED_PERSIST_SESSION 1 (0x1) | The credential persists for the life of the logon session.  It will not be visible to other logon sessions of this same user. It  will not exist after this user logs off and back on.
CRED_PERSIST_LOCAL_MACHINE 2 (0x2) | The credential persists for all subsequent logon sessions on this  same computer. It is visible to other logon sessions of this same user  on this same computer and not visible to logon sessions for this user on  other computers. Windows Vista Home Basic, Windows Vista Home Premium, Windows Vista Starter and Windows XP Home Edition:  This value is not supported.
CRED_PERSIST_ENTERPRISE 3 (0x3) | The credential persists for all subsequent logon sessions on this  same computer. It is visible to other logon sessions of this same user  on this same computer and to logon sessions for this user on other  computers. This option can be implemented as locally persisted credential if the  administrator or user configures the user account to not have roam-able  state. For instance, if the user has no roaming profile, the credential  will only persist locally. Windows Vista Home Basic, Windows Vista Home Premium, Windows Vista Starter and Windows XP Home Edition:  This value is not supported.

Since we save it for CRED_PERSIST_SESSION, user won't be able to reuse these credentials from other sessions.

MikhailArkhipov: OK, Need to study it more though...

MikhailArkhipov:
👍

src/Windows/R/Components/Impl/Resources.resx
AlexanderSher: Error is reported in the watermark. I can add a clickable link to the message. Will that work?

MikhailArkhipov: Sure

src/Windows/Core/Impl/Security/SecurityService.cs
karthiknadig: The case were this is iffy, is when the user says they don't want to save the creds. The problem is if we don't save it (temporarily) any connection attempt will request user to enter credentials. if the save is disabled, you will see credentials prompt show to you at least 4 times (Test connection, repl connection, intellisense, etc), before the connection is ready. Any call to broker (except About and info), will require the credentials and credential window will be presented. The iffy part is when this information is removed. Currently it is removed when user logs-out of windows . But we could remove it on connection switch, if the save flag is not set. But that will be an explicit remove.

PR #3996, opened 2017-09-20
Fix #3980: Show watermark that docker isn't installed on Containers window instead of UI

### LINKED ISSUE #3980: Show watermark that docker isn't installed on Containers window instead of UI

AlexanderSher:
Underlying logic + link


MikhailArkhipov:
👍

PR #4021, opened 2017-09-21
Auto select random host port for a container

### LINKED ISSUE #4000: Unable to create container if broker process is up
It can be up for many reasons...
```
build> SECURITY WARNING: You are building a Docker image from 
Windows against a non-Windows Docker host. All files and directories added to build context 
will have '-rwxr-xr-x' permissions. It is recommended to double check and reset permissions 
for sensitive files and directories.
create> 23b672db28103d7a087f83cd024c9779e6a94da0cd8db60c65d603f569e19305
start 23b672db28103d7a087f83cd024c9779e6a94da0cd8db60c65d603f569e19305> 

ERROR: Error response from daemon: driver failed programming external connectivity on
 endpoint AA (ddad806e29e341c8d98dabf6f7700a05ec619fd6ed27535695214cdfdb13d9c9): 
Error starting userland proxy: Bind for 0.0.0.0:5444 failed: port is already allocated
Error: failed to start containers:
23b672db28103d7a087f83cd024c9779e6a94da0cd8db60c65d603f569e19305
```

karthiknadig:
Fixes #4000

### DIFF
--- a/src/Common/Core/Impl/Net/PortUtil.cs
+++ b/src/Common/Core/Impl/Net/PortUtil.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Net;
+using System.Net.Sockets;
+
+namespace Microsoft.Common.Core.Net {
+    public static class PortUtil {
+        private static Random _rand = new Random();
+        public static int GetAvailablePort(int minPort, int maxPort) {
+            while (true) {
+                int port = _rand.Next(minPort, maxPort + 1);
+                try {
+                    TcpListener listener = new TcpListener(IPAddress.Loopback, port);
+                    listener.Start();
+                    // port is available
+                    listener.Stop();
+                    return port;
+                } catch (SocketException ex) when (ex.SocketErrorCode == SocketError.AddressAlreadyInUse){
+                    // port busy, try another port
+                }
+            }
+        }
+
+        public static bool IsPortAvailable(int port) {
+            try {
+                TcpListener listener = new TcpListener(IPAddress.Loopback, port);
+                listener.Start();
+
+                // port is available
+                listener.Stop();
+                return true;
+            } catch (SocketException ex) when (ex.SocketErrorCode == SocketError.AddressAlreadyInUse) {
+                return false;
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Common/Core/Impl/Net/PortUtil.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Net;
+using System.Net.Sockets;
+
+namespace Microsoft.Common.Core.Net {
+    public static class PortUtil {
+        private static Random _rand = new Random();
+        public static int GetAvailablePort(int minPort, int maxPort) {
+            while (true) {
+                int port = _rand.Next(minPort, maxPort + 1);
```
AlexanderSher: Can we do it for several ports in parallel?

src/Windows/R/Components/Impl/ContainerManager/Implementation/View/ContainerManagerControl.xaml
```diff
@@ -152,7 +153,15 @@
                              ToolTip="{x:Static components:Resources.ContainerManager_CreateLocalDocker_VersionTooltip}"
                              AutomationProperties.Name="{x:Static components:Resources.ContainerManager_CreateLocalDocker_VersionTooltip}" />
 
-                    <WrapPanel Grid.Row="5" Grid.Column="1" Grid.ColumnSpan="2" Margin="0,0,0,0">
+                    <imaging:CrispImage Grid.Row="4" Grid.Column="0" Style="{StaticResource ValidityIndicatorStyle}"
+                                        cwpf:AttachedProperties.IsValid="{Binding Path=IsPortAvailable}" />
+
+                    <TextBox Grid.Row="4" Grid.Column="1" Style="{StaticResource TextBoxSingleLineStyle}"
+                             Text="{Binding Path=Port, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
+                             ToolTip="{x:Static components:Resources.ContainerManager_CreateLocalDocker_PortTooltip}"
```
AlexanderSher: You may want to add `rcontrols:Watermark.TextBoxHint="Local port"` or something like that

src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
```diff
@@ -82,7 +82,7 @@ internal class ContainerManager : IContainerManager {
             }
         }
 
-        public async Task<IContainer> CreateLocalDockerAsync(string name, string username, string password, string version, CancellationToken cancellationToken = default(CancellationToken)) {
+        public async Task<IContainer> CreateLocalDockerAsync(string name, string username, string password, string version, int port, CancellationToken cancellationToken = default(CancellationToken)) {
             if (string.IsNullOrWhiteSpace(version)) {
                 version = "3.4.1";
```
MikhailArkhipov: Hm, why it is hardcoded? We can get latest from interpreter list perhaps? The list is local, but gives an idea what user is working with. Also, we could prepopulate with that number.

karthiknadig: It should be "latest". I will update this.

MikhailArkhipov:
👍

PR #4027, opened 2017-09-22
Three bug fixes

### LINKED ISSUE #3938: Containers "start" button should be disabled for busy ports
`IContainer` has information about its ports. We can disable "start" button for containers with same ports if one of them is running.

### LINKED ISSUE #4006: Renaming connection to container makes it go to Local
Create container
Rename connection
Went to local

### LINKED ISSUE #4007: Container connection properties should show container info
OR it should not be possible to rename it - see #4006. Better if it is possible to rename it but user would see name of the container.

AlexanderSher:
- Fix #3938: Containers "start" button should be disabled for busy ports
- Fix #4006: Renaming connection to container makes it go to Local
- Fix #4007: Container connection properties should show container info

### DIFF
--- a/src/R/Components/Impl/ConnectionManager/IConnection.cs
+++ b/src/R/Components/Impl/ConnectionManager/IConnection.cs
@@ -10,6 +10,8 @@ namespace Microsoft.R.Components.ConnectionManager {
 
         Uri Uri { get; }
 
+        string ContainerName { get; }
+
         /// <summary>
         /// If true, the connection is to a remote machine
         /// </summary>
--- a/src/Windows/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
+++ b/src/Windows/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
@@ -326,13 +326,12 @@ namespace Microsoft.R.Components.ConnectionManager.Implementation.ViewModel {
                 _localDockerConnections.Clear();
                 _remoteConnections.Clear();
                 foreach (var connection in ConnectionManager.RecentConnections.OrderBy(c => c.Name)) {
-                    var cvm = CreateConnectionViewModel(connection);
                     if (connection.IsRemote) {
-                        _remoteConnections.Add(cvm);
+                        CreateRemoteConnectionViewModel(connection);
                     } else if (connection.IsDocker) {
-                        _localDockerConnections.Add(cvm);
+                        CreateLocalDockerConnectionViewModel(connection);
                     } else {
-                        _localConnections.Add(cvm);
+                        CreateLocalConnectionViewModel(connection);
                     }
                 }
             }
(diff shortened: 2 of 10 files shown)


src/R/Components/Impl/ConnectionManager/IConnection.cs
```diff
@@ -10,6 +10,8 @@ public interface IConnection: IConnectionInfo {
 
         Uri Uri { get; }
 
+        string ContainerName { get; }
```
MikhailArkhipov: Does it need to be pesistent? May need to go to `IConnectionInfo`

src/Windows/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
```diff
@@ -326,13 +326,12 @@ internal sealed class ConnectionManagerViewModel : ConnectionStatusBaseViewModel
                 _localDockerConnections.Clear();
                 _remoteConnections.Clear();
                 foreach (var connection in ConnectionManager.RecentConnections.OrderBy(c => c.Name)) {
-                    var cvm = CreateConnectionViewModel(connection);
                     if (connection.IsRemote) {
```
MikhailArkhipov: Prob will benefit from some class derivation and virtual factories, but OK for now

src/Windows/R/Components/Impl/Resources.resx
```diff
@@ -348,6 +348,9 @@
   <data name="ConnectionManager_EditLocalTooltip_Format" xml:space="preserve">
     <value>Edit local connection '{0}'</value>
   </data>
+  <data name="ConnectionManager_EditLocalDockerTooltip_Format" xml:space="preserve">
+    <value>Edit local docker connection '{0}'</value>
```
MikhailArkhipov: Docker is trademark so probably needs to be capitalized. ex Wikipedia `Because Docker containers are so lightweight...`

MikhailArkhipov:
Need to fix `docker` -> `Docker`
👍

src/R/Components/Impl/ConnectionManager/IConnection.cs
AlexanderSher: Dunno. We can always find out if it is a docker connection or not by inspecting docker ports.

src/Windows/R/Components/Impl/ConnectionManager/Implementation/ViewModel/ConnectionManagerViewModel.cs
AlexanderSher: Not just factories, but actually types that handle different types of operations. #4029

src/Windows/R/Components/Impl/Resources.resx
AlexanderSher: Ok

Issue #4051, opened 2017-09-25
Bug in parsing of scopes


MikhailArkhipov:
In
```
x <- function() {
    a <- !b

    if (c) {
    }
}
```

last closing brace is squiggled. Removing nested `if` or `!` removes the squiggle

MikhailArkhipov:
Regression after https://github.com/Microsoft/RTVS/issues/3818

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/54601da3a4e4f0e7315f768e601a1f70d2bee8a3

Issue #4050, opened 2017-09-25
RxSpark function help wrong


Hong-Revo:
This is with the unsigned build that @MikhailArkhipov sent me.

![image](https://user-images.githubusercontent.com/11328890/30818126-06813194-a25e-11e7-81d8-d12f0db84fbc.png)

MikhailArkhipov:
There are probably multiple signatures and we pick the first one. We don't fully support multiple signatures at the moment. However... RD data parser should be just showing the bigger one.

MikhailArkhipov:
This is old. The signature contains
```
hdfsShareDir = paste( "/user/RevoShare", Sys.info()[["user"]], sep="/" )
``` 
and `[[` are causing RD/R hybrid parser trouble.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/54601da3a4e4f0e7315f768e601a1f70d2bee8a3

Issue #4046, opened 2017-09-24
Rmarkdown preview should respect code chunk options


Hong-Revo:
In particular, for something like

<pre>```{r, eval=FALSE}
# example code: do not execute
load("bigdataset.rdata")
rxNeuralNet(bigformula, data=bigdataset)
```
</pre>
it should just copy the code to the preview pane.

Hong-Revo:
(I'm referring to the green triangle button, "Run current code chunk and update preview pane".)

MikhailArkhipov:
It does read chunk options, just doesn't do anything with `eval`. So should then just copy output. OK. With comment or just the code?

Hong-Revo:
Yes, it should have the comment as well. You would generally have `eval=FALSE` if the chunk is meant as an example, so the code is meant for the reader.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/54601da3a4e4f0e7315f768e601a1f70d2bee8a3

Issue #4042, opened 2017-09-22
Hang in function quick info


MikhailArkhipov:
```
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskContext.NoMessagePumpSynchronizationContext.get()	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskContext.OnJoinableTaskStarted(Microsoft.VisualStudio.Threading.JoinableTask task)	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTask.JoinableTask(Microsoft.VisualStudio.Threading.JoinableTaskFactory owner, bool synchronouslyBlocking, Microsoft.VisualStudio.Threading.JoinableTaskCreationOptions creationOptions, System.Delegate initialDelegate)	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.RunAsync(System.Func<System.Threading.Tasks.Task> asyncMethod, bool synchronouslyBlocking, Microsoft.VisualStudio.Threading.JoinableTaskCreationOptions creationOptions, System.Delegate entrypointOverride)	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.Post(System.Threading.SendOrPostCallback callback, object state, bool mainThreadAffinitized)	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTask.Post(System.Threading.SendOrPostCallback d, object state, bool mainThreadAffinitized)	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.RequestSwitchToMainThread(System.Action callback)	Unknown	No symbols loaded.
 	Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.MainThreadAwaiter.OnCompleted(System.Action continuation)	Unknown	No symbols loaded.
[...]

PR #4055, opened 2017-09-25
Assorted fixes

MikhailArkhipov:
- Fix tests (added container service to the fixture)

#4031 Potential NRE in PackageIndex 
- most probably index was still being built when session got disposed

#4050 RxSpark function help wrong 
- old issue, [[ and ]] in RD confused parser. Fixed and added test.

#4036 Change local Docker connection icon in Workspaces 
- changed style
- aligned `Containers...` to the right and made it hyperlink to match `Add`

#4046 Rmarkdown preview should respect code chunk options 
- handle eval
- update test baselines

#4051 Bug in parsing of scopes 
- parser continued past line break
- do terminate expression parsing at the line break by returning proper last operation type
- added test

#4042 Hang in function quick info 
- could not repro, but added reentrancy check

### DIFF
--- a/src/R/Editor/Impl/Completions/Providers/ParameterNameCompletionProvider.cs
+++ b/src/R/Editor/Impl/Completions/Providers/ParameterNameCompletionProvider.cs
@@ -49,7 +49,10 @@ namespace Microsoft.R.Editor.Completions.Providers {
             // Collect parameter names from all signatures
             IEnumerable<KeyValuePair<string, IArgumentInfo>> arguments = new Dictionary<string, IArgumentInfo>();
             foreach (var signature in functionInfo.Signatures) {
-                var args = signature.Arguments.ToDictionary(x => x.Name);
+                var args = new Dictionary<string, IArgumentInfo>();
+                foreach (var arg in signature.Arguments.Where(x => !string.IsNullOrEmpty(x.Name))) {
+                    args[arg.Name] = arg;
+                }
                 arguments = arguments.Union(args);
             }
 
--- a/src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
+++ b/src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
@@ -35,7 +35,7 @@ namespace Microsoft.R.Editor.Data {
         private readonly Dictionary<string, IRSessionDataObject> _topLevelVariables = new Dictionary<string, IRSessionDataObject>();
         private bool _updating;
 
-        public WorkspaceVariableProvider(IServiceContainer services) : 
+        public WorkspaceVariableProvider(IServiceContainer services) :
             base(services.GetService<IRInteractiveWorkflowProvider>()) {
             _services = services;
         }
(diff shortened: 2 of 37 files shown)


MikhailArkhipov:
OK, review post mortem, need build for BB/integration.

src/R/Editor/Impl/Completions/Providers/ParameterNameCompletionProvider.cs
```diff
@@ -49,7 +49,10 @@ public class ParameterNameCompletionProvider : IRCompletionListProvider {
             // Collect parameter names from all signatures
             IEnumerable<KeyValuePair<string, IArgumentInfo>> arguments = new Dictionary<string, IArgumentInfo>();
             foreach (var signature in functionInfo.Signatures) {
-                var args = signature.Arguments.ToDictionary(x => x.Name);
+                var args = new Dictionary<string, IArgumentInfo>();
```
AlexanderSher: It could've been `signature.Arguments.Where(x => !string.IsNullOrEmpty(x.Name)).ToDictionary(x => x.Name);`, but up to you

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
```diff
@@ -90,16 +90,16 @@ internal sealed class WorkspaceVariableProvider : RSessionChangeWatcher, IVariab
 
                 IReadOnlyList<IREvaluationResultInfo> infoList = null;
 
-                async Task FillInfoList() {
+                Task.Run(async () => {
```
AlexanderSher: What was the issue?
`TryEvaluateAndDescribeAsync` is always executed on background thread, so the reason for `Task.Run` can be if current method should proceed while evaluation is running. But since we're synchronously waiting for the task, I don't understand the purpose.

src/R/Editor/Impl/Extensions/AstRootExtensions.cs
```diff
@@ -208,7 +208,7 @@ public static class AstRootExtensions {
         /// </summary>
         public static bool IsInObjectMemberName(this ITextProvider textProvider, int position) {
             if (position > 0) {
-                for (var i = position - 1; i >= 0; i--) {
+                for (var i = position - 1; i >= 0 && i < textProvider.Length; i--) {
```
AlexanderSher: `i` doesn't increment in the loop. So is it `textProvider.Length` that changes?

src/R/Editor/Impl/Functions/PackageIndex.cs
```diff
@@ -108,14 +108,17 @@ public sealed class PackageIndex : IPackageIndex {
                     Debug.WriteLine("R function host start: {0} ms", stopwatch.ElapsedMilliseconds);
 
                     // Fetch list of package functions from R session
+                    _disposableBag.ThrowIfDisposed();
```
AlexanderSher: Ideally we should've provide a `CancellationToken` to this method which is linked to both disposing and timeout. But that is a larger change.

src/R/Editor/Impl/Data/WorkspaceVariableProvider.cs
MikhailArkhipov: Performance regression per Hong. Type `mtcars$` - 2 sec delay, then nothing.

src/R/Editor/Impl/RData/Parser/RdFunctionSignature.cs
```diff
@@ -104,7 +105,7 @@ internal static class RdFunctionSignature {
                 }
             }
             // Should be past \method{...}{...}. Now skip signature
-            var bc = new BraceCounter<char>(new [] { '(', ')' });
+            var bc = new BraceCounter<char>(new[] { '(', ')' });
```
AlexanderSher: Can be just `var bc = new BraceCounter<char>('(', ')'); `

src/R/Editor/Impl/Extensions/AstRootExtensions.cs
MikhailArkhipov: No, when `textProvider.Length == 0` it throws when position is 1. Basically should not be dependent in position value.

src/R/Editor/Impl/Functions/PackageIndex.cs
MikhailArkhipov: Yes, I have it with ct in the LS branch.

PR #4089, opened 2017-09-28
Fix #4053: Occasionally connection status is displayed as yellow

### LINKED ISSUE #4053: Occasionally connection status is displayed as yellow
Doesn't happen always, but I do see it regularly when switching local connections or when container loses and then restores connection

image

AlexanderSher:
Triggers are replaced by bindings.


AlexanderSher:
Ping

MikhailArkhipov:
👍

PR #4106, opened 2017-10-02
Moving docker file contents to a template file.

### LINKED ISSUE #3947: Move Local Docker Template into file
`<content Update="Dockerfile" CopyToOutputDirectory="Always" />` doesn't work in `Microsoft.R.Components` project. `AfterBuild` task can copy file, but it isn't added to project artifacts.

karthiknadig:
fixes #3947


AlexanderSher:
👍

PR #4117, opened 2017-10-05
Adding install script to simplify and allow automation of linux installation

karthiknadig:


karthiknadig:
@int19h can you take a look at this one?

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,179 @@
...
+
+    -s, --silent                 perform a silent, unattended install
+    -l, --no-locale-gen          does not call locale-gen during install
+    -h, --help                   print this help text
+
+EOF
+}
+
+function parse_options {
+  local TEMP=$(getopt -o -lsuh --long no-locale-gen,silent,help -n "${PROG_NAME}" -- "$@")
```
int19h: "u" (in short options) is unused, looks like.

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,179 @@
...
+  if [[ "$(id -u)" != "0" ]]; then
+      error  "This script must be run with sudo or as root."
+      exit 1
+  fi
+}
+
+function check_getopt {
+  if [[ -z ${GETOPT_BIN} ]]; then
+    echo ""
+    echo "This script requires GNU getopt to run. Install this utility and rerun this script."
```
int19h: `getopt` is a standard POSIX utility, so it's always going to be there. But it might not be GNU getopt specifically. The use of long options above is a GNU extension, so this does indeed need GNU getopt. 

I think a better way to check for this is to invoke something like `getopt -o -x --long x`, and see if it fails or not (from error code).

int19h: (Or, alternatively, drop the long options, and use the strict POSIX subset of `getopt`)

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,179 @@
...
+  echo ""
+  exit 1
+}
+
+function install_deb_dpkg {
+    local package=$1
+    local log=$2
+
+    message "Installing ${package}..."
+    dpkg --install ${package} > ${log} 2>&1
```
int19h: It's probably best to show the output of dpkg for easy diagnostics. You can use `tee` to additionally make a copy in the log.

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,179 @@
...
+    else
+      message "ERROR!\n"
+      exit_error
+    fi
+}
+
+
+function install {
+  if [[ ${USE_DPKG} == true ]]; then
+    mkdir -p ${SCRIPT_DIR}/logs
```
int19h: This will cause the directory to be owned by root (since we require them to run this script as root). If they run it with `sudo`, that directory won't be deletable. It might be worth doing `chmod` on it to make it world-writable. 

Also, log location should probably be configurable. It might also be best to default to `/var/log`, per usual conventions.

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,179 @@
...
+    
+    if [[ ${DOTNET_INSTALLED} == false && ${DOTNET_SOURCE_ADDED} == false ]]; then
+      message "Adding dotnet package to source list ..."
+      sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ $(lsb_release -cs) main" > /etc/apt/sources.list.d/dotnetdev.list' > ${SCRIPT_DIR}/logs/source.list.txt 2>&1
+      apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893 > ${SCRIPT_DIR}/logs/apt-key-add.txt 2>&1
```
int19h: Adding a new trusted signing key to the list is an important point. We should definitely tell the user we're doing this, and it's probably best to actually prompt them whether this is okay (with an option to approve in advance for automated scenarios).

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,179 @@
...
+    apt-get update > ${SCRIPT_DIR}/logs/update.txt 2>&1
+    message "done\n"
+    
+    if [[ ${DOTNET_INSTALLED} == false ]]; then
+      message "Installing dotnet package..."
+      apt-get install -y dotnet-dev-1.0.4
+      message "done\n"
+    fi
+
+    message "Installing apt package dependencies ${DEBIAN_DEPENDENCIES}..."
```
int19h: What's our baseline Ubuntu version? If it's 16.04 (which is the most recent LTS version), then we can use `apt install ./foo.deb`, and let it take care of dependencies.

src/Linux/Setup/rtvs-install
```diff
@@ -0,0 +1,199 @@
...
+  echo "There was an error installing RTVS Daemon."
+  echo "Please check the logs at:"
+  echo "${LOG_DIR}"
+  echo ""
+  exit 1
+}
+
+function prompt_yes_no {
+    echo ""
+    echo "Dotnet package installation requires us to add a new trudted signing key."
```
Neil-Schneider: trusted*

PR #4118, opened 2017-10-06
UI fixes

### LINKED ISSUE #4110: Container window message may benefit from some margins
Text is a bit too close to right/left edges

image

### LINKED ISSUE #4111: Container window UI adjustments
- From containers window, have a link to bring up the dockerfile template in an editor
so that people dont have to hunt it down.  

a. usable for pointing the dockerhub url base image
b. usable for adding default pkgs
c. usable for a link to our docs

Also
- Link back to Workspaces and we have with Containers... there
- Some sort of indication that Create is clickable?
- Double click on container connects to it if it is running? Not sure if it is feasible though.

See also #4110

AlexanderSher:
- Fix #4110: Container window message may benefit from some margins
- Fix #4111: Container window UI adjustment

### DIFF
--- a/src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
+++ b/src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
@@ -18,6 +19,7 @@ using Microsoft.R.Containers;
 
 namespace Microsoft.R.Components.Containers.Implementation {
     internal class ContainerManager : IContainerManager {
+        private const string LatestVersion = "latest";
         private readonly IContainerService _containerService;
         private readonly IFileSystem _fileSystem;
         private readonly CountdownDisposable _containersChangedCountdown;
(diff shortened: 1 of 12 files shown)


src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
```diff
@@ -88,16 +90,37 @@ internal class ContainerManager : IContainerManager {
...
+            return CreateLocalDockerFromContentAsync(dockerImageContent, name, version, port, cancellationToken);
+        }
+
+        public async Task<IContainer> CreateLocalDockerFromFileAsync(string name, string filePath, int port, CancellationToken cancellationToken = default(CancellationToken)) {
+            var dockerImageContent = await LoadDockerImageContent(filePath, cancellationToken);
+            return await CreateLocalDockerFromContentAsync(dockerImageContent, name, LatestVersion, port, cancellationToken);
+        }
+
+        private static async Task<string> LoadDockerImageContent(string filePath, CancellationToken cancellationToken) {
+            using (var client = new HttpClient()) {
```
MikhailArkhipov: Do we handle HttpException etc somewhere?

MikhailArkhipov:
Hard to see without screenshot  :-) Make sure we don't crash if network is down. 👍

AlexanderSher:
Ok, let me add couple more exception handlers

PR #4121, opened 2017-10-06
More UI fixes

AlexanderSher:
- Add network exception handling 
- Fixed button styling
- Added "Workspaces..." button
- Made double-click on container name start container and switch to its default connection

### DIFF
--- a/src/Windows/R/Components/Impl/ContainerManager/Implementation/View/ContainerManagerControl.xaml.cs
+++ b/src/Windows/R/Components/Impl/ContainerManager/Implementation/View/ContainerManagerControl.xaml.cs
@@ -42,10 +42,17 @@ namespace Microsoft.R.Components.ContainerManager.Implementation.View {
 
         private void ButtonFromFile_Click(object sender, RoutedEventArgs e) => ViewModel.ShowLocalDockerFromFile();
 
+        private void ButtonShowWorkspaces_Click(object sender, RoutedEventArgs e) => ViewModel.ShowConnections();
+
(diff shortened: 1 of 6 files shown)


src/Windows/R/Components/Impl/ContainerManager/Implementation/View/ContainerManagerControl.xaml.cs
```diff
@@ -42,12 +42,19 @@ public partial class ContainerManagerControl {
...
 
+        private void ButtonShowWorkspaces_Click(object sender, RoutedEventArgs e) => ViewModel.ShowConnections();
+
         private void Container_PreviewKeyUp(object sender, KeyEventArgs keyEventArgs) {
             if (keyEventArgs.Key == Key.Delete && !(keyEventArgs.OriginalSource is TextBox)) {
                 ViewModel.DeleteAsync(GetContainer(keyEventArgs)).DoNotWait();
             }
         }
 
+        private void Container_MouseDoubleClick(object sender, MouseButtonEventArgs e) {
```
MikhailArkhipov: Enter/Space probably should do the same

MikhailArkhipov:
👍

src/Windows/R/Components/Impl/ContainerManager/Implementation/View/ContainerManagerControl.xaml.cs
AlexanderSher: You can't navigate to text itself with keyboard, only buttons can be selected. Workspaces window has similar behavior.

PR #4136, opened 2017-10-11
Updating dockerfile image name.

karthiknadig:
1. Image updated to the final image name.
2. Removing installation of rtvsd. This will be done in the base image.


AlexanderSher:
@karthiknadig , can `EXPOSE 5444` be moved to the base image as well?

karthiknadig:
Yes, I will Move that.

AlexanderSher:
👍

Issue #4141, opened 2017-10-12
Local process support for Linux broker


karthiknadig:
Linux broker currently supports only remote connections via run as user. For local client scenario this will need support in broker to start process as current user.

PR #4142, opened 2017-10-13
Local linux RHost scenario for VSCode.

karthiknadig:
This will be needed to run VSCode R on linux
#4141

### DIFF
--- a/src/Linux/Host/Broker/Impl/Services/Utility.cs
+++ b/src/Linux/Host/Broker/Impl/Services/Utility.cs
@@ -26,6 +26,24 @@ namespace Microsoft.R.Host.Broker.Services {
         private const string RtvsResult = "rtvs-result";
         private const string RtvsError = "rtvs-error";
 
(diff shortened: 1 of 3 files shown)


src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -26,6 +26,10 @@ public class Utility {
         private const string RtvsResult = "rtvs-result";
         private const string RtvsError = "rtvs-error";
 
+        public static IProcess RunAsCurentUser(ILogger<Session> logger, IProcessServices ps, string arguments, string rHomePath, string loadLibPath) {
```
AlexanderSher: RunAsCur**r**entUser

src/Linux/Host/Broker/Impl/Services/Utility.cs
```diff
@@ -156,5 +160,23 @@ public class Utility {
 
             return ps.Start(psi);
         }
+
+        private static IProcess CreateRHostProcess(IProcessServices ps, string arguments, string rHomePath, string loadLibPath) {
```
AlexanderSher: Do you need this as a separate method?

AlexanderSher:
👍

PR #4143, opened 2017-10-13
Container UI Fixes

### LINKED ISSUE #3940: Use dropdown instead of textbox for local docker creation

### LINKED ISSUE #4127: Use ARG in Dockerfile instead of {x}
This solves 2 problems:
- We can open a copy of the file to the user
- We don't have to create copies of docker templates for each version/username/password combination

AlexanderSher:
- Fix #4127: Use ARG in Dockerfile instead of {x}
- Fix #3940: Use dropdown instead of textbox for local docker creation

image


src/Debugger/Impl/project.json
```diff
@@ -7,7 +7,7 @@
     "Microsoft.VisualStudio.SDK.EmbedInteropTypes": "15.0.10",
     "Microsoft.VisualStudio.Shell.15.0": "15.0.26606",
     "Microsoft.VisualStudio.Shell.Interop.10.0": "10.0.30319",
-    "Microsoft.VisualStudio.Text.Data": "15.0.26703.11111"
```
MikhailArkhipov: Do key shortcuts work in the editor? I.e. Backspace for example. I remember that was the issue we've reverted for.

src/R/Components/Impl/Containers/Implementation/ContainerManager.cs
```diff
@@ -57,6 +55,21 @@ internal class ContainerManager : IContainerManager {
 
         public IReadOnlyList<IContainer> GetContainers() => _containers;
         public IReadOnlyList<IContainer> GetRunningContainers() => _runningContainers;
+
+        public async Task<IEnumerable<string>> GetLocalDockerVersions(CancellationToken cancellationToken = default (CancellationToken)) {
+            await TaskUtilities.SwitchToBackgroundThread();
+            const string imageName = "microsoft/rtvs"; // change to microsoft/rtvs;
```
MikhailArkhipov: Nit: comment is obsolete

src/Windows/R/Components/Impl/Resources.resx
```diff
@@ -446,37 +446,43 @@ If you want to delete all data on the remote machine you must delete your user p
...
   <data name="ContainerManager_ContainerServiceIsNotInstalled_FourthPart" xml:space="preserve">
     <value>refresh</value>
   </data>
+  <data name="ContainerManager_ContainerServiceIsNotInstalled_FifthPart" xml:space="preserve">
+    <value>.</value>
+  </data>
   <data name="ContainerManager_ContainerServiceIsNotRunning_FirstPart" xml:space="preserve">
     <value>Docker service isn't running.
-To start working with containers, run Docker for Windows and click</value>
+To start working with containers, run Docker for Windows and click </value>
```
MikhailArkhipov: Some languages do not allow construction of sentences from parts. I.e. 'refresh' may or may not appear at the end of the sentence depending on the language.

src/Windows/R/Components/Impl/Resources.resx
```diff
@@ -446,37 +446,43 @@ If you want to delete all data on the remote machine you must delete your user p
...
   </data>
   <data name="ContainerManager_ContainerServiceIsNotRunning_SecondPart" xml:space="preserve">
     <value>refresh</value>
   </data>
+  <data name="ContainerManager_ContainerServiceIsNotRunning_ThirdPart" xml:space="preserve">
+    <value>.</value>
```
MikhailArkhipov: Some languages do not have periods.

src/Windows/R/Components/Impl/Resources.resx
```diff
@@ -446,37 +446,43 @@ If you want to delete all data on the remote machine you must delete your user p
...
+  </data>
   <data name="ContainerManager_ContainerServiceHasErrors_FirstPart" xml:space="preserve">
     <value>Docker service fails because of the following error:
 </value>
   </data>
   <data name="ContainerManager_ContainerServiceHasErrors_SecondPart" xml:space="preserve">
     <value>
-Please click</value>
+Please click </value>
   </data>
```
MikhailArkhipov: As above. Pref to have complete sentence. Perhaps put 'Refresh' as a separate button rather than word and then have complete sentence like 'Click Refresh button below'

MikhailArkhipov:
Probably too late for 15.5 due to feature-like changes. Not sure about 15.6 timeline yet.

src/Debugger/Impl/project.json
AlexanderSher: Works fine with 15.5

src/Windows/R/Components/Impl/Resources.resx
AlexanderSher: yes, that's why there is `ContainerManager_ContainerServiceIsNotRunning_FirstPart` that goes before refresh and `ContainerManager_ContainerServiceIsNotRunning_ThirdPart` that goes after refresh. In other languages any of those strings can be empty, which doesn't affect formatting in general.

MikhailArkhipov:
👍

AlexanderSher:
So, should I merge it into 1.3 branch?

MikhailArkhipov:
Not yet. Telemetry change is OK to merge. 

@jflam. @smortaz  - opinions? The change is *small* feature level so we'd need to get it through VS Shiproom as 15.5 P2 is past free-for-all phase and there is no 15.5 P3, just RTW.

jflam:
What is the risk here? Is it just the change from textbox to a combo box?

MikhailArkhipov:
There is more - fetching list from the NET, handling various associated cases. Not a big risk, but it is a feature work. We'd have to describe it properly and get DevDiv approval for another signed build and PR.

jflam:
Is the workaround that the customer needs to lookup the list of containers manually under our account?

MikhailArkhipov:
Something like that or we list them in docs. You need them only if you need specific version. Otherwise the default is 'latest'.

PR #4186, opened 2017-10-30
Update MainThreadAwaiter

AlexanderSher:
Changes are required to match current behavior of VS `ThreadHelper`. Since it sets SynchronizationContext in `JoinableTaskFactory.MainThreadAwaite.GetResult`, we need to call it from our own awaiter.

This change may also fix some of the rare issues when VS gets hanged on connection switching.


MikhailArkhipov:
👍

Issue #4187, opened 2017-10-30
REPL may become slow after large data output


MikhailArkhipov:
From the user
```
I have noticed that if I accidentally print a large amount of data to the screen, 
like a large data.frame or vector, the interactive window becomes extremely slow.  
If you type something, there is a long delay before it appears in the interactive window.  
However, if I clear the interactive window (CTRL-L) the problem is resolved. 
```

AlexanderSher:
Is it somehow related to the size of the buffer?

MikhailArkhipov:
Maybe. Or number of projections.

MikhailArkhipov:
So it is a combination of the above. After dumping a lot of data the buffer becomes large so accessing snapshots becomes expensive. This is compounded by the fact that view is projected so accessing *view* snapshot requires buffer to construct it via the projection graph.

I'll see if we can use R document buffer instead of the view buffer. `IsPositionInNamespace` is the offender but it may not be the only one.

MikhailArkhipov:
https://github.com/Microsoft/RTVS/commit/8c804aa1c25c77605af1c52341ab04bbde1b70da

PR #4188, opened 2017-10-30
4187 REPL may become slow after large data output

MikhailArkhipov:
https://github.com/Microsoft/RTVS/issues/4187

Map caret down and use the active editor buffer instead which is  just a few lines typically and does not include all the output.

### DIFF
--- a/src/R.sln
+++ b/src/R.sln
@@ -130,9 +130,6 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.R.Containers.Windows.Test", "Windows\Containers\Test\Microsoft.R.Containers.Windows.Test.csproj", "{62E1CA8E-8398-46FC-B14F-DE0C5759DD2D}"
 EndProject
 Global
-	GlobalSection(Performance) = preSolution
-		HasPerformanceSessions = true
-	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Debug|x64 = Debug|x64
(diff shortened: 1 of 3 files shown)


src/R.sln
```diff
@@ -130,6 +130,9 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.R.Containers.Windows.Test", "Windows\Containers\Test\Microsoft.R.Containers.Windows.Test.csproj", "{62E1CA8E-8398-46FC-B14F-DE0C5759DD2D}"
 EndProject
 Global
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
```
AlexanderSher: ?

src/R/Editor/Impl/Completions/RCompletionContextExtensions.cs
```diff
@@ -30,12 +38,20 @@ public static bool IsCaretInNamespace(this IEditorView view, out bool tripleColo
         }
 
         public static bool IsCaretInLibraryStatement(this IRIntellisenseContext context)
-            => context.Session.View.IsCaretInLibraryStatement();
+            => context.Session.View.IsCaretInLibraryStatement(context.EditorBuffer);
```
AlexanderSher: Is it possible that `EditorBuffer` is null?

MikhailArkhipov: Nope.

src/R.sln
MikhailArkhipov: Oh boy, VS adds this when you use profiler... Its harmless, but lemme try and undo it.

PR #4213, opened 2017-11-08
4212 Remove hard fail when version of remote does not match

MikhailArkhipov:


src/Host/Client/Impl/Host/RemoteBrokerClient.cs
```diff
@@ -47,13 +47,13 @@ public RemoteBrokerClient(string name, IRSessionProvider sessionProvider, Broker
...
-
-            var aboutHost = await GetHostInformationAsync<AboutHost>(cancellationToken);
-            var brokerIncompatibleMessage = aboutHost?.IsHostVersionCompatible();
-            if (brokerIncompatibleMessage != null) {
-                throw new RHostDisconnectedException(brokerIncompatibleMessage);
+            if (connectionInfo.IsInteractive) {
+                var aboutHost = await GetHostInformationAsync<AboutHost>(cancellationToken);
+                var brokerVersionMessage = aboutHost?.IsHostVersionCompatible();
+                if (brokerVersionMessage != null) {
+                    _console.WriteError(brokerVersionMessage);
```
AlexanderSher: We don't handle protocol incompatibility anywhere. If we're still returning a host, this will be broken in a different place.

AlexanderSher:
👍

PR #4257, opened 2018-01-02
Deleting a container deletes the stored credentials.

karthiknadig:
#4215


src/Windows/R/Components/Impl/ContainerManager/Implementation/ViewModel/ContainerManagerViewModel.cs
```diff
@@ -295,6 +295,7 @@ private void ContrainersChangedMainThread()
 
             try {
                 await _containers.DeleteAsync(container.Id, cancellationToken);
+                _security.DeleteUserCredentials(BrokerConnectionInfo.GetCredentialAuthority(container.Name));
```
AlexanderSher: Can `DeleteUserCredentials` throw anything?

karthiknadig: No, it does not.

AlexanderSher: Ok, then let's move it below the `try/catch` block and add `return` after `ShowMessage`.

Issue #4255, opened 2018-01-02
Update docker images and download links


karthiknadig:
1. Add dockerfile for the latest R 3.4.3
2. Update MRO download links (download location has been updated)

karthiknadig:
Looks like there is a issue with dotnet installation on linux. Investigating.

Neil-Schneider:
Not sure what issues you are running into or what your current dockerfile looks like, but this is working on my system. Hope it helps. 

```
# Get Packages
RUN \
    apt-get update && \
    apt-get install -y \
        ca-certificates \
        curl \
        lsb-core \
        apt-transport-https && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

## Get key for dotnet-dev install
## hkp:// on port 80 is to avoid some local firewall issues.
RUN sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ xenial main" > \
    /etc/apt/sources.list.d/dotnetdev.list' && \
    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893 && \
    rm -rf /var/lib/apt/lists/* && \
    apt-get clean

RUN apt-get update --fix-missing && \
    apt-get update && \
    apt-get install -y \
        dotnet-dev-1.0.4 \
        libexplain51 \
        libzip4 \
        libc6 \
        lshw \
        ssl-cert \
        wget && \
    rm -rf /var/lib/apt/lists/*

# install rtvs-daemon
RUN wget -O rtvs-daemon.tar.gz https://rtvs.blob.core.windows.net/download/rtvs-daemon_1.3.31103.1002.tar.gz && \
    tar -xvzf rtvs-daemon.tar.gz && \
    ./rtvs-install -s
```

karthiknadig:
@Neil-Schneider Thanks for letting me know :)

PR #4269, opened 2018-01-24
Update RTVS-daemon installation instructions

Neil-Schneider:
I was running into strange issues with permissions. I believe this may also be related to the dotnet installation issue mentioned in #4255. I was getting a `Failed to initialize CoreCLR, HRESULT: 0x80004005` when running `rtvsd` and `dotnet`. 

I traced the error down to the `tar` command. It appears that since I was un-tarring the rtvs-daemon package into `/tmp`, that it was overwriting the permission on `/tmp`, which triggers this error. This is because the tar-ball included `.\` in the structure with permissions that were different than `/tmp`. 

One option to avoid this was to include `--no-overwrite-dir` in the `tar` command, but I personally felt it was safer to place it in it own folder.


Neil-Schneider:
Since this was done from work, I am working on getting official CLA approval. Even though this was a small addition.

MikhailArkhipov:
Thanks!

PR #4300, opened 2018-03-26
Remove unnecessary binding redirects

AlexanderSher:
From user's email:

_The users of our tools (HDI, ASA, and Data Lake tools for VS) are running into assembly resolution issues for System.Net.Http, and we finally discovered that the root cause is that R Tools for VS changes the VS assembly binding for 4.0.0.0 to 4.1.1.1 for everything in VS.  In our apps, we have some app domains that fail to find 4.1.1.1 when they try to resolve System.Net.Http (looks that version is not standard) so it breaks those scenarios for us.  Could this be changed so that R Tools doesn’t everything else running in VS?_

### DIFF
--- a/src/Package/Impl/Properties/AssemblyInfo.cs
+++ b/src/Package/Impl/Properties/AssemblyInfo.cs
@@ -4,12 +4,8 @@
 using Microsoft.VisualStudio.Shell;
 using System.Runtime.CompilerServices;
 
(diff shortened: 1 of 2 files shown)


MikhailArkhipov:
👍

MikhailArkhipov:
I can deal with updating VS.

PR #4303, opened 2018-03-30
Use different package versions for .net64 and .netstandard builds

AlexanderSher:
.netstandard packages `Microsoft.Extensions.FileSystemGlobbing` and `Microsoft.Extensions.Logging.Abstractions` of version 1.1.x aren't compiled specifically against .net46, and so they require separate version of `System.Net.Http`. Newer version of those packages are compatible only with .netstandard 2.0, so we need to use older versions (1.0.1) with .net46 and 1.1.x with .netcore


MikhailArkhipov:
👍

PR #4319, opened 2018-05-11
Fix #4317: Recent VS update killed R Tools plugin

### LINKED ISSUE #4317: Recent VS update killed R Tools plugin
After a standard procedure of installing new updates - as prompted in notifications - pretty much every single operation that R Tools performs and that involves file manipulations - fails.
So far it is confirmed on two different x64 machines with the same update, but different set of installed tools (both however include at least .NET & RTVS tools).

R Tools is 1.3.40402.1027.
Studio is 15.7.1.

The exception thrown displays a MessageBox with following string:
"Could not load file or assembly 'Microsoft.Extensions.FileSystemGlobbing, Version=1.0.1.0, Culture = neutral, PublicKeyToken = adb9793829ddae60' or one of its dependencies. The system cannot find the file specified."

This message is generated by **every** shortcut command available through icons in the interactive window (like Reset, Source files, attach debugger, etc), by almost all subitems of "R Tools" menu in the main window Menu Bar (I did not check every single option, but out of all checked every single one displayed error message and did nothing). Same goes for all the hotkeys I use (like simple Ctrl + Shift + S).

R session seems to be fine, all commands entered directly into the command prompt or through sourcing an *.R file appear to work, plots are generated and displayed in the R Plot window.

All other menu items and options not related to RTVS work fine, at least one other .NET Framework project loads, compiles, and runs without any errors and VS accepts all hotkeys. So the issue is related to RTVS plugin.

Here is a screenshot:

rtvs_bug

AlexanderSher:
There will be matching change in setup.
Broker and UserProfile is moved into subfolder to avoid conflicts with VS assemblies
With this change, we probably need to run bug bash.


MikhailArkhipov:
I will schedule bug bash next week when I get the change built. Maybe we'll add tool window fix too by then.

